PR #2, opened 2015-09-12
Update README.txt

bapch:
Updated URLs and corrected some typography errors.


danomatika:
A modified version of this was accepted upstream. Thanks!

PR #7, opened 2015-12-15
Use long long for PD_LONGINTTYPE on 64-bit Windows

lukexi:
This matches the pointer size (regular longs are size 4 on 64-bit Windows rather than 8) and yields, in my testing so far, a perfectly working 64bit Pd on Windows.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -83,7 +83,11 @@ typedef unsigned __int64  uint64_t;
 
 /* signed and unsigned integer types the size of a pointer:  */
 #if !defined(PD_LONGINTTYPE)
+#if defined(_WIN32) && defined(__x86_64__)
+#define PD_LONGINTTYPE long long
+#else 
 #define PD_LONGINTTYPE long
+#endif
 #endif
 
 #if !defined(PD_FLOATSIZE)


umlaeute:
Please not that the proper way to contribute to Pd is by sending patches via sourceforge.

https://sf.net/projects/pd/patches

danomatika:
@millerpuckette Any thoughts on this? Otherwise I'll close this issue and ask for a resubmit via SF.

hjek:
@umlaeute If this issue tracker is not to be used, maybe it should be disabled (in repository settings)?

danomatika:
It's not that the issue track is *not* used, it's more that this is an unofficial mirror so PR's submitted here are not directly merged into the upstream repo. Submitting a PR here is not a bad idea as it's a way to show the work you're doing but the best method is to make a post on pd-dev and include a link to this PR.

It's mainly that we don't have a completely GitHub only workflow but, in some ways we're working on it.

millerpuckette:
It looks like I intended for this to be managed by defining PD_LONGINTTYPE via a
compilation flag.  (I don't think any finite amount of ifdeffing will get it right; cf. the alloca
mess :)

umlaeute:
@millerpuckette in any case, the patch here only sets `PD_LONGINTTYPE` if it is not already set (e.g. via a compilation flag).
also, the size of integers on various 64bit platforms is [pretty well known](https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models), so I would vote to actually include the patch (without having tested the actual result)

danomatika:
This is probably good and needed. We can check for sure when we update the autotools build for windows.

danomatika:
>It looks like I intended for this to be managed by defining PD_LONGINTTYPE via a compilation flag

How do we set it in CFLAGS? I'd think something like -DPD_LONGINTTYPE="long long" would become a string.

umlaeute:
see [INSTALL.txt](https://github.com/pure-data/pure-data/pull/275/commits/a353f18f0678c44f1992c40f401ebbda3013a149) or the [AppVeyor config](https://github.com/pure-data/pure-data-ci/blob/master/AppVeyor/appveyor.yml)

oh, and you should set it in `CPPFLAGS` rather than `CFLAGS` :-)

danomatika:
I think this PR is a good idea. In our Windows build testing, it basically always needs to be set. The same is true for libpd builds for 64 bit Windows.

millerpuckette:
merged to master branch

PR #5, opened 2015-12-05
live dialog updates

danomatika:
**_This is a reference PR. Do not merge._**

Updates to dialog boxes for live updates on OS X since the Apply button has been removed to better match Apple's Human Interface Guidelines. Also includes updates for better keyboard editing.

Approach:

* entry boxes apply changes when Return is pressed after editing
* checkbuttons & radio buttons send changes
* focus is removed from Ok button (aka Return to exit) when editing an entry box
* OK button takes focus when entry editing is finished
* remove widgets from focus list if they do not highlight on focus

TODO:

* find a way to highlight check buttons & radiobuttons when focused 

Dialogs done so far:

* dialog_atom.tcl
* dialog_canvas.tcl

Related discussion:

* [[PD-dev] Canvas Dialog Apply](http://lists.puredata.info/pipermail/pd-dev/2015-11/020478.html)
* [number and symbol atom properties dialog width text field not selected](http://sourceforge.net/p/pure-data/patches/565/)

[WIP screengrab video](https://www.youtube.com/watch?v=PZCjgIFMc9g):

[WIP screengrab video](http://www.youtube.com/watch?v=PZCjgIFMc9g)

PR #4, opened 2015-12-05
overall iemgui dialog cleanup

danomatika:
**_This is a reference PR. Do not merge._**

Overall cleanups to dialog_iemgui.tcl:

* replace header strings with label frames
* generate label strings in tcl based on uitype (currently using $mainheader)
* tighten up dynamic placement (aka less UI movement with changing button sizes, etc)
* removal of most dummy spaces in favor of better dynamic placement/anchoring
* edge alignment and better consistency with dialog_gatom.tcl
* fixed bug where example label font sizing was inconsistent from creation to font change, leading to "jumping" UI sizes
* fixed too small VU scale/no_scale button width

Todo:
* remove header & label string sending from C (these are currently simply overriden for proof of concept)
* update translation files to reflect removal/changes to strings 

[WIP screenshots](https://www.flickr.com/photos/danomatika/23423140732):

WIP screenshot

dialog_gatom for reference:

screen shot 2015-12-05 at 12 03 01 pm

PR #9, opened 2016-03-15
Dialog updates

danomatika:
Here's the merged work for the overall dialog cleanups from #4 & #5.

I essentially tidied up all of the dialogs and:

* added live edit support on OS X in lieu of the Apply button where appropriate: dialog_gatom, dialog_canvas, dialog_array, dialog_iemgui
* added better focus handling by removing focus from widgets that don't highlight so tabbing between entries & the OK button is much easier, this is currently enabled on OS X
* updated pdtk_textwindow buttons to handle OS X Command key, also changed OK button modifier to Mod+D and renamed button to "Done" as the original Mod+T is overshadowed by the dialog_font global binding

I tried to change as little functionality as possible, just fix and clean things up. I made a few cosmetic changes so let me know what should stay and what should revert (ie. no highlight on text window & list boxes, dialog_data background now white, etc).

Comparison screenshots can be found in a [Flickr album](https://www.flickr.com/photos/danomatika/albums/72157663617342144).


danomatika:
@millerpuckette Also, something could go along with this: sending name & title from text & plist. Opening the text editor for either always sets the title to "myname: text". It'd be nice if at least the "myname" was the name of the object type: "qlist" or "text". Even better would be if the name & title were something like: "qlist: editing text", etc.

danomatika:
Should be ready to go. I changed dialog_data & pdtk_textwindow to use the default gray background, as per feedback on pd-list. This way all dialogs are gray while canvas windows are white.

danomatika:
Closing as this work has been merged. Thanks!

PR #11, opened 2016-03-27
[WIP] clang warning fixes

danomatika:
Proposed updates to fix clang warnings when compiling on OS X:

* wrapped = assignments within statements (if, for, while, etc) in parens
* wrapped && comparisons in parens when used in conjunction with ||
* added default: break; to switch statements not explicitly handling all cases
* changed alloca stddef.h include to stdlib.h

Note: these are only files used by libpd. I haven't been able to figure out how to compile Pure Data itself on OSX...


danomatika:
-Wno-parentheses is the quick fix for clang

danomatika:
Accepted upstream.

PR #12, opened 2016-03-27
[WIP] proposed updates to alloca include

danomatika:
This is a proposed set of updates to replace the existing alloca if/def platform includes with platform checks themselves. I'm proposing this mainly since compilation leads to warnings when forgetting to define HAVE_ALLOCA_H on OS X.

This also fixes a bug where BSD needs stdlib.h instead of stddef.h.

There is also a small fix that uncomments the string.h include on OS X in x_list.c to fix an unknown strlen warning.

Also, what's the purpose of the HAVE_ALLOCA define in the following files? 

* x_array.c line 27
* x_list.c line 28
* x_text.c line 30
* m_binbuf.c line 571

It seems like it's overridden, so maybe it's not needed anymore?

### DIFF
--- a/src/d_fft_fftsg.c
+++ b/src/d_fft_fftsg.c
@@ -21,13 +21,13 @@ for another, more permissive-sounding copyright notice.  -MSP
 /* ---------- Pd interface to OOURA FFT; imitate Mayer API ---------- */
 #include "m_pd.h"
 
-#ifdef HAVE_ALLOCA_H        /* ifdef nonsense to find include for alloca() */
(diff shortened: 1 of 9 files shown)


umlaeute:
what's wrong with *not* forgetting to define `HAVE_ALLOCA_H` on a platform that has *alloca.h*?

danomatika:
Nothing is wrong. This is a proposal to make that step unnecessary to being with. If the above can be translated to "it's working, you're doing it wrong, fix your makefile" then fair enough.

At the very least, I'll make a patch for the stdlib & string fix.

danomatika:
Accepted upstream.

PR #15, opened 2016-04-26
Spell Cygwin correctly

hjek:


millerpuckette:
Changing spelling of Cygwin (thanks) but not including the URL because I didn't for any
of the other requirements for compilation.

PR #16, opened 2016-04-27
[WIP] Autotools updates

danomatika:
Here's what I've been doing far with the auto tools build system:
- cleaned up the configure formatting (largely following what [I use in my autotools projects](https://github.com/danomatika/joyosc/blob/master/configure.ac))
- added a large summary print with all the available options etc
- portaudio is no longer configured if not enabled
- upgraded the port audio configure.in from the portaudio git which adds the -Wno-deprecation override needed to build pa on newer OSXs
- fixed automake subdir-objects warning
- fixed autoconf INCLUDES warnings, was renamed to AM_CPPFLAGS long before the required version of autoconf in the configure.ac
- fixed portmidi build due to missing .h file
- added fink & macports paths if they exist on OS X (homebrew uses /usr/local)

There are a few more things I'd like to do, but I wanted to throw what I have up.

Here's what the configure summary looks like (default OS X options):

```

    pd 0.47.0 is now configured

    Platform:             macosx
    Universal build:      no
    Translations:         yes
    Source directory:     .
    Installation prefix:  /usr/local

    Compiler:             gcc
    CFLAGS:               -g -O3 -funroll-loops -fomit-frame-pointer
    LDFLAGS:               
    INCLUDES:              

    External extension:   d_fat
    External CFLAGS:      -fPIC
    External LDFLAGS:     -bundle -undefined dynamic_lookup

    alsa:                 no
    asio:                 no
    jack:                 no
    portaudio:            yes
    portmidi:             no
    fftw:                 no
```


danomatika:
Well, the CI failed, so obviously a few things to fix :)

danomatika:
One issue right now is that portmidi is required to build pd, yet is not enabled by default in configure. Building without `--enable-portmidi` gives me:
~~~
Undefined symbols for architecture x86_64:
  "_midi_getdevs", referenced from:
      _sys_mididevnametonumber in pd-s_midi.o
      _sys_open_midi in pd-s_midi.o
      _sys_listmididevs in pd-s_midi.o
      _sys_get_midi_devs in pd-s_midi.o
      _glob_midi_properties in pd-s_midi.o
      _glob_midi_dialog in pd-s_midi.o
      _sys_mididevnumbertoname in pd-s_midi.o
      ...
  "_sys_close_midi", referenced from:
      _sys_bail in pd-s_inter.o
      _glob_quit in pd-s_inter.o
      _glob_midi_setapi in pd-s_midi.o
      _glob_midi_dialog in pd-s_midi.o
  "_sys_do_open_midi", referenced from:
      _sys_open_midi in pd-s_midi.o
  "_sys_poll_midi", referenced from:
      _sys_pollmidiqueue in pd-s_midi.o
  "_sys_putmidibyte", referenced from:
      _sys_pollmidioutqueue in pd-s_midi.o
      _sys_queuemidimess in pd-s_midi.o
      _outmidi_byte in pd-s_midi.o
  "_sys_putmidimess", referenced from:
      _sys_pollmidioutqueue in pd-s_midi.o
      _sys_queuemidimess in pd-s_midi.o
ld: symbol(s) not found for architecture x86_64
~~~

In a similar way, if you disable portaudio and do not enable any other audio api, you can't build Pd. Maybe portmidi should be enabled by default like port audio?

The real solution is to fall back and compile with the DUMMY interfaces. I tried using `-DUSEAPI_DUMMY` while disabling portaudio, but no change.

umlaeute:
is the "depends on portmidi" a misfeature of your autotools-updates? because ti works her with Pd-0.47.0test2...

danomatika:
No. It was like this on OS X before I even started. Thanks for assuming I don't bother to check anything.

danomatika:
I think it's related to the portmidi=yes being set in the OS detected in configure.ac, but then being overwritten by the AC_ARG_ENABLE i.e. portmidi=$enableval. So if the --enable-portmidi option is not set, portmidi gets set to no.

umlaeute:
working on this (falling back to dummy)

danomatika:
Ok cool. Here's what I have so far.

Configure.ac, after all the other options around line 312:

~~~
# check if there is no audio api
if [ test x$alsa = xno ] && [ test x$asio = xno ] && \
   [ test x$jack = xno ] && [ test x$portaudio = xno ] ; then
    AC_MSG_WARN([No audio API specified or detected, using dummy api])
    dummyaudio=yes
fi
AM_CONDITIONAL(DUMMYAUDIO, test x$dummyaudio = xyes)

# check if there is no midi api
if [ test x$alsa = xno ] && [ test x$jack = xno ] && [ test x$portmidi = xno ] ; then
    AC_MSG_WARN([No MIDI API specified or detected, using dummy api])
    dummymidi=yes
fi
AM_CONDITIONAL(DUMMYMIDI, test x$dummymidi = xyes)
~~~

src/Makefile.am, around line 104 and replacing the FIXME part:

~~~
# if building without an audio api, use the dummy
if DUMMYAUDIO
pd_CFLAGS += -DUSEAPI_DUMMY
pd_SOURCES += s_audio_dummy.c
endif

# if building without a MIDI api, use the dummy
if DUMMYMIDI
pd_SOURCES += s_midi_dummy.c
endif
~~

It's not working yet, but kind of what I was thinking would be easiest/make the most sense.

danomatika:
Looks good, thanks! I will test later today.

danomatika:
Some updates:

* dummy api building now works and runs, fixed related small tcl bug in audio dialog
* default configure settings now working aka portaudio & port midi
* building localizations is now optional with a configure argument, but is enabled by default if get text found
* updated universal.m4 to check *each* given arch and throw out those not allowed by the system linker, this means the same arch list will work on older Mac OS X with ppc and on newer OSXs which do not support ppc anymore (basically 10.8+)
* added x86_64 as a universal target as well as a linux 32 & 64 bit arch
* universal builds working, I'm able to build for 32bit and load externals now :)
* removed core audio argument since it's implicitly required for PortAudio on OS X
* added Language line to po files to fix gettext warning
* added install step to build version string from m_pd.h and set it in 1.introduction.txt (adapted from makefile.mac), could be replaced with$ VERSION or $PACKAGE from AC_INIT

danomatika:
Now passes `make distcheck`.

umlaeute:
yes, but travis needs to be tolds to run `make distcheck` manually.

danomatika:
I added a more advanced script to create the Pd OS X .app bundle which can use the system Wish.app by version or the included Wish.app. The default is to build using the current version of TK's Wish.app that's found on the system. The script and needed resource files are now added to the distribution tarsal.

I also added a "app" make target so the following will build Pd & create an app bundle with the current $VERSION:

    ./autogen.sh
    ./configure
    make
    make app

tcl/Makefile.am
```diff
@@ -2,22 +2,44 @@ AUTOMAKE_OPTIONS = foreign
...
+    pd_menucommands.tcl \
+    pdtk_text.tcl \
+    pdtk_textwindow.tcl \
+    scrollbox.tcl \
+    pd_guiprefs.tcl \
+    pd_deken.tcl \
+    pd.ico
 
 etags: TAGS
-	etags --append --language=none --regex="/proc[ \t]+\([^ \t]+\)/\1/" *.tcl
```
umlaeute: oops, Makefile targets need to be tab-indented...

danomatika: fixing...

danomatika: fixed

danomatika:
I'm thinking of changing indentation to tabs for all the Makefile.ams to avoid the previous issue. Mixing spaces and tabs throws my text editor out of whack so sometimes copy paste/updates are dropping the tabs. Bad for in Makefiles anyway...

umlaeute:
is `distcheck` supposed to succeed now (after e490908)?

i was going to add the `distcheck` target to travis, but noticed that it fails locally.

danomatika:
No, you need the commit that fixes the permissions on those files. The it will pass.

umlaeute:
hmm, no.

i just did a fresh clone (9532fdc2), and it fails:


```
./autogen.sh && ./configure && make distcheck
[...]
make[1]: Entering directory '/tmp/src/pure-data/pd-0.47.0/_build/sub'
ERROR: files left after uninstall:
./include/pa_jack.h
./include/pa_linux_alsa.h
Makefile:745: recipe for target 'distuninstallcheck' failed
make[1]: *** [distuninstallcheck] Error 1
make[1]: Leaving directory '/tmp/src/pure-data/pd-0.47.0/_build/sub'
Makefile:689: recipe for target 'distcheck' failed
make: *** [distcheck] Error 1
```

danomatika:
Looks like a platform difference as it's working for me. Would be nice if portaudio updated their setup to include dest targets :P

Also, why would it be installing portaudio? We only want a convenience lib right or?

(anyway, bed)

umlaeute:
because `distcheck` is not about deploying binaries, but about creating a complete source tarball (which needs the source code for the convenience libs).

danomatika:
Sure. But I'm referring more to the fact that running `make install` is currently installing the portaudio headers and libs, at least on my system. Is that expected?

Judging from your earlier issue, it looks like it may have installed *all* the portaudio headers and then did not uninstall them in the distcheck cycle.

umlaeute:
yes this is correct.
i don't know whether this is expected (but i think that the use case of a convenience-only library with their own autotools chain is considered exotic and not really handled by autotools).

i think the reason why i see the problem and you don't is that my portaudio  detects jack and installs some header-files accordingly (but doesn't uninstall them).

umlaeute:
there's an (unanswered) [StackOverflow question](http://stackoverflow.com/questions/25065337/) about this.

danomatika:
Maybe the better solution is to move to building portaudio as a more manual libtool lib like portmidi?

enohp ym morf tnes
--------------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
I put together a Makefile.am to build portaudio using libtool like how we currently build portmidi. This now means, installing pd no longer installs portaudio which, more probably than not, would install a different version of PA over whatever the user currently has installed. The better solution would be to use a system portaudio & portmidi, if available, at least in my thinking for Linux. As it is now, it's nice that pd comes with the main dependencies so you can build it relatively easily.

This portaudio lib tool build, of course, will need to be checked on other platforms, but it's working on OSX.

Also, I added some build targets which populate the `bin` folder with symlinks to the pd binaries. This way, a `make all` will result in a similar setup as with the src/makefiles.

umlaeute:
hmm, are we sure this is heading into the right direction? while i understand the motivation behind all these changes, every change to portaudio/portmidi will make it more complicated to update the bundled libraries for @millerpuckette (who i understand not to be a loving autotools head).

e.g. would it be more worthwhile to invest the time to get upstream portaudio behave properly and then lobby towards including a newer portaudio release into Pd?

danomatika:
Sure.

But I always wondered why we bothered having the auto tools build at all if:

* there are 2 competing ways to build pd
* the main users of pd are told to use the makefiles

The was my attempt to make the auto tools to build work as easily as the regular makefile build. It now does, essentially with a lot more configurability.

This work, of course can be cleaned up, squashed etc in a new branch if the number of commits is the issue.

In fact, the new portaudio approach is using the same approach as how pd is built in the makefile build. So the maintainability factor is no different.

As it was before today, updating the included portaudio involved manual modifications to portaudio/Makefile.in & portaudio/configure.in (those of which were not documented and took me many hours to understand). As of the last few commits, updating portaudio involves copying all the contents of the portaudio source distribution *except for Makefile.am* into the portaudio directory. That seems far simpler and more maintainable to me.

If I follow your approach, I need to fix portaudio, portmidi, etc for newer OS X's as they are still using Carbon APIs which have been deprecated for many years. So now I need to rewrite the backends for those as well .. .weee. Dannenberg, et al have grad students that should be doing that.

All of this is moot for the fact that until I started working on this, I *COULD NOT BUILD PD* on my OS X machine. Now that I can and can actually help with lower level development, you say it's not needed. 

Wee again.

danomatika:
Also, this branch is meant to promote discussion as, so far, it seems like I can't get very much feedback until I actually do some work...

danomatika:
Ok, finished up what I was working on earlier:

* Added --without-local-portaudio & --without-local-portmidi configure options so pd can be optionally built with system installed libraries. The default is to build with the included libs as before.
* Fixed AM_CPPFLAGS (former INCLUDES) so it actually propagates down the individual makefiles. This fixes include paths were being added but showing up when compiling so the optional OS X includes now work. This means I can now build pd using the portaudio & port midi installed via homebrew, macports, etc.

That's about all the things I had on my list for this branch. Should be good for testing as it passes distcheck for me.

danomatika:
Also, as to portaudio & port midi maintainability, it occurs to me that putting both libs into subfolders would make this even easier:

~~~
portaudio/Makefile.am
portaudio/portaudio/* <-- portaudio distribution

portmidi/Makefile.am
portmidi/portmidi/* <-- portmidi distribution
~~~

Separating the generated makefiles from the distribution this way means @millerpuckette only needs to extract the new version of either lib into it's respective subfolder without needing to worrying about overwriting or modifying any of the build files. I argue that this would end up being far *easier* than the existing situation (master branch).

Obviously, we'll need to modify the paths in the makefiles but that's easy.

danomatika:
Hah, one more :)

Finished a TODO listed in configure.ac: --enable-static option now passes -static to the LDFLAGS.

Tested by building pd with static linking and system installed portaudio & port midi via:
~~~
./configure --without-local-portaudio --without-local-portmidi --enable-static
~~~

which gives me a pd without dynamic linking to the libs (only to the OS X frameworks & C lib):

~~~
danomatika:pure-data dano$ otool -L src/pd
src/pd:
	/System/Library/Frameworks/CoreAudio.framework/Versions/A/CoreAudio (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/CoreMIDI.framework/Versions/A/CoreMIDI (compatibility version 1.0.0, current version 73.0.0)
	/System/Library/Frameworks/AudioUnit.framework/Versions/A/AudioUnit (compatibility version 1.0.0, current version 1.0.0)
	/System/Library/Frameworks/AudioToolbox.framework/Versions/A/AudioToolbox (compatibility version 1.0.0, current version 492.0.0)
	/System/Library/Frameworks/Carbon.framework/Versions/A/Carbon (compatibility version 2.0.0, current version 157.0.0)
	/System/Library/Frameworks/CoreFoundation.framework/Versions/A/CoreFoundation (compatibility version 150.0.0, current version 1258.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1226.10.1)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 120.1.0)
	/System/Library/Frameworks/CoreServices.framework/Versions/A/CoreServices (compatibility version 1.0.0, current version 728.9.0)
~~~

This means you can build pd with a self installed portaudio or portmidi and run it on another system.

danomatika:
Never mind on the static link option. It breaks the --enable-universal option and is not really portable. I tried some individual static link options but they don't translate from gcc to clang anyway, so not really portable.

umlaeute:
sorry if i upset you with my last comment.
i'm all for the changes you propose, my concern was mainly about the applicability of them.

if, as you say, the import of new portaudio will actually be *easier* after your changes have been merged, then this is about the best outcome i can think of :yum:

danomatika:
It was late last night and I had worked on it all day by that point... ;)

That being said, I feel I haven't communicated some things effectively. I've been doing more than fixing things for OS X:

1. I understand that this overall diff looks pretty big.
2. I've spent *alot* of time making sure that the changes I am making do not break current functionality and honor previous defaults while at the same time handling platform differences, ie. building Pd on old versus new OS X.
3. My approach is to make the setup both more readable and maintainable ie. make your job easier.
4. The configure script should be as *helpful* as possible in providing useful overrides as well as useful output.
5. I'm most probably not going to disappear if this work does get accepted at some point. :)

My thought is that if the auto tools build *just* works, then it will be much easier for people to test non-trivial work done in branches. This could also take some of the weight off of @millerpuckette when the test build times come again.

Also, now you can generate a distribution package easily as well an OS X app (while choosing which Tk/Wish version to use). I feel the easier and more helpful the build process is, the easier it is for people to contribute by testing etc.

millerpuckette:
Is this ready for me to merge into Pd?  (In particular, I'm hoping this doean't break debian package building, which I think Iohannes can check :)

danomatika:
Not just yet. There is one or two issues I want to address, one of them being the portaudio / portmidi subfolder option.

danomatika:
The latest commit moves portaudio & portmidi into subfolders which should greatly aid in future maintenance. This means the auto tools makefiles are now in a separate place, so no need to worry about overwriting something we *need*. I added readmes for both that explain what's going on an how to update.

A few earlier commits handle things like:

Making sure *all* folders installed by pd are removed during an uninstall. I noticed /usr/local/lib/pd was left behind with all of it's folders but no content, so now it removes that folder as well. I'm not 100% sure that's desired right now though as it would remove anything extra that's there aka manually installed externals. I think I may change it so it only removes if all the subfolders are empty...

Installing also installs the lib tool .la files for the built in externals (bonk~, etc). Since these are built as dynamic modules (aka plugins), no linking info really needs to be kept so I removed the .la files in a post install step.

 Symlinks for externals are created from the hidden .libs lib tool folder so running pd locally (aka via cd bin && ./pd) now loads the externals correctly.

Last, I realized I screwed up and was installing the localization .po files when making the OS X app. Now we correctly load the .msg files instead.

umlaeute:
cool.

however, i do not think that `make uninstall` should/must remove any folder that contains data not installed by `make install`. (i'm not even sure it should remove the empty folders, though i'm not opposed either)

danomatika:
Yeah I agree.

It's curious, though, that it was leaving the entire dir empty, including all the subfolders within `/usr/local/lib/pd/doc`. Maybe that's something happening on my system but not on GNU/Linux? Anyway, it might be related to the non-standard install path aka:

docs -> $(prefix)/lib/pd instead of $(prefix)/doc/pd etc

danomatika:
As a follow up, we're not thinly ones seeing this ... http://stackoverflow.com/questions/19521335/how-can-i-get-make-uninstall-to-remove-empty-directories-as-well-as-files

danomatika:
Doh. One reason: removing the lib tool .la files means the uninstall leaves behind the build externals. If I *do not* remove them after install, then uninstall works correctly in extra. So maybe we leave them :P

umlaeute:
yes that is one problem i also just discovered.
the other is, that the unconditional removal of the directories breaks the idempotency of `make uninstall`:

```
$ rm -rf /tmp/foo
$ make install DESTDIR=/tmp/foo >/dev/null 2>&1 && echo "ok"
ok
$ make uninstall DESTDIR=/tmp/foo >/dev/null 2>&1 && echo "ok"
ok
$ make uninstall DESTDIR=/tmp/foo >/dev/null 2>&1 || echo "KO"
KO
$ make uninstall DESTDIR=/tmp/foo 2>&1 | tail
make[1]: Entering directory '/root/pd/src'
 ( cd '/tmp/foo/usr/local/bin' && rm -f pd pdsend pdreceive )
 ( cd '/tmp/foo/usr/local/include' && rm -f m_pd.h )
 ( cd '/tmp/foo/usr/local/lib/pd/bin' && rm -f pd-watchdog pd )
/bin/bash: line 7: cd: /tmp/foo/usr/local/lib/pd/bin: No such file or directory
Makefile:709: recipe for target 'uninstall-libpdbinPROGRAMS' failed
make[1]: *** [uninstall-libpdbinPROGRAMS] Error 1
make[1]: Leaving directory '/root/pd/src'
Makefile:483: recipe for target 'uninstall-recursive' failed
make: *** [uninstall-recursive] Error 1
$
```

danomatika:
That's weird, uninstall-hook is supposed to happen *after* all the main uninstall stuff.

umlaeute:
yes, this is what it is doing.

however, when i *re-run* `make uninstall`, then it will fail (in the `src/` subdirectory).

for whatever reasons it does a `cd /path/ && rm -f file` rather than `rm -f /path/file`.
while this would succeed if the file was not there (due to the `-f`), it blatantly fails if the directory is missing.

danomatika:
To confirm, with the current commit, the following folders are left behind after:

    make install
    make uninstall

`find /usr/local/lib/pd` gives me:
~~~
/usr/local/lib/pd
/usr/local/lib/pd/doc
/usr/local/lib/pd/doc/1.manual
/usr/local/lib/pd/doc/2.control.examples
/usr/local/lib/pd/doc/3.audio.examples
/usr/local/lib/pd/doc/4.data.structures
/usr/local/lib/pd/doc/5.reference
/usr/local/lib/pd/doc/6.externs
/usr/local/lib/pd/doc/7.stuff
/usr/local/lib/pd/doc/7.stuff/soundfile-tools
/usr/local/lib/pd/doc/7.stuff/synth
/usr/local/lib/pd/doc/7.stuff/tools
/usr/local/lib/pd/doc/sound
/usr/local/lib/pd/doc/startup
/usr/local/lib/pd/doc/startup/disabled
/usr/local/lib/pd/extra
/usr/local/lib/pd/extra/bob~
/usr/local/lib/pd/extra/bonk~
/usr/local/lib/pd/extra/choice
/usr/local/lib/pd/extra/fiddle~
/usr/local/lib/pd/extra/loop~
/usr/local/lib/pd/extra/lrshift~
/usr/local/lib/pd/extra/pd~
/usr/local/lib/pd/extra/pique
/usr/local/lib/pd/extra/sigmund~
/usr/local/lib/pd/extra/stdout
/usr/local/lib/pd/tcl
~~~

danomatika:
Also a different question: what should the external extension be for Mac? I see both d_fat & pd_darwin. It seems like d_ppc & d_fat are old now and pd_darwin matches better with pd_linux, etc. I have configure setting pd_darwin but I want to double check. It seems like pd opens both anyway.

umlaeute:
but the line that is failing is

     ( cd '/usr/local/lib/pd/bin' && rm -f pd-watchdog pd )

and i don't see `/usr/local/lib/pd/bin` in your list of existing directories.

please run `make uninstall` two-times:

    make install
    make uninstall
    make uninstall

danomatika:
Works fine, well by that I mean, it uninstalls bin fine without error, but still leaves the empty folders.

I think it think it needs touninstall pd-watchdog but, I believe, pd-watchdog is used on Mac but not used on Linux. I'll check the src/Makefile.am.

umlaeute:
`pd-watchdog` is used on both Darwin and Linux.

danomatika:
Yeah I see that now. It's not used on Win.

umlaeute:
which `automake` are you using?

danomatika:
Ah, I think it's related to `libpdbindir = $(pkglibdir)/bin` being set for Linux but not for Mac. It's installing the BINS to mu /usr/local/bin folder but not yours...

danomatika:
automake (GNU automake) 1.15
autoconf (GNU Autoconf) 2.69

umlaeute:
ah, your explanation about `libpdbindir` sounds very reasonable

umlaeute:
regarding your question about extensions: `d_fat` is the current one and supersedes `pd_darwin`.
this is mainly backed up by my memory and the fact that `src/makefile.mac` builds to `d_fat`.

also, 'd_ppc` nicely aligns with `l_arm` and `m_i386`

danomatika:
Ok, so d_fat it is. It just seems like all the 3rd party externals compile to pd_darwin (GEM for instance) but, if Pd loads either fine, then it doesn't matter to much. :)

umlaeute:
another thing: you have `--disable-portaudio` even though the *site-wide* default for portaudio is `no`.

danomatika:
>another thing: you have --disable-portaudio even though the site-wide default for portaudio is no.

That's related, I think, that the underlying configure variable is actually --enable-portaudio which is by default set to "no" -> aka --disable-portaudio.

umlaeute:
todo:
- [ ] configure should refuse to link with both OSS-midi and portmidi

danomatika:
Got the todo via:

    test x$oss = xyes || portmidi=no

danomatika:
Ok, put in an oss test which should disable portmidi if oss is being used. I guess rumors of OSS's demise were greatly exaggerated in the Linux community?

danomatika:
Good catches.

I've narrowed the uninstall issue down to those files which are given to automate using the "nobase_" prefix.

danomatika:
Latest commit resolves the uninstall issue by doing a recursive removal of empty dirs for:

* $(prefix)/include/pd
* $(prefix)/lib/pd

I feel this is the safest & sanest method, judging from this Automake list discussion: https://lists.gnu.org/archive/html/automake/2004-03/msg00101.html

Basically, Automake doesn't know who owns what dirs so doesn't remove dirs by default, i.e.. in the case of /usr/local/bin. As the discussion notes though, it makes the most sense to instruct Automake what it should remove manually rather than rely on auto detection. I feel the uninstall-hooks fit that regard and fall back silently if another package (i.e. external) was installed to the given location.

umlaeute:
unfortunately a double `make uninstall` still fails.

danomatika:
Not for me. What's the error?

umlaeute:
the same as before.
please check this on a linux system.

umlaeute:
```
zmoelnig@umlautQ:~/src/pd $ git describe --always
12a5b99
zmoelnig@umlautQ:~/src/pd $ make uninstall DESTDIR=/tmp/foo 2>&1 | tail
make[1]: Entering directory '/home/zmoelnig/src/pd/src'
 ( cd '/tmp/foo/usr/local/bin' && rm -f pd pdsend pdreceive )
 ( cd '/tmp/foo/usr/local/include' && rm -f m_pd.h )
 ( cd '/tmp/foo/usr/local/lib/pd/bin' && rm -f pd-watchdog pd )
/bin/bash: line 7: cd: /tmp/foo/usr/local/lib/pd/bin: No such file or directory
Makefile:746: recipe for target 'uninstall-libpdbinPROGRAMS' failed
make[1]: *** [uninstall-libpdbinPROGRAMS] Error 1
make[1]: Leaving directory '/home/zmoelnig/src/pd/src'
Makefile:494: recipe for target 'uninstall-recursive' failed
make: *** [uninstall-recursive] Error 1
zmoelnig@umlautQ:~/src/pd $
```

danomatika:
Looks like it's related to Linux using both bin_PROGRAMS and libpdbin_PROGRAMS while Mac only uses bin_PROGRAMS in `src/Makefile.am`. Why do we need to install the binaries to 2 locations?

danomatika:
I can confirm the issue if I add the libpdbin lines (aka install some bins to $(prefix)/lib/pd/bin) to the MACOSX section:

~~~
libpdbindir = $(pkglibdir)/bin
libpdbin_DATA =
libpdbin_PROGRAMS = pd-watchdog pd
~~~

umlaeute:
actually i think that OSX should install `pd-watchdog` to `$libpdbindir` as well: it is a helper-binary that is not to be used by any user, and such a thing should not go to `$bindir`

millerpuckette:
I agree with that... maybe pdsend and pdreceive should be in $bindir though.

cheers
Miller

danomatika:
Yeah, I think it should be:

$(prefix)/bin -> pd, pdsend, pdreceive  
$(prefix)/lib/pd/bin -> pd-watchdog

Aso, right now it looks like (on Mac at least), pd-gui.tcl is also installed to $(prefix)/bin.

umlaeute:
which is probably ok, as the users are expected to start the GUI.

while the "normal" way might be to start the pd-core which will then launch the pd-gui, the other way round is also possible (pd-gui will launch the pd-core if started as standalone).
there are subtle differences to how Pd behaves when started from the GUI:
- on OSX, the menu will behave properly (iirc)
- launching Pd-GUI multiple times will just re-connect to the (first) running instance, whereas starting Pd-core will launch multiple instances. (e.g. on Debian all the file-associations are with the Pd-GUI; so you can "open" patches from the file-browser, without ending up with zillions of Pd-instances, all competing for the same audio device).


TL;DR i think both `pd` & `pd-gui` should go into `${prefix}/bin`, along with the utilities `pdsend`/`pdreceive`; iirc it made "everything" simpler to have the *other part* (`pd-gui` resp `pd` as launched as a secondary process) in `$(libpdbindir}`, that's why they are installed doubly.

danomatika:
OK, the following?

~~~~
$(prefix)/bin/pd
$(prefix)/bin/pdsend
$(prefix)/bin/pdreceive
$(prefix)/bin/pd-gui.tcl
*
$(prefix)/lib/pd/bin/pd
$(prefix)/lib/pd/bin/pd-gui.tcl
~~~

Is the send for required since launching another pd is looking for a path relate to the tcl folder?

umlaeute:
not sure i understand your suggestion (where's pd-watchdog gone to?).

checking with `src/makefile.gnu` i think that the following should be OK (and that i was mistaken about the 'everything easier' part)

```
$(prefix)/bin/pd
$(prefix)/bin/pd-gui.tcl
$(prefix)/bin/pdsend
$(prefix)/bin/pdreceive

$(prefix)/lib/pd/bin/pd-watchdog
```

danomatika:
I edited the post after the initial send, it's there.

Just double checking, here's my commenting on this:

In `src/Makefile.am`:
~~~
# install watchdog to $(libdir)/pd/bin as it's not a user facing program,
# install a second pd core to $(libdir)/pd/bin so the tcl scripts can
# launch it if the gui is started first
libpdbindir = $(pkglibdir)/bin
libpdbin_PROGRAMS = pd-watchdog
~~~

In `tcl/Makefile.am`:
~~~
# install a second pd-gui to $(libdir)/pd/tcl so the pd core can
# launch the gui if the core is started first
dist_libpdtcl_SCRIPTS = pd-gui.tcl
~~~

danomatika:
Note, if I do try running the pd-gui.tcl that's installed to `/usr/local/bin`, I get an error that it can;t find all the other tcl scripts. 

~~~
$ pd-gui.tcl
Error in startup script: can't find package pd_connect
    while executing
"package require pd_connect"
    (file "/usr/local/bin/pd-gui.tcl" line 29)
~~~

danomatika:
Hah, I think we're out of sync (at least I am). Sorry about that.

Using your reply with the following now:

~~~
$(prefix)/bin/pd
$(prefix)/bin/pd-gui.tcl
$(prefix)/bin/pdsend
$(prefix)/bin/pdreceive

$(prefix)/lib/pd/bin/pd-watchdog
~~~

umlaeute:
so here' some observations, with the following setup:

```
${prefix}/bin/pd
${prefix}/bin/pd-gui.tcl
${prefix}/bin/pdreceive
${prefix}/bin/pdsend
${prefix}/lib/pd/bin/pd-watchdog
```

- `pd-watchdog` is found without problems :ok:
- starting `${prefix}/bin/pd` will start `${prefix}/lib/pd/tcl//pd-gui.tcl` as the GUI (actually it will use `TCL_LIBRARY=${prefix/lib/pd/lib/tcl/library TK_LIBRARY=${prefix/lib/pd/lib/tk/library ${prefix/lib/pd/tcl/pd-gui.tcl 5400` :ok:
- starting `${prefix}/bin/pd-gui.tcl` will start will bail out with an error:
```
Error in startup script: can't find package pd_connect
    while executing
"package require pd_connect"
    (file "${prefix}/bin/pd-gui.tcl" line 29)
```
- starting `${prefix}/lib/pd/tcl//pd-gui.tcl` will bail out with the *same* error
- starting `TCL_LIBRARY=${prefix/lib/pd/lib/tcl/library TK_LIBRARY=${prefix/lib/pd/lib/tk/library ${prefix/lib/pd/tcl/pd-gui.tcl` (which is taken from the *pd-core starts pd-gui* case), bails out with another error:
```
Error in startup script: couldn't execute "${prefix}/lib/pd/tcl/../bin/pd": no such file or directory
    while executing
"exec -- $pd_exec -guiport $::port &"
    (procedure "main" line 19)
    invoked from within
"main $::argc $::argv"
    (file "${prefix}/lib/pd/tcl/pd-gui.tcl" line 763)
```

danomatika:
Right, I just did the same test and got the same results.

The following layout works for me where starting pd via `$ pd` and the gui `$ /usr/local/lib/tcl/pd-gui.tcl` both work:

~~~
$(prefix)/bin/pd
$(prefix)/bin/pdsend
$(prefix)/bin/pdreceive

$(prefix)/lib/pd/bin/pd
$(prefix)/lib/pd/bin/pd-watchdog
~~~

Should starting pd via `$(prefix)/bin/pd-gui.tcl` work?

Another thought is that the second pd in the `$(prefix)/lib/pd/bin/pd` could be a symlink to the one installed to `$(prefix)/bin`. Sorry if I'm digging in too deep here on things that "just worked" in the past.

umlaeute:
i wouldn't take my word on it that all those things *just worked*. e.g. it could be that starting via `$(prefix)/bin/pd-gui.tcl` actually never worked (I just checked, and Debian really provides a script `/usr/bin/pd-gui` (lacking the extension):
```
#!/bin/sh
exec /usr/lib/puredata/tcl/pd-gui.tcl $@
```

the `${prefix}/bin/pd-gui.tcl is not installed at all by Debian...

note however, that the Debian *does* install a copy of the main `pd` binary as `/usr/lib/pd/bin/pd`, so that the GUI can find it.

danomatika:
Right, `/usr/lib/pd/bin/pd` is needed. How about we do our own `pd-gui` script with the following layout:

~~~
$(prefix)/bin/pd
$(prefix)/bin/pdsend
$(prefix)/bin/pdreceive
$(prefix)/bin/pd-gui <-- script that launches pd-gui.tcl

$(prefix)/lib/pd/bin/pd <-- possible symlink to $(prefix)/bin/pd?
$(prefix)/lib/pd/bin/pd-watchdog
~~~

umlaeute:
i can confirm that a symlink `$(prefix)/lib/pd/bin/pd` to `$(prefix)/bin/pd`  seems to work fine.

i wonder though how is it is to create that symlink with autotools.

danomatika:
>i wonder though how is it is to create that symlink with auto tools.

Just an install-exec-hook target, easy enough. Doing it now.

umlaeute:
right; i meant without hooks and hoops :-)

danomatika:
>right; i meant without hooks and hoops :-)

Nope. It's a dance we play, but that's why the hooks are there...

danomatika:
Also, what's the `$(prefix)/lib/pd/startup` folder for?

umlaeute:
iirc, Pd-extended used to auto-start patches found in that folder (think *Application Builder*)
it seems that this functionality is not in Pd though.

umlaeute:
and looking at the use of `startup/` in the `po/`-files, it seems that this was though of as a place to put GUI-plugins which were shipped with the Pd-distribution but were kept separately.

danomatika:
Ok, latest commits resolve the install stuff:

* pd, pdsend, pdreceive are isnatleld to `$(prefix)/bin`
* pd-watchdog is installed to `$(prefix)/lib/pd/bin`
* a symlink is created from  `$(prefix)/bin/pd` to  `$(prefix)/lib/pd/bin/pd`
* a new `src/pd-gui` launcher script is installed to  `$(prefix)/bin`, as well as a symlink created in the local `bin` dir for local use

The `pd-gui` launcher script has a relative path which is replaced with the full path to  `$(prefix)/lib/pd` during the `make install` step.

Also, I commented out the  `$(prefix)/lib/pd` cleanup step as it's still causing problems with a double `make uninstall` .

danomatika:
Latest commit fixes the uninstall issue so now the installed folders are removed if they are non-empty. Uses a small kludge that relies on the order the makefile subdirs are executed in order to make sure the folder exists that the later `src/Makefile` needs.

umlaeute:
i moved the pd-gui script to tcl/;

i also wonder why we don't just use `tcl/pd-gui.in`:
```
#!/bin/sh
exec @libdir@/@PACKAGE@/tcl/pd-gui.tcl "$@"
```

danomatika:
Yeah, I thought of that. There are 2 options:

1 do it at configure time via pd-gui.in
2 do it at install time via the install-data-hook

Option 2 allows the local /bin usage of pd-gui to work, but at the expense of a more detailed hook. Perhaps that's not a large use case and not needed?

danomatika:
I'm pretty much done with this branch now. Time for testing.

umlaeute:
- I updated the `.travis.yml` file so the Travis-CI autobuilds are also run on their OSX infrastructure (the updates [already showed](https://travis-ci.org/pure-data/pure-data/jobs/132584187) that  the current `master` branch would not build on OSX...

- I also changed the handling of the pd-gui wrapper script slightly, so the `PD_PREFIX` substitution is done by configure, while keeping the `${builddir}/bin/pd-gui` symlinks working.

po/Makefile.am
```diff
@@ -5,6 +5,16 @@ AUTOMAKE_OPTIONS = foreign
...
+if MSGFMT
+XGETTEXT=xgettext
+MSGMERGE=msgmerge
+MSGFMT=msgfmt
+else
+XGETTEXT=echo fake:xgettext
+MSGMERGE=echo fake:msgmerge
+MSGFMT=echo fake:msgfmt
+endif
+
```
danomatika: Is po/Makefile being called without msgfmt? That shouldn't happen...

danomatika: ie. https://github.com/pure-data/pure-data/blob/autotools_updates/Makefile.am#L32

danomatika:
One thing left, though, is to update the build/install instructions about the important configure options and how to build the Mac app.

umlaeute:
I'm aware of the [Makefile.am:32](https://github.com/pure-data/pure-data/blob/autotools_updates/Makefile.am#L32). However, it seems that this was *not enough* for a `make distcheck` run, as can be seen in the [travis build](https://travis-ci.org/pure-data/pure-data/jobs/132586852).

danomatika:
Should the release build include debug info aka "-g"?

umlaeute:
personally i'd say yes, include `-g` in the debug-builds (though I've seen strong opinions against this, as it is too gcc-specific)

umlaeute:
actually, i just saw that you meant using `-g` in *Release* builds.

i'm still all for this (you can always `strip` them away)

umlaeute:
i guess the question is related to 425f41a17bd9682bd9cf94b22ca8d25162811c2e, which i do not understand.

danomatika:
Ok. I'll re-add the "-g" to the RELEASE_CFLAGS.

As for the last commit, from the autoconf manual:

>If output variable CXXFLAGS was not already set, set it to -g -O2 for the GNU C++ compiler (-O2 on systems where G++ does not accept -g), or -g for other compilers. If your package does not like this default, then it is acceptable to insert the line ‘: ${CXXFLAGS=""}’ after AC_INIT and before AC_PROG_CXX to select an empty default instead.

This means it was always setting CXXFLAGS="-g -O2" even when I wanted just a straight debug build without optimizations. That commit adds the line suggested by the manual to make sure AC doesn't set default optimizations for CFLAGS & CXXFLAGS, since we have our own set in RELEASE_CFLAGS.

danomatika:
Also, do you think -O3 is too aggressive? Should it be -O2 for release instead?

umlaeute:
actually, i think we should not worry too much about these flags.
if autotools defaults to `-g -O2` then I think these defaults are good enough.

if you have special needs (e.g. your debug builds need to be `-O0`), then the common wisdom is to add these special flags to your personal build:

    ./configure CFLAGS="-g -O0" CXXFLAGS="-g -O0"
    make

in theory, autotools can be used with a large range of compilers, and not all of them may accept some flags (e.g. `-g`); autotools should be clever enough to only add `-g -O2` to compilers that support it. the more you customize, the more likely things will break.

danomatika:
So no debug build? I think being able to specify debug settings is a good thing.

I would agree that we can drop the "-g -O3" from the release settings and use the default ones.

umlaeute:
if the *only* thing that `--enable-debug` is going to add is `-O0` to the compiler-flags, then it is probably an overkill.

danomatika:
Actually --enable-debug also makes sure the fast-math & other optimizations are not set. Its more than -O0.

danomatika:
Or, are you arguing from the "[do not abuse the auto tools](http://stackoverflow.com/a/4680578/2146055)" standpoint? :)

umlaeute:
i have to admit that i *did* read that article before, and that while initially i wouldn't follow it's argumentation, i can now see the light :wink:

danomatika:
Yeah, I get where that's coming from. I think auto tools users have a tendency to force things on the system such as this example. I think having a script to set debug settings might make the most sense and break things the least. I've been thinking about not forcing `-g` anyway once I read that it's not completely portable to all compilers.

umlaeute:
btw, if you do want to add various compiler flags, i've once written a [couple of m4-macros](https://git.iem.at/pd/iemmatrix/blob/master/m4/iem_checkflags.m4) to test whether a compiler accepts a given flag...

danomatika:
I did some digging regarding the release optimization settings (which I pulled from the makefiles):

* [-fomit-frame-pointer](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-fomit-frame-pointer-724) is generally enabled by defaults we probably don't need to set it: "Enabled at levels -O, -O2, -O3, -Os"
* [-ffast-math](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-ffast-math-905) disables NaNs and such, so maybe best used only when building the pd core. It's only used in one place in the PortAudio configure.in: when building for Windows with ASIO.
* [-funrool-loops](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#index-funroll-loops-925) also seems to be good for the pd core only

So my thinking now is to have a AM variable called DEBUG which is set by default in _configure_. If DEBUG is not set when compiling, then `-ffast-math -funroll-loops` is used when building the pd core in _src/Makefile_. This way, those specific flags do not propagate to other binaries and libs where we perhaps do not want them.

The _--enable-debug_ configure flag then just sets DEBUG and overrides the default CFLAGS optimizations settings. I figure this is a decent tradeoff since most people won't compile for debug and the default can still be overridden by the user.

danomatika:
>btw, if you do want to add various compiler flags, i've once written a couple of m4-macros to test whether a compiler accepts a given flag...

Yeah, I think we can check if `-g` is accepted and add it.

danomatika:
Just made a quick commit that resolves the debug setting issue by *not* overriding the Autoconf CFLAGS and CXXFLAGS default values. Instead, we append our settings which will naturally override things aka "-O0" *after* a default "-O2." Also, we now append user-supplied CLAGS and LDFLAGS so user settings are always last to allow overriding.

One aspect of this was that I needed to move the platform-specific configure.ac section after the calls to AC_PROG so the defaults are set first.

danomatika:
Sweet. Thanks!

PR #18, opened 2016-05-19
fixing log arg

porres:
argument for log~ wasn't being loaded in the object


millerpuckette:
Should already be done.

PR #20, opened 2016-06-09
added commandline option to disable jack automatic connections

x37v:
Sometimes I don't want pd to automatically connect to any jack clients, this enables that. 

```
  -nojackconnect
```


umlaeute:
as mentioned on the IRC, all "negating flags" use "no" as a prefix, e.g. `-noloadbang` and come with an enabling flavour as well (`-loadbang`).
also, it is unclear what the "auto" stands for: auto-connecting (to the system-devices), auto-start (of jack; something we finally got rid of), auto-mobile,...

i would therefore recommend to rename the flag to `-nojackconnect` and add its sibling option `-jackconnect`.

PR #21, opened 2016-06-22
fix multiple definition of glist_reloadingabstraction

avilleret:
see https://lists.puredata.info/pipermail/pd-list/2016-06/115414.html and #https://github.com/jamoma/JamomaPureData/issues/44


avilleret:
fix with : https://github.com/pure-data/pure-data/commit/07a14c540630bc2344a80da30464405900d556fb

PR #23, opened 2016-11-06
Deken progressbar

chr15m:
@millerpuckette this is a progressbar improvement to the deken plugin from @umlaeute.


chr15m:
@millerpuckette you can get the raw patch for use with `git-apply` from here: https://patch-diff.githubusercontent.com/raw/pure-data/pure-data/pull/23.patch

PR #24, opened 2016-11-21
first try applying canvas local paths to canvas's own directory

Ant1r:
abstraction example:

A test patch is calling an abstraction [dir1/abs1] located in subdir "dir1".
abs1 itself declares a dir1-subdir "dir2" : [declare -path dir2], and tries calling [abs2] located in dir2.
Currently [abs2] won't load (although [dir2/abs2] will). This fixes that.


millerpuckette:
I think there's a bigger issue - patch1 calls patch2 which calls patch3 - what if the search should have
been done relative to the directory holding patch2?  It gets complicated.  My current idea is to search through all of them, from the inside outward - but somehow avoiding making lots of duplicate attempts.  This will take some coding :)

Ant1r:
I don't fully understand:
> patch1 calls patch2 which calls patch3 - what if the search should have
> been done relative to the directory holding patch2
Rightly this is the point of this PR: when looking for patch3, do search the paths [declare]d within patch2 relatively to the directory holding patch2.

To me this feature is really missing, because for now it's quite impossible to build a modular abs library in a clean way:
a lib that would be subdivided in multiple subdirs, but which could be simply declared with one single [declare -path myLib]; abstractions would just have to sub-[declare] local private stuff subdirectories relatively to their own dir (so no change involved if you move the upper lib directory, only change the main [declare]d path).

Ant1r:
I've fixed a case of duplicate path search (that was caused by the first commit): don't try to apply declared path locally to a patch if it's a toplevel patch, because toplevel will be search after anyway.
So finally this PR shouldn't bring duplicate search.

BTW: I wonder if the own patch's dir shouldn't be searched at the very first time, before the declared paths and every owner's ones. This would give a way to locally overload an abstraction contained inside of a shared declared library's directory.

umlaeute:
is this not what `-declare -path .` would do?

Ant1r:
> is this not what -declare -path . would do?

Right, this would add . to the firstly searched paths, but you can't be sure that it will be searched first of all, which you need for overloading. 

-----
I would like to explicit the primary request addressed by this PR: 
For a patch patch1 located into dir1, if dir1 contains a subdir dir2, and if dir2 is declared as path ([declare -path dir2]) by patch1, then patch1 should load patch2 located in dir2. 

Currently this is only true if patch1 is a toplevel patch, but not if patch1 is called by another patch as an abstraction, see the included test.
[test_declared_path.tar.gz](https://github.com/pure-data/pure-data/files/693965/test_declared_path.tar.gz)

millerpuckette:
Looking at this again, I now believe declarations inside an abstraction should _only_ be considered
relative to the abstraction... I've committed a change to that effect.  I hope this doesn't break anything...

Ant1r:
Thanks, I agree with you ; although it will surely break someone's patches, that way it won't break [declare -path]s when a top-level patch is turned to an abstraction.
So I close this PR.

PR #27, opened 2016-12-07
CFBundleVersion updating for macOS builds

orbsmiv:
Apply string substitution on the version number (as typed) to replace `-` and `test` with `.`, then update `CFBundleVersion` key in Info.plist with resulting value using the `plutil` command.

### DIFF
--- a/mac/build-mac64
+++ b/mac/build-mac64
@@ -26,6 +26,8 @@ mv "Wish Shell.app" Pd-$1-64bit.app
 cd Pd-$1-64bit.app/Contents
 rm -f Info.plist
 cp -p  ../../stuff/Info.plist .
+VERSION=${1/-/.}; VERSION=${VERSION/test/.}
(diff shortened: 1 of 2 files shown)


danomatika:
Could you apply this change to the `mac/osx-app.sh` script which was recently merged into the master branch?

orbsmiv:
Done. I've also added the `CFBundleShortVersionString` key with the same value (and updated the older build scripts to include this addition).

danomatika:
Thanks. I'm merging this into a topic branch for now and then it can be pulled in upstream.

orbsmiv:
@danomatika Can you see what's causing the Travis build to fail?

danomatika:
I think it's ok. Looks like it failed because I did the merge before it had pulled the PR: https://travis-ci.org/pure-data/pure-data/jobs/189360750

PR #28, opened 2017-01-05
configure.ac: add test for oss check header

wimmuskee:
Without that test, oss will be enabled if you have sys/soundcard.h on your system, regardless of setting --disabled-oss.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -273,7 +273,9 @@ AM_CONDITIONAL(MSGFMT, test x$HAVE_MSGFMT = xyes)
 AC_ARG_ENABLE([oss],
     [AS_HELP_STRING([--disable-oss], [do not use OSS driver])],
     [oss=$enableval], [oss=yes])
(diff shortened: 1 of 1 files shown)


danomatika:
Good catch.

umlaeute:
otoh, there is little harm in unconditionally enabling it, as OSS doesn't add any additional dependencies.

nevertheless, it's good to have `--disable-oss` do something consistent.

wimmuskee:
currently, you can't have portmidi and oss midi enabled at the same time, having a use case for disabling oss

umlaeute:
i see.

it might make sense to be able to enable/disable sound and midi components of a given backend separately.

danomatika:
When I was revamping configure.ac, my thinking was:

* appropriate default audio/midi backends are chosen based on the platform
* users can manually enable/disable audio/midi backends

With this thinking, if you're building on linux and want OSS, you could disable portaudio (the default) and enabled OSS.

millerpuckette:
merged

PR #30, opened 2017-01-05
various fixes for null-pointer dereferences, ...

umlaeute:
this PR contains a number of fixes for problems discovered by coverity-scan (a static code analysis tool)

https://scan.coverity.com/projects/pure-data-pure-data

the issues fixed are mostly:
- ressource leaks (e.g. not closing socket file descriptors before they leave scope)
- forgotten NULL-checks before dereferencing (or more often: the NULL check is done after dereferencing)
- unused code
- ...

### DIFF
--- a/src/g_readwrite.c
+++ b/src/g_readwrite.c
@@ -402,7 +402,7 @@ void canvas_writescalar(t_symbol *templatesym, t_word *w, t_binbuf *b,
     t_dataslot *ds;
     t_template *template = template_findbyname(templatesym);
     t_atom *a = (t_atom *)t_getbytes(0);
-    int i, n = template->t_n, natom = 0;
+    int i, n = template?(template->t_n):0, natom = 0;
     if (!amarrayelement)
     {
         t_atom templatename;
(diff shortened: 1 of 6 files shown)


umlaeute:
btw, this PR also tries to sneak in a change of how known-but-unknown cmdline arguments are handled.

Currently all unknown cmdline flags lead to a `usage()` message and an early exit.
This is nice to prevent wrong cmdline flags, but it is annoying when it comes to *optional* flags.
E.g. if Pd is compiled with JACK support, I can specify a `-jack` flag. But if it has been compiled without JACK support, then using the `-jack` will prevent Pd from starting up - even though the flag was valid *last week*.

As this is esp. annoying in scripts that are supposed to run on multiple platforms, the behaviour introduced with e0d40517254cbd0e5fa9ed7553d2c543a919d7ee will just ignore optional-but-disabled flags (and print a warning to the stderr).

This allows one to prioritize audio-backends. E.g. the following will use `jack` if available, but fallback to `PortAudio`; if none are available, it will use the system default.

    pd -portaudio -jack

PR #31, opened 2017-01-17
bendin fix

agraef:
This fixes up bendin so that it delivers values starting at -8192 to be consistent with bendout. Fixes ticket #1262: https://sourceforge.net/p/pure-data/bugs/1262/

Caveat: This bug has been there forever, so bug compatibility with older Pd versions may be an issue here. FWIW, I really think that it should be fixed anyway, it's just wrong the way it is, and having the bipolar bend range -8192..8191 throughout just makes patches employing these values a lot simpler.

### DIFF
--- a/src/x_midi.c
+++ b/src/x_midi.c
@@ -356,7 +356,7 @@ void inmidi_pitchbend(int portno, int channel, int value)
     if (pd_this->pd_bendin_sym->s_thing)
     {
         t_atom at[2];
-        SETFLOAT(at, value);
+        SETFLOAT(at, value-8192);
         SETFLOAT(at+1, (channel + (portno << 4) + 1));
         pd_list(pd_this->pd_bendin_sym->s_thing, &s_list, 2, at);
     }


porres:
I'm also into accepting this fix, but miller has considered a different option where a new object with the correct behaviour in introduced.

If this fix is not accepted for pd 0.48, I hope a new "pitchbend" object comes along instead then

agraef:
@millerpuckette Have you thought about this yet? Maybe it is a painful decision, but it's easy to get this fixed once and for all, there are basically two options:

- Accept this pull request and live with the fallout from not being bug-compatible with older Pd versions.

- Add a new object, or pair of objects, which implements the proper value range.

In any case, some kind of feedback would be appreciated. :)

millerpuckette:
There's already xbendin in Cyclone (an/d or it's easy to make an abstraction).
No way can I make an incompatible change at this point.

cheers
Miller

porres:
cyclone's xbendin / xbendout has a range between 0 - 16383

though they are consistent between each other, I was really hopin for a range starting at -8192

porres:
but yeah, I guess pd-l2ork / purr data can offer an extra [pitchbend] abstraction that fixes this - I could put it in a new library I'm working on!

agraef:
> No way can I make an incompatible change at this point.

Thought so. Well at least we tried. ;-)

PR #39, opened 2017-02-03
fix sched_setscheduler(0,SCHED_FIFO,...) error/success logging

Ant1r:
send sched_setscheduler(0,SCHED_FIFO,...) error/success logging to stderr.
post() is not available at this point, as GUI is not yet connected.


millerpuckette:
This should be fixed now in a different way (real-time promotion now happens after GUI is opened; and anyway, posts before opening GUI are automatically rereouted to stderr).  OTOH, it would be even better if pre-GUI posts were buffered and sent up later - this is for another day.

Ant1r:
Ok, I'll check this later ; let's close this for now.

PR #41, opened 2017-02-14
fixing midiclkin / clip-help.pd typo / re-allocate expr/expr~/fexpr~ / moving the gl_zoom member to the end of the _glist declaration

porres:
Hi, this a few commits al linto a single pull request.
---------------------------

the first fixes the [midiclkin] object, which was completely dead...

this has been fixed, tested and merged in Purr Data by Albert Graef, I'm just doing the Pull Request for vanilla... I did compile the vanilla app in Mac OS and could test that the object is indeed working and everything else seems fine ;)

Note that the object's code hasn't been messed with, but I have some thoughts.

I see that it's 24 clock ticks pert quarter note; hence, 1 tick is then 64th tuplets. So... the left output sends a 1 at every 3rd tick, which seems to be a way to trigger on 32th notes and filter out the 64th tuplets ticks, cool...

The problem is: the "1" comes at every 3rd tick... so it's not in sync at the head of the tempo, it is off by two 64th tuplets...

It would make much more sense to me if the first tick is "1" and then the 2nd and 3rd tick are filtered out, which would also be like a simple edit. But if this is the desired behaviour, maybe you could explain it why in the documentation.

see more of this dicussion here: https://git.purrdata.net/jwilkes/purr-data/merge_requests/77/diffs

there's also a ticket I opened about it in here https://sourceforge.net/p/pure-data/bugs/1260/

----------

another commit is just fixing a typo in clip-help.pd 

----------

moreover, I have a commit where I edited the help-intro.pd to re-allocate expr/expr~/fexpr~ in the objects list, since it is now not longer an "extra" object, but part of the vanilla binary internals

----------

lastly, I have a commit that moves the gl_zoom member to the end of the _glist declaration

This is a fix that solves backwards compatibility issues with externals compiled for earlier versions.

here's a ticket discussion: https://sourceforge.net/p/pure-data/bugs/1255/

-------------

cheers

### DIFF
--- a/src/g_canvas.h
+++ b/src/g_canvas.h
@@ -175,7 +175,6 @@ struct _glist
     t_editor *gl_editor;        /* editor structure when visible */
     t_symbol *gl_name;          /* symbol bound here */
     int gl_font;                /* nominal font size in points, e.g., 10 */
-    int gl_zoom;                /* zoom factor (integer zoom-in only) */
     struct _glist *gl_next;         /* link in list of toplevels */
     t_canvasenvironment *gl_env;    /* root canvases and abstractions only */
     unsigned int gl_havewindow:1;   /* true if we own a window */
@@ -190,6 +189,7 @@ struct _glist
     unsigned int gl_hidetext:1;     /* hide object-name + args when doing graph on parent */
     unsigned int gl_private:1;      /* private flag used in x_scalar.c */
     unsigned int gl_isclone:1;      /* esists as part of a clone object */
+    int gl_zoom;                /* zoom factor (integer zoom-in only) */
 };
 
 #define gl_gobj gl_obj.te_g


millerpuckette:
This will take some time for me to sort out.  I can't figure out why Pd has a "clkin" object at all.  It wouldn''t
be an incompatible change to simply delete the object from Pd, which I think is what I should do, considering that MIDI clock messages are of doubtful use.

porres:
Hi Miller. The good thing about this external is that it gives you the time in ms between clocks. I finf it really usefull. I'm just a little weirded out about the behaviour of the left output!

---------

Now, there are other commits here, a couple of them are really simple and silly, related to the documentation.

But there's this last commit, for another fix, which is still running the building test. I didn't know how to make a separate pull request just for that. When I do the commit, it just right in here. Anyway, what can you make of it? It's an old issue I've been pointing out, I'm giving it a try now to see if it fixes it.

cheers

porres:
> MIDI clock messages are of doubtful use.

how come? It's one thing still pretty active in the MIDI world. Anyway, [midirealtimein] can also react to MIDI Clock messages, but [midiclkin] is more powerful, and you can plug it directly into a [metro]

by the way, here's an easy way to test the object, if you make a "loop" with an internal midi device - I do it on the mac via IAC Driver Bus. Here's a screenshot with this object as it is working in Purr Data. The "248" is the midi clock message.

screen shot 2017-06-22 at 16 18 01

porres:
ok, the check is over, and what I did seem to have screwed up the building process, but I was just going for the fix suggested by Matt Barber. Maybe you can figure this out yourself, or I'll try something else with Matt.

cheers

millerpuckette:
I'm not sure what failed here, but anyway, if I understand what midiclkin is doing it's badly designed, as it uses a floating-point number to output time since Pd was started up - this will gradually lose accuracy the longer Pd has been run.  I thik the correct thing to do here would be for me to delete the object altogether so taht it can easily be supplied as an abstraction using midirealtimein.

porres:
Well, I agree that deleting an object that never worked in Pd doesn't really break anything, and somebody may want to create abstractions or externals for whatever it is missing, and we can close this.

Just remember to delete [midiclkin] from the object list in help-intro.pd and midi-help.pd

----------

now, there are other two silly things in here that you could take, the typo in clip's help and how I reorganized the expr family of objects out of the "extra" section in help-intro.pd

-------------

> I'm not sure what failed here

neither did I, but this is for yet another thing, related to https://sourceforge.net/p/pure-data/bugs/1255/

millerpuckette:
leaving out the clkin 'fix' (which I think leaves it broken only in a more subtle way) but the other changes should appear next time I push local repo.

porres:
> leaving out the clkin 'fix' (which I think leaves it broken only in a more subtle way) 

Please confirm your intent to remove it for good, and then we should take it out of the documentation as well. I'm totally fine with letting it go.

Now, the thing is that we had fixed it in Purr Data, and now we're talking about deleting it because you're considering this. So it'd be good to be on the same page... it needs to be clear if it's out for real, so we can go ahead and delete it there as well.

cheers

porres:
as discussed on the dev list, I'm closing this since most of the issues have been taken care of, then I'll either create an issue or another pull request to remove [midiclkin]

PR #42, opened 2017-02-15
[WIP] DejaVu Sans Mono as default font crossplatform

danomatika:
This is a prelim branch with some work regarding the pd-list threads on using DejaVu Sans Mono as the default font across all platforms. My initial recommendation is for a "normal" font weight.

This font is commonly installed on Linux systems but not on macOS or Windows. The default font, as set in s_main.c, for mac is Monaco normal and DVSM bold for Windows & Linux, although I believe it falls back to Courier on Windows from what I can tell in pd-gui.tcl.

See some testing info & further screenshots on this [wiki page](https://github.com/pure-data/pure-data/wiki/Crossplatform-font-metrics-%26-comparisons)

macOS
-------

As the macOS build results in a mac .app bundle, the solution is to simply include the font file as an included bundle Resource. This is already done and working in the first commit.

Linux
------

On Linux, there is no need to distribute the font with Pd as it usually installed and/or can be an install requirement by the distribution's package manager.

Windows
---------

On Windows, the situation is a little trickier. The only option for bundling, it seems, is to use the Windows API to [load the font file into a Private Font Collection and then tell Tk about it](http://code.activestate.com/lists/tcl-core/8159/), otherwise the font has to be installed system-wide (not an ideal option). This could be done using Tcl extensions, like [TWAPI](http://twapi.sourceforge.net), but Pd has historically avoided using anything beyond the base Tcl/Tk.

My initial thought is then to create a tiny, custom Tcl extension that simply calls the Tcl & Win API C code needed to load the font. The pd-gui can then require this custom extension when it starts on Windows. There is no need to build or use this custom extension on any other platform.

Relevant links:

* [Hello World as a C extension](https://wiki.tcl.tk/11153)
* [Extending Tcl](https://wiki.tcl.tk/6276)


umlaeute:
how about moving the task of installing fonts into the *installer*?
that is:
- when using the ZIP-package, no font gets installed (the font could still be shipped within the ZIP, and the README might tell the user to right-click/install it manually)
- when using the NSIS-installer, the user is presented an option to install the font system wide (enabled by default)

i'm pretty sure that NSIS has decent, built-in support for installing fonts properly.

danomatika:
>how about moving the task of installing fonts into the installer?

This is the easiest option, but I think less than optimal, as things like this should *just work* and not require touching the system on any platform. I'm willing to take a stab at a font loading tcl extension on Windows. If it's too hard/too much of a hassle, then yeah, let's fall back on the installer.

danomatika:
For comparison on macOS using the included TK 8.4 Wish:

**Monaco normal (default)**

monaco normal tk 84

** DejaVu Sans Mono normal**

dejavu normal tk84

Lucarda:
>  I'm willing to take a stab at a font loading tcl extension on Windows. If it's too hard/too much of a hassle, then yeah, let's fall back on the installer.

Sure we don`t prefer to fall back to TWAPI?

Lucarda:
>  that simply calls the Tcl & Win API C code needed to load the font. 

We need 3 files from “twapi-bin-4.1.27.zip”:

twapi_base.dll
pkgIndex.tcl
LICENSE

That will do.

porres:
To my surprise, DVSM on the Mac is not too different from Monaco...

I say that because installing DVSM and calling it on startup did compromise my patches. Let me try this now...

porres:
well, did compile it and... ok, actually... it does look corrupted after all

screen shot 2017-02-16 at 14 02 41

porres:
but yeah, comaring that patch doesn't make things look too differently 
screen shot 2017-02-16 at 14 04 50

Lucarda:
>  ok, actually... it does look corrupted after all

Alex did you changed the .tcl?

If not you should.

Especially for your test.

Here is the how-to:

[https://github.com/pure-data/pure-data/wiki/Crossplatform-font-metrics-&-comparisons](url)

porres:
no, I didn't, but shouldn't it be a part of the branch/merge request then?

Lucarda:
> no, I didn't, but shouldn't it be a part of the branch/merge request then?

I don`t know. (surely not)
Folow the how-to & re-test
Remember to restart Pd for the changes in the .tcl to apply.

Lucarda:
And here: [http://lucarda.com.ar/x/macosx.jpg](url)

A screen-shot of your coll-help on OSX.

I did that yesterday.

porres:
i cant see the screenshot, error 404

Lucarda:
I edited the above.

[clik here](http://lucarda.com.ar/x/macosx.jpg) anyway.

porres:
looks great, if this is what we get with this merge, and this is how it looks for all platforms, i couldn't be happier :)

Lucarda:
Try it on your Mac.

I updated the link to the HOW-TO and upload your tests:
[here](https://github.com/pure-data/pure-data/wiki/Crossplatform-font-metrics-&-comparisons)

Lucarda:
More info on the TWAPI implementation for loading fonts on w32 here:

[https://lists.puredata.info/pipermail/pd-list/2017-02/117838.html](https://lists.puredata.info/pipermail/pd-list/2017-02/117838.html)

porres:
hi, I wonder...

is this just about using DejaVu on all platforms, or also about fixing the metrics to the Pd Extended metrics?

Cause for now I tried it and although it makes it consistent across platforms, things change when using it on the Mac, and the text gets wider and things may overlap

Lucarda:
> is this just about using DejaVu on all platforms, or also about fixing the metrics to the Pd Extended metrics?

For your tests fix the metrics to be as Pd-extended.

porres:
> For your tests fix the metrics to be as Pd-extended.

sure, I know, I'm following this... but this doesn't really answer my question yet...

will this branch also focus on geting the same metrics as Pd-Extended or not? Cause, for now, you're asking me to do it myself.

cheers

danomatika:
>will this branch also focus on geting the same metrics as Pd-Extended or not?

Yes it does, it simply doesn't have that change applied yet. I'm not 100% convinced the smaller boxes make the most sense overall, maybe there are some tweaks needed as a couple patches I have seem too slim where the top of the text overlaps the object border. I'll dig that up later.

I'd like to get some feedback from @millerpuckette & @umlaeute in regard to both font & metrics adjustment.

umlaeute:
i have nothing much to say about this. personally i favor **bold**, but could live with *normal* weight :-)
also, I usually don't write patches that rely on pixel-adjustment...

danomatika:
Note: this still doesn't work for Windows yet. It's totally doable, I just need a Windows setup again :P

umlaeute:
jeez, @danomatika is pre-empting my review! get out of my head. now!

danomatika:
>get out of my head. now!

I'm in your time zone now. :)

danomatika:
@umlaeute as mentioned in the main issue info, IMO the best approach for Windows is load the font into a Private Font Collection so Tk can see it. It looks pretty simple based on this: http://code.activestate.com/lists/tcl-core/8159/

My initial thought it to build it as a custom Tk extension just for the Pd gui on Windows. I'm not sure it could be built into the C core since it runs as a separate process. In any case, we should do appropriate checks to make sure the font file exists before trying to load it, etc.

porres:
> i have nothing much to say about this. personally i favor bold, but could live with normal weight :-)

** I personally favor normal weight, but I guess anyone could learn to live with one way or another for general patching - so this could be a matter of just getting used to it. On the other hand, when it comes to documentation/help files and tutorials with lots of text, this really jumps to the eye and creates a very polluted environment.**

Lucarda:
For w32 lets remember that there is something that is working and is very easy: TWAPI:

[(https://lists.puredata.info/pipermail/pd-list/2017-02/117838.html)](https://lists.puredata.info/pipermail/pd-list/2017-02/117838.html)

danomatika:
Latest commits update the fallback font in the GUI tcl to Monaco on macOS. I've tested with both 8.4 and 8.5. I also added a readme explaining the current font defaults & info on how to try other fonts.

rnkn:
Sorry to interject.. Was Monaco chosen over Menlo due to metrics? Afaik, Menlo replaced Monaco as the monospace font de jour on Snow Leopard.

danomatika:
I believe Monaco was chosen by extended as it was the closest in size and appearance to DejaVu Sans Mono without having to install anything. I tried Menlo and it doesn't look or size as "good" IMO (or at least to my expectation).

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

millerpuckette:
This is making the windows version crash (on wine at least; that's how I test it).  Reverting
pd-gui.tcl fixes the crash.  Perhaps the new code should be only be run if we're not windows.

danomatika:
Strange, the changed to pd-gui.tcl are very minimal.

danomatika:
@millerpuckette Also, this PR does not have the font metrics changes, only the font. I have a different branch for that locally. In my testing, there are more tweaks from Pd extended required and I haven't tracked them done yet.

millerpuckette:
I agree - I can't see what difference any of this could make. Maybe this is just wine being unstable somehow.

millerpuckette:
yeah, after an hour of changing things back and forth it started to work.  What a bore...

danomatika:
I can push the metrics branch later on but, IMO, it's not ready for prime time. This font one is good though.

PR #43, opened 2017-02-15
osx-app.sh default Wish should be included Tk 8.4

danomatika:
A small commit to use the included TK 8.4 Wish by default instead of the system's buggy Tk 8.5 Wish. This is something fixed in the osx-retina-support/TK 8.6 branch but did not migrate back yet.


danomatika:
Fixed with the recent merge of the osc-retina-support branch.

PR #44, opened 2017-02-16
calculate window position offsets (rather than hardcoding them)

umlaeute:
on X11, this ensures that windows are always opened at the same position.
so we can do `[vis 0, vis 1(` repeatedly without moving the window around...

the fix uses the pdwindow (once it is visible) to calculate the difference between the decorated window `wm geometry` and the `winfo xroot` ...


umlaeute:
note: like the original code, this currently does not produce the correct position if the window is so narrow, that the menu wraps (and covers more than one line).

PR #46, opened 2017-02-28
Fix and update fr.po

Ant1r:
This fixes some annoying translation bugs, and harmless typos and misspelling (sorry Cyrille ;-).

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -6,8 +6,8 @@ msgstr ""
 "Project-Id-Version: Pure Data 0.43\n"
 "Report-Msgid-Bugs-To: pd-dev@iem.at\n"
 "POT-Creation-Date: 2009-08-18 20:56-0400\n"
-"PO-Revision-Date: \n"
-"Last-Translator: Cyrille Henry\n"
+"PO-Revision-Date: 2017-03-01 00:00\n"
+"Last-Translator: Cyrille Henry / Antoine Rousseau\n"
 "Language: fr\n"
 "Language-Team: \n"
 "MIME-Version: 1.0\n"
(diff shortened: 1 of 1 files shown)


porres:
how do these translations take effect when using Pd? I can't get mine to work in brazilian portuguese... or any other so it seems...

Ant1r:
I really don't know... For me it's in french out the box on various Linux boxes I have.

avilleret:
oups that clashes with mine on #58 
after a quick review of yours, I think both have good things (mine add lots of new translation while yours is more accurate I think)
so we should merge the two I think, what's your opinion @Ant1r ?

avilleret:
@porres translation language selection relies on local settings, so If your system is in portuguese, Pd should be in portuguese if available of course.
For example, you can override local settings temporarily with : 

    LANG=pt_PT.UTF-8 pd 
 
which will start Pd in Portuguese while my system remains in French (tested on Ubuntu 17.04)

Ant1r:
Yep @avilleret  I did notice the clash.
I think (but I'm not sure...) that if you merge my mods into yours, then my PR could be merged first then yours without conflict.
OR: you could propose your fixes to my branch (https://github.com/Ant1r/pure-data/tree/update_fr.po) then I would accept it and my PR to Pd would be updated with your mods.
Maybe there could be a better (simpler) way...

umlaeute:
@ant1r if @avilleret merges your changes into his branch, then your changes are a subset of #58 and merging that PR will automatically merge in your changes.

that is: the two PRs can be condensed into a single one that only needs to be accepted once (and the other PR should be closed)

avilleret:
@umlaeute this have been done (merging @Ant1r changes into my branch) this morning. So this PR could be closed and #58 should be update-to-date.

Ant1r:
I agree that this PR should be closed now.

PR #47, opened 2017-03-02
Gui Preferences

umlaeute:
this PR extends the `tcl/pd_guiprefs` to be more generally useful (the current implementation was tied to the very specific recentfiles functionality).

if you prefer a squashed PR, let me know.

### new features

- simple API that requires a minimum amount of data:

      ::pd_guiprefs::write foo "bla" # write a key/value pair to the default domain
      ::pd_guiprefs::read foo        # read a value associated with the key from the default domain

- consistent use of `$::pd_guiprefs::domain` (it's always a domain-like string, that can be used regardless of the actual storage backend); the default is `org.puredata.pd.pd-gui`
- honours `PD_CONFIG_DIR` environment variable to specify the filesystem location of Pd's config directory
- honours [`XDG_CONFIG_HOME`](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html) (on linux) to specify the filesystem location of the configuration
- easy to switch to filesystem-based storage on OSX and W32 (though the default is still the native storage), simply by setting the `PD_CONFIG_DIR` variable (if set to an empty-value, sane defaults are assumed)

- as a side-effect, Pd now also exposes a `::platform` variable, that can be used to distinguish between OSX/linux/W32/... (rather than abusing the `windowingsystem` for that kind of information)

- some trailing whitespace has also been removed from the edited files

### (in)compatibility with old system
the rewrite retains API-compatibility with the old system.
however, i expect that after applying this fix, any data stored with an old version of Pd might not be availble in the new version.
Afaict, this only affects the recentfiles (which i consider non-critical data).

we are now using a slightly different domain `org.puredata.pd.pd-gui`, which changes the actual storage path of the preferences (compared to older versions of Pd).
the main effect of this is, that the old settings stored via the guiprefs will not be available anymore.

### tested systems
the new implementation has been tested on linux and W32.
i haven't tested on OSX (yet).

also the default-domain on OSX is still `org.puredata`, because i don't know whether the *RecentFiles* functionality is somehow bound to that key.
eventually, that should be changed to  `org.puredata.pd.pd-gui` as well.

### DIFF
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -58,8 +58,8 @@ package require pd_guiprefs
 namespace import ::pd_guiprefs::init
 namespace import ::pd_guiprefs::update_recentfiles
 namespace import ::pd_guiprefs::write_recentfiles
-# make global since they are used throughout    
-namespace import ::pd_menucommands::* 
+# make global since they are used throughout
+namespace import ::pd_menucommands::*
 
 # import into the global namespace for backwards compatibility
 namespace import ::pd_connect::pdsend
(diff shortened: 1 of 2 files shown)


umlaeute:
btw, this is required for pure-data/deken#140 ("remember last download-path between restarts) to work correctly

umlaeute:
i'll just squash and resubmit...

PR #52, opened 2017-04-11
Use info.puredata.pd for macOS CFBundleIdentifier

rnkn:
This PR proposes using `info.puredata.pd` for the macOS CFBundleIdentifier string, as per https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-102070

> The string should also be in reverse-DNS format. For example, if your company’s domain is `Ajax.com` and you create an app named Hello, you could assign the string `com.Ajax.Hello` as your app’s bundle identifier.

i.e. the Pd domain is <http://puredata.info> and app name is `pd` hence `info.puredata.pd`

Please review the changes as this is a pretty dirty find/replace of the source.

### DIFF
--- a/mac/stuff/Info.plist
+++ b/mac/stuff/Info.plist
@@ -22,7 +22,7 @@
 	<key>CFBundleExecutable</key>
 	<string>Pd</string>
 	<key>CFBundleIdentifier</key>
-	<string>org.puredata</string>
+	<string>info.puredata.pd</string>
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
 	<key>CFBundlePackageType</key>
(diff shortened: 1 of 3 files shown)


rnkn:
Upon trying to merge this locally it seems maybe better to branch this from `guiprefs`?

umlaeute:
does this solve any specific problem or is it "just" cleanup, trying to sync the code with common good practice?

since the [guiprefs branch](#48) has not received any feedback yet, its hard to tell whether it would be better use it as the base branch for this PR...

rnkn:
Yeah it's only best practices in my view. Using an app-specific identifier (e.g. org.puredata.pd or info.puredata.pd) rather than a domain-level identifier accommodates multiple apps under the domain, if there ever comes a need for that.

Matching the identifier domain with the actual domain name is just helpful to the user, if they're ever looking for the plist file. But otherwise, yes just cleanup.

Looking at #48 it seems much more prudent of me to make a PR from that branch, otherwise it seems like a recipe for org/info confusion.

PR #53, opened 2017-04-16
Newly designed Retina icons

rnkn:
I've created some Retina icons for both `pd.icns` and `pd-file.icns`. Please take a look. Feedback is most welcome.

I've used Menlo for the "Pd~" in the icon because it's the default font for macOS Terminal.app and quite elegant ~~(Deja Vu Sans by contrast is not a very attractive font)~~.

**Edit**: DejaVu Sans is a nice font. For some reason I was thinking of Roboto. Silly.


rnkn:
I just realised that the .icns files won't preview on the web, so here are the png images:

pd

pd-file

umlaeute:
thanks for taking on on this.
however, some notes:
- the canonical way to spell Pd is always `Pd`, never `PD`
- the font looks not like courier/vera, which is the default font (monospaced, with serifs). Instead it rather looks like arial (or so, i'm not very good at fonts); could that be changed?

rnkn:
You mean the uppercase  "PD" on the file icon? That's not the program name, that's macOS's canonical way of listing the filetype, e.g. here's a .png file:

png

The choice of Courier on the original icon was, I would assume, because the default font for the Pd GUI was Courier. But it appears that `pd_gui.tcl` tests for a collections of other fonts and sets a default: https://github.com/pure-data/pure-data/blob/master/tcl/pd-gui.tcl#L447-L448

This is cool, because I think Courier is a terrible font for code (I associate it more with screenplays actually). But, if the original icon principle of design was to match the "Pd~" icon font with the GUI font then ought not a new icon follow suite and use the foremost test font for each system?

But this raises another thought for me... which is that I think the choice and order of the test fonts could be improved, both from an aesthetic perspective and fitting with a system default. I might do another quick PR with an idea for such an order.

rnkn:
If #42 is merged I will update these with DejaVu Sans Mono. (Menlo is a close relative, so they will look pretty similar, i.e. sans-serif.)

rnkn:
Will rebase & reopen with improved design.

rnkn:
Here's a modern flat-icon style idea... (uses DejaVu Sans Mono)

screen shot 2017-04-18 at 10 31 00 pm

danomatika:
Do you have a version of the first icon using DejaVu Sans Mono or, better yet, an SVG? Can you send it to me or open a PR to <https://github.com/pure-data/pd-icon>? I'd like to see how your initial Pd-extended style icon looks on the other platforms. It would be nice to update the icon for 0.48.0.

rnkn:
Sorry I used the same Pixelmator document and just kept altering it until I got to the latter version.

danomatika:
If you have a vector program, maybe you can modify https://github.com/pure-data/pd-icon/blob/master/masters/icon_svgplugin.svg? I think it really just needs the simple gradients your first vision had. It's still a bit too flat IMO.

PR #54, opened 2017-04-16
New test font stack for default font

rnkn:
This order tests fonts to choose the most recent default monospace font for each system, e.g. Menlo is preferred over the older Monaco for macOS systems, but neither of these should be available on Windows or Linux, etc.

Of course, another way to do this would be to have per-system test font stacks...


rnkn:
On second thoughts I think it's cleaner to bundle DejaVu Sans Mono (#42) and treat that as the canonical font. If you agree feel free to close this PR.

PR #56, opened 2017-04-25
Native style designed icons for macOS

rnkn:
These icons use DejaVu Sans Mono font for the "Pd~" lettering. They are heavily stylized to fit with the macOS environment.

pd

pd-file

context

## To use these icons without this PR merged

Download the binary files `pd.icns` and `pd-file.icns` from https://github.com/pure-data/pure-data/pull/56/files

~~~
$ mv pd.icns /Applications/Pd.app/Contents/Resources/pd.icns
$ mv pd-file.icns /Applications/Pd.app/Contents/Resources/pd-file.icns
$ touch /Applications/Pd.app
~~~

**Edit**: rebased with fresh icons with slightly reduced gradient.


danomatika:
Fixed. I forgot some brackets. Little detail are easy to forget sometimes and perhaps I'm doing too many things right now...

porres:
should we close this now that we have a new PR? https://github.com/pure-data/pure-data/pull/195

rnkn:
@porres this PR is intentionally left open as it has installation instructions for people who prefer this design.

umlaeute:
@rnkn i think that a PR is the wrong place for documentation.

if the PR has been accepted, rejected or *solved otherwise* (which seems to be the case here) it should be `closed`. else we end up with thousands of open PR solving solved problems again and again just because some people prefer one solution over the other.

(note: i personally don't care much about the icon set (although i think i dislike both suggestions, but not too much); i'm merely talking about the use of the issue/PR tracker)

rnkn:
I think #195 is not aiming to solve the same problem. That PR has created an icon that looks the same cross-platform, whereas the designs I provided are intended to give Pd.app on macOS a more native icon feel, not touching other systems. I've renamed the PR to reflect this.

umlaeute:
however, it seems that #195 and this PR are mutually exclusive (could you confirm that?)

if #195 is going to stay, this PR ought to be rejected as `wont-fix`

rnkn:
I think I probably sparked the idea to get a higher resolution icon designed with this (and a previous) PR, so I guess #195 could be inspired by that desire. But my intention was that a macOS style icon would *not* be used cross-platform, and to put it another way, this PR addresses my belief that Pd would be better served if the macOS Pd.app looked at home on a macOS system.

umlaeute:
i understand that; but this doesn't make it any more likely that this PR is going to be accepted (as there are obviously two camps: those that want to have Pd look the *same* on all platforms; and those that want to have Pd look *native*. It seems that the *samers* are currently prevailing, so there's little chance for the *natives*)

probably @danomatika could add his thoughts.

danomatika:
I think the icon should be consistent on all platforms. This is in line with the work to provide a consistent font and object sizing between systems.

umlaeute:
Since there is nobody going to push this to be included, I'm going to close this PR.

@rnkn please add the information on how to make Pd look more native on OSX to the [wiki](/pure-data/pure-data/wiki), which I think is the correct place for any such information.

rnkn:
> so there's little chance for the natives

Can I just draw attention to how funny this line is?

danomatika:
Note: I've asked for feedback and preference on the list multiple times. We could use this icon, for instance, as the basis for a the cross platform icon, however I didn't hear much about it.

Also, IMO I'm not sure how this icon looks "more native" than the other, but I suppose that's a matter of taste.

Issue #60, opened 2017-05-11
keyboard shortcuts and capital letters


LGoodacre:
It would be really super wonderful if Ctrl + s, e, c, v, and d worked with capital letters also. Reason: you can't use these handy shortcuts if caps lock is on. Seems like this would be an easy fix.

umlaeute:
hmm,... <kbd>Ctrl</kbd>+<kbd>s</kbd> (aka `Save`) is already something different as <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>s</kbd> (aka `Save as...`)...and those two keystroke assignments are very common.

And the difference between <kbd>Shift</kbd>+<kbd>s</kbd> and uppercase `S` (<kbd>CapsLock</kbd>-<kbd>s</kbd>) is somewhat...subtle.

OTOH, <kbd>CapsLock</kbd>-<kbd>Ctrl</kbd>+<kbd>s</kbd> seems to be something different from <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>s</kbd>, whereas <kbd>CapsLock</kbd>-<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>s</kbd> seems to be the same as <kbd>Ctrl</kbd>+<kbd>s</kbd>. urgh.

umlaeute:
if you have access to non-linux platforms, it would be great if you could test the PR #61.
(just replace the [pd_bindings.tcl](https://github.com/pure-data/pure-data/blob/capslock-keybindings/tcl/pd_bindings.tcl) file; no need to recompile...)

LGoodacre:
Sorry for the delayed response. I tested the new pd_bindings.tcl on Windows 10 and it seems to work perfectly. I tried Ctrl S, N, C, V, D, and they all seem to work as expected. In addition, Ctrl + Shift + s/S calls save-as regardless of case, as expected. Nice job. I'd love to see this included in the next release.

LGoodacre:
Sorry for the delayed response. I tested the new pd_bindings.tcl on Windows 10 and it seems to work perfectly. I tried Ctrl S, N, C, V, D, and they all seem to work as expected. In addition, Ctrl + Shift + s/S calls save-as regardless of case, as expected. Nice job. I'd love to see this included in the next release.

LGoodacre:
[...]

PR #61, opened 2017-05-11
keybinding for CapsLock cases (same as without CapsLock)

umlaeute:
seems like Tcl/Tk distinguishes between:
- lowercase <kbd>s</kbd>
- <kbd>Shift</kbd>+<kbd>S</kbd> (uppercase)
- uppercase <kbd>S</kbd>: happens if CapsLock is enabled
- <kbd>Shift</kbd>+<kbd>s</kbd> (lowercase): happens if CapsLock is enabled (so the Shift actually lowercases the uppercase `S`)

Closes: https://github.com/pure-data/pure-data/issues/60

This has been tested on linux with tcl/tk 8.6
it might be a good idea to test this on the other platforms first...

### DIFF
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -53,6 +53,25 @@ proc ::pd_bindings::global_bindings {} {
     bind all <$::modifier-Key-w>      {menu_send_float %W menuclose 0}
     bind all <$::modifier-Key-x>      {menu_send %W cut}
     bind all <$::modifier-Key-z>      {menu_undo}
+    # uppercase bindings, for the CapsLock case
+    bind all <$::modifier-Key-A>      {menu_send %W selectall}
+    bind all <$::modifier-Key-B>      {menu_helpbrowser}
+    bind all <$::modifier-Key-C>      {menu_send %W copy}
+    bind all <$::modifier-Key-D>      {menu_send %W duplicate}
+    bind all <$::modifier-Key-E>      {menu_toggle_editmode}
+    bind all <$::modifier-Key-F>      {menu_find_dialog}
+    bind all <$::modifier-Key-G>      {menu_send %W findagain}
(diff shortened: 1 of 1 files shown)


millerpuckette:
I'm not sure whether caps-lock should change the capitalization on command-keys or not.

umlaeute:
i think the underlying issue is, that if you accidentally have CapsLock enabled (e.g. on a keyboard that has no LEDs indicating the Caps-Lock state), then it becomes really hard to use the keyboard shortcuts at all (Pd suddenly starts to ignore shortcuts or doing different things as expected).

millerpuckette:
applied by hand.... hope I got it right.

umlaeute:
looking forward to see this in a public repository ;-)

danomatika:
Looks like not all commits were merged. Maybe double-check this before deleting the branch?

millerpuckette:
I ended up having to edit everything by hand.  I could easily have missed something, but the two versions of the file are way different now so it's hard to tell.

danomatika:
If it was done manually and did not involving pulling this commit into master, then the changes can be there but GitHub simply doesn't see the commit as it's missing. I just wanted to make sure everything was there before deleting the branch.

Also, as we do not commit to the master branch here directly, perhaps a alternative workflow is to work from a clone from GitHub and also push changes to the upstream master? Pressing the Merge button on a Github PR is pretty nice most of the time :)

umlaeute:
most of the time github detects a merge, even if it was not done via the `merge` button (personally, I hardly ever use the `merge` button, because I want to test the merged code locally before publishing it).

danomatika:
>(personally, I hardly ever use the merge button, because I want to test the merged code locally before publishing it)

You can add the submitters repo locally as a remote, fetch the remote branch, then checkout and test. If things are good, you hit the Github Merge button which handles all the rest. This has worked well for me in the past.

PR #64, opened 2017-05-28
Fix memory leaks/memory management improvement 

jpcima:
This is a proposal to improve the memory management of puredata in an embedded setting where libpd is used. My goal overall is to make audio plugins which perfectly wrap puredata patches and also include the Tk GUI.

I want plugins to be able to load/unload without leaking memory, so I did the changes in puredata to support that, and want to discuss getting them upstream. Some changes fix memory leaks, and some introduce modifications in the implementation, so I would appreciate a feedback on the latter.

My fork of libpd is a complement to this PR. I introduced the [libpd_term](https://github.com/jpcima/libpd/blob/58488fb42244f96dbfb13cd917055291f7425ddf/libpd_wrapper/z_libpd.c#L124) API, libpd_init counterpart to free resources of puredata.
I keep this freeing logic outside of pd itself because if depends on **pdinstance_free** and features available only with **PDINSTANCE** defined.

My implementation changes are:

- ~~making the **class_list** accessible to libpd, making it EXTERN~~
  introducing public API class_getfirst()
- fixing how symbol tables are freed, Some **t_symbol** there are not malloc'd; I add an attribute **s_allocd** to indicate whether the structure is a result of dynamic allocation.
- fixing how **pdinstance_free** updates **c_methods**. (made no sense and looked like copy-paste from pdinstance_new)
- ~~splitting **pd_init()**, introducing **pd_init_systems()** and counterpart **pd_term_systems()** (could possibly use some better vocabulary)~~
 ~~**pd_term_systems()** is to free the resources of any systems (aka *_init() functions) which need, beside the registered classes. (creating classes what they most do, but I release everything in **class_list** at the end)~~
  introducing **class_free()** and allowing classes to have a destructor callback
- keeping the instances of **t_template** in a linked list of the pdinstance. As I understand templates are like globals which are bound to symbols. Like the canvases, I free them in **pdinstance_free** but after the canvases since these may have references.
- ~~putting the **cos_table** in static memory since it has a fixed size~~
- merging PR #40, fixing a memleak I have encountered


millerpuckette:
This will take me some time to go through.  I have only one question to raise right now, which is,
can the canvas_list be kept hidden and accessed through a function instead (I think all you need it for is to delete them all, correct? That could be done in a class_term() function or something.)

Incidentally, I want to offer cosine tables of various lengths.  I don't see why they can't be kept static, as long as it's OK to fix the list of offered lengths in advance somehow.

jpcima:
Thanks for considering this pull request.

> can the canvas_list be kept hidden and accessed through a function instead

I assume you are speaking of the class list, not the canvases, which I have made visible via the extern keyword. I fully agree with your remark. For now, I was trying to get to the point with as little changes in PD as possible.

I knew little to PD internals prior to doing this, so It has let me familiarize with the concepts and how all things are articulated with each other. I am myself not entirely satisfied with the way how I do the cleanup, because I put many internals details into libpd which may better belong in puredata itself.
I am happy to discuss and make any change to make this patch accepted.

A detail which I have found regrettable is, because pd classes aren't objects, they can't be pd_free'd/have free methods. Looking at my version of pd_term in libpd, I clean up classes by invoking freebytes() successively on each element of class_list.
If there was a cleanup protocol for classes, it would possible to avoid the *_term() circus which I have proposed, and each class would be able to release their own resources when class_free'd (fft~'s buffers, cos~ tables, and so on).

millerpuckette:
yep, class list, not canvas list, oops.  I think OTOH that one can write a class_free() routine independent
of its being a first-class object in Pd (that's only necessary if we need regular Pd messages to be sent to classes which so far I don't see as being useful).  Class_free() could always make a callback for getting
rid of cosine tables etc.

jpcima:
I have introduced class_free with commit 03750d5.
It is a version which is callback based like you describe. Do you approve of this API and implementation?
Please do not merge anything yet, until I will port other cleanups to this API.

jpcima:
It seems one of the CI is failing for reasons I think are unrelated to my programming..
Anyway. I converted fft and got rid of the previous system. There are 5 classes which used the fft so I implemented a reference counter.
However, and that is a thing I missed on the first time, I have not implemented the cleanup for the fftw variant. I will look into that tomorrow.

jpcima:
I have a final observation about the editor's _copy_binbuf_. That buffer's reference is stored in the _pd_this_, yet it's a variable that's initialized globally, not per instance. So presumably it's only the _pd_maininstance_ that's going to get this reference. Because it causes trouble in the cleanup, is there any objection to making _copy_binbuf_ a global instead?

Beside that, I rechecked the program in valgrind since the last modifications, and confirm no memory leaks or problem whatsoever.

millerpuckette:
I think copy_binbuf needs to remain per-instance - but I don't see the problem you're referring to - as I read it there's a new one created every time objects are "copy"ied.  OTOH, there needs to be stuff
added to binbuf_free() added to g_editor_freepdinstance() - I've put in an attempt in commit
c7816c2b5ad77148a6b2161c5513c8f7d62c8bcf

jpcima:
>there's a new one created every time objects are "copy"ied
With your explanation I understand now. It 's my mistake to have thought this binbuf was used like a global when in fact it is not.

~I would love to have a look at this commit unfortunately I don't find where it is.~
Commit link works now, I'll look.

To answer about the reason of this binbuf commit.
From libpd's view, my global lifecycle of pd is not exactly symmetric, I did it like this:
- init: (1)main instance > (2)classes > (3)secondary instances
- cleanup: (i)secondary instances < (ii)main instance < (iii)classes

I have considered it would be prudent to destroy classes(iii) after pdinstances(i, ii), so that would ensure no objects still being around.

In this model, if g_editor_cleanup attempts to free the copy_binbuf of g_editor_setup, a crash happens because all pdinstances were destroyed at that point. This was the problem I encountered and tried to solve.

copy_binbuf being initialized in g_editor_setup() has tricked me to believe it was really a pseudo global.
Since it's not, do secondary instances never get their binbuf initialized? isn't this problematic?

I look at your commit now.

jpcima:
Your commit c7816c2b5ad77148a6b2161c5513c8f7d62c8bcf has a typo in it.
canvas_findbuf was probably meant instead of copy_binbuf.

+    if (EDITOR->canvas_findbuf)
 +        binbuf_free(EDITOR->copy_binbuf);

I'll revert my binbuf commit, merge master and recheck with your fix.

jpcima:
It's just to say I rechecked in valgrind having pulled the latest master, and it's good.

millerpuckette:
I'm at teh end stage of trying to get Pd 0.48 out so will have to let this wait until after that, sorry.  I agree in principle with these changes but will have to look at them in detail.

millerpuckette:
I was able to apply the first 10 or so of these, but once class freeing set in it got very hard to merge so I put that off for later.

danomatika:
Can you merge this with master? That would make it easier to consider applying for 0.49, if @millerpuckette can get to it.

danomatika:
I would also add that this would indeed be good of libpd. :)

PR #66, opened 2017-06-25
Soundfiler fix

Ant1r:
soundfiler write operation needs some fixes:
- 'skip' doesn't work properly on 64bit
- 'normalize' currently computes the biggest sample on the entire pd array, while it should only take account of the range selected with skip/nframes, in order the sound file to be actually normalized.


millerpuckette:
commit 000c8bd..fabc919 should fix this - I ended up cleaning up a bunch of stuff.

Ant1r:
Great! Closing.

PR #67, opened 2017-06-25
Po update

Ant1r:
update template.pot
update fr.po (again...)


danomatika:
Looks good. I can confirm that, with the recent changes, the template needs to be updated.

millerpuckette:
losing -- accepted

Issue #69, opened 2017-06-27
Struct crash 


pierreguillot:
I don't find a way (or create a simple patch) to reproduce the error but Pd crashes  when I perform some actions on a abstractions that manage a graphical data structures system.
I don't know if it's relevant but for example my patch contains one or more [struct] objects and it automatically draws  at loading (mostly with [append] objects) inside a canvas (given as an argument). If I make a patch that I can share (with no dependencies) that crashes I'll add it to this issue. 
I tried with Pd 0.47.1-32 & 64bits on MacOS 10.12.5.
Here the crash report:

> Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
> 
> 0   pd                            	0x0000000103c0a82f pd_findbyclass + 63 (m_pd.c:191)
> 1   pd                            	0x0000000103bde102 template_findbyname + 66 (g_template.c:489)
> 2   pd                            	0x0000000103bdc9eb scalar_vis + 107
> 3   pd                            	0x0000000103bdc35f scalar_doredraw + 31
> 4   pd                            	0x0000000103c13b75 sys_flushqueue + 293 (s_inter.c:809)
> 5   pd                            	0x0000000103c12956 sys_poll_togui + 86 (s_inter.c:832)
> 6   pd                            	0x0000000103c128e7 sys_pollgui + 39 (s_inter.c:893)
> 7   pd                            	0x0000000103c0bfde m_pollingscheduler + 462 (m_sched.c:523)
> 8   pd                            	0x0000000103c0bd32 m_mainloop + 50 (m_sched.c:597)
> 9   pd                            	0x0000000103c15a83 sys_main + 595 (s_main.c:363)
> 10  pd                            	0x0000000103c0f282 main + 34 (s_entry.c:32)
> 11  libdyld.dylib                 	0x00007fffb43f0235 start + 1

It seems that the problems come from the name (_sc_template_) of the structure in a scalar. Sometime the symbol is not valid. I tried to add conditions that check if the symbol isn't _NULL_ in _template_findbyname_ and _scalar_vis_, it seems to solve some of the crashes but not all.

pierreguillot:
Here patches if you want to test the crash (open the help file)
[Archive.zip](https://github.com/pure-data/pure-data/files/1106867/Archive.zip)

umlaeute:
the crash-test is missing some abstractions:
- `xel.2d.get.nelems`
- `xel.2d.get.itorc`

pierreguillot:
[...]

PR #70, opened 2017-06-27
Fix structure crash

pierreguillot:
I think _sys_unqueuegui_ must be called from _scalar_free_ to avoid crashes (see #69)
For info: _sys_queuegui_ is called [here](https://github.com/pure-data/pure-data/blob/master/src/g_scalar.c#L334) that's why it is necessary to call _sys_unqueuegui_ in the free method otherwise _scalar_doredraw_ (and _scalar_vis_) can be called when the object is already deleted.


millerpuckette:
closing - accepted

PR #71, opened 2017-06-27
mac script updates

danomatika:
As suggested on the pd-list...

osx-app.sh:

* version arg now optional (defaults to Pd.app)
* plist version string now taken from configure --version output
* added /Library/Frameworks to System Wish.app search paths

updated both script references to macOS to reflect newer OS naming


millerpuckette:
closing -- accepted

Issue #57, opened 2017-05-05
Make Jackmp.framework optional on macOS


reduzent:
Currently, Pd on macOS connects to jackd only when Jack  was installed with the JackOSX package, which installs something called Jackmp.framework. When using jackd from homebrew or a self-compiled binary (with traditional lib format), Pd won't detect it and complains:

"Can't open Jack (it seems not be installed on this Mac)"

If it's technically feasible, it would be nice, if Pd would talk to jackd without Jackmp.framework.

umlaeute:
the nice thing about the framework is, that is allows to *easily* do *weak-linkage*, so that the framework need not be present on the target system at all, and Pd will still load/function.

it might be possible to do that with ordinary libraries, i just don't know (yet).

virusys:
I have Jackmp.framework and libjack.dylib installed on my system, but running `make` still doesn't seem to find either:

```
[...]

PR #72, opened 2017-06-28
optional JACK framework linking

danomatika:
On macOS, the default behavior when building with JACK support is to weak link to Jackmp.framework. Some users, however, prefer to link to a traditional jack dynamic lib such as that installed by homebrew (see #57).

This PR adds the `--disable-jack-framework` option to configure so this default behavior can be turned off and the traditional library search and linking can be used (as on Linux). The default weak linking is retained unless explicitly disabled by this option and a configure message is printed when weak linking is taking place.

Please test.


millerpuckette:
closing -- accepted

PR #73, opened 2017-06-28
Autoconf clean-pdlib missing target warning fix & small Windows configure tweaks 

danomatika:
This is a simple bugfix for clean-pdlib missing targets warning thrown by autoconf.

Also included are some small tweaks to configure.ac for windows, mainly that port midi should be built by default.


millerpuckette:
closed -- accepted

PR #74, opened 2017-06-28
Added [fudiformat]/[fudiparse]

umlaeute:
similar in spirit as [oscformat]/[oscparse]

this allows us to transmit Pd-messages over binary interfaces.
it also allows us to reparse messages (think: symbol "12")


millerpuckette:
I can't see why anyone would ever need this (except for the symbol-to-int parsing which should be done in the "float" object or something) - but anyhow, since nobody seems to know about FUDI, perhaps  this works as a gentle advertisement.

Ant1r:
I find this cool personally, I think it will be helpful.
Helpfiles are not distributed yet though, because undeclared in doc/Makefile.am ; see #84

PR #77, opened 2017-06-29
fixing wrap~ to make it consistent to wrap

porres:
[wrap~] has a bug, when you feed it "0", it becomes "1", unlike [wrap], which outputs 0. As a workaround, I need to use 2 [wrap~] object so I can turn the generated "1" back to zero.

So I just made [wrap~]'s code exactly like [wrap], which solves this issue.

### DIFF
--- a/src/d_math.c
+++ b/src/d_math.c
@@ -257,7 +257,7 @@ static t_int *sigwrap_perform(t_int *w)
     while (n--)
     {
         t_sample f = *in++;
-        *out++ = f - floor(f)    
+        *out++ = f - floor(f);    
     }
     return (w + 4);
 }


Ant1r:
Travis said : src/d_math.c:260:30: error: expected ';' after expression
You should better try compiling BEFORE submitting ;-)

porres:
I can't compile pd on the mac yet, I've been trying to sort it out on the pd list to no success, so I was just counting on travis for this kind of thing. Anyway, it is now fixed, thanks.

pierreguillot:
I'm agree with you [wrap~] should work like you say but this correction will break a lot of patches I guess.

porres:
I guess this is a recurrent concern, you have an object supposed to give an objective mathematical result, but gives a different one, and if you fix it, the thing will change, so will this cause issues?

Well, I can't think of a case where someone is relying on this wrong mathematical of the audio rate version result in order to make a patch work... so I think such a concern is exaggerated. By analyzing the issues in practice, I can't see how this behaviour would give you any desired feature, cause it's just so inconsistent that it's just an annoyance that you actually just have to create workarounds for it. 

Say every time you are in the beginning of a cycle it stays there, but not when it is zero, how can this be good for you? This is clearly a bug, I think there is no discussion there... and it seems to me that this particular concern ends up promoting the idea no bug should be fixed because things change. By the way, this a math function, it's like [+~] giving a wrong result... and you also have two objects that are inconsistent between each (wrap and wrap~).

cheers

porres:
> Say every time you are in the beginning of a cycle it stays there, but not when it is zero, how can this be good for you? 

I have to stress how I see this only generates inconsistent annoyances, like the example I gavem where I need to use two [wrap~] objects so it behaves consistently. Now, fixing this, would only allow you not to create ugly workarounds for it.

pierreguillot:
I think you're right. I just wanted to notice that a lot of people uses it anyway so the correction will break their patches and their current workaround will generate wrong values. I mean , [wrap~] is not really a well defined mathematical operation and right now when you open the help file the 1st thing you see is that 0 gives 1 so I guess for some people this is the right behavior (or at least some people has accepted it). Anyway, I'm totally for this kind of changes but perhaps just a warning in the help file explaining that the behavior has changed in 0.48, why and how, could ensure a minimum of compatibility.

porres:
> [wrap~] is not really a well defined mathematical operation

Well, the Wrapping Function is actually pretty straightforward, known and well established/defined. it is present in many programming languages, so everywhere else you find it, let's say SuperCollider or Max/MSP to stay within the realm of computer music, this is how they work. Perhaps people may not know it... but that is something else.

Anyway, it is a circular function, so when you reach the ending/wrapping point, it goes back to the starting point. In the case of Pd, the starting point is "0" and the wrapping point is "1", which makes it kind of a "modulo of 1" operation, and in fact this is even mentioned in the help file, which even gives more mathematical details on what it is or should do: "wrap~ gives the difference between the input and the largest integer not exceeding it". And this means it is the difference between its input and its "floor".

> now when you open the help file the 1st thing you see is that 0 gives 1 so I guess for some people this is the right behavior (or at least some people has accepted it). 

Well, it is pretty straight forward what it should do and output in the help file. Perhaps if you do not read it and pay attention to it, you may not realize that it has a bug... but this is different than saying the help file should give you the idea that this is how the object works.  So if you feed it "0", considering the description of the help file, you should not expect a "1" out as a result.

> I'm totally for this kind of changes but perhaps just a warning in the help file explaining that the behavior has changed in 0.48, why and how, could ensure a minimum of compatibility.

Anyway, I think the documentation is fine, it's pretty descriptive, and add more information just makes it confusing. There's not asctual room else to explain a "change", or "why and how", if anything just mention "hey, this was buggy and now it's fixed", but usually this kind of stuff is presented in the bugfixes changelog of a release, not in the documentation.

cheers

pierreguillot:
I guess you're right a lot of people knows how the object should work, but there is no wrap object in Max, nor in C, C++ or JS and the one in SC uses low and high values. So it wasn't obvious for me to use a [wrap~] object instead of a [%~] when I started with Pd and to know exactly how it should have work but I managed to use it despite the bug and I guess I'm not the only one. So I think a note at the bottom of the help file (like for the other objects) is not too intrusive (or a sub-patch for further information). For example "updated for Pd version 0.48 (correct the output when input equals 0)". I can be wrong and perhaps we should think of a better place to put this kind of information :)

porres:
In max it is [pong], or [pong~] that does the wrapping, it is also for 2 values though, which I think is really useful.

porres:
No, there is not one in C, but it is still something quite known and established, you can find people online discussing how to implement "wrap" in several languages. It is not an unknown term/concept, especially when it comes to dealing with "angles", because it is circular, so sometimes it's wrapping from -180 to 180, 0 to 2pi, and I believe in Pd it's 0 to 1 because it's how we usually deal with angles in Pd - 0 to 1 is the output of a phasor~, the input of a cos~ is from 0 to 1 and it even wraps internally - and maybe that information could be in the help file ;)

millerpuckette:
I just tried teh existing wrap~ - given 0 as input it outputs 0 on my machine.  Also, using "floor"
(even "floorf"i)  doubled the execution time.  So until I can see wrap~ malfunction I'm going slow on this one :)

porres:
Funny, maybe it depends on the OS? Here on my Mac it is malfunctioning, see attachment.

I'm on pd 64 bits, I can try everywhere else and see how it goes. But I can also confirm that the same issue happens for other versions for mac, namely the old 32 bit Pd Extended 0.42-5 and the new Purr Data 64 bits.

screen shot 2017-07-14 at 15 23 31

porres:
well, got the same thing on linux, ubuntu studio, 64 bits, pd vanilla 0.47-1!

porres:
testing on windows gave me the same thing as well, I'm on windows 10, 64 bits, tried on the latest vanilla, Extended 0.42-5 and the new Purr Data.

So, what machine and system do you have? 

@pierreguillot , you got this wrong behaviour as well, right? What is your setup?

porres:
as far as the current code goes, we have:

int k = f;
if (f > 0) *out++ = f-k;
else *out++ = f - (k-1);

So "f" is the input and "k" is its integer part. Hence, for an input of "0", we get:

 0 - (0-1), which is the same as 0 + 1 = 1, 

So I can't see how you'd get a 0 as the output, unless it is computing, somehow, the integer of "0" as "1", is it possible?

> Also, using "floor" (even "floorf"i) doubled the execution time.

well, I guess that if you have it as 
if (f >= 0) *out++ = f-k;

it'd work, right?

Or even, 
int k = f;
if (f == 0) *out++ = 0;
else if (f > 0) *out++ = f-k;
else *out++ = f - (k-1);

millerpuckette:
Oh, I see.  I'm using vanilla.  Try that :)

porres:
> I'm using vanilla. Try that :)

well, me too :) I tried it in Mac, Windows and Linux, but I only got 64 but machines.

I was mentioning I also tried Extended and Purr Data, but they just have the same version of wrap~ and I get the same things in it.

porres:
the image I posted, by the way, is a screen shot of vanilla, showing how it is malfunctioning.

What exactly is your setup?

millerpuckette:
here's the code in vanilla:

        t_sample f = *in++;
        if (f < x->x_lo) f = x->x_lo;
        if (f > x->x_hi) f = x->x_hi;
        *out++ = f;

I guess the code you cite above is from somewhere else.

porres:
Ok, now I'm really confused... 

All I did was download the source from this repository and then searched for wrap~ in it, I found it in d_math.c, so I changed it and made a pull request. You can see here in this pull request that I changed the "static t_int *sigwrap_perform(t_int *w)" function from line 253 of d_math.c, and that you'll see we have in it this bit of code that I changed.

"    while (n--)
    {
        t_sample f = *in++;
        int k = f;
        if (f > 0) *out++ = f-k;
        else *out++ = f - (k-1);
    }
"

Now, by searching for:  

"if (f < x->x_lo) f = x->x_lo;
if (f > x->x_hi) f = x->x_hi;"

What I find is the code of "clip~", from line 45 of d_math.c

So can you please double check it?

millerpuckette:
You're right... I must be going nuts or something.  And now I can get it to malfunction on a Mac (but it works on linux).  There's something going on that I don't understand at all :)

porres:
> And now I can get it to malfunction on a Mac (but it works on linux). There's something going on that I don't understand at all :)

Well, it malfunctions everywhere for me... I got a 64 bits PC with windows 10, and my linux is ubuntu studio 17.04 - 64 bits, which already ships with pd 0.47-1

what is your setup? really curious here ;)

millerpuckette:
Linux fuzz.localdomain 3.15.6-200.fc20.x86_64 #1 SMP Fri Jul 18 02:36:27 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux

porres:
well, I hope you find the reason why the bug doesn't happen on your machine :) 

but one way or another, and if you're concerned if using "floor" compromises efficiency significantly, how about my earlier suggestion of doing something like

"if (f == 0) *out++ = 0;
else if (f > 0) *out++ = f-k;
else *out++ = f - (k-1);"

?

Seems it would fix this where it malfucntions, keep the same efficiency, and not compromise or add a bug where it works already, huh?

cheers

millerpuckette:
I think that one returns -2 if you give it -1... but after 2 or 3 more tries I ended up with this one:

        t_sample f = *in++;
        int k = f;
        if (k <= f) *out++ = f-k;
        else *out++ = f - (k-1);

porres:
cool, tried it as a pd patch and it works indeed :)

porres:
well, since I can see this is working in the new 0.48 test release, I'm closing this ;) - thanks

PR #78, opened 2017-06-29
appveyor support

umlaeute:
this PR adds support for the [appveyor](https://appveyor.com) Continuous Integration service.

this is run a new test-build using Microsoft Visual Studio whenever something is pushed to the github repository (or whenever a pull request is made), allowing to find build failures early (given that nobody builds with MSVC regularily these days...).

The build is executed both in a 32bit and a 64bit environment.

The build logs can be found on the [appveyor project page](https://ci.appveyor.com/project/umlaeute/pure-data).
There are also artifacts attached to each (successful) build, which contain the produced binaries (the `bin/` folder), which can be downloaded for easy testing.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -2,7 +2,7 @@
 #
 # this automatically builds Pd for W32 using VisualStudio,
 # whenever a new commit is pushed to github...
-#   https://ci.appveyor.com/project/pure-data/pure-data
+#   https://ci.appveyor.com/project/umlaeute/pure-data
 
 environment:
   matrix:
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,0 +1,49 @@
+## AppVeyor configuration for Pure Data
+#
+# this automatically builds Pd for W32 using VisualStudio,
+# whenever a new commit is pushed to github...
+#   https://ci.appveyor.com/project/pure-data/pure-data
+
+environment:
+  matrix:
+  - platform: x86
+    AFLAGS: "/D__i386__"
+  - platform: x64
+    AFLAGS: "/D__x86_64__  /DPD_LONGINTTYPE=\\\"long long\\\""
+
+matrix:
+  allow_failures:
+    - platform: x64
+
+shallow_clone: true
+
+install:
+  - echo init
+  - mkdir bin
+
+# the version included with the Pd binaries,
+# is actually ftp://sourceware.org/pub/pthreads-win32/pthreads-dll-2002-03-02
+  - curl ftp://sourceware.org/pub/pthreads-win32/pthreads-w32-2-9-1-release.zip -o pthreads.zip
+  - 7z x -opthreads pthreads.zip  Pre-built.2/dll Pre-built.2/lib Pre-built.2/include
+  - set pthreaddir="%APPVEYOR_BUILD_FOLDER%\\pthreads\\Pre-built.2"
+  - cd "%pthreaddir%\\include" && copy "*.h" "%APPVEYOR_BUILD_FOLDER%\\src\\"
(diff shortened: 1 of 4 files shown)


danomatika:
I remember reordering configure.ac so the basic libs check section came before the os detection section for a specific reason and I see it's been changed back. Of the top of my head, I think it's related to needed to define the Makefile variables via AC_SUBST *before* setting the flags. I'll look through the commits to find when I did it.

danomatika:
Ok I found it. I reordered things in 38fb31a so user overrides for CFLAGS & LDFLAGS end up at the end so they can actually override things. This is done via appending instead of prepending in the OS detection and after all the autoconf libs checks, some of which add to the flags.

umlaeute:
ah ok, i'll have a closer look then.

umlaeute:
hmm: I've checked and double checked the output of configure in both the `master` and the `appveyor` branch with something like:

     CFLAGS="-DMYCFLAGGGGGGGGGGGGGGG" LDFLAGS="-Wl,--as-needed" ../configure --enable-jack --disable-portaudio --disable-portmidi --enable-debug

and 

     ../configure --enable-jack --disable-portaudio --disable-portmidi --enable-debug CFLAGS="-DMYCFLAGGGGGGGGGGGGGGG" LDFLAGS="-Wl,--as-needed"

i compared both the output of the `./configure` run and the generated top-level `Makefile`, and they are virtually identical (the only difference I found is, that with the re-ordered OS-code will obviously print out earlier)

~~~bash
$ diff Makefile.master  Makefile.appveyor
$ diff  config.log.master  config.log.appveyor
8a9,10
> configure: iPhone SDK only available for arm-apple-darwin hosts, skipping tests
> configure: Android SDK only available for arm-linux hosts, skipping tests
191,192d192
< configure: iPhone SDK only available for arm-apple-darwin hosts, skipping tests
< configure: Android SDK only available for arm-linux hosts, skipping tests
$
~~~

also, i've never noticed that the location of the `AC_SUBST` *declaration* mattered.

danomatika:
Ok cool. Just wanted to double check. I think the issue I was running into was wanting to manually specify -O3 and it was, for some reason, didn't seem to override the defaults. I think this was fixed more by removing the overrides in configure.ac then in moving things around.

Ant1r:
greetings for this work (although I never work on Windows myself...).
You certainly noticed that appveyor currently fails building; is it related to the fact this PR isn't yet merged ?

umlaeute:
@Ant1r yes, the failing builds are related to this.

I haven't managed yet to tell appveyor to *not* attempt a build if there is no `.appveyor.yml` file (as is the case in practically all branches/PRs apart from *this* one.

millerpuckette:
i sure would be happier not having "dotfiles" in the Pd source.  In general, Pd seems to be accreting compiler- and distribution-related files - it would be nice (perhaps in the far future) to be able to consider the various build gunk to be exerior to the Pd repo itself (and thus pluggable).  But for now I'll just put up with it...

danomatika:
@millerpuckette I agree and have been trying not to add more than what is already there. The appveyor support will be pretty useful in finding cross platform issues though.

How do you create the src distribution tarballs? The auto tools can generate one which contains only the base stuff:

    ./configure
    make dist

millerpuckette:
I'm using a homemade script to build src taralls.  At some point I want to
simplify things and just use "git archive".  Pobably not this release though :)

M

danomatika:
Well, I mention the "make dist" target as that exactly what it does: creates a distribution Pd-#.#.#.tar.gz that contains the files marked for distribution (and ignores those which aren't). Easy peasy :) Even though the autotools may be a bear to set up, once it is, it's easy to use.

danomatika:
Here's an example tarsal I just built: http://docs.danomatika.com/pdbuilds/pd-0.47.1-07112017.tar.gz

umlaeute:
turns out that appveyor already supports external config-files, so there's really no need to include the `.appveyor.yml` in the repository if this is making people  unhappy... see #98

PR #79, opened 2017-06-29
removing [midiclkin] / updating the help files of [float], [value] and [send]

porres:
About the first commit: I simply removed all about 'midiclkin' from Pd, and I was able to compile it in MacOS Sierra...


porres:
btw, I was able to compile it in MacOS Sierra...

porres:
about the second commit: I	updated the help files of [float], [value] and [send] to include "hidden" information about the relationship between these objects - that is how a [send] can talk to a [value], and then how [float] can also send a number to a [value] object

porres:
why is the "AppVeyor build" failing?

umlaeute:
i think you should start doing pull-requests that only deal with a specific issue: **either** fix `[midiclkin]` **or** update the help-patches.
it makes it much easier for the person who merges the PRs to pick only the features they want and not those they don't want to.

the *AppVeyor* build fails because it tries to build even though it appveyor is not configured for your branch (the build consequently shouldn't even be attempted. I'm going to file a support request with appveyor).
for now, this can safely be ignored.

porres:
> i think you should start doing pull-requests that only deal with a specific issue: either fix [midiclkin] or update the help-patches.

Agreed, only problem is that I really don't know how to do it - I'm really adventuring myself here in this field... 

What happens is that I do the first pull request, then the next thing I do and change gets automatically merged into the same pull request... I tried finding a way to split it and organize this but completely missed it.

cheers

umlaeute:
you might want to [read the documentation](https://help.github.com/articles/creating-a-pull-request/).

basically each PR is attached to a single `branch`. whatever you push to this branch, will end up in the PR.
On *this very page*, github tells you this explicitely. If you look just below my post, you should see something like:
> Add more commits by pushing to the `master` branch on `porres/pure-data`.

IIRC I already tried to [explain this in detail](https://github.com/pure-data/pure-data/pull/17#issuecomment-220300776) a while ago...

porres:
> you might want to read the documentation.

thanks for your help, as I intend to do more and more of this, I definitely should learn it. But I'll leave this as it is cause it seems to be related, as the changes of midiclkin also ask for documentation updates.

Now, any chance this could be checked and incorporated in the upcoming release any time now? Seems like a simple and straightforward.

By the way, I just made a new commit for help-intro.pd, because if you accepted my previous one, I thing the new objects like [fudiformat] would be lost. So now all is good, I just removed [midiclkin], and the new objects are there.

cheers

PR #80, opened 2017-07-01
make WISH settable from cmdline

umlaeute:
this PR aims at making the actually used `wish.exe` settable during the build-process.


danomatika:
Looks good to me. As you've tested the restructured configure, should be fine.

umlaeute:
so I've added support for the autotools chain as well (where you can specify an alternative `wish` executable via a configure-argument: `--with-wish=wiff`

I've also renamed the define from `WISHAPP`  to just `WISH` because I actually think that Dan is right about the name :-)

the define is honoured by both W32 and Linux, but not by OSX, since the Darwin code for selecting the proper tcl/tk interpreter is much too complicated to not break it. maybe @danomatika wants to give it a try...

danomatika:
You can set the Wish used when you build the app with the mac/osx-app.sh script. I'm not sure if anything needs to be changed in the load code.

Issue #68, opened 2017-06-27
FFT crash with large blocks sizes


pierreguillot:
When I use a block size >= 2^20 (and sometimes >= 2^19) with [rfft~], pd crashes. 
I tried with Pd 0.47.1-32 & 64bits on MacOS 10.12.5. 
Here the console report:

> Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
> 0   pd                            	0x00082bd4 mayer_realfft + 36
> 1   pd                            	0x0007a46c sigrfft_perform + 28
> 2   pd                            	0x0006ce55 dsp_tick + 37
> 3   pd                            	0x00059ef6 sched_tick + 214
> 4   pd                            	0x0005ad63 m_mainloop + 1875
> 5   pd                            	0x0005e8b9 sys_main + 1577
> 6   pd                            	0x00001f19 _start + 208
> 7   pd                            	0x00001e48 start + 40

Here the files to test it (_FFT-tests-1.pd_ makes Pd crash and _FFT-tests-2.pd_ is alright - remove the extension _.txt_):
[FFT-test.pd.txt](https://github.com/pure-data/pure-data/files/1104623/FFT-test.pd.txt)
[FFT-tests-1.pd.txt](https://github.com/pure-data/pure-data/files/1104624/FFT-tests-1.pd.txt)
[FFT-tests-2.pd.txt](https://github.com/pure-data/pure-data/files/1104625/FFT-tests-2.pd.txt)

pierreguillot:
I think the problem comes from the function _alloca_ in the _mayer_realfft_ and _mayer_realifft_ that fails to allocate memory for a large block size. 
The main problem is that the function alloca doesn't return a NULL pointer if the allocation fail but for me there are 2 solutions:

- In the DSP method, if the size is higher to a specific limit the function generates a warning and avoid to add the perform routine. The problem is that the limit size of alloca is not defined and depends on the OS. Perhaps 2^19 or 2^18?

- Replace the alloca function with malloc or whatever.

umlaeute:
sidenote: I thought that *mayer_fft* has been replaced by *ooura*...

pierreguillot:
The mayer_fft interface is used as wrapper for the ooura (_d_fft_fftsg.c_) or the fftw (_d_fft_fftw.c_) libraries.

Spacechild1:
fixed by #81

PR #81, opened 2017-07-02
Fix: fft size for MacOsX in d_fft_fftsg

pierreguillot:
In **d_fft_fftsg.c**, replace the **alloca** functions with the use of a global pointer allocated with **t_getbytes** in the **ooura_init** function (like the other global pointers) to avoid crashes when **alloca** fails. See #68 for further information of the problem.


umlaeute:
i haven't had a close look, but this seems to breed havoc.

recently there has been quite some work on making Pd usable from multiple instances/multiple threads.
how is *introducing* a global variable not breaking thread-safety?

also, why can't `alloca()` be replaced by a simple `malloc()/getbytes()` call and some `goto cleanup` to free the allocated memory before the function exits?

pierreguillot:
I tried to follow the same approach of the code. The `alloca` function is used because it allocates memory in the stack frame of the caller. The direct replacement of the `alloca` function by the functions `malloc` and `free` reduces a lot the efficiency (IMHO it's really a bad practice). In this approach the allocation is called at the first DSP tick - and until there are no bigger FFT size it will never be called again so this seems to be a good compromise. 
Perhaps I'm wrong but I think that this whole part of the code is not thread safe and need to be modified. And I think it's worst practice to add a new bad approach that will be possibly forgotten because it compiles and supports the multi instances and multi threads. In my approach, the global pointer is near the other global pointers that need to be changed and is allocated in the same function so the day the code is changed you can't ignore it.

umlaeute:
I really wonder about the order of magnitude of the efficiency decrease: if you make any such claims, you should add references/numbers.

I also believe that global memory allocation is **much worse** practice than heap allocating memory.

pierreguillot:
> I really wonder about the order of magnitude of the efficiency decrease: if you make any such claims, you should add references/numbers.

You can read [this](http://www.rossbencina.com/code/real-time-audio-programming-101-time-waits-for-nothing) and [this](https://github.com/harryhaaren/fypRealtimeCppPrograming/blob/master/realtimeAudioApplicationsInCpp.pdf) and you can find a lot of documentation on internet. I don't think that numbers are relevant because the results can change drastically depending on the patches. 

> I also believe that global memory allocation is much worse practice than heap allocating memory.

I never said that global memory allocation is the **best** practice. It depends on the context. In this context I think it's better than using `malloc()`.

umlaeute:
i still don't see how this is related to replacing `alloca()` to avoid crashes when it fails.

allocating memory in an rt-thread is considered bad practice, but there is no difference between using `malloc()` and `alloca()` in this respect. the question is not weather the memory is allocated on the *heap* or on the *stack*, but whether it is allocated at all (rather than pre-allocated).

If you want to get rid of memory allocation in the rt-thread in Pd, then you have to rewrite Pd entirely.
Practically each time a message is generated, memory is allocated and freed.

Maybe yo ushould just change the description of your PR, so it says what you are really trying to achieve...

pierreguillot:
I know that pd runs entirely on the rt-thread... That's not the problem here, I just want to avoid a dirty crash... :(

If you read what I wrote in the issue #68: 
> The main problem is that the function alloca doesn't return a NULL pointer if the allocation fails

so if `alloca()` fails, Pd crashes directly. It's not really specified but it seems that the amount of memory that can be allocated by `alloca()` is pretty small compared to the amount of memory that can be allocated by `malloc()`. And I think it depends on the OS. Anyway on my computer, the allocation succeed with `malloc()` and fails with `alloca()` when the FFT size >= 2^19. 

You can also consider that 2^19 is too big and Pd should just avoid the computation. But as this limit size (the amount of memory that can be allocated by `alloca()`) seems to depend on the OS, it's difficult to determine when to compute or not...

 What I wrote before is that with my approach
> the allocation is called at the first DSP tick - and until there are no bigger FFT sizes it will never be called again 

I only suppose this is more efficient than allocating memory at each DPS tick and furthermore this is the same approach that is already in the code. I explained why the multi-instances/multi-threads problem is, for me, not relevant in this case. 

Anyway, if you have a better implementation I'll be happy to test it on my computer and if you want to change the name of the issue, feel free to do it.

millerpuckette:
I think a good approach would be to use malloc() if the size is greater than some reasonable limit (64kbytes?) and alloca() otherwise.  Once th FFT gets big its own execution time will dwarf that of malloc().  But this might be something I should be doing everywhere I'm using alloca() ?

pierreguillot:
I guess you're right, with large size the allocation time of **malloc()** is small compared to the time of the FFT (furthermore big FFT is not usable in real-time ,I only achieved to generate a lot of pops...). 

Anyway if the multi-thread support continues, I guess that the memory management of the FFT process will change and certainly a better place to allocate this memory can be found. I mean, the buffers **ooura_costab** and **ooura_bitrev** are not thread safe, they must become specific to the instance (and why not to an FFT process ?). As they are used together with **buf** (allocated with **alloca()**), why not doing the same for all the buffers?

I tried to have a look of the other uses of **alloca()** and it seems that the sizes are small enough to prevent any problem (even if we can never know) but some times it can be replaced by **malloc()** without too many diffficulties (like in g_clone.c or x_misc.c). If you want I can have a closer look at it.

umlaeute:
i guess there is no sane default for a *reasonable* size to switch from `alloca()` to `malloc()`.

how about, using `malloc()` as a fallback if `alloca()` fails?

pierreguillot:
The main problem is 
> There is no error indication if the stack frame cannot be extended.

from this [doc](http://man7.org/linux/man-pages/man3/alloca.3.html)
I can't find an ideal solution... perhaps what you suggested was the best one in fact :)

danomatika:
Is this something that can be cleaned up for the 0.49 release? If it fixes a bug, it might be good to have.

pierreguillot:
I can merge the master and fix the conflict. I will work on it ASAP!

pierreguillot:
I tested on MacOS 10.13.4 with Pd 0.49 (master branch - commit c958dca) and the application still crash when the block size is higher to 2^20. I merged the master branch into my branch and I adapted the code to the changes of the master branch. It fix the problem and it should be be compatible with the multiinstance/multithread.

So if the approach fits you, you can merge it.

millerpuckette:
merged

PR #84, opened 2017-07-03
register fudi helpfiles to doc/Makefile.am

Ant1r:
else they won't install.


Ant1r:
Was just done by Miller. Closing.

PR #86, opened 2017-07-03
Correct help for text object

pierreguillot:
1st commit: change large number to append a new line from 2e+15 to 2e+09 because of this [modification](https://github.com/pure-data/pure-data/commit/b563fd427f6ce729ae021eaa584dc0ee16be2649#diff-a99cb8752b27f71246b1996f90c32179R684) i guess in b563fd427f6ce729ae021eaa584dc0ee16be2649.

2nd commit: improve the documentation at the bottom of the help file
from 
> if inlet 2 is supplied to specify a starting field, the line is not resized - instead, as many items are replaced as were in the list at left. Otherwise, the line is resized to the size of the list.

to 
> if inlet 2 is supplied to specify a starting field (field number != -1), the line is not resized - instead, as many items are replaced as were in the list at left. Otherwise (field number == -1), the line is resized to the size of the list. Also if inlet 2 is supplied to specify a starting field (field number != -1), new line can't be appended.


millerpuckette:
I think the 1e15 isn't a problem - at least on my machine that's working fine and there should be no limit to what one can specify to mean "infinity".  I updated the text in my own words to try to correct the problems flagged in the second commit.

pierreguillot:
Thank for the text change ! And you're right, changing the help file doesn't solve the problem, it is like putting a bandage on a broken leg :) I guess this is the c code that must change. Here is my problem on MacOS with Pd-0.47 (master branch)
ezgif-3-573ab5a01e
Perhaps I can close this PR and create a new issue specific to this problem?

millerpuckette:
Hmm, that bug should have been fixed in this commit:

b563fd42 src/x_text.c  (Miller Puckette    2017-06-28 22:44:20 -0700  691)
   lineno = (x->x_f1 > (double)0x7fffffff ? 0x7fffffff : x->x_f1);

I just tried it on my 10.6 mac and it seems to be working there.  Can
you verify that your source has a line like the above?

thanks
M

pierreguillot:
Yes I've just tried and I'm sure I have this line. I investigated the problem a bit more (after testing your new binaries that work just fine!) and it's perhaps my version of GCC (or Clang because it seems that on Mac, GCC links to Clang). The problem is caused by a bad cast on `x->x_f1`

`lineno = (x->x_f1 > (double)0x7fffffff ? 0x7fffffff : x->x_f1);`
should be 
`lineno = (x->x_f1 > (double)0x7fffffff ? 0x7fffffff : (int)x->x_f1);`

For information, I'm on mac 10.12.5 and 
> Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
Apple LLVM version 8.1.0 (clang-802.0.42)
Target: x86_64-apple-darwin16.6.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

millerpuckette:
Cool - I grabbed that one commit (finally figured out how to grab a csingle
commit from a PR :)

M

PR #88, opened 2017-07-05
develop updates for 0.48

danomatika:
Various updates: (started with the GUI, then went into translation & build system):

* raise audio, midi, & font dialogs on creation, this fixes any of these dialogs opening *behind* the other
* changed default macOS Preferences menu entry & key binding to open the audio dialog (previously
opened path dialog), I was looking to remove the default Preferences entry but there isn't really a way to do it from Tk
* added generated tcl/pd-gui script to .gitignore
* removed a couple stale Shift-k bindings (message dialog key bindings were previous changed from accelerator-Shift-K to accelerator-Shift-M)
* helpbrowser:
  * added opening files & folders via the keyboard (Enter & right arrow keys)
  * hide while refreshing list boxes to avoid ugly black window flash
* find dialog:
  * switched back to withdraw on cancel instead of destroy, this should fix some behavior such as the labels not updating and other possible bugs (maybe this? https://lists.puredata.info/pipermail/pd-list/2017-06/119098.html)
  * added safety mechanism for dialog changes while search is in progress, this was in response to a comment about changing search windows while a search is in progress. I actually implemented live updates of find's target window but it was indeed confusing. Decided to leave the is_searching mechanism in place.
  * no longer hide after search on macOS, this better matches the behavior on other platforms & the contextual label text can now be seen again
  * search string in label now properly truncated, this keeps overly long search strings from overflowing the contextual label
  * match button placement now left aligned and no longer fills, this should get the layout to play more nicely but needs to be tested on other platforms as the button now aligns horizontally with the find button
  * entry selection now highlighted on window focus in for keyboardists, this means you can switch between windows with the keyboard and immediate update the find query if the dialog is active
* other small fixes: whitespace, typos, etc

Tested on macOS in Tk 8.4 and (partially) 8.5.


danomatika:
@Ant1r Since we have all the strings now, I'll remove the xgettext --join-existing option when building the template.

All it was doing was *appending* the new line number comments under the old ones leading to confusion since we're seeing both the new and historical positions of all the strings. We could either keep this functionality and drop the line comments or, as I think is best, simply don't append. The only reason why would want to append would be if there are strings added by hand but we seemed to have that fixed now.

danomatika:
Also, don't merge yet as I'm writing up a Readme on how the translations work, partially in order to understand it better myself. I'm noticing that we need to update the .po files whenever the template is changed in order to merge in any new strings and mark changed strings as fuzzy aka possibly needing translator updates.

danomatika:
Ok, I fixed the template and po file relationship. It was commented out but wouldn't run correctly even if it wasn't. The basic idea is that if the tcl files change, the template is rebuilt and looks for new or different translation strings. If the template is updated, then the translation .po files should also be updated to reflect and (possible) changes. This now works.

Now, I can understand the comment says it's annoying for this to be run whenever any tcl file is changed, but that means it's much easier for the translations to get out of sync with the current source.

If we want to have this working but without it happening automatically whenever *any* tcl file's timestamp is change, I can rework the make targets to it becomes an explicitly manual step, say something like:

    cd po
    make update-po

Also, the last commit fixes the make distcheck breakage.

danomatika:
I'm getting errors with the previous commit when running make after a template change:

    cp: ./vi.po and ./vi.po are identical (not copied).

You can trigger a template hang by touching any of the tcl files:

    touch tcl/dialog_find.tcl
    make -C po

umlaeute:
hmm, weird that you are getting errors: the `-` prefix should ignore any errors.
(or does the build-succeed despite getting errors?).

i think the out-of-tree builds have the very same use-case as in-tree-builds (with the only additional constraint, that they must not pollute the out-of-the-build directory; i use this regularly to work both on mingw and native linux builds in parallel).

it also seems that the out-of-tree builds are not fixed anyhow, as I get

~~~
$ touch ../tcl/dialog_find.tcl 
$ make -C po
[...]
xgettext: error while opening "../tcl/pd_deken.tcl" for reading: No such file or directory
~~~

when re-running `make` the error goes away...

danomatika:
Well I guess "error" in this case is subjective as it literally prints "error" even if the build succeeds. I would be nicer/better if there aren't any extra errors or warnings in the make output as it is. I'll add a redirect so it will fail silently ala:

    -cp $(srcdir)/$@ . 2>>/dev/null

I'm also noticing that the bin/pd symlink fails in the out of tree build as it's expecting the path to pd-gui.tcl to be ../tcl/pd-gui.tcl. I'm wondering how to make this work. Easiest option would be to copy the tcl files into the out of tree build's tcl dir, but I'm not 100% sure this is the "right way".

To double check, this is how I'm testing the out of tree build:

    mkdir -p build
    ../configure
    make

I thought that running make distcheck was enough to test this but I guess not...

danomatika:
Damn, adding the redirect doesn't silence the make error anyway ;P:

    make: [vi.po] Error 1 (ignored)

danomatika:
Overview of the last set of commits:

**po**

* Building translations now have proper dependencies (msg -> po -> template). This fixes an issue where you'd need to run make twice to update the msg file since that process was happening before the po update.
* The template string line location comment generation was disabled and POT-Creation-Date template header line is now scrubbed. This should reduce template & po updates so they only happen on translation string changes, not *any* tcl file timestamp change.
* Since the template line location info is removed, you can see where things are using the "locations" make target. This info has been added to the po readme.

**Out of Tree Builds**

* The po & template generation now stays in srcdir with generated .msg files ending up in builddir. I see the template and translation files as source files and the .msg files as part of the build process, so there's no need to rebuild the po files in each build tree locally.
* symlink tcl files to builddir so launching pd locally works
* osx-app.sh has been updated to handle a different builddir and make app now works out of tree.
 
And a few small things:

* fixed clang warning
* helpbrowser now shows c files, so the example external sources are now listed
* improved pd~ help file placement

Ant1r:
Thanks a lot for all of this! However some translation strings still are problematic:

- I think "Show Hidden Files and Directories" should be "Show &Hidden Files and Directories" ('&' here specifies the letter taken for key accelerator), though even if I fix it (into pd_dynamic_strings.tcl then update in fr.po), it doesn't change the string in the file dialog; and it seems this translation is not reported in fr.msg.
Actually I was unable to make this work again with the present branch, although it worked before (when manually added to fr.po).

- if I try to translate the "file edit put find media window help" string, pd's GUI doen't show up anymore (and doesn't print any related error message to stdout/stderr, even with -stderr -verbose -d 3). Maybe
the simplest would be to add each of these words separately in pd_dynamic_strings?

Will you be strong enough this time ;-) ?

danomatika:
>though even if I fix it (into pd_dynamic_strings.tcl then update in fr.po)

Try making the change & running `make all`. It should now regenerate the template and then the po files form the template.

>the simplest would be to add each of these words separately in pd_dynamic_strings?

That won't work as get text uses the entire string as a unique identifier. Each name needs to be separated into a separate string. We can look into this.

danomatika:
I think the problem with the hidden files string was also that I made a typo and uses "pus" instead of "puts". This should hopefully be fixed in the latest commit.

danomatika:
Got the menus:

screen shot 2017-07-10 at 2 44 20 am

danomatika:
@umlaeute I think maybe both together? I was responding the Roman's info that weakjack.h usually never needs to be included on Linux.

umlaeute:
I'm afraid you are right and probably both should be used together (though I'm still convinced that [his is a bug somewhere](https://stackoverflow.com/questions/44967115/) and we shouldn't add kludges to our codebase to work around them...; but the pragmatic solution is certainly to just not include `weakjack.h`

danomatika:
@umlaeute @millerpuckette Last commit is the beginning of the documentation update. Most work was done on greatly expanding the INSTALL.txt instructions. I've not built Pd on Windows, so maybe someone who has can either send me info or add it to the relevant section.

INSTALL.txt
```diff
@@ -1,64 +1,255 @@
...
-(universal) build, etc:
+## Requirements
 
-    $ ./configure --help
+Core build requirements:
 
-Pd can now be run out of the bin directory via:
+* Unix command shell: bash, dash, etc
+* C compiler chain: gcc/clang & make
+* Tcl/Tk: Tk wish shell
```
umlaeute: technically, tcl/tk is not a build requirement (it is a requirement to *run* Pd with a GUI). this is probably slightly blurred if by "building" you mean "creating a (self-contained) package that you can put somewhere and run Pd" (e.g. a Pd.app)

INSTALL.txt
```diff
@@ -1,64 +1,255 @@
...
+For libraries, you will need to install the "development" versions which include
+the source code header files. In Debian, the ALSA development package is called
+"libasound2-dev":
+
+    sudo apt-get install libasound2-dev
+
+Similarly, optional development libraries can be also be installed to for
+additional features. Install the JACK development files on Debian:
+
+    sudo apt-get install libjack-jackd2-dev
```
umlaeute: you might also want to mention `libjack-dev` as an alternative

umlaeute: which i did just now (see below)

INSTALL.txt
danomatika: Agreed. Now marked as a core runtime requirement.

danomatika:
Looks like the Appveyor build is failing as $WISH is not being treated as a string. Do we need to add explicit quotes on Windows or?

millerpuckette:
Is this in a state where I can merge it?  (I.e., if I do so, will it cause confusion?)  I imagine it's still underway but i believe I can leave it open and just track new commits as they arrive (?)

danomatika:
> Is this in a state where I can merge it? (I.e., if I do so, will it cause confusion?)

It could be merged. @umlaeute has been merging master and small PRs and it's largely up to date. I will do another merge with master now.

> I imagine it's still underway but i believe I can leave it open and just track new commits as they arrive (?)

It's mostly done except for the following things:

* fixing current build error on Windows (WISH string escaping?)
* updating install info in the HTML instructions to match a subset of that in INSTALL.txt
* finishing mac/README.txt update
* double checking ' in recent files filename on macOS

We could wait a day and/or these could be rolled into another, smaller PR.

This next one I think is important to fix for libpd as it is a build breaker. It could be done in it's own PR but involves more deeper architecture stuff, so I was waiting for you to weigh in:

* libpd: disable preferences saving & loading (remove s_file.c) -> #95

danomatika:
@millerpuckette Actually, there are some things that have changed in master I need to merge in manually. Will do that soon.

millerpuckette:
I discovered (commit b9b06bf9) that simply changing WISH to WISHAPPNAME in s_inter.c seems
to fix windows compilation problem.

millerpuckette:
OK, all merged in... maybe time to close this and start another.

danomatika:
Sweet. Thanks. Will do for the last couple things.

PR #89, opened 2017-07-05
binary mode for [stdout]

umlaeute:
this PR adds a few options to `[stdout]` to switch between various modes useful outside of the use with `[pd~]`.

namely:
- *binary mode*: allows to send arbitrary strings to the stdout as list of bytes
- *no-semicolon mode*: omits the trailing semicolon (which is rather annoying if you want to communicate with any other program (which typically does not speak FUDI) over the `stdout`)
- *no/force fflush mode*: allow to override the default fflush mode (on for W32, off for the rest of the world)

i think with these additions, `[stdout]` becomes an object of general usefulness.

btw, is there a reason why `[stdout]` is kept in *extra* (rather than built-in)

### DIFF
--- a/extra/stdout/stdout-help.pd
+++ b/extra/stdout/stdout-help.pd
@@ -1,19 +1,38 @@
-#N canvas 121 60 539 339 12;
+#N canvas 124 51 758 493 12;
 #X msg 126 203 walk the dog;
 #X msg 117 156 1;
 #X obj 117 240 stdout;
-#X text 269 287 updated for Pd version 0.42;
 #X obj 14 13 stdout;
 #X text 67 14 - write messages to standard output;
 #X msg 122 179 1 2;
-#X obj 119 291 pd~;
-#X text 44 291 see also:;
+#X obj 119 431 pd~;
+#X text 44 431 see also:;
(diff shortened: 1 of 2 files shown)


umlaeute:
I also wonder about this in the docs:
> there seems to be no canonical way a `stdin` object should act

how would that be different than e.g. a `netreceive` object? assuming there is a *binary* mode, I'm pretty sure that there is a *general usefulness* of such an object (though I guess I'm missing some pitfalls)

millerpuckette:
hmm, your 'binary' write will fail if any of the bytes are zero.  I'm also making changes now to stdout so will fix this in passing :)

umlaeute:
thanks for taking a look on this.

as for `0`-bytes: well yes, this is how `printf()` works, i don't think that there is anything inherently wrong with the current behaviour. cf.

     printf("hello\0world");

The only thing to keep in mind is, that without a terminating `\n`, the line will not get flushed (so it doesn't print immediately).

forcing fflush mode will do that for you:

     [64 0 64(
     |
     [sprintf -b -f]

in *binary* mode, the object will also `fflush()` when receiving an empty list (resp. bang), so the same works as well:

      [ 64 0 64, bang(
      |
      [sprintf -b]

Issue #90, opened 2017-07-06
SIGFPE crash when Tk is installed without xft support


simonvanderveldt:
Whilst I was packaging Pd (0.47-1 as well as [current master](https://github.com/pure-data/pure-data/commit/3381b5f1d025c9ee35fa810a5125026eee2c4127))  for gentoo I ran into the following issue: Pd compiled and ran fine but it would crash because of a SIGFPE when I tried to add any of the 5 default objects.
This is the accompanying backtrace
```
#0  0x000000000045c62a in rtext_senditup ()
#1  0x000000000045d454 in rtext_height ()
#2  0x000000000046b6cf in text_vis ()
#3  0x000000000043f06c in gobj_vis ()
#4  0x0000000000448e19 in glist_add ()
#5  0x0000000000468715 in canvas_objtext ()
#6  0x0000000000468ae5 in canvas_obj ()
#7  0x0000000000483a08 in pd_typedmess ()
#8  0x000000000044ccc9 in guiconnect_anything ()
#9  0x0000000000483edc in pd_typedmess ()
#10 0x000000000047dacf in binbuf_eval ()
#11 0x000000000048ddb7 in socketreceiver_read ()
#12 0x000000000048d0cb in sys_domicrosleep ()
#13 0x000000000048d152 in sys_microsleep ()
#14 0x000000000048881e in m_pollingscheduler ()
#15 0x000000000048895f in m_mainloop ()
#16 0x000000000049107b in sys_main ()
#17 0x000000000048b071 in main ()
```

After some digging I found out this was caused by my Tk being installed without xft/libXft support.
Now I did get a warning in the Pd console that Pd couldn't load it's font, but no errors.

AFAIK this is a compile time dependency that should be catched at configure time.
This can be done by passing `--static` and `--libs` to `pkg-config`, like this
```
pkg-config --static --libs tk
```
And then checking if `-lXft` is in the list.

In case that's problematic checking at runtime is, according to [Tk's docs](https://wiki.tcl.tk/9015#pagetoc101e06f4), relatively easy as wel
```
tk::pkgconfig get fontsystem
```

This should return `xft`

umlaeute:
hmmm, Pd-gui and Pd-core are two separate processes that communicate via a network socket.
the Pd-gui is a pure Tcl/Tk program (basically running `wish pd-gui.tcl`), the Pd-core is a pure C program.
[...]

PR #92, opened 2017-07-06
make font-handling robust against invalid font-metrics

umlaeute:
sometimes the GUI is unable to come-up with proper font-metrics, and returns `0` instead (see #90 ).

This patch checks the font-metrics as received by from GUI, and falls back to the built-in `sys_fontspec` if the values look fishy.

this will give slightly off font-metrics for broken tcl/tk builds (which i think is better than segfaulting or refusing to work).

### DIFF
--- a/extra/pd~/GNUmakefile.am
+++ b/extra/pd~/GNUmakefile.am
@@ -10,7 +10,7 @@ OTHERDATA =
 pd__la_SOURCES = pd~.c
 pdsched_la_SOURCES = pdsched.c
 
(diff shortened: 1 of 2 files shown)


simonvanderveldt:
This works for me, I can create all object and I'm not getting a `SIGFPE`.

umlaeute:
which error message?

simonvanderveldt:
Sorry, that was a bit unclear. I meant the error message that was added in this pr
`error("Ignoring invalid font-metrics from GUI!");`

PR #93, opened 2017-07-06
check for weak jack

umlaeute:
we might have weak jack symbols on non-apple platforms as well (i just stumbled upon this on Ubuntu-16.04LTS, where Pd would crash when switching to jack), so the check whether the jack-functions are non-null should be unconditional.

(the overhead is minimal anyhow)

### DIFF
--- a/extra/pd~/GNUmakefile.am
+++ b/extra/pd~/GNUmakefile.am
@@ -10,7 +10,7 @@ OTHERDATA =
 pd__la_SOURCES = pd~.c
 pdsched_la_SOURCES = pdsched.c
 
-EXTRA_DIST = makefile notes.txt
+EXTRA_DIST = makefile notes.txt binarymsg.c
 
 #########################################
 ##### Files, Binaries, & Libs #####
--- a/src/s_audio_jack.c
+++ b/src/s_audio_jack.c
@@ -12,7 +12,9 @@
 #include <string.h>
 #include "m_pd.h"
 #include "s_stuff.h"
+#ifdef __APPLE__
 #include <jack/weakjack.h>
+#endif
 #include <jack/jack.h>
 #include <regex.h>
 
--- a/src/s_audio_jack.c
+++ b/src/s_audio_jack.c
@@ -307,13 +307,11 @@ jack_open_audio(int inchans, int outchans, int rate, t_audiocallback callback)
     int srate;
     jack_status_t status;
 
-#ifdef __APPLE__
     if (!jack_client_open)
     {
-        error("Can't open Jack (it seems not to be installed on this Mac)");
+        error("Can't open Jack (it seems not to be installed)");
         return 1;
     }
-#endif
 
     jack_dio_error = 0;
 


umlaeute:
most likely, this fixes the long-standing https://sourceforge.net/p/pure-data/bugs/1228/

simonvanderveldt:
For me this fixes the linking issue (as expected), but I get the following segfault when actually connecting to jack from Pd (i.e. when I click on Media > jack):
```
(gdb) run
Starting program: /usr/bin/pd 
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
[New Thread 0x7ffff7faf700 (LWP 18367)]
[New Thread 0x7ffff7f2e700 (LWP 18368)]

Thread 3 "pd" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7f2e700 (LWP 18368)]
0x00007ffff78df1fa in Jack::JackLinuxFutex::Connect (this=0x781c20, 
    name=name@entry=0x7ffff7f2dd74 "system", server_name=server_name@entry=0x76eaa0 "default")
    at ../linux/JackLinuxFutex.cpp:177
177	../linux/JackLinuxFutex.cpp: No such file or directory.
(gdb) thread apply all bt

Thread 3 (Thread 0x7ffff7f2e700 (LWP 18368)):
#0  0x00007ffff78df1fa in Jack::JackLinuxFutex::Connect (this=0x781c20, 
    name=name@entry=0x7ffff7f2dd74 "system", server_name=server_name@entry=0x76eaa0 "default")
    at ../linux/JackLinuxFutex.cpp:177
#1  0x00007ffff78dfcfe in Jack::JackLibClient::ClientNotifyImp (this=0x76e970, refnum=0, 
    name=0x7ffff7f2dd74 "system", notify=0, sync=<optimized out>, message=<optimized out>, value1=0, 
    value2=0) at ../common/JackLibClient.cpp:169
#2  0x00007ffff78c455f in Jack::JackClient::ClientNotify (this=0x76e970, refnum=0, 
    name=0x7ffff7f2dd74 "system", notify=0, sync=1, message=0x7ffff7f2ddcc "", value1=0, value2=0)
    at ../common/JackClient.cpp:175
#3  0x00007ffff78e1f03 in Jack::JackSocketClientChannel::Execute (this=0x76ec10)
    at ../posix/JackSocketClientChannel.cpp:143
#4  0x00007ffff78dd0d1 in Jack::JackPosixThread::ThreadHandler (arg=0x76ed38)
    at ../posix/JackPosixThread.cpp:59
#5  0x00007ffff769e394 in start_thread () from /lib64/libpthread.so.0
#6  0x00007ffff62e8abd in clone () from /lib64/libc.so.6

Thread 2 (Thread 0x7ffff7faf700 (LWP 18367)):
#0  0x00007ffff76a3f4f in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1  0x00007ffff78ddce9 in Jack::JackPosixProcessSync::Wait (this=this@entry=0x78e4e8)
    at ../posix/JackPosixProcessSync.cpp:81
#2  0x00007ffff78d4d68 in Jack::JackMessageBuffer::Execute (this=0x7862b0)
    at ../common/JackMessageBuffer.cpp:104
#3  0x00007ffff78dd0d1 in Jack::JackPosixThread::ThreadHandler (arg=0x78e4c8)
    at ../posix/JackPosixThread.cpp:59
#4  0x00007ffff769e394 in start_thread () from /lib64/libpthread.so.0
#5  0x00007ffff62e8abd in clone () from /lib64/libc.so.6

Thread 1 (Thread 0x7ffff7fb1740 (LWP 18276)):
#0  0x00007ffff76a6fdd in read () from /lib64/libpthread.so.0
#1  0x00007ffff78de5d6 in read (__nbytes=4, __buf=<optimized out>, __fd=<optimized out>)
    at /usr/include/bits/unistd.h:44
#2  Jack::JackClientSocket::Read (this=0x773b80, data=<optimized out>, len=4)
    at ../posix/JackSocket.cpp:187
#3  0x00007ffff78cee7a in Jack::JackResult::Read (trans=0x773b80, this=0x7fffffffd060)
    at ../common/JackRequest.h:143
#4  Jack::JackClientOpenResult::Read (this=0x7fffffffd060, trans=0x773b80)
    at ../common/JackRequest.h:303
#5  0x00007ffff78cccbf in Jack::JackGenericClientChannel::ServerSyncCall (this=this@entry=0x76ec10, 
    req=req@entry=0x7fffffffd080, res=res@entry=0x7fffffffd060, result=result@entry=0x7fffffffd15c)
    at ../common/JackGenericClientChannel.cpp:68
#6  0x00007ffff78ccf0c in Jack::JackGenericClientChannel::ClientOpen (this=0x76ec10, 
    name=name@entry=0x7fffffffd160 "pure_data", pid=18276, uuid=uuid@entry=-1, 
    shared_engine=shared_engine@entry=0x7fffffffd150, shared_client=shared_client@entry=0x7fffffffd154, 
    shared_graph=0x7fffffffd158, result=0x7fffffffd15c) at ../common/JackGenericClientChannel.cpp:113
#7  0x00007ffff78df866 in Jack::JackLibClient::Open (this=0x76e970, 
    server_name=0x7ffff78e2410 "default", name=<optimized out>, uuid=-1, options=JackNoStartServer, 
    status=0x7fffffffd3fc) at ../common/JackLibClient.cpp:114
#8  0x00007ffff78e0b54 in jack_client_open_aux(const char *, jack_options_t, jack_status_t *, typedef __va_list_tag __va_list_tag *) (client_name=client_name@entry=0x77a300 "pure_data", 
    options=options@entry=JackNoStartServer, status=status@entry=0x7fffffffd3fc, 
    ap=ap@entry=0x7fffffffd2e0) at ../common/JackLibAPI.cpp:155
#9  0x00007ffff78e0e9a in jack_client_open (ext_client_name=0x77a300 "pure_data", 
    options=JackNoStartServer, status=0x7fffffffd3fc) at ../common/JackLibAPI.cpp:176
#10 0x00000000004d97c8 in jack_open_audio (inchans=2, outchans=2, rate=44100, callback=0x0)
    at s_audio_jack.c:339
#11 0x000000000048ae67 in sys_reopen_audio () at s_audio.c:452
#12 0x000000000048c005 in glob_audio_setapi (dummy=0x718538 <glob_pdobject>, f=5) at s_audio.c:1001
#13 0x0000000000485012 in pd_typedmess (x=0x718538 <glob_pdobject>, s=0x74fb50, argc=0, 
---Type <return> to continue, or q <return> to quit---
    argv=0x7fffffffd700) at m_class.c:949
#14 0x000000000047ee1f in binbuf_eval (x=0x755b40, target=0x718538 <glob_pdobject>, argc=0, argv=0x0)
    at m_binbuf.c:773
#15 0x000000000048f2be in socketreceiver_read (x=0x750d10, fd=44) at s_inter.c:594
#16 0x000000000048e5d2 in sys_domicrosleep (microsec=5000, pollem=1) at s_inter.c:223
#17 0x000000000048e659 in sys_microsleep (microsec=5000) at s_inter.c:247
#18 0x0000000000489b6e in m_pollingscheduler () at m_sched.c:541
#19 0x0000000000489caf in m_mainloop () at m_sched.c:596
#20 0x0000000000492582 in sys_main (argc=1, argv=0x7fffffffdb08) at s_main.c:365
#21 0x000000000048c578 in main (argc=1, argv=0x7fffffffdb08) at s_entry.c:32
```

umlaeute:
@simonvanderveldt hmm, I wouldn't expect this PR to actually fix any *linking* issue. instead it should fix the crash when weak-linking against libjack and then not having a full libjack.

I'm not sure what causes the crash in your setup, but it sounds pretty unrelated (please file a different issue)

simonvanderveldt:
> @simonvanderveldt hmm, I wouldn't expect this PR to actually fix any linking issues. Instead it should (only) fix the crash when weak-linking against libjack and then not having a full libjack.

@umlaeute You're right, it doesn't fix the linking, but it does fix the crash.
The only strange thing from a UX perspective is the Audio Settings dialog shows JACK as selected even though it actually isn't.
image

Looking at the code I did notice there's already a [real/actual call to `jack_client_open()`
](https://github.com/pure-data/pure-data/pull/93/files#diff-6b29bcc8fc427fa42604ec51dfdd4d93R339), maybe the separate check if `jack_client_open()` exists can be combined with that?

I'll create a separate issue for the segfault I'm seeing.

umlaeute:
regarding `jack_client_open()`: I'm not sure I understand.

In practice we have a hard dependency on that function if we want to use JACK (because we *must* register a jack-client).
If the function does not exist, we cannot use it in [line 339](https://github.com/pure-data/pure-data/blob/weak-jack/src/s_audio_jack.c#L339). So we must check for it's existence beforehand, which we do in [line 310](https://github.com/pure-data/pure-data/blob/weak-jack/src/s_audio_jack.c#L310).
We use this check as an early exit from the entire JACK backend (as it is useless without `jack_client_open()`). I don't know why it would be better to do the check later (which I think is what you suggest)

simonvanderveldt:
@umlaeute Makes sense, thanks for the explanation

Issue #100, opened 2017-07-12
Bug with the size of the subpatches


pierreguillot:
When a subpatch has his size specified (after resizing it for example), the size is well saved at the end of the **restore** line  (`#X restore 15 16 pd zaza 10, f 55;` for example). But when the parent patch is reloaded (from the file) this is the last object in the graph of the subpatch (an **[outlet]** for example) that is resized. Then if the patch is saved, the size is saved for this object (`#X obj 22 18 outlet, f 55;` for example) and ignored for the subpatch (`#X restore 15 16 pd zaza 10;` for example).

![ezgif-3-573ab5a01e](https://user-images.githubusercontent.com/1409918/28122825-a1d6f57c-671f-11e7-8cc8-4eaa0037d018.gif)
Behind, this is the patch in the text format. 
The video starts when the subpatch is resized.

pierreguillot:
Fixed

PR #99, opened 2017-07-11
Resize several objects together

pierreguillot:
- Allow several objects to be resized together based on the selection.
- The resizing doesn't use incrementation anymore but it computes the new size with the position at clicking - it offers a better precision and this is the only solution to resize several objets because the width is based on the number of characters. 
- Set the minimum width to 4 and the minimum height to  6 for the GOP otherwise it is impossible to select it anymore.
- Use shift to restrain the height of GOP by the width

The subject has been discussed here #82 (see the GIF to have an overview). It goes beyond the issue so if you prefer I can create another PR that only solve this specific issue.

Issue #82, opened 2017-07-03
Confusion between objects while resizing them


pierreguillot:
On MacOs with Pd 0.47.1

When I try to resize an object close to another one (but doesn't overlaps it visually), there is a confusion between the objects about which one must be resized. 
I think this happens when one is above another, when you want to resize the one at the bottom this is the one at the top that is resized. 
I guess the good object to select can be found easily because it is selected anyway - even if the wrong object is resized.
See the example with comments:

![giphy](https://user-images.githubusercontent.com/1409918/27783001-d2d07704-5fd4-11e7-9e45-00a6fe446427.gif)

danomatika:
Looks like even though the comments seems to be separate, there is a slight overlap and the first one receives the change probably because it's earlier in the graph order.

pierreguillot:
I tried again and you're right it's slight but there is an overlap ! Nevertheless, for me it's still a strange behavior because I really expect the 2nd comment to be resized. And it seems that the current width of the objects is not considered by the resizing operation. Here another example if it helps...
![giphy](https://user-images.githubusercontent.com/1409918/28007962-d0022a82-6555-11e7-9df2-1dbb7ff0633f.gif)

pierreguillot:
And what about resizing several objects depending on the selection (like Max)? 
![ezgif-3-573ab5a01e](https://user-images.githubusercontent.com/1409918/28064846-678c3a2e-6636-11e7-98a7-c72e77bda3e5.gif)
Here my [branch](https://github.com/pierreguillot/pure-data/tree/dev/obj-multi-resize) (I can create a PR if you want it)

Ant1r:
> I can create a PR if you want it

I'd say : please do it :-)

Spacechild1:
the issue at hand has been fixed with #99

pierreguillot:
In fact, the issue is not fixed. I will create another MR to fix it ASAP. If it's ok for you, please reopen the issue.

Spacechild1:
Are you sure? I've actually tried and couldn't reproduce the problem, so I assumed it has been fixed already. Can you double check?

pierreguillot:
You're right. Sorry, I get confused with the #254 I think.

PR #101, opened 2017-07-12
Fix the bug with the size of subpatches

pierreguillot:
The bug is explained in the issue #100.
I'm not sure that's the best approach, the loading of pd files is a bit opaque to me. 
I solved the problem by checking if the canvas is loading in `canvas_f()`. If not, it means that the size must be applied to this canvas and not to one of his objects.


umlaeute:
i think this doesn't work correctly if:
- the subpatch gets resized via the mouse (e.g. `, f 42`)
- and after that the subpatch gets GOP enabled.

the result will ignore the GOP-width and instead use the `, f` width.

~~~pd
#N canvas 6 59 759 309 10;
#N canvas 552 90 450 300 (subpatch) 0;
#X coords 0 -1 1 1 80 40 1 100 100;
#X restore 166 125 pd, f 50;
~~~

pierreguillot:
I think this is a second bug that already occurs on the master branch. Here is what I got on the master (tell me if it's a different problem):
ezgif com-video-to-gif

The only difference on my PR is that the the **f** width of the GOP is preserved after reopening the patch (that's not on the video due to size of the GIF). What I can try is to add another correction for this 2nd bug.

PS: Here is a patch will almost all the possibilities I suppose (GOP or not, resized with mouse or not)
[test-size-subpatch.pd.txt](https://github.com/pure-data/pure-data/files/1481539/test-size-subpatch.pd.txt)

pierreguillot:
After investigation, it seems that this second bug is caused by another part of the code (in **graph_getrect()** when **canvas_showtext()** is true). What is corrected by this actual PR is a problem of reading and interpreting the patch in its textual form (see #100). In my opinion, it would be better to correct this bug first in this specific PR and to correct the others with other PR because they are related to size checking etc. and not with the reading of the text file.

For this other PR: Here is another other bugs due to the size of objects and subpatches (#82). One solution is here (#99). Another bug is, for example, if you rewrite the name of the subpatch, the text can exceed the limits of the rectangle and you have to move the GOP to update the rectangle boundaries.

millerpuckette:
I think this is fixed now.

Issue #103, opened 2017-07-15
single quotes added to macOS guiprefs strings


danomatika:
I've been noticing that with the new GUI prefs, the recent files entries do not work after restarting Pd since they all have a single quote char suffix ala `some-patch.pd'`.

In digging into pd_guiprefs.tcl, I see embedded single quotes in the string are escaped properly but that the string is enclosed by single quotes when passed to the defaults command to update the macOS preferences plist. From what I can tell, Tcl treats single quotes as regular string characters so the single quotes are not actually treated as marking the beginning and ending of a string but as being added to the string itself aka this in Tcl:

    set path /path/to/some-patch.pd
    set quoted '$path'
    puts $quoted

prints:

    '/path/to/some-patch.pd'

So this string now contains the quotes and those are then added as the string into the defaults file. When Pd restarts and loads the recent files list, the single quotes are there and the ' suffix means the file path is invalid.

I'm digging into how to pass the single quotes along to exec so they are treated as actual string delimiters and not as part of the string. This is all down to the fact the defaults man page states that single quotes should be used, annoying.

PR #104, opened 2017-07-15
fixed guiprefs saving single quotes around string values on mac

danomatika:
fix for #103 

Invokes /bin/sh -c to run defaults write command as a string so single quotes are treated as expected by the shell instead of being concatenated by tcl.

### DIFF
--- a/tcl/pd_guiprefs.tcl
+++ b/tcl/pd_guiprefs.tcl
@@ -112,14 +112,17 @@ proc ::pd_guiprefs::init {} {
                 if {[catch {exec defaults write $adomain $akey -array} errorMsg]} {
                     ::pdwindow::error "write_config $akey: $errorMsg\n"
                 }
+                # invoke /bin/sh -c to pass command as a string, this ensures
+                # the shell interprets the single quotes around the value so the
+                # quotes don't end up being saved as part of the string itself
                 if {$arr} {
                     foreach filepath $data {
                         set escaped [escape_for_plist $filepath]
-                        exec defaults write $adomain $akey -array-add $escaped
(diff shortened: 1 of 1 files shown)


tcl/pd_guiprefs.tcl
```diff
@@ -112,14 +112,17 @@ proc ::pd_guiprefs::init {} {
...
                     ::pdwindow::error "write_config $akey: $errorMsg\n"
                 }
+                # invoke /bin/sh -c to pass command as a string, this ensures
+                # the shell interprets the single quotes around the value so the
+                # quotes don't end up being saved as part of the string itself
                 if {$arr} {
                     foreach filepath $data {
                         set escaped [escape_for_plist $filepath]
-                        exec defaults write $adomain $akey -array-add $escaped
+                        exec /bin/sh -c "defaults write $adomain $akey $escaped"
```
umlaeute: urgh. this seems to be easily exploitable....

danomatika: If you can come up with a better way to ensure the quoting is evaluated by the shell & not tcl, I'd like to know.

danomatika: In looking into this, we could use the -r option for [restricted mode](https://en.wikipedia.org/wiki/Restricted_shell).

danomatika: I just did a test and -rc works fine. I can update this if that makes sense. At least it's only used on 1 platform.

danomatika: I went ahead and added it. I tried a number of different quoting schemes but invoking the shell was the only one that worked. Definitely not ideal but was a suggestion I found on StackOverflow.

Issue #105, opened 2017-07-16
Asymmetric cos table


brbrofsvl:
The cos table in d_osc.c is constructed in one shot using cos() from math.h. This table is slightly asymmetric, which means it is unstable for FM using [osc~]. [tabosc4~] fares a little better with a similar table constructed in Pd due to cubic interpolation, but for real symmetry, the segment of the cosine table from phase 0 to pi/2 should be copied. See attached Pd patch for an illustration; I have a working d_osc.c modification that fixes the problem, which I can submit if desired.

(This issue was raised by Porres on the list some time before PdCon16, but I think it got lost among everything else)

[FM-instability.pd.zip](https://github.com/pure-data/pure-data/files/1150829/FM-instability.pd.zip)

porres:
the end of that discussion was that miller was willing ro fix this and keep a backwards compatibility flag for old behaviour, so I guess we can send this as a Pull Request

brbrofsvl:
Sure. I don't think I have permissions to push to this git repo, but I can attach the new source file here for someone else to.

porres:
well, it should be easy for anyone to fork and make the pull request online. I'm not an expert though

porres:
hi @brbrofsvl , can we make sure this also affects "sinesum", I'm not really being able to get the same behaviour with tabosc4~ right now, anyway, I mentioned this also in another issue here: https://github.com/pure-data/pure-data/issues/371

porres:
by the waym testing the above test patch, I don't see a bad result with the cos-BAD table!!!

brbrofsvl:
It is unstable on my machine, but it takes much longer to show, presumably because the interpolation is better. It's useable at least. Let it run for 15 minutes or so and you should see a much more extreme difference, or see what happens when you set the cos-BAD table with [cosinesum 64 0 1(. With a much larger table the effect all but goes away, but it does show up eventually so it could cause problems with e.g. around-the-clock installations.

porres:
I am closing this as it has been fixed by https://github.com/pure-data/pure-data/commit/8b013118170e81f7eef7efd998868908f589d35d

porres:
I am reopening it as it isn't really "simmetric" yet, at least not using perfect guard points of "1" and "0" and "-1" as discussed on the dev list

porres:
> (This issue was raised by Porres on the list some time before PdCon16, but I think it got lost among everything else)
> 
[...]

PR #106, opened 2017-07-16
Make cosine table symmetric

brbrofsvl:
This is a fix for issue #105.

It replaces the one-shot cosine table construction with one that copies the segment from 0 to pi/2, making FM using [osc~] stable. I did not include a backwards compatibility flag, but one could be added very easily.

### DIFF
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -205,7 +205,8 @@ static void cos_dsp(t_cos *x, t_signal **sp)
 static void cos_maketable(void)
 {
     int i;
-    float *fp, phase, phsinc = (2.0 * M_PI) / COSTABSIZE;
+    double phase, phsinc = (2.0 * M_PI) / COSTABSIZE;
+    float *fp;
     union tabfudge tf;
 
     if (cos_table) return;
@@ -213,7 +214,7 @@ static void cos_maketable(void)
 	phase = 0;
 	fp = cos_table;
 	for (i = COSTABSIZE/4; i--; fp++, phase += phsinc)
-		*fp = cos(phase);
+		*fp = (float)cos(phase);
 	*fp++ = 0.;
 	for (i = COSTABSIZE/4; i--; fp++)
 		*fp = -cos_table[i];
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -10,6 +10,9 @@
 
 #define UNITBIT32 1572864.  /* 3*2^19; bit 32 has place value 1 */
 
+#ifndef M_PI
+#define M_PI 3.14159265358979323846264338327950288
+#endif
 
 #if defined(__FreeBSD__) || defined(__APPLE__) || defined(__FreeBSD_kernel__) \
     || defined(__OpenBSD__)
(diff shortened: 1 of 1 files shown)


porres:
I dug up the discussion about this in the Pd list, which I had brought up by the way. With the way [osc~] is now, FM just can't be stable over time... that's a flaw that always bugged me, specially when teaching FM in Pd :/

So, as usual, there was a discussion about the concern of changing the behaviour of the oscillator. Miller's conclusion, as I quote, was _"I think in the case of this particular design flaw,
the correct thing to do is to correct it but provide back compatibility (as I
did earlier with hip~ and inlet~).  It's quite possible nobody will ever care but it
would be there in case someone did."_ 

here's the link to that response
https://lists.puredata.info/pipermail/pd-list/2015-11/112306.html

porres:
I also love the quote by IOhannes "hey, why doesn't Pd crash any more? my entire performance relied on that..."

millerpuckette:
will leave this til the next release - too many things are already going wrong!

porres:
like 0.48-1? :)

danomatika:
@millerpuckette something to consider for 0.49?

src/d_osc.c
```diff
@@ -202,15 +205,22 @@ static void cos_dsp(t_cos *x, t_signal **sp)
...
+    float *fp, phase, phsinc = (2.0 * M_PI) / COSTABSIZE;
     union tabfudge tf;
 
     if (cos_table) return;
-    cos_table = (float *)getbytes(sizeof(float) * (COSTABSIZE+1));
-    for (i = COSTABSIZE + 1, fp = cos_table, phase = 0; i--;
-        fp++, phase += phsinc)
-            *fp = cos(phase);
-
+     cos_table = (float *)getbytes(sizeof(float) * (COSTABSIZE+1));
```
umlaeute: hmm, could you please fix the indentation?

porres:
@millerpuckette replied about this one on the mail list with considerations once, but it's gonna be hard to find it.

porres:
well, here it is

"Claude Heiland has a much better cosine implementation that's faster and
more accurate - but I need to do some profiling to see how to compile it
so that it gets the speeds he gets.  Someone else could look at it I guess
but I'll certainly have to spend some time with it myself too.  So it's for
0.49.

cheers
Miller"

https://lists.puredata.info/pipermail/pd-list/2017-12/121085.html

porres:
For reference -> here...

https://mathr.co.uk/blog/2015-04-21_approximating_cosine.html

with an update here...

https://mathr.co.uk/blog/2016-07-20_approximating_cosine_update.html

danomatika:
This seems simple enough and could be setup to take a compatibility mode to choose between the old and new methods, ie. passing compatibility mode < 0.54 would use the old method.

sebshader:
Imo it makes sense to get this in until the polynomial version is worked out.
Hopefully there's a version that doesn't require a very specific set of compiler flags..

danomatika:
Yeah I think this is a no brainer after the math double-checked by Miller.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

porres:
I am closing this as it has been fixed by https://github.com/pure-data/pure-data/commit/8b013118170e81f7eef7efd998868908f589d35d

PR #107, opened 2017-07-17
mac app bundle README git usage clarification + template.pot version bump

danomatika:
* updated mac/README.txt based on feedback
* template.pot version was bumped, this gets picked up when using the auto tools build after the version string is changed in configure.ac


danomatika:
Since template.pot get's updated when there version is updated, maybe this step should be added to the release checklist? Basically have the regenerate the build files then make in po:

    ./autogen.sh
    ./configure
    make -f po

PR #108, opened 2017-07-18
bugfix for windowpositions

umlaeute:
this is a follow-up fix for the window position problem as resolved in #44 (but which was effectively undone by merging in the `osx-retina-support` branch).

the PR also includes a `.gitattributes` file to exclude all the various dotfiles from the *source package* (as created via `git archive`)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+# make sure that .gitignore, .travis.yml,... are not part of a
+# source-package generated via 'git archive'
+.git*      	export-ignore
+/.*		export-ignore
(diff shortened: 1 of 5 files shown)


danomatika:
Should be fine. I'm not sure how I broke it before, I guess it's a creation order issue? (as these things tend to be)

umlaeute:
yes.
> this [`attaching the menu to the pdwindow`] needs to happen *after* the main menu is created, otherwise the default Wish menu is not replaced by the custom Apple menu on OSX

this change somehow made sure that the window frame bias calculation happened before the menu was attached to the window, so the menu was not taken into account. unfortunately the menu adds the most bias to the window frames, so the window-position fix was for naught.

i just wanted to make sure (by requesting a review) that i didn't accidentally revert *that* part (but then, i refactored my fix a bit, and it seems that the relevant parts are not touched any more)

PR #109, opened 2017-07-19
removed menulist strings from translations

danomatika:
This PR removes the menu list string from the translations as it leads to a crash on startup when anything is actually changed. The individual entries are available for translation already.


danomatika:
Note: The commits in this PR are now pulled into the larger development PR #113

PR #111, opened 2017-07-19
fixed tcl error when using Undo and Redo menu entries

danomatika:
This fixes an error with the menu binds which were calling menu_undo and menu_redo with the focused window argument. The problem is that neither of these procs requires an argument.


danomatika:
Note: The commits in this PR are now pulled into the larger development PR #113

PR #112, opened 2017-07-19
fixed undo/redo calls from menu

umlaeute:
as reported on the pd-list, the "Undo" and "Redo" entries in the menu throw an error (Undo/Redo works fine via the accelerator keys).

That's because the `menu_undo` (and `menu_redo`) do not accept any arguments, but the menu-entries supply one.

this PR fixes this.


umlaeute:
i wonder though, why undo/redo use a global variable for the focused window, rather than getting the current window-id from the menu itself.

danomatika:
Hah, I beat you to it: #111

umlaeute:
hrmpf!

danomatika:
Maybe we should do another larger PR with just bug fixes. I can create one.

danomatika:
Ok, it's PR #113 and the branch is `update/feedback-fixes`

PR #113, opened 2017-07-19
0.48 test feedback bug fixes

danomatika:
This is a larger development branch / PR for feedback bug fixes from the 0.48.0 test1 release.

* [x] undo/redo menu tcl errors
* [x] window position issue on Linux
* [x] help browser double click not working
* [ ] prefs save/load not reloading dialogs?
* [ ] prefs paths prepending a + on close/reopen
* [x] spelling errors

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,5 @@
+# make sure that .gitignore, .travis.yml,... are not part of a
+# source-package generated via 'git archive'
+.git*      	export-ignore
+/.*		export-ignore
+/debian/	export-ignore
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ src/pd
 src/pd-watchdog
 src/pdreceive
 src/pdsend
+*.app
 
 # autotools
 autom4te.cache
--- a/configure.ac
+++ b/configure.ac
@@ -432,8 +432,8 @@ AS_IF([test x$oss = xyes], [midi_backends="OSS ${midi_backends}"])
 
 ##### ALSA #####
 AM_CONDITIONAL(ALSA, test x$alsa = xyes)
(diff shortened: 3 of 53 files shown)


danomatika:
I added a TODO list of what we have so far. Feel free to add new issues/check off others they are fixed in this PR.

danomatika:
I discovered that the system was automatically storing contextual info (open pane location, sizing, etc) to the default domain on macOS. I'm thinking we want to keep the core settings separate from any GUI specific settings like this, so I changed the app bundle domain to org.puredata.pd.pd-gui so the system uses the GUI only settings instead. This should have no discernible effect to users.

danomatika:
In looking into the Path dialog, adding a new Path, closing the window, & reopening prepends a + character to the path.

danomatika:
The + prepared is coming from saving the tcl $::sys_searchpath global in pd string encoding. This isn't really needed as the list can be encoded when using pdsend. I'm seeing the same issue with the listbox entries in the startup dialog ie. being (re)encoded each time the dialog is closed. I have a fix for this and it should be fine as long as other parts of the GUI don't rely on this encoding. From a what my search have pulled up, they don't.

We still need to get these values from pd when loading preferences.

danomatika:
Still missing one Todo: $::sys_search paths in tcl is not being populated from the core when loading preferences, AFAICT.

danomatika:
$::sys_searchpaths issue now fixed in #116

umlaeute:
jeez, i stumbled upon this in #120 and have been debugging it for the last 30 minutes.
only to find out it's already fixed...

danomatika:
Sorry :) Does that fix the -path issue? I was digging around yesterday and didn't;t find what was going on. Today I did a source text search for sys_searchpaths and that showed up...

umlaeute:
yes, it fixes the issue...

PR #115, opened 2017-07-19
translation fixes

umlaeute:
oops, sorry; it seems that the gui/updates branch got merged while i was busy fixing (German) translations.

this also includes some minor fixes of the translatable-strings so the engine can actually create meaningful transaltions (e.g. translate phrases rather than single words)


danomatika:
Merged with `update/test2-fixes`.

PR #116, opened 2017-07-19
test2 feedback fixes

danomatika:
This is a development branch for 0.48.0-test feedback fixes.

### DIFF
--- a/po/pt_br.po
+++ b/po/pt_br.po
@@ -276,7 +276,7 @@ msgstr "Dispositivo de saída"
 
 #, fuzzy
 msgid "(same as input device)..."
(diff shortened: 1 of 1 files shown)


umlaeute:
seems like you missed #115

danomatika:
@millerpuckette I fixed the path dialog issue, looks like an errant find/replace from the multi instance support udpates: 33f0f14

danomatika:
Ok, I've fixed all the major issues I had from the feedback. The only thing I can think of is the Linux+JACK crash report.

danomatika:
I am seeing an issue with the recent files/open recent menu not saving the actual open files... Looking into it now.

millerpuckette:
merged

PR #126, opened 2017-07-21
added es.po

fdch:
Hi everyone, here's a spanish version (following the steps of @porres, and @danomatika's recommendations on the list). 

I tested everything and it does seem to work fine. I'd suggest double-checking with spanish speaking users, just to be sure. I can make quick changes if anything, all done in OmegaT

Please, let me know if i messed something up or whatever. I couldn't find previous spanish versions.

PR #127, opened 2017-07-21
added es.po to update/test3-fixes

fdch:
Spanish GUI translation, with recent deken strings.
Hopefully i got it right this time (https://github.com/pure-data/pure-data/pull/126)


danomatika:
Looks good. I'm surprised we didn't have an ES translation already.

PR #133, opened 2017-07-22
[soundfiler] info right outlet

danomatika:
Scratched an itch...

This PR adds a right outlet to [soundfiler] that sends out an info list on a read or write:

* samplerate
* headersize 
* number of channels
* bytes per sample
* big or little endian ("b" or "l")

The order is meant to match the argument order to the write -raw flag. The help patch was also updated.

I've tested with both the [soundfiler] & [readsf~] help patches as well as with files of varying types and sample rates. This should be safe as it's mainly only accessing values which were already parsed when reading the files. The only part I needed to add was parsing the AIFF 80 bit samplerate to an int but that was found online easily.

The last commit is only refactoring to use the new t_soundfile_info struct to cleanup some of the overly long function argument lists. If you don't want that, you can cherry pick the other commits for the new functionality with more minimal code changes.

### DIFF
--- a/doc/5.reference/soundfiler-help.pd
+++ b/doc/5.reference/soundfiler-help.pd
@@ -1,4 +1,4 @@
-#N canvas 84 44 1102 576 12;
+#N canvas 87 38 1102 576 12;
 #N canvas 0 22 450 300 (subpatch) 0;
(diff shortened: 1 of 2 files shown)


porres:
Wow, thank you so much!

PR #136, opened 2017-07-24
Update expr-help.pd

porres:
removing the remark that atanh() and others were not working on windows, cause they in fact are now!

also added other Pd Vanilla objects related to expr in "see also" section


fdch:
Alex, thanks for this help file, it really brings expr into the surface for me, and opens up a whole other level of patching! THANKS!

PR #140, opened 2017-07-24
Set deken install path in the Path dialog

danomatika:
This PR adds an entry box to the Path dialog for the deken install path. A browse button lets you choose a new directory. This functionality is only added if the `::deken` namespace is is loaded.

This should eliminate the ambiguity as to where people might want to install things & puts this option into the "standard place" with the other paths.

~~Currently, the deken path is empty in a new install so this path shows nothing. That is normal behavior as deken, by default, installs to the ../extra dir for the first time and gives the user a prompt.~~

EDIT: We now show the default path if nothing is set, see updated screenshot

screen shot 2017-07-25 at 10 34 50 am

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -595,10 +595,19 @@ msgstr ""
 msgid "Pd search path for objects, help, fonts, and other files"
 msgstr ""
 
+msgid "Install externals to:"
+msgstr ""
+
+msgid "Browse"
+msgstr ""
+
 msgid "Use standard paths"
 msgstr ""
 
 msgid "Verbose"
+msgstr ""
+
+msgid "Install externals to directory:"
 msgstr ""
 
 msgid "Add new library"
@@ -647,9 +656,6 @@ msgstr ""
 
 #, tcl-format
 msgid "[deken] Platform detected: %s"
-msgstr ""
-
-msgid "Install libraries to directory:"
 msgstr ""
 
 msgid "Show all"
--- a/po/az.po
+++ b/po/az.po
@@ -591,10 +591,19 @@ msgstr ""
 msgid "Pd search path for objects, help, fonts, and other files"
 msgstr ""
 
+msgid "Install externals to:"
+msgstr ""
+
+msgid "Browse"
+msgstr ""
+
 msgid "Use standard paths"
 msgstr ""
 
 msgid "Verbose"
+msgstr ""
+
+msgid "Install externals to directory:"
 msgstr ""
 
 msgid "Add new library"
@@ -643,9 +652,6 @@ msgstr ""
 
 #, tcl-format
 msgid "[deken] Platform detected: %s"
-msgstr ""
-
-msgid "Install libraries to directory:"
 msgstr ""
 
 msgid "Show all"
--- a/po/be.po
+++ b/po/be.po
@@ -591,10 +591,19 @@ msgstr ""
 msgid "Pd search path for objects, help, fonts, and other files"
 msgstr ""
 
+msgid "Install externals to:"
+msgstr ""
+
+msgid "Browse"
+msgstr ""
+
 msgid "Use standard paths"
 msgstr ""
 
 msgid "Verbose"
+msgstr ""
+
+msgid "Install externals to directory:"
 msgstr ""
 
 msgid "Add new library"
@@ -614,7 +623,7 @@ msgstr ""
 
 #, fuzzy
 msgid "Help Browser"
-msgstr "Агляд.."
+msgstr "Агляд..."
 
 msgid "(Tcl) MISSING CLOSE-BRACE '}': "
 msgstr ""
@@ -643,9 +652,6 @@ msgstr ""
 
 #, tcl-format
 msgid "[deken] Platform detected: %s"
-msgstr ""
-
-msgid "Install libraries to directory:"
 msgstr ""
 
 msgid "Show all"
--- a/po/bg.po
+++ b/po/bg.po
@@ -592,10 +592,19 @@ msgstr ""
 msgid "Pd search path for objects, help, fonts, and other files"
 msgstr ""
 
+msgid "Install externals to:"
+msgstr ""
+
+msgid "Browse"
+msgstr ""
+
 msgid "Use standard paths"
 msgstr ""
 
 msgid "Verbose"
+msgstr ""
+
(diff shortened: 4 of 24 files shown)


porres:
Wow, this is so great, I'm so excited. This really makes installing externals much easier!!! I'll take it as it is :)

But... if I remember correctly, there were long discussions in the Pd list about where to install externals and wether if Pd or Deken would create the standard path folders for downloading (which I think makes sense).

> Currently, the deken path is empty in a new install so this path shows nothing. That is normal behavior as deken, by default, installs to the ../extra dir for the first time and gives the user a prompt.

So, apparently not. Seems deken will search for a valid path and prompt you for it if it finds it. For Mac, it does find the /extra folder if you got admin powers (which is what you have by default, if the machine is yours).

The problem is that in Windows this folder is not writable, so deken wasn't giving any valid option for download. The end of the discussion generated a fix where deken could prompt you for a folder, even if it doesn't exist (such as the standard user path). Then, if you agree to download there, this folder is created automatically, and things are downloaded there. But I never see this happening for mac since I have my powers to download to extra... anyway, here's the fix for that https://github.com/pure-data/deken/commit/8625a443917f25448b5b1199b0507f5fe8c59607

The point is, there was a caution to not automatically create any folder for you, unless you wanted to or agreed so. This is a great enhancement, but there was something missing. The current situation isn't ideal, as I believe the extra folder should not be the best option to install externals. This is "application specific", and means that it gets lost when you upgrade to a new Pd. Only advanced users who know what they're doing, and have multiple Pds would use that feature, not all users, who are mostly not savyy enough. 

> This functionality is only added if the ::deken namespace is is loaded.
> Currently, the deken path is empty in a new install so this path shows nothing. 

Yeah, so, I tried it, and that entry path is empty by default, because of the above.

My idea then is to set the user standard path as a default, for the first deken download, and that this path would show up in this entry out of the box. Since it is the desired feature for the current deken anyway, to prompt if users are ok with downloading to the user specific folder, and then creating the folder there if they say so...

But again, it's just wonderful enough as it is :)

thanks

danomatika:
As noted, there was a long discussion on this so I am hesitant to change the default path behavior. The fix you mention is what gives you the "../extra" folder as it's the first system search path.

All this does is display the current deken installer and let you change it. Default path locations etc can be done in a different PR.

porres:
> Default path locations etc can be done in a different PR.

fair enough, one step at a time ;)

danomatika:
>My idea then is to set the user standard path as a default, for the first deken download, and that this path would show up in this entry out of the box.

It does that, I believe.

porres:
> It does that, I believe.

well, I built it from this branch, and it was empty by default, only after downloading from deken is that it showed up. So I read through and was led to believe it was the desired behaviour, and now I'm just confused as usual.

danomatika:
This is the current behavior AFAICT:

* the path shown in the box is whatever path deken has saved to the guy preferences
* if the gui preferences are empty, the box will be empty
* once you choose a new path, the path should be saved
* if you do not choose a new path, I don't think anything is saved

Again, this is all baed on what deken does and should be figured out in another PR.

porres:
yeah, it does all that. By "out of the box" I meant "at first", when you first run it... so it is empty.

Anyway, I just love this as I said, any further discussion can wait.

thanks

danomatika:
Yeah, let's just say I don't want to fight IOhannes or Chris if I want this PR to be accepted...

fdch:
This is great, man! It makes getting externals much clearer. 
Maybe it makes sense to have a 'Find Externals' link/alias button on that dialog now as well as on the Help menu. But maybe this is adding too much.

porres:
> Yeah, let's just say I don't want to fight IOhannes or Chris if I want this PR to be accepted...

I know, and, anyway, it's kind of a separate issue altogether, sorry for getting excited and bringing it here.

Nonetheless, what I'm suggesting is something very simple and small. And it doesn't really go against what has already been recently done in deken, it kinda just makes it more consistent. Anyway, I opened an issue/request in the deken repository, hopefully we can discuss this over there and reach a conclusion: https://github.com/pure-data/deken/issues/156

cheers

chr15m:
@danomatika I haven't run this yet but the code looks good, idea is sound, screenshot looks good. If the path is blank I think it would be good to communicate the default to the user somehow as suggested by @porres or some other way.

danomatika:
I was thinking to populate it with the "system default" for deken. I have a small update to deken which adds a proc to return it. The issue now is I can't just drop in easy $::deken::installdir variable to the entry widget, so I think the better fix is to check if the installdir is empty when deken loads, if so, then set it to the system path, but don't save it to the guy prefs until the user explicitly does so.

danomatika:
Actually, the easiest solution is to use what deken already has: `::deken::prompt_installdir` The pop could appear if you open the path dialog and the installdir is empty. Maybe that's annoying, I'll check.

danomatika:
Ok this is fixed with the latest commit. I basically took the install path searching stuff and made it a proc we call after the plugin is loaded to set the $::deken::install path variable. This is also sued in it's original place, so there should be no difference unless there are parts of the code which expect $::deken::installpath to be an empty string.

Note: This does not *save* the installpath at startup. The guiprefs value is only saved on user interaction, as before.

porres:
great, now that we have this cool feature merged. Seems we need a new string to translate, I mean: "install externals to:".

Is it somewhere already/.

porres:
found it, sending a new translation update any minute

umlaeute:
Quoting Dan Wilcox <notifications@github.com>:

> @danomatika requested your review on: pure-data/pure-data#140 Set  
> deken install path in the Path dialog.

i've been starting a separate preferences page for deken (accessible  
via the deken window; but that's a different issue), that would allow  
to "set all things deken", including the default install path.
the idea was, that it would suggest all standard paths (with an option  
to create them, if they don't exist) in addition to the "let the user  
chose their own".

the deken prefs would also manage things like filtering out non-archs  
(completely), or automatically opening embedded READMEs.

i still think that this would be a preferrable solution.

however, don't expect an implementation before the 2nd half of august.

umlaeute:
cross-referencing some related issues (so this doesn't get lost)

https://github.com/pure-data/deken/issues/156
https://github.com/pure-data/pure-data/pull/139

PR #143, opened 2017-07-25
helpbrowser: col output now updates based on key navigation

danomatika:
This is a PR based on feedback from the 0.48 test releases. It mainly adds "live" interaction by refreshing the columns when navigating with the keyboard. This behavior is modeled on the macOS 
Finder column view (inherited from NextStep).

This builds on browser key bindings by Sebastian Shader: https://sourceforge.net/p/pure-data/patches/579/

Also, libdir display based on the current user search paths was brought back and exact path duplicates are now filtered. The core object help files no longer show in the root column as they can be found in Pure Data -> 5.reference.

Current TODOs:

* [x] update child col when moving from one col to the next with the right key
* [x] reintroduce user libdirs
* [x] filter *exact* path duplicates (keep warning for similar basenames)

Demo Video: <https://www.youtube.com/embed/2hL1brds1XA>

### DIFF
--- a/tcl/helpbrowser.tcl
+++ b/tcl/helpbrowser.tcl
@@ -28,11 +28,19 @@ proc ::helpbrowser::open_helpbrowser {} {
             .helpbrowser configure -menu $::dialog_menubar
         }
 
+        # FIXME wrap frame in a canvas with a horz scrollbar,
+        # currently we simply add cols to the left but only need 4 levels so far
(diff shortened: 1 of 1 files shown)


danomatika:
Ok, I think this is more or less ready.

danomatika:
Note: Test on windows to make sure the path separators are correct. I'm *pretty* sure they are by using [file normalize] and [file separator].

PR #149, opened 2017-07-28
keybinding bugfixes & cleanups

danomatika:
This PR addresses some keybinding issues & feedback from 0.48.0test5 (as well as other stuff I found):

* changed hsl bindings back to MOD-Shift-J, this fixes the MOD-Shift-H binding triggering the default "hide Pd" behavior on macOS

* cleaned up extraneous bindings (non-shift bindings don't need a double bind with uppercase, checking for TK 8.5 not really needed for most things), this was a vestige of my previous work updating things for TK 8.5+ on macOS & the upper case shift key bindings from IOhannes

* menu_print now uses the canvas window name + .ps extension for the default filename

* make sure find_dialog is raised on creation, this fixes Find sometimes opening behind a canvas window

* fixed find_dialog print, save, & saveas key binding pass through & update the menus to reflect this, also no longer set the print pass through bindings if the target search window is the Pd Window

* window focus is now shifted to target/next available after find window is closed

* show Pd window on macOS when the dock icon is clicked and there are no windows visible

Some TODOs ~remain~ and done:

* [x] opening a find window for a patch & then closing the patch leaves stale Find keybindings
* [x] Pd window saves log output is not fully escaped and contains the buffer Tcl list braces
* [x] find dialog sometimes moves & closes wth canvas window
* [x] Pd gui crash on occasional find creation after open/close cycle

### DIFF
--- a/tcl/apple_events.tcl
+++ b/tcl/apple_events.tcl
@@ -72,7 +72,7 @@ proc ::tk::mac::OpenApplication {args} {
 proc ::tk::mac::ReopenApplication {args} {
     ::pdwindow::verbose 1 "::tk::mac::ReopenApplication $args ++++++++++\n"
     # raise pdwindow if it's hidden and there are no other windows visible
-    if {![winfo viewable .pdwindow] && [llength [wm stackorder .]] == 0} {
+    if {[winfo exists .pdwindow] && ![winfo viewable .pdwindow] && [llength [wm stackorder .]] == 0} {
         ::pd_menucommands::menu_raise_pdwindow
     }
 }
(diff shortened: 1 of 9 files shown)


danomatika:
@millerpuckette Ok, this is pretty much done.

If you want to take the important fix for the release and leave the rest for later, cherry-pick the first commit.

millerpuckette:
Hmm.. I think the easiest thing woudl be for me to merge the whole hotdog and put out a 'test6' - will do that later this eve if I don't think better of it.

millerpuckette:
sweet!  I've always wished that worked a bit better.

danomatika:
I'm doing some testing with bleeding edge Tk 8.7 alpha on mac and the font positioning is good in the PS output now. However, some of the box sizing is not for longer object boxes. Oh well. At least the font is found.

danomatika:
I am seeing the "walking windows" issue Roman was talking about on the list with the bleeding edge Tk. Basically, whenever a dialog is opened, it's slighter lower and the right than before. Moving has no effect as it continues to "walk" relative to the previous pos on the next reopen. Weird. I saw this before with the buggy TK 8.5.9 on macOS but it seemed to be fixed in newer versions and now appears to be back in the latest form git.

PR #153, opened 2017-07-31
st_temppath for -path on commandline, matches -helppath behavior

danomatika:
This PR is a proposal to change the behavior of the -path command line argument.

Current behavior:

* `-path` adds a path to the st_searchpath user search paths variable
* paths added via the command line are then saved to settings
* requires making sure not to save and/or manually *remove* it each time, even when in startup flags (more likely occurrence)
* doesn't match `-helppath` behavior which is not saved

Proposed behavior:

* `-path` adds to a separate st_temppath variable for temporary/non-saved user search paths
* these paths are *not* saved to the settings
* requires explicitly adding and saving if the paths need to be in settings (less likely occurrence)
* matches `-helppath` behavior

My use case is running Pd from the command line with changeable paths set via shell scripts. As I am explicitly specifying a different path each time, I expect the given path to be added to the list of paths to search but *not* to be saved which could lead to lots of temporary paths being saved to settings.


millerpuckette:
Reasonable idea... I think this is an enhancement so would like to wait on it.

danomatika:

danomatika:
As a followup, I think this approach should apply to the command line flags which deal with settings as well. As I wrote on the pd-list:


>IMO the command line flags should *not* save to the settings, but override what's in the settings at start. That matches command line flag behavior in pretty much all command line programs I've ever used and I see Pd saving those flag settings automatically an anomaly.
>
>As stated in the PR, I vote we change that behavior and perhaps add a new flag like "-save" to allow saving the settings *with* the command line flags.

danomatika:

millerpuckette:
merged

PR #160, opened 2017-08-02
Added canvas_background option from pd-extended

rnkn:
This is a super simple PR to add the `canvas_fill` option pulled straight from Pd-extended.

Any chance to get this in 0.48??

### DIFF
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -234,7 +234,7 @@ set ::redo_action "no"
 set ::undo_toplevel "."
 
 # color scheme
-set ::canvas_fill "white"
+set ::canvas_background "white"
 
 namespace eval ::pdgui:: {
     variable scriptname [ file normalize [ info script ] ]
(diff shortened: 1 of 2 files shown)


millerpuckette:
I'm down to trying to fix bugs for 0.48, but can take this on for 0.49.  Is there a particular reason for the name "canvas_fill"?  Seems to me it should be "canvas_background" or even "canvas_backgroundcolor".  Doesn't matter much at first, but if we end up with a few more optins like this we'll want to name them in a systematic way.

rnkn:
It's literally just yanked from Pd-extended.... maybe it was chosen because the tcl `pack` command appears to take the flag `-fill`? Other than that I guess extended people would appreciate the consistency. But I am not invested in the name at all.

rnkn:
Yeah I think `canvas_background` makes more sense.

danomatika:
You could also make this PR to the feature/font-metrics branch where we could roll in various extended ports to vanilla. I could then rename to branch :)

rnkn:
@danomatika opened #161

danomatika:
Sweet. Definitely join in porting some of the visual stuff from extended into vanilla. I will rename the branch.

PR #165, opened 2017-08-04
[WIP] open dialogs over parent windows

danomatika:
This PR adds changes which ensure dialogs open over their parent windows.

This fixes bugs with newer versions of TK on macOS where dialogs move SE each time they are opened. This should probably be tested on other platforms.


danomatika:
This branch has been brought up to date. Still needs testing overall.

Ant1r:
Tested on Ubuntu16.04, seems working and pleasant.

PR #171, opened 2017-08-10
HTML manual documentation revamp

rnkn:
This PR replaces the existing HTML manual with a single HTML file, generated with [Pandoc](http://pandoc.org) from a new `pdmanual.md` file (which was generated from the existing HTML manual).

A render the HTML manual can be found here: https://htmlpreview.github.io/?https://github.com/rnkn/pure-data/blob/doc-polish/doc/1.manual/index.htm

The proposed future workflow for updating the HTML manual would therefore be to make changes to `pdmanual.md` and regenerated with Pandoc.

The command to convert `pdmanual.md -> index.htm` is:
~~~
$ pandoc pdmanual.md -r markdown-subscript-strikeout -s -S -N --toc -c pdmanual.css -o index.htm
~~~

(The `-f markdown-subscript` option disables the use of "~" for subscript formatting.)

### Advantages

- a single file
- sections can be linked by name, e.g. `(see [Audio and MIDI])` links to **3.1 Audio and MIDI**
- sections are automatically numbered
- table of contents is automatically generated
- smart typography
- easier to discern object names by wrapping with `` `backticks` ``
- (arguably) easier to keep up to date
- (arguably) easier to spot errors
- easier to convert to other formats
- can be rendered nicely on GitHub

### Changes

- a single file
- I fixed all sections headings to use title case
- given that sections can now be anchored by name and are automatically numbered, I replaced some links e.g. (see section 1.2 ) -> (see Other Resources); this has the advantage of never again needing to know/update the section number when updating the manual
- fixed a bunch of obvious errors/typos
- fixed the unclosed hyperlink in the release notes
- fixed Windows paths, which I'm pretty sure are `C:\Program Files` not `C:Program Files`

### Todo

- [x] remove unnecessary backslash escapes
- [x] changing lots of "objects" and "--flags" to `objects` and `--flags`
- [x] maybe some nicer CSS for e.g. code blocks, figure captions


rnkn:
And here's a PDF generated from `index.md` with:
~~~
$ pandoc index.md -s -N -S --toc -o manual.pdf
~~~

[manual.pdf](https://github.com/pure-data/pure-data/files/1216868/manual.pdf)

danomatika:
I would add a makefile target in Makefile.am to generate the html and/or pdf. This is how I generally use pandoc when writing papers (and my thesis).

Also, I would *highly* suggest adding a README.txt to the `doc` directory that includes information on how all of this works. Not everyone knows what pandoc is and when proposing an update like this, its an easier sell when everything is spelled out so *anyone* can easily understand and make changes themselves.. Look to the other subfolder README.txt files for examples.

danomatika:
I would also limit the TOC depth as the Release versions make the TOC excessively long.

danomatika:
Also, want to throw in a responsive css layout too? :)

millerpuckette:
I think it's too late in the 0.48 release cycle for me to make this change just yet, but I'm interested.  I need to know more about pandoc (how stable is it, for instance) before starting to rely on it.

danomatika:
I can say it's pretty stable when coupled with a minimal template. It's when you want to get more advanced with lots of customizations, you can run into edges cases. For this usage however, I think it would be pretty reliable.

millerpuckette:
cool... perhaps this is for 0.48-1 - I'm just at the point of putting out 0.48-0.

porres:
cool, I'm hoping to do more changes to the manual as well, to revise some typos, conflicting information and address the new GUI for path and externals - plus talk about 'deken'

cheers

rnkn:
Added minimally responsive CSS. (For some reason it's working in Firefox's responsive design mode but not Safari's...)

My thinking was that anyone building Pd should not be required to rebuild the manual (and thus not make Pandoc a dependency for building), rather, only those who want to edit the manual need to rebuild it. So the HTML manual ought to be built and checked in to git (and PDF manual if that's popular). Thoughts?

I've added a note to `1.introduction.txt` about editing the manual, but it might be worth renaming this file to `README.txt` to make it clear that this file contains info not just for users.

I can add to the `Makefile.am`... I just gotta get my head around how `Makefiles{.am,.in}` work first...

Agree that the long list of releases in the TOC is ugly, but in this case does function outweigh form, i.e. ability to skip to a specific release notes? Do people do this? (I had fun reading over Pd's history actually...)

I'll keep an eye on merges to the master `index.htm` and port these over to `pdmanual.md` for the time being; once there's agreement to go with this PR or not it would be awesome to direct new PRs that alter the manual to this PR. Checks will fail while there are newer commits to `x#.htm` since these are goners with this PR.

rnkn:
As for the stability of Pandoc, it has been around for a while, and I'd say is one of the most common tools for Markdown converting.

That said, I'm disabling a lot of extensions Pandoc makes to regular Markdown for the conversion process (e.g. the use of `~` for subscript or `$` for math, since these are relevant to Pd), therefore there should be no lock-in. Ideally building the manual should be achievable by any Markdown tool in widespread use.

danomatika:
>but in this case does function outweigh form, i.e. ability to skip to a specific release notes?

Yes.

>My thinking was that anyone building Pd should not be required to rebuild the manual (and thus not make Pandoc a dependency for building), rather, only those who want to edit the manual need to rebuild it.

Right. The generated HTML can live in the repo and only needs to be rerun now and then, so it doesn't need to be an automatic makefile target. Another, fancier, option is to *indeed* have an automatic generation which is run when the markdown file has been updated. This is how the translations are built. We can check for pandoc and only run the generation if it's installed.

>I can add to the Makefile.am... I just gotta get my head around how Makefiles{.am,.in} work first...

I some makefiles for working with pandoc I can share.

danomatika:
Actually, adding your existing pandoc command to doc/Makefile.am is the easiest way to start:

    manual:
    	pandoc pdmanual.md -r markdown-subscript-strikeout -s -S -N --toc -c pdmanual.css -o index.htm

This adds a "manual" target so the following can be run:

    cd doc
    make manual

When editing Makefile.am, you may need to regenerate the makefiles in the main pd source directory with:

    ./autogen.sh
    ./configure

rnkn:
Okay, added `make manual` and `make pdfmanual` targets to `doc/Makefile.am`. I'm a little shaky on make, but it seems solid. (It seems a good test of a Makefile is to call from another directory with the `-C` flag.)

rnkn:
Ideally I'd like it if the master Makefile imported all the targets... 😎

danomatika:
>Ideally I'd like it if the master Makefile imported all the targets

You can add them to the makefiles in the higher layers. See the bottom of the main Makefile.am "app" and "po" targets.

danomatika:
If you want to get fancy, here is how I would expand the target. (Note, this is not tested.)

~~~
MANUAL_FILES = pdmanual.md
MANUAL_HTML = index.htm
PANDOC_FLAGS = -r markdown-subscript-strikeout -s -S -N --toc -c pdmanual.css

manual: $(MANUAL_FILES)
	if command -v pandoc 2>&1 ; then \
		pandoc $(MANUAL_FILES) $(PANDOC_FLAGS) -o $(MANUAL_HTML) \
	else \
		@echo "pandoc not found, ignoring"
	end \
~~~

Notes:

* Separating the files & the flags make things cleaner and a bit easier for others to know where to edit without worrying about breaking the whole command as much.

* Adding $(MANUAL_FILES) after the initial target tells make to only run the target if *any* of the files have changed aka been edited and saved, etc. This cuts down on the amount of times Make needs to run things and removes redundancies.

* The shell script test with `command -v` checks if the "pandoc" command exists on the system so people who don't have it installed can get some information instead of errors messages.

danomatika:
Addendum: The MANUAL_FILES make variable can also be a list of markdown files, so you could separate the relevant sections into separate files.

umlaeute:
i've never seen the `command` command in my linux live. (i see that it is available on OSX; it's probably a standard BSD command)
afaik, the proper way to detect the presence of a command is to use `which`.

the more proper way would be to test for the program in `configure.ac` and make that part of the makefile optional.

danomatika:
I chose `command` as it's POSIX whereas `which` has system differences, at least as far SO is concerned: https://stackoverflow.com/questions/592620/check-if-a-program-exists-from-a-bash-script

umlaeute:
i see. for whatever reasons `command` did **not** work last night. it seems to work now.
consider my comment bogus.

rnkn:
@danomatika just to be crystal...
> > but in this case does function outweigh form, i.e. ability to skip to a specific release notes?
>
> Yes.

You agree that the releases ought to be included in the TOC, as the utility outweighs the aesthetic?

danomatika:
>as the utility outweighs the aesthetic?

It does. The TOC becomes ridiculously long. It's not likes it's hard to scroll once you get to the release notes section.

millerpuckette:
I would argue that utility and aesthetic are inseparable -- in this case, having a well-proportioned
TOC makes it easier to navigate.  This isn't the place for a long discussion about software design (!) but in another context I'll happily talk anyone's ears off about the more general question :)

rnkn:
Sorry for the radio silence on this. I'll fix up the makefiles soon. In the interim, it might be a good idea to merge this into a new branch in the repo, to allow others wanting to make changes to the manual to branch off that branch? This will prevent a lot of duplicate work.

Lucarda:
- Why the need of Pandoc to create a single file?
This can be made with a text editor and can be done easily.
- Why the need to keep the 6 html files if this has been done?
- Whats the benefit of one single file?

rnkn:
@Lucarda these questions are covered above. Hopefully this PR is merged to a Pd feature branch soon so you can base your work from it rather than double-handling.

Lucarda:
There might be some misunderstanding, I think Pandoc is not needed to create a single file. Also I don't think is necessary to have a `pdmanual.md`. I believe Pandoc is totally unnecessary.
But if others think there's a need include Pandoc in the building of a simple html (that's already there) I laugh. 
: - )

rnkn:
I’m always happy to amuse. As explained in the workflow above, the HTML file is not edited by hand. The Markdown file is edited and Pandoc is used to generate the HTML. The HTML is only included in the repository to ease the process of building Pd when not changing the manual (e.g. Pandoc is then not required). This is all in attempt to make editing the manual more friendly, both to the reader and the maintainers.

Lucarda:
To create a single html file you just copy everything inside `<body>` of the 6 files and then delete all ocurrencies of ::::  index.htm, x1.htm, x2.htm, x3.htm, x4.htm, x5.htm, x6.htm. (you do this once, if this get merged you don't do it again)

> The Markdown file is edited and Pandoc is used to generate the HTML.

Why would you generate a markdown file? Html is a markup language and maintainers can handle that without the need of any special software, and without the need to regenerate nothing.

Lucarda:
What is the benefit of editing by hand an `pdmanual.md` that then you have to regenerate?

We can edit by hand the html and that's all.

rnkn:
@Lucarda do you have anything constructive to add to the discussion or are you just trolling?

Lucarda:
I`m not trolling.
You learned how to join 6 htmls into one with just a text editor.
And if you can handle .md is most likely you can handle .html and vice versa.

PR #173, opened 2017-08-11
Update x3.htm (entry about path mechanism was wrong)

porres:
I'm updating this manual entry.

It was originally saying Pd **does not** use the "Path" mechanism to search for files and externals, if you use a slash declaration: "/" - such as "../sounds/sample1.wav" or "../cyclone/cycle~"

But it actually does use the Path, so I changed it.

### DIFF
--- a/doc/1.manual/x3.htm
+++ b/doc/1.manual/x3.htm
@@ -444,12 +444,12 @@ I think you just download the macOS driver and follow directions.
 audio devices to open, what font size to use, and so on.  Most of
 these may also be changed using the various dialogs you can open from Pd's
 menus.  Others take effect only when Pd starts up; some of these appear
-on the ``startup" dialog and some of them, too cranky to put in a GUI, must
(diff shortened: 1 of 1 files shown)


porres:
I'm also not sure about this bit 

_Pd does not automatically look in the current directory however; to enable that, include "."_

This seems to be wrong, as Pd will automatically look in the current directory regardless if you are including "." or not

rnkn:
How do you feel about #171 ?

porres:
are you asking me?

well, it's not up to me, and I'm not sure I get it...

first, is it ready to merge or not?

If Miller takes this changes before yours, you can just be sure and include them in your pull request as well

rnkn:
Yeah just your opinion. Considering you've taken the time to edit the HTML documentation, hypothetically would you have found it easier to edit Markdown?

It is ready to merge now; the incomplete things on the TODO list can be done after a merge.

rnkn:
Ported changes to #171

PR #175, opened 2017-08-11
expr mention in the release notes not accurate

porres:
hi, I removed this mentioning of the "value" object, because that had in fact already been mentioned before, see notes for 0.47-0 release.

"Shahrokh Yadegari has updated the
source and made manifold improvements in the objects.  Notably, they now allow
access to variables in Pd defined via the "value" object."

The problem is that it wasn't then that it was able to access variables defined via "value" objects. In 0.47-0, expr version was upgraded to version 0.5, and this feature was added in expr version 0.4, which seems to pretty old, but older than that.

Anyway, maybe not a big deal and just leave as it is, but I thought it was pretty weird to mention this feature twice, in close versions in the release notes, saying it hadn't been mentioned before, when it actually had.

cheers


rnkn:
Ported changes to #171

PR #183, opened 2017-08-15
Remove additional "standard paths" in conflict with Pd's documentation

porres:
I just commented out additional "standard paths" that are not the "extra" folder. There's a long discussion that led to this proposal of mine in an earlier Pull Request that got rejected - see https://github.com/pure-data/pure-data/pull/139

I wanted to add a better additional "standard path" than what we had (~/Documents/Pd besides ~/Library). As I understand, the rejection came from the notion that "Standard Path" should only refer to the "extra" folder and not anything else, but somehow these additional paths were added. Pd's manual actually makes it clear that the "standard path" is only the "extra" folder, and teaches you how to use the Path mechanism to add other "paths" to Pd. 

From the discussion in that earlier Pull Request, the ~/Documents/Pd folder has actually been included as a conventional location for externals, but not as a "Standard Path", reinforcing the notion that we should not have another "Standard Path" other than the "extra" folder.

Thus, in order to avoid confusion and better sort Pd's path mechanism, I propose we get rid of these additional "Standard Paths" that are not even mentioned in Pd's manual and in fact present themselves in a conflicting matter to Pd's Manual and all the reasoning behind earlier discussions that did not lead to add yet another "Standard Path" location.

### DIFF
--- a/src/s_path.c
+++ b/src/s_path.c
@@ -241,7 +241,8 @@ void sys_setextrapath(const char *p)
 #ifdef __gnu_linux__
     sys_expandpath("~/.local/lib/pd/extra/", pathbuf, MAXPDSTRING);
     STUFF->st_staticpath = namelist_append(0, pathbuf, 0);
-    sys_expandpath("~/pd-externals", pathbuf, MAXPDSTRING);
+#endif
+/*    sys_expandpath("~/pd-externals", pathbuf, MAXPDSTRING);
     STUFF->st_staticpath = namelist_append(STUFF->st_staticpath, pathbuf, 0);
     STUFF->st_staticpath = namelist_append(STUFF->st_staticpath,
         "/usr/local/lib/pd-externals", 0);


porres:
I can foresee some concern on making such changes, in respect to backwards compatibility and the like. But I don't think this "breaks" pd's mechanism, it just "fixes" it and sorts out the very confusing system we currently have.

Basically, we have a conscious design and decision on how this mechanism should work, but there is a breach on it. One that is not intentional, and one that goes against with what's currently in the manual. So I just say get rid of this breach and make it simpler and clearer.

People who still would like to keep using those paths as before, for some reason, will still have the liberty to do so. They'll just have to add it to Pd's path, or choose it instead of the location Pd currently suggests.

porres:
I'm also aware of a "user case" with the current state, and I tried using that as an argument so that we'd have ~/Documents/Pd as a "Standard Path", so I hope we don't go over that discussion here again. I can quote Miller's final word on it from this thread on the Pd list

_I always meant "standard path" to refer to "stuff distributed with Pd", and "Path" to be adjustable to point to libraries that are not part of the Pd distribution (...) I think the way forward is to somehow get the "path" mechanism to do what people are asking for._

see full message at: https://lists.puredata.info/pipermail/pd-list/2017-07/119839.html

 So, based on that, I'm not hoping to discuss changes and enhancements to the Pd path on this Pull Request, we can have another issue for that, so we don't loose the focus.

porres:
>  I'm not hoping to discuss changes and enhancements to the Pd path on this Pull Request, we can have another issue for that, so we don't loose the focus.

As far as user cases and relative discussions goes, I opened a new issue at https://github.com/pure-data/pure-data/issues/184 - which would be an addition that would completely free us from relying on the additional "Standard Paths"

megrimm:
yes +1...

 I'm a fan of minimizing and discarding unnecessary garbage... so removing any additional, unneeded, undocumented standard paths would be a nice clean up.

porres:
Anyway, the issue I opened ( see #184  ), which is a parallel issue to this PR, has just received a Pull request to fix it in https://github.com/pure-data/pure-data/pull/205 - so, if this is accepted, I can't see any other reason, parallel or not, to not fix this as this PR suggest.

umlaeute:
i see little value in removing things just because they are "undocumented".

the principle might be great if you have *documentation-driven development*, but the current model in Pd is not (it rather follows the (dubious) OS model of building documentation long after the fact).

porres:
> i see little value in removing things just because they are "undocumented".

ok, but that is not my point.

The fact is that in previous long discussions (which you may have missed) @millerpuckette  made it clear that Standard Path should only be the "extra" folder. I'm only referring the documentation to say that this is even clearly stated in the documentation as the "official" intent.

> the current model in Pd is (...) building documentation long after the fact

I'd be fine with updating the manual myself, I actually did point to some fixes for some issues it had. This is not a matter of conflict with the documentation, in which we would have to decide to either fix the documentation or fix the behaviour.

My initial motivation here was actually to update the manual and document this thing, and also improve its design and behaviour.  Like I said, I wanted to add a better additional "standard path" than what we had (~/Documents/Pd besides ~/Library). But the outcome is that it has been included as a conventional location for externals, but not as a "Standard Path", corroborating it as a design choice, that we should not have another "Standard Path" other than the "extra" folder.

Thus, if these extra "standard paths" are not intended, we should NOT document it... and we might as well just fix this. Anyway, I repeated myself a little, but I hope my point is clearer now.

cheers

porres:
I see now how my first subject title was confusing, hence I changed to **_Remove additional "standard paths" in conflict with Pd's documentation_**

porres:
Some of the concerns I raise here can now be found also in https://github.com/pure-data/pure-data/pull/320

porres:
I couldn't edit this PR's commits, so I opened a new one https://github.com/pure-data/pure-data/pull/321

Issue #181, opened 2017-08-15
Fat macOS app bundle


danomatika:
The osx-app.sh script only copies the minimal requirements into the Pd macOS app bundle: the binaries, help files, translations, etc. The source files are left out in favor of the headers only.

Recently, it's come up that the entire source tree *should* be included inside the app bundle: https://github.com/pure-data/pd-lib-builder/issues/33

A solution to this could be to add a "--fat" flag to the script which optionally copies in the full source tree.

PR #186, opened 2017-08-15
osx-app.sh: install headers and/or source tree

danomatika:
This PR is in response to #181.

The initial commit installs the headers to `Contents/Resources/src` instead of `Contents/Resource/include`. This should fix pdlibbuilder not finding them.

TODOs:
* [ ] add option to install full source tree?


umlaeute:
merged into #191

PR #187, opened 2017-08-16
removing [midiclkin] and updating documentation accordingly

porres:
When removing [midiclkin] from midi-help.pd, I also took the opportunity to add more details, cause I've seen people getting really confused and not really grasping the information. So this also makes it clearer.


danomatika:
Note: This should be handled in #214 as well.

danomatika:
#214 has been merged which contains this fix, so this PR is no longer needed.

porres:
awesome, will check it out!

I also did some changes to the help file I liked, maybe I'll propose a new help update with some of the things I did, where I added more details.

cheers

porres:
>  maybe I'll propose a new help update with some of the things I did, where I added more details.

and there we have it https://github.com/pure-data/pure-data/pull/263

PR #192, opened 2017-08-17
Untitled filename keyword

### LINKED ISSUE #156: Patch names beginning with "Untitled" always open saveas when saving
I know this is not much a priority but thought I would log issue anyway since maybe there is easy fix:

when creating a new doc and saving it without retitling it (ex. Untitled.pd) when the user "save" doc the "save as" dialog appears rather than just resaving "Untitled.pd"

I am not sure I am even explaining this correctly. I think this has been brought up on the list but considered a minor issue, which I agree with, but and issue none the less.

danomatika:
This PR adds an explicit untitled filename keyword "PDUNTITLED" which is less likely to be used as a normal patch name. This keyword is replaced by "Untitled" in the GUI patch window title and as the initial filename when using save as.

"Untitled" is currently matched for *any* patch which starts with that string.

This should fix #156 without removing the save as initialfile as in b6b19c7afbdba37701d38cd4cac98dc31fb116ea.

Also, this keyword is consolidated into one place in both the core and the GUI each.


umlaeute:
Quoting Dan Wilcox <notifications@github.com>:
> @danomatika requested your review on: pure-data/pure-data#192  
> Untitled filename keyword.

i wonder why you did not implement empty filename (rather than an  
"unlikely" (but possible) one like PDUNTITLED).
note: i'm currently "offline", so i can't do a proper assessment of  
the problem.

the "make sure that the translations don't contain spaces" is a bit  
disturbing, i'd rather have this fixed before inclusion.

danomatika:
>i wonder why you did not implement empty filename (rather than an
"unlikely" (but possible) one like PDUNTITLED).

The core needs something as it's expecting FILENAME-NUMBER when a new patch window is created. I didn't feel like changing that. You get an error if you send only -NUMBER.

>the "make sure that the translations don't contain spaces" is a bit
disturbing, i'd rather have this fixed before inclusion.

That is an important note as there was a previous note about possibly translating the "Untitled" string. With the changes, it would be possible to do this now but I'm pretty sure we would want to make sure the translation for the filename would be without spaces as it's a filename and not a simple label string. I did not add this to a gettext call as I also wanted to avoid matches to existing translation strings for "Untitled".

danomatika:
Also, I didn't consider changing the pd core new file message handling as who knows what relies on the current behavior...

danomatika:

danomatika:

millerpuckette:
I think it's now necessary to actually type out "Untitled" into the saved name to get the stupid behavior... if I'm right this isn't too important.  But perhaps I'm missing something...

danomatika:
I think the problem came from someone *wanting* to be able to use "Untitled" as the file name. This handles that by using PDUNTITLED as the initial string passed from core to gui.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:

danomatika:
@millerpuckette Also, The current behavior will match anything that *starts* with "Untitled", so "Untitled Piece 1.pd" will be matched as well. I merely changed the default to "PDUNTITLED" and added some checking to filter this name out and replace it with "Untitled" in the GUI. This way, the behavior is the same but opening a previously saved file starting with "Untitled" no longer treats it like a new, unsaved patch.

Of course, you could save a patch called "PDUNTITLED.pd" and run into the same issue, but I think that's much more unlikely.

danomatika:
Brought branch up to date.

danomatika:
It might be too late for this for 0.49, but I'll push it one more time :)

danomatika:
Actually, this might be more simply fixed by checking for "Untitled" string AND checking the length. If "Untitled" matches and the length is exactly 8, then at least filenames starting with "Untitled" would no longer be matched.

I will do this in the updates/0.50 branch at the very least.

PR #193, opened 2017-08-17
soundfiler read -info

### LINKED ISSUE #179: Request: query number of channels from [soundfiler] or something else
I need to get the number of channels to automatically load multi channel files into a set of tables according to the number of channels in the original sound file.

But maybe [soundfiler] is not the best place for this, as it would spit it out after the file has already been put into a table. If so, maybe the only way out is with another object that gives us info.

cheers

danomatika:
This PR adds an -info flag to the [soundfiler] read message which reads the header info without loading samples. Resizing works as well and the items read value is estimated based on the known header info.

There wasn't too much work to add this now that we already have the header info.

This should close #179.


porres:
Do we need an argument for the table when using the -info flag?

It would seem more intuitive to me that it wouldn't need it. But I guess that's needed because of the "read" message type, so maybe it would require it being another message method rather than a flag - which is what I thought you meant in the beginning anyway. An info message could also output the sample size without requiring -resize.

But I can live with things the way they are ;)

porres:
ohm be sure to set the help file to 
"updated for Pd version 0.48-1" ;)

danomatika:
You don't need the table argument, it does nothing unless -resize is specified.

porres:
> You don't need the table argument

I tried it, didn't work

screen shot 2017-08-18 at 05 09 03

> it does nothing unless -resize is specified.

what do you mean?

danomatika:
Ok, I'll check.

danomatika:
Fixed with last commit.

danomatika:
Working on printing usage when using `-info -resize` without table name.

danomatika:
Ok, fixed most issues I found and the -info flag returns the same info as when reading a file, ie. number of samples is truncated to table size when a table name is given without -resize. It also works without a table name (as it should) and there is a smarter check on the number of arguments.

danomatika:
Also, for testing sound files, you can print the header info using `soxi` if you have sox installed. TO install on macOS, use homebrew and

    brew install sox

The you can use `sox`:

    soxi doc/sound/bell.aiff

prints:

~~~

Input File     : 'doc/sound/bell.aiff'
Channels       : 1
Sample Rate    : 44100
Precision      : 16-bit
Duration       : 00:00:03.54 = 155944 samples = 265.211 CDDA sectors
File Size      : 312k
Bit Rate       : 706k
Sample Encoding: 16-bit Signed Integer PCM

~~~

The info print returns the same info for this file:

~~~
left outlet: 155944 (number of samples)
right outlet: 44100 124 1 2 b (info)
  * samplerate: 44100 samplerate 
  * header size in bytes: 124 (not important here)
  * number of channels: 1
  * bytes per sample: 2 -> 16 bit
  * byte order: b -> big endian (probably from an old Next system or mac)
~~~

porres:
All looks perfect now ;) - nailed it!

danomatika:
It was a good suggestion and fit within what I did the last time. I can also use query the info without actually loading, especially on mobile devices using libpd.

porres:
and I can update cyclone/buffer~ without having to create another object for cyclone ( sfinfo~ ) :D

danomatika:

danomatika:

danomatika:
I'll do another merge later.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
@millerpuckette One issue I see with the current implementation of this in the master branch, is that the entire file is read, even if we only want the header info.

In the future, I think it's easier for both of us if you post what changes you'd like to see in this PR, I can make some commits, then you can do a clean merge without having to do anything manually. You can see the [changes](https://github.com/pure-data/pure-data/pull/193/files) already and use that for discussion.

As it is now, I can make another PR as a followup from here.

danomatika:
As a follow up, I can see the positive aspects of reading the entire file being that reading a raw file and a file with bad header info would then provide the correct number of samples. The negatives would be the disk access involved with reading a (possibly large) file just to get the header info.

millerpuckette:
Idea: what if I just suppress reading the samples if there aren't any arrays specified?  Sounds like a
one-liner.  In fact,without any of this, one could have read onto a one-sample array (or perhaps even
with a zero requested sample count) so this migth be unnecessary to start with.  But that's a morning-after insight.

Larger issue - how can I most efficiently (and respectfully) deal with 40-ish mutually conflicting PRs?  I think that's one for pd-dev.  Some of the problems are probably stemming from my clumsy git-craft.

danomatika:
I will close this for now and, possibly, revisit in a different PR/commit.

porres:
> I will close this for now and, possibly revisit in a different PR/commit

so maybe not getting into 0.49 yet?

porres:
"If no array name is given, no samples are read but the info is provided anyway."

<3

umlaeute:
should the `feature/soundfile-info-only` branch be removed, now that this PR is closed?

danomatika:
yup

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

PR #162, opened 2017-08-02
[WIP] Pd-extended styling

danomatika:
This is a WIP/test PR for experimenting with porting the Pd-extended styling, font metrics, and object sizing into Pd vanilla.

So far, these changes are pulled from the extended source code by looking for meaningful differences with the pd vanilla source code in the "right places." Painful.

Note: The "Pd-extended styling" doesn't mean the coloring and filled object backgrounds, just that the coloring and styles become accessible. It is *not* my intention to change the default vanilla styling, but this is something that could become accessible via a plugin etc.

**Current Status: Not There Yet**

PR #196, opened 2017-08-19
add tcl color variables to pd

sebshader:
create tcl variables for colors in pd
colors in the tcl array ::pd_colors :

canvas_fill - color of the canvas (patch background)
 gop_box - color of the GOP rectangle (when editing GOP patches)
 obj_box_text - text in an object box
 msg_box_text - text in a message box
 comment
 selected - color of a selection
 obj_box_outline_broken - color of a "broken" object box (that failed to create)
 obj_box_outline
 msg_box_outline
 msg_box_fill - fill of a message box
 obj_box_fill - "                 "   object box
 signal_cord - color of signal cord and outline of signal inlets
 msg_cord - color of message cord and outline of message inlets
 msg_iolet - message inlet/outlet fill color
 signal_iolet - signal inlet/outlet fill color
 graph_outline - color for the outline of arrays and GOP patches in the parent patch
graph_text - color of the names of GOP patches in the parent patch
 selection_rectangle - selection rectangle color in edit mode
 txt_highlight - color text is highlighted in (the "background") when selected
array_name - color of array names
array_values - color of array contents
atom_box_fill - fill of gatoms (number box, symbol box)
atom_box_text - text of gatoms
atom_box_label - label of gatoms
atom_box_outline - outline of gatoms
text_window_fill - [text] window background
text_window_text - [text] window text color
text_window_highlight - [text] window highlight
text_window_cursor - [text] window cursor color
pdwindow_fill - background of post window
pdwindow_fatal_text - text color for fatal errors
pdwindow_fatal_highlight - highlight color for fatal errors
pdwindow_error_text - text color for errors
pdwindow_post_text - text color for posting
pdwindow_debug_text  - text color for verbose logs


danomatika:
Before you spend more tie on this, see #162

sebshader:
ah, I'm not comletely sure what this means but these are all of the colors/inlets/outlets stuff from Pd-extended and it's working I have a build of it here https://github.com/sebshader/pdnext/releases

danomatika:
Ok, I guess I'm saying this may sit a bit and possibly be integrated into that branch instead of master.

src/g_canvas.h
```diff
@@ -184,6 +184,8 @@ struct _glist
     unsigned int gl_willvis:1;      /* make me visible after loading */
     unsigned int gl_edit:1;         /* edit mode */
     unsigned int gl_isdeleting:1;   /* we're inside glist_delete -- hack! */
+    unsigned int gl_isselected:1;	/* SS: attempt at keeping graphs selected
```
umlaeute: this will most likely break binary (ABI) compatibility with GUI-externals

umlaeute (changes requested):
i think there are two groups of changes here
- simply replacing hardcoded colors with GUI-variables
- adding additional checks on the Pd-core side (inlet/outlet highlighting, ...)

the first group is pretty much a no-brainer (and i'd vote for inclusion, even for a minor release).
the second group is not so, so i'd downvote an inclusion before 0.49

the ABI compat issue by adding new fields to `_glist` is a no-go.

sebshader:
I fixed the binary compatibility. Should I take out all the inlet/outlet stuff and make a new PR with just the GUI-variables?

umlaeute:
On 2017-08-22 07:56, Seb Shader wrote:
> I fixed the binary compatibility. Should I take out all the inlet/outlet stuff and make a new PR with just the GUI-variables?

i think it would be best to have a new PR that consists of multiple
commits (separating the no-brainers from the changes that might break
things.
this would allow @millerpuckette to cherry-pick the commits he likes
(and leave out those he dislikes).

btw, you can also re-set your local branch (`git reset --mixed master`),
then create disentangled commits and finally do a force-push, thus
updating *this* PR with a new set of commits (without the old ones)

gmtf
IOhannes

sebshader:
so that one takes the inlet/outlet highlighting out but leaves the inlet/outlet colors... I changed the sizes of the outlets because they were too small to see on OS X but I'm not sure if it will be consistent across platforms

sebshader:
Cool, I put just the colors in the first commit. The second commit is an attempt to get the outlets the same size as the inlets so you can see the color. (In vanilla they are smaller on OS X). This isn't tested on anything other than OS X though..

sebshader:
updated to 0.49, should be ABI compatible and only colors

porres:
colors for which items? could we use this to design themes for Pd (including background color?

danomatika:
I think I'd like to see the individual color variables within some sort of parent object or namespace. I think Pd already has too many globals, so it would be good to avoid adding more.

umlaeute:
i only glanced at the code, and it seems that this makes objects non-transparent.

i'd very much prefer if objects could stay transparent if requested (of course i'm not opposed to *allowing* to set a non-transparent fill color; it's just that both should be possible)

sebshader:
Ok, I think I can figure out how to get the boxes transparent (by making the "fill" value empty). There's no alpha option for objects on the canvas unfortunately..
Should I put all of the tcl color variables in a tcl array? that's what Pd-L2Ork does.. but they obviously have to be accessible to plugins..

danomatika:
>Should I put all of the tcl color variables in a tcl array? that's what Pd-L2Ork does.. but they obviously have to be accessible to plugins..

Something like that or a hash. It can still be accessible, just not in the global namespace and/or reachable by plugins through a get/set function. Perhaps @umlaeute has a good suggestion for that.

sebshader:
>  It can still be accessible, just not in the global namespace and/or reachable by plugins through a >get/set function
wait, isn't that the whole point? so you can set the colors?

porres:
oh, fill colours, cool! I miss having a white fill against a background canvas like I did in extended :)

umlaeute:
@porres   filled object seriously break a number of my performances. so there's obviously two sides of the coin and some like it and some dislike it. so if we are going to change the behaviour, it should accomodate the need for both groups and not introduce a regression

@sebshader i think you misunderstood what @danomatika meant. afaict, hist statement should really read:
- colours should be accessible
- gui-plugins should be able to access the colours via some encapsulation mechanism
   - either via a setter/getter
   - and/or something else (e.g. a namespace,...)

danomatika:
Yes. My point, after having working on the GUI tcl files, is that there are far too many GLOBAL variables. As in, it's too easy to trample over them accidentally within plugins and far too hard to change them once they are set. If the colors are encapsulated within a pd name space and accessed, it would be possible to change things internally later on for whatever reason without breaking the plugins that would access it.

One of the problems with working on the GUI right now is the fact that I'm afraid to reorganize or change some things (for the better!) as it might end up breaking *whatever* random plugin or object relies on grabbing/setting a global variable or function call.

umlaeute:
here's a suggestion:

- how the colours are stored is an implementation detail (but should not pollute the *global* namespace)
- use a getter **for Pd** that returns the required colour.
- plugins *could* override the getter to change the colour handling (e.g. colour scheme per window).

~~~tcl
proc ::pdtk_canvas::getcolor {type {window 0} {
    # ...
    return "black"
}
~~~

and then use something like the following on the C-side:
~~~C
sys_vgui(".x%lx.c itemconfigure l%lx -fill [::pdtk_canvas::get_color signalcord_select .x%lx]\n",
    x, x->gl_editor->e_selectline_tag, x);
~~~

danomatika:
Oooh, that makes the tcl string even longer. I'm also thinking of how to make the GUI communication more generic as opposed to more TCL-ish. I would like to look into integrating the GUI messaging approach from Purr Data, so being less explicitly TCL like this is good.

danomatika:
I was thinking of something more like "::pd::color.cord", "::pd::color.outline", "pd::color.inlet_stroke", "pd::color.inlet_fill", etc. Although this still becomes more explicit TCL.

One reason I was wary to adopt this PR right away was the use of globals, as I've said, but also to enforce a paradigm that would make generic GUI message wrapping *hard* in the future. This might be a little off topic, but I would really like to be able to add GUI functionality to libpd without having to reverse engineer "too much" TCL. :)

umlaeute:
i'm totally for being less tcl, but i think it would require a more radical approach.

*on the long run* i want something like:

~~~tcl
proc ::pdtk_canvas::create {name canvas type posx posy objectname {args}} {
   # ...
}
proc ::pdtk_canvas::add_inlets {name inlettypes} {
   # ...
}
proc ::pdtk_canvas::connect {name1 outlet1 name2 inlet2} {
   # ...
}
~~~

a proc like ` ::pdtk_canvas::create` would then call `::pdtk_canvas::get_color signalcord_select $canvas` internally, and that command would no longer be sent over the wire (the Pd->GUI connection).

optimizing out a few tcl-calls in the Pd->GUI connection does not necessarily bring us closer to *not* having *any* tcl-calls on the wire, and might give away the opportunity to make something easily usable *later*.

danomatika:
Right! That looks great. That's similar to what I was thinking for libpd where the drawing details are almost completely up to whatever API is receiving the message. The args would be an atom list the receiver could walk over.

Again, I'm taking this off topic, but I hope everyone can agree we don't want to adopt something too quickly that limits us later on. Hence my desire for a technical discussion.

porres:
> if we are going to change the behaviour, it should accomodate the need for both groups and not introduce a regression

Sure! I was thinking about it as a non-default option

umlaeute:
@danomatika exactly. i made the `proc getcolor` proposal because i think it this gives us the most options for the future.

sebshader:
I guess the best solution is to have a getter for pd source to use and on the tcl side have the colors exist in a map anyways both in order to avoid a big switch statement and so that people that don't know much tcl can easily create color schemes without having to rewrite the function (and a big switch statement) every time (plus people probably wouldn't use the window-specific thing too much I'd imagine)

sebshader:
I ended up putting the colors in the global array pd_colors. There's only 1 "select" color, and gatoms are unchanged because I thought it might mess up GOP abstractions. kind of a cool random color get_color proc in doc/7.stuff/colors-plugin.txt

sebshader:
sorry for all the "force pushing", I think it's done now

oskude:
whats the status here?

from where extactly does (current) pd read the (linux) system colors from?

especially patch back/text color, cause i made all my (linux) system color "themes" dark, and pd apparently took some of them, but sadly left patch window text black:
image

umlaeute:
@oskude while your problem *seems* to be related to *this* issue, it is really not. please create a [new issue](/pure-data/pure-data/issues/new) (and I'll delete both your post and this/my response)

oskude:
ill create new ticket later. feel free to delete my comments here. (they dont contribute to the issue, sorry)

porres:
what's the status of this? Is it ok to be merged?

I want to ping @millerpuckette , to see if he checked this and considered it, since he went through all of the oldest PRs

cheers

umlaeute:
> what's the status of this? Is it ok to be merged?

the PR **cannot** be merged due to merge conflicts.

sebshader:
I'll try to update this in the next couple of days if people are interested..

tcl/pd-gui.tcl
```diff
@@ -233,6 +233,28 @@ array set redo_actions {}
...
+msg_box_outline "#ccc"
+msg_box_fill ""
+obj_box_fill ""
+signal_cord "#558"
+msg_cord "#121"
+msg_nlet "#fff"
+graph_outline "#000"
+signal_nlet "#558"
+selrect_color "#000"
+txt_hilight "#37e744"
```
danomatika, line 255: Should this be "txt_highlight" ?

sebshader: could be.. I forget where I got these names from, but signal_nlet could also be signal_inlet

danomatika: Maybe it's nitpick, but these are the kinds of details that get fixed *in stone* once they go in... :)

sebshader: true, though maybe names like msg_box_fill should remain (instead of message_box_fill?).. not sure about selrect_color either (the color of the selection box for clicking & dragging in edit mode)

sebshader: I think the names originally come from extended, so if people still have color configurations from way back they would be back-compatible. But it's been so long now I don't know if that should be a consideration

danomatika: What do you think @umlaeute?

Ant1r: I think "_nlet" is not a replacement for "_inlet", but is more an equivalent to something like "_xlet" (inlet or outlet).

umlaeute: - i would prefer consistency over compatibility with Pd-extended
- i agree with @danomatika that we should use proper spelling
- i think that `msg_` is a common enough prefix (no need for `message_`)
- use either `_xlet` or `_iolet` (but not `_nlet`)
- drop any `_color` suffix (e.g. `comment_color`) it's totally redundant since those vars all live in the `::pd_colors` array anyhow
- what's `txt_highlight` anyhow (as opposed to e.g. `msg_txt` and `obj_txt`)? shouldn't that be something like `msg_txt_highlight` instead?
- same for `dash_outline`, shouldn't that be `obj_box_outline_dash`; or even `obj_box_outline_broken` (that is: use a semantic approach. we want to specify the color for "not-created objects", rather than for "dashed lines")

sebshader: @umlaeute  sounds good to me, txt_highlight is the color that text is highlighted in. I wasn't sure what default to set it to so I went with the osx green system highlight but thinking now maybe blue is better. Right now it gets set to whatever the system highlight is when pd starts (on osx at least).

umlaeute: while i still don't understand what `txt_highlight` exactly means (is it the background colour of the selected text within an objbox if you click it?), t's definitely not *green* on my system (Debian).

if "Right now tcl/tk uses whatever the system highlight color is when pd starts", then i think this should stay the default.

porres:
I tried compiling this to check it out and I couldn't open it... it also compiles as "49". So I guess it might be because of conflicts... if theres's a working compiled version for macOS, please let me know... otherwise I'll wait until I can compile to test this and check it out.

tcl/pd-gui.tcl
sebshader: hmm I couldn't figure out how to get the system color from tcl so I guess I'll add a check to see if it's set and only pass the option to canvas if it is..

I don't think adding more cross-platform consistency in pd's behavior/appearance would be a negative though

danomatika: If `txt_hilight` is using the Tk `systemHighlight` or `systemHighlightText` colors, those are macOS-only, duding from [this](https://www.tcl.tk/man/tcl8.5/TkCmd/colors.htm).

sebshader: yes, it seems that systemHighlightText is also available on Winiows. I'm trying to peruse/decrypt the tk source to figure out what they do for other systems

sebshader:
@porres should be up to date now (to where it was a year ago), it compiled for me ok

tcl/pd-gui.tcl
sebshader: seems like they just use gray (#C3C3C3)

porres:
cool, got it running now!

So yeah, I see the looks of old Pd Extended / Purr Data style! Signal inlets are differentiated from control ones. 

I see objects gets a green highlight in the background when selected. Don't know why green and where it came from... doesn't look that great to me by the way.

Now, I got the idea and was expecting that the object would have a white background when placed in front of a background canvas, for instance... why did I assume that?

I also recall that there was an idea of letting the user be able to choose the style, but I can't see where anybody could set any of this this anywhere.

sebshader:
@porres odd, should be blue, (hex #00f) not green..
anyhow, as mentioned above I put the example color scheme I use (as well as a wacky random color hook) in doc/7.stuff/colors-plugin.txt. Change .txt to .tclw put it in your path and then you can edit the valuesv

porres:
it's quite greenish to me

Screen Shot 2020-04-10 at 21 06 09

porres:
>  Change .txt to .tcl, put it in your path and then you can edit the values

The final idea is that people will have to do this manually by editing .txt files and stuff?

sebshader:
@porres that's the text highlight color we were discussing above, should be changed in a few hours

porres:
> should be changed in a few hours

to the same way as in Vanilla? Cause I think that looks fine...

anyway, what about being able to have a white background when the object is in front of a colored GUI canvas object ([cnv])? Was that ever discussed or did I dream of it?

sebshader:
@porres once the variables are included/exposed it would be possible to make a gui in tcl or something to edit them, this has to be done first though.

personally, I'm content enough to edit .tcl files that I'm not gonna write a plugin for it, but it shouldn't be too difficult

I didn't want to change the number or symbol gui objects precisely bc they are used on canvases, as older patches won't display the way they were intended. You can put a small white canvas behind the box to get a similar effect. This way is more flexible because you can put whatever (color?) you want behind it. (and might be able to get the shape right with data structures too? idk). If someone that preferred transparent boxes opened a gui from someone that didn't they might not be able to see it well. Honestly that was was somewhat frustrating in the days of extended imo.

I think the only solution would be to make them into iemgui-style objects so you rould decide on a per-object basis

however, you can decide on fill for object and message boxes now

porres:
> You can put a small white canvas behind the box to get a similar effect.

Of course, I know I can do that, but that's a lot of work and I asked because I think an option of being able to set this color is useful. So, will this provide it or not? I don't get it, the answer was a bit confusing for me, sorry.

Simple question: Can the user set such color schemes? If it is not by default, then fine, but could we or not?

And I wonder if in the end we'll have an interface to set and load color themes. That seems necessary to me

sebshader:
@porres no, this doesn't change the number or symbol gui objects at all from how they are now. The problem is that they are used both for patching and for guis.

users can set color schemes by editing the variables in a .tcl file. (which is technically a plugin like how it was in extended)

porres:
I dont get it, I never asked about number or symbol gui (atom boxes?)

sebshader:
@porres I am unsure what you're asking then.. In the editor you can have a filled object box in front of a [cnv] but once that is a gop object you won't be able to see it anymore (like it is right now)

Lucarda:
These are the "blue" colors sampled from current `master` on Windows :

```
select_color "#0000ff"
txt_hilight "#3399ff"
``` 

Here shown used on this PR:

pd-colors-windows

sebshader:
ok @umlaeute I think all is good 2 go now

porres:
Cool, nice, it looks now just like in vanilla here on my macOS!

Screen Shot 2020-04-11 at 01 30 22

porres:
> @porres I am unsure what you're asking then.. In the editor you can have a filled object or message box in front of a [cnv] but once that is a gop object you won't be able to see it anymore (like it is right now)

I'm afraid I don't understand your answers either.

All I'm asking is if we can have a filled color for the object so it is not transparent, and so it can be placed in front of a [cnv] object with a white filled color, for instance. This is somethin we could do in Pd Extended and I miss it. See screenshot.

Screen Shot 2020-04-11 at 01 34 35


Now, it seems that this PR allows this somehow. But it is not clear to me how yet. I guess I can dig in and learn it, sure, but I'm also asking if we could have a nice interface to set all these colors and design themes that people can share and load. It could be part of the "preferences". That's all.

I know objects disappear in a GOP, I don't know why that was brought up, I never mentioned 'GOP'.

and even though I didn't mention atom boxes, it's nice you did, cause I can see now that Pd Extended has this grayed filled in color for them, while in Pd they're transparent, so I can ask if we can also choose colors for them.

Anyway, thanks a lot for your work on this one. It looks really nice and promising.

porres:
Oh, actually, by comparing Extended to this, the filled signal inlets/outlets seem to have a different color instead of the gray one from Extended.

sebshader:
@porres if you put this in your path you`ll get the behavior I think you want.
[colors-plugin.zip](https://github.com/pure-data/pure-data/files/4464080/colors-plugin.zip)

porres:
it did, for objects! Cool... I can still ask you about filled colors for atom boxes.

And anyway, yeah, I was assuming this was possible, but I'm still asking if there's any desire of allowing an interface to set these colors, save them as a preferences file and load them?

sebshader:
@porres to be honest, personally I don't think it's worth the energy for the added benefit..

porres:
ok, from a user perspective, I have the personal opinion that this is a lot of energy for no real benefit if we're not letting people set themes in an interface and easily load them.

sebshader:
I feel that if users open the example file and change the hex colors and names it's not the most user-unfriendly thing.. you don't need to know how to script, all you really need is a color picker that gives hex values..

I just updated the helpfile so people can see how to set one color per line

Lucarda:
@sebshader  when compiling with autotools `colors-plugin.txt` does not get copied in `outputdir/doc/7.stuff/`

I think you must add it to the list on `pure-data/doc/Makefile.am`

porres:
> all you really need is a color picker that gives hex values..

yeah, and what providing all that? that's my point, offering an interface so people don't have to "hack it"... but you really really have to see this under the perspective of a user, otherwise it'll always seem quite trivial and silly, but it isn't

sebshader:
@Lucarda ah, thanks! hopefully it is now

Lucarda:
@sebshader  is just me that I'm not sure so I ask:

deleting this lines from `pd-gui.tcl`:

```
namespace eval ::pdgui:: {
    variable scriptname [ file normalize [ info script ] ]
}
```

was intentional or by mistake?

umlaeute:
@sebshader I think it is plain wrong to assume that the highlight colour on Debian (or Windows, or macOS, ...) has any specific value. so you don't need to try to reverse-engineer tcl/tk.

Pd should use the system defaults *unless* the user choses to override them.

i think that applying your patch **must not change anything in the appearance of Pd**.
everything should stay exactly as it is. if some user has tweaked their X11 and/or tk settings yo display patches in a weird way, then that should continue to work as is.

what this PR is supposed to do is *add the **possibility** to change the colours*, rather than changing them.

sebshader:
@Lucarda whoops! not even sure how that happened.. another good catch

umlaeute:
and to add on my previous comment:

 i think, all the talk about missing colour pickers and the like is totally besides the point.
once wr have the infrastructure for user-defined colour schemes, it's *trivial* to create a gui-plugin that exposes this in a user-friendly way.


once the infrastructure is there, you might want to add a feature-request to add this gui-plugin to pd-vanilla.
but that's another feature.


meanwhile I think it is pointless and counterproductive to fill the PR-discussion with such requests, as it makes the discussion hard to read up (eg when considering whether to merge or not).

please dont forget that this is an *issue tracker*, not a forum, chat or mailinglist.

porres:
makes sense to me:
- to have a PR that offers an infrastructure to allow color change as a first step;
- that no colors should change by default as a result of merging

porres:
so, again, how about changing colors of atom boxes, can we do that?

sebshader:
@umlaeute ok.. I just thought that someone opening pd for the first time could use the distinction between message and signal i/olets but changing it back is simple..

On the subject of system defaults: I have the system defaults for osx and windows "plugged in". The wish binary has no inherent awareness of an X11/gtk etc. system highlight, so the only practical way  make it aware would be through tcl or some dynamic library anyways. It seems to me that it's worth avoiding bloat for a primary distro at the expense of an extremely rare case that might require a couple lines of code on their end. (plus, after making txt_highlight accessible it should be way easier to plug system theming in that isn't natively supported by tk.)

tldr: from what I can tell this is hard-coded into tk anyways and if that ever changes we could easily change it as well. So mimicking what tk does seems better than some hacky solution imo.

danomatika:
>all the talk about missing colour pickers and the like is totally besides the point.

Yes. Let's get the colors in there first and in a accessible way so that adding a UI layer to edit them will be easy. So far, I think Seb has done that.

>i think that applying your patch must not change anything in the appearance of Pd.
everything should stay exactly as it is. if some user has tweaked their X11 and/or tk settings yo display patches in a weird way, then that should continue to work as is.

This a million times over. You'd be surprised what tiny details people seem to rely on. In general, I stick to the defaults to avoid this, so I have to remind myself that many people heavily customize.

>and even though I didn't mention atom boxes, it's nice you did, cause I can see now that Pd Extended has this grayed filled in color for them, while in Pd they're transparent, so I can ask if we can also choose colors for them.

Object background fill should be selectable, in my opinion, as there are definite uses to the Pd-extended style:
* option to set object background fill: boolean fill or no fill
* option to set object background fill color: color

I think atom boxes fall into this category as well, so maybe they need a boolean fill option to?

umlaeute:
>> if some user has tweaked their [...]
>> then that should continue to work as is.
>
>This a million times over.

not sure I understand what you mean here (is it: "yes, these use cases are important" or rather "there's too many tweaks people could have applied to cater for all of them"?)

> I think atom boxes fall into this category as well, so maybe they need
> a boolean fill option to?

how about: if the fill-color is set to a non-empty string, this should be used to fill. if the fill-color is unset/empty, no filling is done

danomatika:
> > if some user has tweaked their [...] > then that should continue to work as is. This a million times over.
> not sure I understand what you mean here (is it: "yes, these use cases are important" or rather "there's too many tweaks people could have applied to cater for all of them"?)

"then that should continue to work as is" -> "yes, these use cases are important"

danomatika:
>how about: if the fill-color is set to a non-empty string, this should be used to fill. if the fill-color is unset/empty, no filling is done

I'm rethinking the boolean option.

Does Tk have a concept of a "clear" color? If so, the simplest option is to set the background color to clear, otherwise I'm not sure where putting the logic will go as we are still sending raw Tcl strings. Ideally, we would send a "draw an object here" message to the GUI, but we are not there yet...

sebshader:
most canvas objects are transparent if the fill is given an empty string.

if every user has different gatom fills then GOP patches will change when distributed depending on what the user's preferences are.. that's why if any color is used for them it should be decided per-object like the other gui objects imo.

umlaeute:
> On the subject of system defaults: I have the system defaults for osx and windows "plugged in". 

no. afaics, you are mimicking the system defaults. why?
honestly i don't understand why you think you **must** set the `txt_highlight` color in all cases (because you only need to reverse engineer the actually used color if you must set it at all cost and want to appear it as if you didn't set it in the first place).

that is: rather than having something like https://github.com/sebshader/pure-data/blob/9ed3bb318b8fa81a9fc79998f4b25665718d6edb/tcl/pdtk_canvas.tcl#L121-L122

why notuse this  instead?

~~~tcl
    canvas $tkcanvas -width $width -height $height \
        -highlightthickness 0 -scrollregion [list 0 0 $width $height] \
        -xscrollcommand "$mytoplevel.xscroll set" \
        -yscrollcommand "$mytoplevel.yscroll set"

    set tmpcol [::pdtk_canvas::get_color txt_highlight $mytoplevel]
    if {$tmpcol ne ""} {
        $tkcanvas configure -selectbackground $tmpcol
    }
    set tmpcol [::pdtk_canvas::get_color canvas_fill $mytoplevel]
    if {$tmpcol ne ""} {
        $tkcanvas configure -background $tmpcol
    }
~~~

if `txt_highlight` as empty (as should be the default), you don't even touch the `-selectbackground` property of the created canvas.

sebshader:
@umlaeute That's not too unreasonable, I guess I just didn't get why we care so much about hardcoding a default

So now it looks like pd by default as well.
I think that's everything unless we want to tackle gatoms this PR (personally I think that should be a seperate/bigger project)

umlaeute:
ah well, some more nitpicks (in case you feel bored):
- i haven't yet figured out the order of appearance of the colours (i haven't looked to hard though)
- consistent spelling of colours (e.g. all-uppercase hexcolours)
- fix the accidental deletion of `::pdgui::scriptname` by amending the original commit and force-push (avoiding commits like fee6a3d which mingle unrelated things)

porres:
> I think that's everything unless we want to tackle gatoms this PR (personally I think that should be a seperate/bigger project)

Why? I guess all we could add is the atom boxes, then this is done. 

What else could we do in a "bigger" later project? I guess it would be a next step on how to edit, save and load themes.

porres:
>  I guess all we could add is the atom boxes, then this is done.

maybe there's also one other thing we could think about, which is the background and text/font color of Pd's terminal window (and perhaps also color of error messages for something other than red, but that might be too much) . For reference, here's a theme from Purr Data that changes the color of both canvases and the terminal window. This one is particularly useful for patching in the dark caves in the middle of the night.

Screen Shot 2020-04-11 at 21 28 17

sebshader:
@porres idk if this is the forum to discuss it but like I said before the issue is that 2 people opening the same GOP abstraction will see different things, which seems less trivial to solve design-wise.

perhaps priority could be given to an individually editable color like for other guis, and for non-GOP patchers a user preference would take over as default, whereas black could take over as default for GOPs

at any rate the design and implementation of that would be a big enough to do as a separate project

porres:
we're discussing wether we should include filled colors for atom boxes in this PR, so yeah, this is the place for it. 

> if every user has different gatom fills then GOP patches will change

> 2 people opening the same GOP abstraction will see different things,
> which seems less trivial to solve design-wise.

I don't get your issue with GOP at all. What's special about GOP? Why do you mention specifically it and not a regular patch canvas without GOP?

Anyway, yeah, if every user has different fills and colors for whatever, like objects, background, messages, etc, same patch will look different in different systems, sure. So what?

Like, Pd Extended used to have a different fill for atom boxes. That's the same wether in GOP or not... a patch opened in Extended looks different than in Vanilla.

Anyway, I just don't get it, so please explain.

sebshader:
@porres because when we create a GOP abstraction in a library we want it to look the same no matter who opens it. (at least I do)

even in the case where users can only change the background color consider this situation:
user A makes a GOP abstraction with number boxes on a black [cnv]. user A's preferences make a gray fill for the number boxes, and everything looks good.

user B dowloads user A's library and opens the GOP abstraction. Because user B has no fill set, user B cannot see the number boxes at all against the black [cnv]

porres:
What you say is regardless of 'GOP' and abstractions. So I'll just forget about the specificity of 'GOP', as I don't see one. Any shared patches, be it an abstraction with GOP or not, will be impacted by what the user has set its preferences to. Like you said, help files (which are not GOP abstractions) can also be affected if you change the color of comments... 

Yeah, it is what it is, so what?

If you want to open patches from others, maybe don't go too crazy on your personal color scheme. No one will be able to know how crazy someone else might have gone in setting up their system, so what?

And if you care about sharing your stuff, it's a good reasoning to assume people will be using Pd's default Vanilla style. It's not a good idea to design your abstraction in a fancy personal color scheme and expect people to be able to load it worry free in another one.

What you seem to want is the ability to expand Pd's GUI items such as comments and atom boxes to have individual colors, and I can actually see a good point in that, but I don't think it is necessary.

If one wants to design a nice interface with a number box with a background color, there's [nbx] for that... for symbols with background colors, one can also use the label on a [cnv]. There are alternatives.

As for an option for comments with colors, you can use externals like cyclone/comment for that too...

sebshader:
@porres the difference is that in a GOP almost all of the objects that are drawn have individually settable colors. Comments (which are rarely used) and gatoms (which are used extensively) are the only visible things that don't have individually settable color. And when created all of these will have (and should have) the same color no matter the user variables (except for comments now, but like I say those are rare.)
If message and obiect boxes were visible in GOPs those would be an issue also (but they aren't) 

I guess I specifically mention GOPs bc I don't feel like it's that common for people to distribute patches with GUIs other than in GOP abstractions + the fact that they don't draw objects/messages

sebshader:
I guess I could implement it if everyone thinks it's a good idea.. just adding blanket colors wouldn't be that difficult. I could also do the array-values-line if we're doing stuff that will go on other objects.. maybe there should be a warning about compatibility in some file tho

umlaeute:
Am 12. April 2020 05:58:55 MESZ schrieb Seb Shader <notifications@github.com>:
> I guess I could implement it if everyone thinks it's a good idea.. 

i agree that this should go in a separate PR.

porres:
> I guess I specifically mention GOPs bc I don't feel like it's that common for people to distribute patches

But it is... and this is my point. And I'm not only talking about help files. People share patches all of the time. Have you seen Pd's patch repository for instance? Tons of patches for sharing. I also have a tutorial, with almost 400 Pd files that I share...

And I guess my point is also: you can't let people change colors and then worry "what if people change colors?". Yeah, it's risky, that's why there's a standard Vanilla color setting that will keep the same as the default, and users change at their own risk. And also, like I said, there are ways to avoid this ([nbx], [cnv)].

Moreover, comments are shown in GOP, and this PR already allows its color to be changed. Why is it ok to do that and not for atom boxes? There's no logic. You're just making assumptions and basing on supposedly hierarchy so it seems "ok" (comments are less common in GOP, it's more common not to share patches).

> i agree that this should go in a separate PR.

What exactly? And why?

porres:
> I could also do the array-values-line if we're doing stuff that will go on other objects.

Now that's something I'd agree it'd be good if you could just set the array line color as a parameter for each one. And this would be a separate idea/PR. This would allow one to put many arrays in a single GOP and give them different colors. 

Setting atom boxes and comments separately wouldn't be that bad, I just think that maybe is not worth the hassle. 

Anyway, I don't know, maybe a separate PR makes sense here because we cannot seem to reach an agreement on what to do and this discussion might go on forever.

Spacechild1:
by @sebshader 
> I don't feel like it's that common for people to distribute patches with GUIs other than in GOP abstractions

I have to agree with @porres here! Many user facing patches - besides help patches - have GUIs outside of a GOP, so the issue is certainly not specific to GOP patches.

But I also see that it's more of a problem with GOPs, because when I want to use someone else's GOP abstractions in my own patch, it might force me to use a certain color scheme. So ideally, GOP abstractions should be made so that they are not affected by the color scheme at all. This boils down to:

a) set the background explicitly, e.g. with a [cnv]
b) don't use non-IEM GUI objects

Maybe we could introduce an easier way to set the background color of a canvas and save it in the patch, e.g. something like a [color( message. If not set, the canvas would just have the default color. This would be an independent PR, of course, and would work independently of your color scheme PR.

Anyway, if someone sets his/her global background color to black, existing patches might become unusuable, but then again the user can either fix the patches or choose another color scheme... Don't know a better solution, atm.

---

> Now that's something I'd agree it'd be good if you could just set the array line color as a parameter for each one.

https://github.com/pure-data/pure-data/pull/711

sebshader:
@Spacechild1 well hopefully by keeping gatoms black this will keep all kinds of GUI usable for the moment

> Maybe we could introduce an easier way to set the background color of a canvas and save it in the patch, e.g. something like a [color( message. If not set, the canvas would just have the default color. This would be an independent PR, of course, and would work independently of your color scheme PR.

In this PR the assumption is that it's the user's responsibility to set a color scheme where everything is visible for them while patching. This means that gatoms (and everything else) not on a [cnv] will be seen against the color of canvas_fill, which the user will presumably have set to something able to see all objects against (and I don't think it's wrong for GOP to reflect the color values of the user's patcher when not on [cnv]). So the issue is mainly when theres a combo of something from a user defined color variable on top of a [cnv] that has consistent/settable color. (and GUIs with no [cnv] should still reflect the user's colors for their editor imo)

Spacechild1:
> So the issue is mainly when theres a combo of something from a user defined color variable on top of a [cnv], which has consistent/settable color. 

Yes, that's tricky. But the problem can also occur the other way round: IEM GUI objects colliding with a user defined background color. This is especially problematic with [cnv] used for displaying text (without a background rectangle). If the label color is the same as the user defined canvas background color, then the label effectively becomes invisible :-)

I think this is a fundamental problem, but it shouldn't affect your PR. For now, it's simply the user's responsibility to set the colors so that they don't collide with other people's patches and GOP abstractions. In the long run, widespread use of color theming might simply change how people design patches and abstractions.

porres:
> But I also see that it's more of a problem with GOPs, because when I want to use someone else's GOP abstractions in my own patch, it might force me to use a certain color scheme. 

Ok, using other people's GOP abstractions might be more common, but as I see it, an argument is only valid if it considers 100% of the cases, so even if 99.9% of the time something happens, not accounting for the 0.01% case ruins it. Changing colors will always be a problem. And using any other patch or loading any help file might force you to change the color scheme.

The only solution that I see to avoid any issue altogether is by being able to save in the patch ALL colors, so each patch has a color scheme. But this is something else, and totally kills the point of letting a user set its own color theme.

> hopefully by keeping gatoms black this will keep all kinds of GUI usable for the moment

"Hopefully" isn't good enough. This can and will be ruined eventually. There's no escape. Here's an example of a patch with two different color schemes (I used Purr Data, not this PR, only because it was easier, but it's the same thing).

Vanilla color theme:
Screen Shot 2020-04-12 at 13 13 08


Inverted Vanilla color theme: 
Screen Shot 2020-04-12 at 13 13 26

This is not a GOP, but could be. Now, imagine I could deal with this but also setting a black background to atom boxes. That would clearly solve the issue. But you're not allowing this because you are trying to avoid issues. Not only you're not avoiding issues, you're preventing people to fix issues.

And allowing people to set color for comments, which are visible in GOP, which can also ruin the design, but not allowing a fill color for atom boxes makes no sense. You're not solving or avoiding any issue.

> Yes, that's tricky. But the problem can also occur the other way round

Exactly. This is what I've been emphasizing, changing colors will affect things one way or another... 

> it's simply the user's responsibility to set the colors so that they don't collide with other people's patches and GOP abstractions. 

Of course, that's simply it, this PR and people that makes patches and abstractions to share cannot be held responsible for how some other user wants to change color settings. 

If you change the color setting, you should be aware something funny might happen if you're opening other people's patches or sharing your own, there's just no escape. And making the cut at "atom boxes cannot have a fill color" is pointless.

sebshader:
@porres your example is clearly using a white symbol box so that wouldn't occur w/ this pr

porres:
Anyway, this is also why I've been talking about a later addition to this PR => https://github.com/pure-data/pure-data/issues/931 <=  which is a way to easily save and load color themes. In my idea, the interface would have a few most common presets (Vanilla (default), Extended, inverted vanilla) and people would also design user defined presets that could be saved in a preferences file.

As such, people can  share patches and their color schemes. And when loading patches and abstractions from others (made in vanilla theme, for example), people could easily switch between themes. And deals deals with the issues around.

> it's simply the user's responsibility to set the colors so that they don't collide with other people's patches and GOP abstractions.

Or... if a user likes his own wacky theme and likes to stick with it, he should know it might ruin and collide with other people's patches, which then also makes his/her responsibility to edit abstractions and patches so they match to that so beloved, beautiful, artistic masterpiece color theme.

porres:
> @porres your example is clearly using a white symbol box, so that wouldn't occur w/ this pr

Oh, so symbols aren't touched at all by this PR, sorry! But no problem, I can find something else then. What if you want a color theme that has a black background for an inverted like theme. That means I can't do that? Cause atom boxes are always black? That's no good then... an inverted like theme is what makes me most excited about this PR... 

When trying to avoid issues, you're now avoiding people to have the liberty to explore and set their color schemes...

And anyway, comments can have colors set by this PR, and the same issue would arise, as we've already discussed...

it's all very arbitrary and doesn't really cover all cases, so I can't see the point.

porres:
> an inverted like theme is what makes me most excited about this PR...

A Pd extended like theme is also impossible by the way. And I know MANY MANY MANY people would  REALLY REALLY REALLY love to have that...

Spacechild1:
I think it's very easy: every GUI object which can't be configured explicitly (i.e. all IEM GUI objects) must be themeable.

@sebshader can you update/expand the PR description, so people can immediately get a notion of what the PR does without having to look at the actual code?

Spacechild1:
@sebshader just to be sure: not only the title but also the actual description ;-)

porres:
> I think it's very easy: every GUI object which can't be configured explicitly on the patch level must be themeable.

yup, makes sense, and I now realize arrays aren't affected either, right? So what else could be missing besides atom boxes and arrays?

What I'd also like to add is background color and text color of Pd's terminal window.

To be clear, when I say array, it's basically just the graph outline and array name. As for the line of the array, yup, we can leave as it is in favor of https://github.com/pure-data/pure-data/pull/711, which I was pretty excited about but forgot about :/

sebshader:
@Spacechild1 what do you mean? the first comment? this pr has changed a lot since it was proposed (originally had in/outlet highlighting.)

What I'm thinking is the best solution is to add individually settable colors for gatoms (and maybe comments if I can figure it out), and if no color is set then to display according to the user tcl variable/theme.

Which I am happy to do but I agree that it should go into a separate PR bc at this point I think we should get this merged, as the gatom stuff will be different/separate implementation-wise (like making a color picker for gatoms in the editor} band this PR already has lots of changes.

sebshader:
@porres the name and outline can be set by graph_outline which is also used for GOP object outlines (pd considers both a "graph"

Spacechild1:
> the first comment?

yes

> this pr has changed a lot since it was proposed (originally had in/outlet highlighting.)

exactly ;-) people shouldn't have to read through the whole thread or the code to figure what the PR is about. The PR description should tell people what the code does. Here's a good example: https://github.com/pure-data/pure-data/pull/577

Spacechild1:
> I'm thinking the best/simplest solution would be to add individually settable colors for gatoms (and maybe comments if I can figure it out), and if no color is set then to display according to the user tcl variable/theme.

by individually you mean per-object? I don't think this is a good idea. gatoms and comments should be themeable. We already have [nbx] and [cnv], which can be configured per object basis. What's probably missing is a configurable symbol box, similarly to [nbx].

gatoms and comments are for patching. If I choose a black background, I probably also want to set gatoms and comments to white. Of course, I want to do this *globally*, not on an per-object basis. This would be quite annoying.

Spacechild1:
I still stand by my assertion:

> every GUI object which can't be configured explicitly on the patch level must be themeable.

And I don't think the consequence is to add color pickers for comments and gatoms, for the reasons stated above.

Spacechild1:
@sebshader thanks for updating! :-) AFAICS, the only thing missing is the gatom. Please at least consider it. Otherwise, the PR is very cool!

porres:
> What's probably missing is a configurable symbol box, similarly to [nbx].

agreed: https://github.com/pure-data/pure-data/issues/932

Spacechild1:
ah, and problably also default value for garray vertices. this could be later overridden with https://github.com/pure-data/pure-data/pull/711, but for now I think we have to be able to theme it, otherwise things like dark themes are not possible... @porres already mentioned it and I think he has a point.

sebshader:
@Spacechild1 
> gatoms and comments are for patching. If I choose a black background, I probably also want to set gatoms and comments to white. Of course, I want to do this globally, not on an per-object basis. This would be quite annoying.

You could do it globally. All gatoms without set color would still display according to your global variables/settings. but people creating guis/helpfiles could also set the color individually (like iemguis) to override this in order to have the patch appear how the creator intended. But all gatoms with unset color would display per your global settings.

Spacechild1:
@sebshader ok, this sounds reasonable. But then it would be cool to add the global colors *now* and the individual colors *later*. Then at least we can create fully functional color schemes.

sebshader:
@Spacechild1 splitting it up like that would work.. I guess I've been convinced by you all. I'll try to add it by later today

porres:
> and problably also default value for garray vertices. this could be later overridden with #711, but for now I think we have to be able to theme it, otherwise things like dark themes are not possible...

good point, theme this and then override it later if needed.

And again, for a proper dark theme, being able to set the color of Pd's terminal window and its text is also needed. Any objection in that?

sebshader:
tis done. In addition to colors for atom box fill, label, outline and text I also added color for the name of a GOP patch in its parent and for the name of an array
Screen Shot 2020-04-12 at 3 51 23 PM

Spacechild1:
this is great! would you mind tackling the default garray plot color as well? Then we would have everything we need for dark modes, afaict *). I can also give it a shot, if you don't have time.

*) @porres also mentioned the console window. it would be nice, but it's not super important, I guess.

porres:
>  @porres also mentioned the console window. it would be nice, but it's not super important, I guess.

At least for something like a dark theme, needed mostly for patching in the dark in the middle of the night (as I do very often), whenever you need the Pd window around, it pretty much ruins the whole idea and point of the dark theme with that shiny and terribly bright rays from the completely white canvas piercing your retina.

porres:
So, I tried last commit and I see you did catch the color settings of [text define]. And well, it also works for the [qlist] text window, so I assume this is a global thing. In your provided example, it has a black background and white text. 

This looks good and is a good example of a dark theme. We never thought of this and it wasn't requested, but it's a nice catch I think!

I'm not sure if you were trying to set Pd's terminal window with this one, which I see it still is unaffected, and that one was suggested. Anyway, if we're also doing the background and text color of [text define], and I think we should, I can't see why not do the terminal window. 

Please let me know if you're about to do that too.

thanks

sebshader:
@porres I'm looking into it, I'm not all that knowledgable about tk's options/database (I don't know if I even have to be)

I succeeded in setting the background, but another thing I'm not sure about is how to deal with is the different log levels (for instance one of them seems to use 20 different shades of gray by using the tk names gray1, gray2 etc.)

sebshader:
@porres alright I think I've got that sorted too
Screen Shot 2020-04-13 at 3 56 27 AM

sebshader:
@umlaeute I had to add a tag to the array name (in g_graph), hope that's alright

porres:
> @porres alright I think I've got that sorted too

yeah, tested it, looks great, thanks

> (I just left the "all" log level of gray-dient alone)

I don't know what you mean, but fine I guess? :)

Now, I tested arrays, and the color only works for points, if you set the display for polygon or belzier curve, it goes back to hardcoded black!

danomatika:
> > @porres alright I think I've got that sorted too
> 
> yeah, tested it, looks great, thanks
> 
> > (I just left the "all" log level of gray-dient alone)
> 
> I don't know what you mean, but fine I guess? :)
> 
> Now, I tested arrays, and the color only works for points, if you set the display for polygon or belzier curve, it goes back to hardcoded black!

danomatika:
> > @porres alright I think I've got that sorted too
> 
> yeah, tested it, looks great, thanks
> 
> > (I just left the "all" log level of gray-dient alone)
> 
> I don't know what you mean, but fine I guess? :)

In the main Pd window, select Log: 4. Notice that not all log lines are the same color (black by default), but also shades of gray, depending on their level (verbose, debug, error, etc).

sebshader:
@porres yeah I discovered that last night... I'm unfamiliar with that part of the code (how garrays interact with scalar, template, plot classes) so it might take a bit  to come up w/ a fix.

It seems like there are some other objects that use plot colors (use plot as a member class) and those colors are stored as a number in a feilddesc.. anyway I have to learn about that stuff

Spacechild1:
@sebshader this PR of mine might be helpful: https://github.com/pure-data/pure-data/pull/711 If you have any question, just ask!

sebshader:
@Spacechild1 Thanks, I did look at that to do the points colors, it was helpful (I wouldn't have known where to begin). But I think the issue now is knowing when/how to use the tcl variables instead of the value in the template. (And I think your PR sets the color of the data structure in pd as well)

Spacechild1:
Yes. that is indeed tricky. After all, garrays are really data structure arrays.

Maybe instead of trying to change the plot color, we could add an option to set the garray background fill color? For a dark theme the user might choose a white fill color (instead of a transparent background), so the actual plot can stay black.

Another trick could be to set the "color" variable in the `float-array` template to a special reserved value, which is recognized by `numbertocolor` (used in `curve_vis`), so it won't write a hex color but a Tcl function call.

sebshader:
@Spacechild1 
>Another trick could be to set the "color" variable in the float-array template to a special reserved value, which is recognized by numbertocolor (used in curve_vis), so it won't write a hex color but a Tcl function call.
thank you, this seems the most promising..

Spacechild1:
BTW, instead of choosing a reserved float value, it might be better to make the color field in the `float-array` template a *symbol*. `numbertocolor` could then become something like `fieldtocolor` and a symbol "default" would create the Tcl function call.

This could later even officially documented, so people can do data structure array plots with the default color.

I'm actually planning to make a PR where all data structure colors can be alternatively specified as symbol fields, so we can set them as hexcolors instead of the low resolution int colors.

porres:
> so we can set them as hexcolors instead of the low resolution int colors.

YES PLEASE!!!!!!

sebshader:
@Spacechild1 I feel like being able to set hex colors is more of an all-datastructures issue for a separate PR so I'll use the number trick. I think negative numbers are a good choice since they are already simply clipped to 0 right now, (black) which will be the default color.

Spacechild1:
> I feel like being able to set hex colors is more of an all-datastructures issue for a separate PR

certainly, that's why I've said I want to make a PR for that ;-)

The reason why I've thought about using a symbol is that some users could have relied on the fact that negative values are always clipped to black. But that's probably a) very rare and b) only manifests when tweaking the color scheme (because the default is black anyway).

So I guess you're right that negative floats are good enough for now.

sebshader:
ok that should be pretty much every color (except for array list view) ;-)

porres:
cool, havent's tested yet. Anyway, feels like there's all kind of liberties here. One can set a different background color for Pd' Terminal window, for patch background, for [text define] background. For help browser backgeround. And maybe that's ok, but wouldn't it also make sense to group these as a single background color? Same goes for select color. Seems it makes sense that all selections are the same color.

sebshader:
@porres I agree that data structures should have the same selection color as other things in the patcher (I changed it)

but different things (like text window, console, and helpbrowser) should definitely be able to be colored differently imo

porres:
> it would be easy to group all e.g. background colors together in other tcl plugins in order to set them as a group

yeah, I thought so... and I just realized how an interface to set and load color themes can be a tcl plugin external like the completion plugin (D'OH)

Spacechild1:
> well, it costs t_float worth of stack if the compiler doesn't catch it :-)

Any modern compiler will, otherwise it can go straight to the trash bin :-) Anyway, thanks for the great work! I really hope this gets included.

porres:
> Anyway, thanks for the great work!
+1000

tcl/helpbrowser.tcl
```diff
@@ -268,10 +278,15 @@ proc ::helpbrowser::make_doclistbox {dir count {select true}} {
...
-                             -height 20 -width 24 -exportselection 0 -bd 0]
+        -yscrollcommand "$b-scroll set" -highlightbackground white \
+        -highlightthickness 5 -highlightcolor white -selectborderwidth 0 \
+        -height 20 -width 24 -exportselection 0 -bd 0 \
+        -background [::pdtk_canvas::get_color helpbrowser_fill .helpbrowser] \
+        -foreground [::pdtk_canvas::get_color helpbrowser_text .helpbrowser]]
+    set tmpcol [::pdtk_canvas::get_color helpbrowser_highlight .helpbrowser]
+    if {$tmpcol ne ""} {
+		$current_listbox configure -selectbackground $tmpcol
+	}
```
umlaeute, line 289: why aren't the `-background` and `-foreground` properties made optional (like `-selectbackground`)?

sebshader: @umlaute I suppose I was again working under the assumption that we wanted consistent background and foreground color (black text on white background) for the helpbrowser

sebshader: @umlaeute I'm testing vanilla on dark theme osx mojave with tk 8.6.10. The default canvas is almost black! and I can barely see objects.. similarly, text in a text_define window is white on a white background (invisible)! it seems to me that this is an example of why the pdtk_canvas should draw white by default, and other colors should have a default too except for maybe for using the system highlight.. maybe this could be fixed through options in pd-gui if we really want the canvas subject to system themes but then we have to color many other things like object boxes according to system themes as well to make it work.

umlaeute:
i'm sure i cannot follow.

in any case: rather than forcing the use of the colour when creating the GUI even if it is unset, you can make it optional but define a non-empty colour in the standard colour-array.

this way you get what you want, but people can use system-colours by unsetting those they don't like.

if "we" decide that we want to keep the old behaviour rather than use consistent colours across all platforms, it would've a one-line change.

and I'm still favouring this PR, to implement a new (infrastructure) feature rather than "also fix bugs on the way".
at the very least those bug-fixe commits should be clearly separated from the feature commits (but I may just be anal)

sebshader:
@umlaute ok that makes sense, I was just thinking that if someone's using "" they could just as easily use whatever variable corresponds to their system theme and we could avoid checks for most people that don't want that.

Anyway, I've set everything to be black-on-white by default (in the global color array) except for highlight color. And there are checks in helpbrowser, text_define window and canvas to use colors only if they're non-empty.

sebshader:
@umlaeute is there anything else this needs? Was it ok to define DEF_ARRAY_COLOR to be -1 in g_array.c to use as a plot color "hack"? (so that in the plot code in g_template.c the color from the tcl variable is used if fielddesc_getfloat is < 0)?

porres:
I see this still has "changes requested", what is still in line? Or is it the review?

sebshader:
@umlaeute do you think you might have a chance to re-review? I think I made the changes requested

porres:
@sebshader , can you resolve the conflcts?

sebshader:
@porres these conflicts are pretty recent I think. Unfortunately I'm having issues with my laptop charger right now so it will probably be an least a week before I can resolve.
It might be better to include this pr earlier in the release in the future imo.

porres:
> It might be better to include this pr earlier in the release in the future imo.

@sebshader  I guess we're still not sure about this being included right away, and I didn't mean to tell you that you need to solve the issues so this can be included now. By the way, have a look at the discussion here => https://github.com/pure-data/pure-data/issues/1482

sebshader:
@porres ok I fixed my laptop charging cable and fixed conflicts

sebshader:
Before I try to update this to the new message format, does anyone have suggestions by any chance? @danomatika @umlaeute @Spacechild1

My first thought is to create a proc on the tcl side that will add the colors from there. (instead of calling ```create``` directly on the canvas, send it to some proc that adds appropriate colors to outline and fill etc. Before it sends to canvas)

I guess ideally tcl/TK would be aware of the class for a given draw command..

Of course it seems like I could still send the ```get_color``` proc string over with the canvas like it is now but that seems like it doesn't follow the concept here..

porres:
I'd really like to see this get in someday soon...

recent discussions in the mail list and also the fact that plug data deals with color themes that also affect iemguis I wonder if we can also thing of ways to change color of iemguis by changing themes

i know it'd involve some hackery and surgery

sebshader:
There was discussion of 'color theme palettes' and that they could be set/ unset on specific elements.

Imo the thing that makes the most sense would be not to have color palettes, but choosing whether to honor individually-set colors based on a checkbox.

That way UI elements could have 'set' colors but elements in the patch would use colors from the theme.

Still trying to figure out an elegant way to update to the new message format.. may have to default to a non-elegant way..

danomatika:
> There was discussion of 'color theme palettes' and that they could be set/ unset on specific elements.
> 
> Imo the thing that makes the most sense would be not to have color palettes, but choosing whether to honor individually-set colors based on a checkbox.
> 
> That way UI elements could have 'set' colors but elements in the patch would use colors from the 'theme'.
> 
> Still trying to figure out an elegant way to update to the new message format.. may have to fall back to a non-elegant way..

This was my suggestion. It would work similar to when @millerpuckette added drag-sizing to objects in edit mode: default theme colors are used unless the color was set manually. Like sizing a comment: default wraps at 60 chars, changing the size manually now wraps at the set width. If we know which UIs now have manually set colors, then we can know which objects could be mode-aware, ie. light colors or dark colors. Manually set colors would be kept the same. Border and iolet colors would be theme and mode aware.

Technically, it involves adding an additional key/value pair after the main object entry when saving the file. As Miller already added this to accommodate manual object sizing in the GUI, we could use the same mechanism. Currently, those objects which are resized save with ", f #" in the patch file.

Example: patch with a single comment:
~~~
#N canvas 92 117 450 300 12;
#X text 116 66 this is a comment;
~~~

Same patch resaved with comment width decreased manually by drag sizing:
~~~
#N canvas 92 117 450 300 12;
#X text 116 66 this is a comment, f 9;
~~~

The comma separates the line into two lists with this first used for object creation (text = comment), and the second is used for the object width. We could think about adding something similar.

This patch contains two IEM bang guis:

When the background color was changed to red, the object would know it no longer uses the default colors. Saving the patch would add something like a "c 1" list to denote manual color is set:
~~~
#N canvas 92 117 450 300 12;
#N canvas 92 117 450 300 12;
#X obj 134 91 bng 19 250 50 0 empty empty empty 0 -10 0 12 #fcfcfc #000000 #000000;
#X obj 194 91 bng 19 250 50 0 empty empty empty 0 -10 0 12 #ff0400 #000000 #000000, c 1;
~~~

Now we would be able to tell the GUI which objects use the default color which could come from a palette and/or mode-aware color (border color, background color, control color, ...) and which colors are set (border color,  #ff0400, #000000, ...).

porres:
just pinging here again as I'd like to suggest this to get in now that we're in an update cycle again... is something still on the way?

sebshader:
I think the PR works but it relies on some hacky weird stuff now since the gui refactor. I asked for better implementation suggestions at the time, but what I came up with was a new tcl proc that intercepts specifically positioned arguments and replaces them with lookups into the tcl color table.
now that I think about it might be better to create a more explicit api for 'function calls' in tcl.. might have to involve adding a 'function call' api to the framework-agnostic gui messages.

porres:
Like I said in that thread

We could also have these basic color settings in the preferences, not only via startup flags. Or maybe it's just preferable to have it in the settings (say, under the 'misc' tab), so users can pick the colors and also save it as part of a whole preference settings.

We could start with the basic few colors and maybe later add more options
for configuring more specific items.

chr15m:
Yes sounds good and Miller also indicated preferences would be good.

sebshader:
> @sebshader I know you have been at this for a long time now, but I was wondering how hard it would be to put together a new minimal patch that only does this one change to those 4 colors, with a view to more involved theming changes mentioned in #1482 down the line? In other words, making a small start in the right direction that would get users the much requested dark mode in the short term.

Imo fewer colors isn't really the same as having fewer features.
I also think it's odd to group colors based on existing (non-semantic) colors. what would the color names be? 'black-alt: purple', 'blue-alt: green', 'red-alt : white', 'white-alt: marigold'? It makes way more sense to label classes as a certain color like "object_fill: violet"
By the way, I also think one of the main benefits of this is the ability to distinguish between signal and message cords/iolets

I have to remember but I think the remaining ugliness of this PR would be present with fewer colors anyways (primarily getting the color values from the tcl table into the values passed to tcl procs/commands in a gui-backend-agnostic way, such as with some function call api)

The thing is, the C code would have to be modified in all the same places with just 4 colors as well anyways so you get fewer options for basically the same amount of complexity. I think the cleanest way to do this is first add some 'structured/stackful call' api for generic gui backends, then modify the calls to use that

porres:
So, we now do have color variables in Pd, 4 of them (and the 4th is optional). I wonder how this PR sits now in the new scenario. Maybe is the idea now to request and add more color options?

sebshader:
my solution for the moment was going to be:
the new variables on the c-side have the values of the tcl pd_colors array by default (e.g. comment, msg_box_text, gop_box etc)
when the colors are set by message, for the text/foreground color the corresponding variables in the colors array in tcl are set (for instance comment and msg_box_text, obj_box_text etc.). For the other colors they are set as normal.
when the draw command is sent to the tcl color lookup proc, it looks in the pd_colors array and if the key is not there uses the value received from pd as the value

This is a bit messy, maybe the corresponding tcl colors should be set for the canvas_fill and gop_box as well for consistency
or maybe for now it would be simplest just to have the pd message set the tcl variables and not use the new c-side data structure fields at all..

millerpuckette:
Perhaps best not to put too much stateful stuff on the TK side, since we might well someday migrate to some other GUI framework such as GTK or QT.

umlaeute:
i tend to agree with @sebshader.

there is state and state: if the GUI somehow f*cks up with keeping it's color state consistent, then i think there is little harm.

i think skinning and theming is none of the business of the Pd-core;so the core shouldn't get *more* state in this regard.

the ability to display colors is heavily depending on the actual GUI, so it should also handle this on its own. (it's probably somewhat academic, as i'm having a hard time to think of any meaningful example where a GUI could only render to a monochrome display)

(the same goes for zooming and the like (#1726), where I see no benefit in for adding all this logic to the core)

porres:
@sebshader I see Pd next has "canvas_text_cursor" which changes the color of the text caret on canvas items (objects, messages, etc). I think this is also set in here but not listed in https://github.com/pure-data/pure-data/pull/196#issue-251401447

Anyway, could you help us tweaking Pd so we can also do that and set it to the current foreground color? This is one of the issues being discussed in https://github.com/pure-data/pure-data/issues/2674

thanks

sebshader:
@porres yes it was added later
trying to keep the colors in the doc/7.stuff/colors-plugin.txt file should as the up-to-date 'documentation', but really I guess where the array is set in pd-gui.tcl is the real source of truth

anyways it's set here when the canvas is created:
https://github.com/pure-data/pure-data/blob/5681fde43df64f49b728f2565604b06edfdc0356/tcl/pdtk_canvas.tcl#L145..
I guess it's canvas -insertbackground option?

porres:
Yeah, I was aware of where it was set, but I thought there was more to it and... it's simpler than I thought :)

PR #200, opened 2017-08-21
allow [symbol] object to convert float messages to a symbol message

porres:
now that [float] can convert a symbol message that looks like a float to an actual float, I hope we can also make [symbol] convert a float message to a symbol, as discussed in https://github.com/pure-data/pure-data/issues/198

Though I still hope we could convert lists and anythings to a symbol, this is my first step towards that direction.

Anyway, here's the proposal. This allows the float to symbol conversion, here's a screenshot:

screen shot 2017-08-21 at 01 21 23

### DIFF
--- a/src/x_connective.c
+++ b/src/x_connective.c
@@ -156,6 +156,13 @@ static void pdsymbol_symbol(t_pdsymbol *x, t_symbol *s)
     outlet_symbol(x->x_obj.ob_outlet, x->x_s = s);
 }
 
+static void pdsymbol_float(t_pdsymbol *x, t_float f)
+{
+    char s[14];
+    sprintf(s, "%g", f);
+    outlet_symbol(x->x_obj.ob_outlet, gensym(s));
+}
+
 static void pdsymbol_anything(t_pdsymbol *x, t_symbol *s, int ac, t_atom *av)
 {
     outlet_symbol(x->x_obj.ob_outlet, x->x_s = s);
@@ -184,6 +191,7 @@ void pdsymbol_setup(void)
     class_addbang(pdsymbol_class, pdsymbol_bang);
     class_addsymbol(pdsymbol_class, pdsymbol_symbol);
     class_addanything(pdsymbol_class, pdsymbol_anything);
+    class_addfloat(pdsymbol_class, pdsymbol_float);
 }
 
 /* -------------------------- bang ------------------------------ */


porres:
perhaps we should also allow it to load a float value as an argument and output it as a symbol, I can try that (though I just did my first try at it and failed miserably)

megrimm:
+1 on this... would be nice.

porres:
I just stole the code and idea from cyclone's [tosymbol] object., so I don't really have a good argument, and I do indeed consider you'd all know better on how to do this ;) 

Now, I also tried to make it be able to load a float argument and spit it as a symbol, but failed. That would also be nice with this.

Further on, I would like as well to make it be able to convert lists of any kind to symbol messages, and also load lists as an argument.

umlaeute:
i think the "list to symbol" functionality might break patches that rely on [symbol] converting only the first atom.
(but then i just checked, and noticed that any `list` message will output the `list` selector; which is probably not what anybody wants - invalidating what i just wrote)

porres:
no, I don't think anybody uses that feature :) anyway, even if not in symbol, this kind of conversion from any message to a symbol is really important and I think it should be in vanilla, even if this functionality is achieved in another object, but I just don't see why it couldn't be in [symbol]

cheers

umlaeute:
well, you *can* already do that in vanilla, so there is no real necessity to have that in an object.

also, i think i remembered what the actual use case was: `[symbol]` currently will transform any message with a non-standard ("anything") selector into a symbol wit hthat selector. e.g. `[foo bar(` will become `symbol foo`. this is something i *do* think that people are actually using.
however, your idea would break this (unless you want to have `[list foo bar(` turned into symbol `foo bar` and `[last foo bar(` into symbol `last` which is kind of inconsequent.

src/x_connective.c
```diff
@@ -156,6 +156,13 @@ static void pdsymbol_symbol(t_pdsymbol *x, t_symbol *s)
     outlet_symbol(x->x_obj.ob_outlet, x->x_s = s);
 }
 
+static void pdsymbol_float(t_pdsymbol *x, t_float f)
+{
+    char s[14];
+    sprintf(s, "%g", f);
+    outlet_symbol(x->x_obj.ob_outlet, gensym(s));
+}
```
umlaeute: this fails to implement the core functionality of `[symbol]` - that is: to store the actual symbol.
this kind of "breaks" the object.

porres: oops... yeah, this just seems like a timid starting point for implementing a proper way to deal with floats, far from done. You also had remarks on this code as well, huh? 

I can go ahead and try to make this a proper PR myself if I get some help on what to do.

cheers

porres:
> well, you can already do that in vanilla, so there is no real necessity to have that in an object.

what can you do? and how?

src/x_connective.c
umlaeute: i thought this was already meant as a "proper" PR. if you just want a branch for discussion, please add `[WIP]` at the beginning of the title.

porres: what does [WIP] mean?

umlaeute:
closing this, as the implementation is incomplete and the PR branch has been removed (so it's kind of impossible that it will receive more love).

PR #204, opened 2017-08-21
Pd Help menu now opens manual

### LINKED ISSUE #172: extra redundant "pd Help" option in Help Menu
There's a new menu entry in "Help" named "Pd Help" which just opens the browser, so it the same thing as the "Browser" option below, thus, completely redundant.

Was this supposed to open something else?

cheers

danomatika:
This PR changes the Help menu on macOS when TK > 8.5:

* Pd Help menu now opens manual
* HTML Manual... entry removed

The Pd Help entry is created automatically by Tk Cocoa and cannot be changed or removed. This solution simple uses it in lieu of the HTML Manual... entry to avoid duplicate menu entries at the slight cost of a different menu entry name on 1 platform.

This fixes #172.

### DIFF
--- a/tcl/apple_events.tcl
+++ b/tcl/apple_events.tcl
@@ -57,7 +57,7 @@ proc ::tk::mac::Quit {args} {
 # is provided by default and cannot disabled or removed 
 proc ::tk::mac::ShowHelp {args} {
     ::pdwindow::verbose 1 "::tk::mac::ShowHelp $args ++++++++++++\n"
-    ::pd_menucommands::menu_helpbrowser
+    ::pd_menucommands::menu_manual
 }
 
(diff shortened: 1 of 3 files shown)


porres:
+1

danomatika:
Here's a screenshot of the Help menu on macOS using TK 8.5:

screen shot 2017-08-21 at 8 44 41 pm

megrimm:
+1

looks good... i think this is our best bet for eliminating confusing redundancy.

thanks!

danomatika:
I hope that the missing "HTML Manual..." entry is also not confusing...

porres:
I think this is less confusing and simpler ;)

PR #211, opened 2017-08-24
[WIP] MIDI updates

danomatika:
This PR contains updates to the Pure Data MIDI handling:

* [x] fix portmidi infinite loop bug when receiving active sense during running status messages on macOS
* [x] all messages are now received
* [x] slight code cleanup favoring MIDI status byte defines for readability and some commenting
* [x] enable [midiclkin] object
* [ ] decide on [midiclkin] tick range and possible reset message (for use with tick position counting)
* [x] possibly add [songposin]/[songposout] & [songin]/[songout] objects
* [ ] MIDI time code handling

The work so far has been tested with a simple utility program [miditester](https://github.com/danomatika/miditester) and Logic Pro X on macOS. This will need testing on other platforms.

### DIFF
--- a/portmidi/portmidi/pm_mac/pmmacosxcm.c
+++ b/portmidi/portmidi/pm_mac/pmmacosxcm.c
@@ -70,6 +70,7 @@
 #define VERBOSE_ON 1
 #define VERBOSE if (VERBOSE_ON)
 
+#define MIDI_CLOCK      0xf8
 #define MIDI_SYSEX      0xf0
 #define MIDI_EOX        0xf7
 #define MIDI_STATUS_MASK 0x80
@@ -199,9 +200,16 @@ process_packet(MIDIPacket *packet, PmEvent *event,
 		/* since there's no more data, we're done */
 		return;
 	    }
-	    m->last_msg_length = cur_message_length;
-	    m->last_command = cur_packet_data[0];
-	    switch (cur_message_length) {
+        if (cur_packet_data[0] < MIDI_SYSEX) {
+            /* channel messages allow running status */
+            m->last_command = cur_packet_data[0];
+            m->last_msg_length = cur_message_length;
+        } else if(cur_packet_data[0] < MIDI_CLOCK) {
+            /* system messages clear running status */
+            m->last_command = 0;
+            m->last_msg_length = 0;
+        }
+        switch (cur_message_length) {
 	    case 1:
 	        event->message = Pm_Message(cur_packet_data[0], 0, 0);
 		break; 
@@ -220,8 +228,8 @@ process_packet(MIDIPacket *packet, PmEvent *event,
 	        return; /* give up on packet if continued after assert */
 	    }
 	    pm_read_short(midi, event);
-	    remaining_length -= m->last_msg_length;
-	    cur_packet_data += m->last_msg_length;
+	    remaining_length -= cur_message_length;
+	    cur_packet_data += cur_message_length;
 	} else if (m->last_msg_length > remaining_length + 1) {
 	    /* we have running status, but not enough data */
 #ifdef DEBUG
--- a/src/s_midi.c
+++ b/src/s_midi.c
@@ -84,7 +84,7 @@ static double sys_newdactimeminusrealtime = -1e20;
 static double sys_newadctimeminusrealtime = -1e20;
 static double sys_whenupdate;
 
-void sys_initmidiqueue( void)
+void sys_initmidiqueue(void)
 {
     sys_midiinittime = clock_getlogicaltime();
     sys_dactimeminusrealtime = sys_adctimeminusrealtime = 0;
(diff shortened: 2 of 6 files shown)


porres:
Hi, the deal with [midiclkin] is that I already sent a PR to enable it and discussed fixes/enhancements. The outcome is that @millerpuckette thought it was best not to have a [midiclkin] object, so now I have a new PR on removing it, which also updates midi-help.pd https://github.com/pure-data/pure-data/pull/187

danomatika:
@porres I am aware of the PR. My understanding of the exchange boils down to:

* you: "This object is documented but doesn't do anything."
* Miller: "Oh. Maybe we should remove it then."

I'm offering a alternative where it *does* do something. At the very least, you can now test it and help decide if it's behavior should be modified or it should in fact be removed. For instance, it provides a delta time between click messages. Perhaps this is useful.

porres:
Actually there was more to that. I also thought it would be useful (thus I proposed a fix) and pointed the delta time feature, but miller didn't accept the fix and thought the delta thing was badly implemented, and one could do it with a timer object.

> "This object is documented but doesn't do anything."

funny fact, it was even documented, until I pointed "hey, there's this object but it isn't even mentioned in the help files", so it was included, then I tested it and saw it wasn't working at all...

And even by "fixing it" (that is, bringing it onboard), it certainly does have some weird design choices that even miller really couldn't figure out the purpose. 

Let me just say I don't really care that if it stays or not. It's up to you and Miller ;) and I'm fine either way. But I figure we should look into it if it's actually brought in.

danomatika:
Yeah, it is a little weird. It *seems* like handling the clock message from [midiin] would be enough. I do think knowing the exact tick within the 24 "pulse per quarter note" count (from the MIDI spec) could be useful, but you could also build that yourself using a counter in Pd. I experimented with adding a reset message so you could reset the counter manually when a MIDI songpos message comes in.

porres:
what I thought it was weird is that it was counting every third clock tick... I agree a dedicated object would be cool, and I'm usually excited with the addition of any new object and function into vanilla ;)

danomatika:
Well, my understanding is that the MIDI clock message is 24 times per quarter note, so it could be useful if [midiclkin] sends a 1 for every 1/16 note and 0 for all other ticks.

porres:
yeah, but why the third? and not the first one? the head of tempo, huh?

danomatika:
Yes, it should be the first one. I had changed that in my experiments.

porres:
what I think it'd be cool is that if it'd output a BPM value from the ticks

danomatika:
Added the following objects:

* [songposin] & [songposout]: song position 0-16383 (14 bit)
* [songin] & [songout]: song select 1-127 (7 bit)

As these are system messages which do not take a channel, they use the port number like [midiin], [midirealtimein], & [midiout]

danomatika:
I've added working [timecodein] and [timecodeout] objects which handle the MIDI Timecode Quarter Frame message. Now that I'm reading into the MIDI spec more, the MTC Full Frame message is sent as a sysex message plus a couple of other things. I may rework how these objects work in order to support the full MTC protocol since this kind of byte handling involves bit masking which is difficult, if not impossible using only core objects.

porres:
as long as we're at it, doing this major revision, I'd like to bring attention to this new specification http://expressiveness.org/2015/04/24/midi-specifications-for-multidimensional-polyphonic-expression-mpe

which is already included in Max and other softwares - I hope we could also make it happen for Pd

cheers

porres:
this link is better https://www.midi.org/articles/midi-polyphonic-expression-mpe

porres:
the way this is handled in max seems to be via a [mpeparser] object, receiving raw data from [midiin] https://support.roli.com/article/max-tutorial-1/

So maybe we have [midiin] sending the same raw things, or just other dedicated objects (which seems like a better design choice)

danomatika:
If it takes data from [midiin], then it can be implemented as an external. I'm trying to get the original MIDI spec fleshed out, not addons right now.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

porres:
I'm not sure Pd's [midiin] gives us that already, that's the point. Maybe it needs to be updated to also send it. Then we can even consider having an external like in cyclone and stuff...

danomatika:
>I'm not sure Pd's [midiin] gives us that already

It does now. That was partially the point of this PR. [midiin] + [midirealtime] now give you *all* the raw byte values received by Pd, as they should.

porres:
are you sure it handles MPE though? that's what I suspect we'd have to do something new to add it

danomatika:
> are you sure it handles MPE though? that's what I suspect we'd have to do something new to add it

It does *not*. [midiin] gives you the raw MIDI bytes without any parsing. I imagine [mpeparse] does that, so a Pd version of [mpeparse] would probably work right now with the output from [midiin].

I am not going to implement a Pd [mpeparse]. It's a proposed protocol built on top of the existing MIDI specification and not something that needs to be included with the Pd core. It's a perfect candidate for an external.

porres:
yeah, my question was if the MPE thing was available "raw" in midiin, I know the concept of how [mpeparser] works in max, but I just wonder if we can filter it out from [midiin] as is... if even the realtime messages were missing, I thought these messages could be missing too, you know?

porres:
> filter it out

not a good term, I meant "retrieved", as in a 'parser' object

danomatika:
Doh, of course there are bitwise operators in Pd, I've just never used them. In that case, we may not need the song, songs, & timecode objects as they can be implemented within Pd itself. I will probably move some of this work to a different branch and keep the MIDI fixes only.

danomatika:
Closing this as it got a bit too complicated. The added objects can be implemented as externals which parse bytes form [midiin]. The bugfixes and cleanups are available in PR #214 instead.

PR #214, opened 2017-08-29
midifixes

danomatika:
This PR is an offshoot of #211 which only contains the MIDI bug fixes and code cleanups:

* [x] fix portmidi infinite loop bug when receiving active sense during running status messages on macOS (submitted and accepted upstream)
* [x] fix port midi string encoding bug on macOS (via upstream)
* [x] all message types should now be received and sent
* [x] slight code cleanup favoring MIDI status byte defines for readability and some commenting
* [x] remove unused [midiclkin] object ala #187 
* [x] midiout should send float lists

The work so far has been tested with a simple utility program I wrote ([miditester](https://github.com/danomatika/miditester)) and Logic Pro X on macOS. This will need testing on other platforms.


src/s_midi_alsa.c
```diff
@@ -112,57 +126,69 @@ void sys_alsa_do_open_midi(int nmidiin, int *midiinvec,
...
-            channel = a-160;
-            snd_seq_ev_set_keypress(&ev,channel,b,c);
-        }
-        else if (a >= 144)      // note on
-        {
-            channel = a-144;
-            snd_seq_ev_set_noteon(&ev,channel,b,c);
-        }
-        else if (a >= 128)      // note off
+        status = a & 0xf0;
```
umlaeute: so `status <= 0xf0` (240)...

src/s_midi_alsa.c
```diff
@@ -112,57 +126,69 @@ void sys_alsa_do_open_midi(int nmidiin, int *midiinvec,
...
+                snd_seq_ev_set_pgmchange(&ev, channel, b);
+                break;
+            case MIDI_AFTERTOUCH:
+                snd_seq_ev_set_chanpress(&ev, channel, b);
+                break;
+            case MIDI_PITCHBEND:
+                /* b and c are already correct but alsa needs to recalculate them */
+                snd_seq_ev_set_pitchbend(&ev, channel, (((c<<7)|b)-8192));
+                break;
+            case MIDI_TIMECODE:
```
umlaeute: and MIDI_TIMECODE is 0xf1 (241); so how on earth is `status == MIDI_TIMECODE` ever going to be `true`?

umlaeute: the same holds for `SONGPOS` and `SONGSELECT`

src/s_midi_alsa.c
danomatika: The Status value is the upper nibble and the channel is the lower nibble for those messages which have them. This masks out the channel.

danomatika: AH I see what you're saying. That should be:

    status = a & 0xFF;
    if (status < MIDI_SYSEX) status = status & 0xF0;

src/s_midi_alsa.c
danomatika: Good catch. Fixed.

danomatika:
@umlaeute As you detest MIDI, I'm guessing you're not interested in running some tests?

danomatika:
I added the midi test patch I've ben using to `docs/7.stuff/tools`. It sends & receives all MIDI messages.

umlaeute:
On 08/30/2017 05:37 PM, Dan Wilcox wrote:
> @umlaeute As you detest MIDI, I'm guessing you're not interested in running some tests?

i honestly wouldn't know how.
(that is: i don't have any MIDI gear ready at hand to preform some real
world tests)

danomatika:
@umlaeute Neither do I right now. That's why I wrote a test utility: https://github.com/danomatika/miditester

It you run it with the `-n -d` flags, the values are pretty much the same so comparing the output in the `doc/7.stuff/tools/miditester.pd` patch is relatively easy. The only caveats are that sending from Pd results in controller and program numbers off by 1 due to Pd adding 1.

src/x_midi.c
```diff
@@ -586,11 +586,21 @@ static void midiout_float(t_midiout *x, t_floatarg f)
     outmidi_byte(x->x_portno - 1, f);
 }
 
+static void midiout_list(t_midiout *x, t_symbol *s, int ac, t_atom *av)
+{
+    for (int i = 0; i < ac; ++i)
```
umlaeute: this is only valid in C99; since adding support for ANSI C is pretty straightforward, i would suggest to use:
~~~C
int i;
for (i = 0; i < ac; ++i)
{
   // ..
}
~~~

danomatika: Got it. Is C99 still a hard limit? I suppose so if we want to continue supporting Windows XP builds? ;P

PR #219, opened 2017-09-16
allow float input in left inlet of fexpr~

porres:
I'm adding "CLASS_MAINSIGNALIN(fexpr_tilde_class, t_expr, exp_f);" to fexpr~ creation. This allows a float input into fexpr~ left inlet, which is the general expected behaviour of pd objects.

This also makes it consistent to [expr~], which can also allow float input in its left inlet


porres:
I've brought this up to the list and including Shahrokh to see if this is actually cool or not ;) https://lists.puredata.info/pipermail/pd-list/2017-09/120306.html

porres:
maybe we should also update the version to something like "0.56"

I should also fix documentation if this is accepted

porres:
this was done in 0.48-1

PR #221, opened 2017-09-17
update french translation

Ant1r:


po/fr.po
```diff
@@ -30,25 +30,25 @@ msgid "-----------output-range:-----------"
 msgstr "---------échelle_de_sortie:--------"
 
 msgid "------selectable_dimensions(pix):------"
-msgstr "-------dimensions_a_choisir(pix):------"
+msgstr "-------dimensions_sélectionnable(pix):------"
```
danomatika: This needs to be the same width as the string above. Easiest option is to remove some of the "-" characters.

po/fr.po
```diff
@@ -6,16 +6,16 @@ msgstr ""
...
+"Last-Translator: Cyrille Henry / Antoine Rousseau / Antoine Villeret / Baptiste Chatel\n"
 "Language-Team: \n"
 "Language: fr\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
 
 msgid "-------dimensions(digits)(pix):-------"
-msgstr "-------dimensions(digits)(pix):-------"
+msgstr "-------dimensions(chiffres)(pix):-------"
```
danomatika: This needs to be the same width as the string above.

danomatika:
Why were spaces added before semicolons? Is this a language convention I'm unaware of? Otherwise, I think the semicolons should be adjacent to the text as this affects the label sizing.

Ant1r:
> Why were spaces added before semicolons? Is this a language convention?
Yes, french typography requires a space before and after any "double" sign, like: : ! ? ;

Anyway I'll fix other things before re-submitting.

Ant1r:
I simplified this PR and reopen as #250, directed to master.

PR #222, opened 2017-09-17

Ant1r:


danomatika:
This will probably break things which rely on the current dialog communication with Pd and will lead to inconsistency with the Pd forks as well. We're kind of stuck with original decisions like this.

Also, a few of your PR's include a commit from IOhannes, so the diff is not clean.

Ant1r:
Since I've updated src/s_path.c as well as tcl dialogs, I think that it should be fully compatible, but let me know if you see something I don't; anyway I will clean this PR (removing IO's commit, sorry...) and propose it again, in master branch if you prefer.

PR #223, opened 2017-09-20
clang implicit conversion warning fixes

danomatika:
This PR mostly adds explicit casts to fix conversion warnings in clang:

* long -> int
* t_int -> int
* etc

Some of the warnings come from `atom_getintarg()` returning a t_int but many functions taking int as args. The more complicated update came in d_soundfile.c where I erred on using longs for buffer sizes.

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -708,7 +708,7 @@ static t_int *tabsend_perform(t_int *w)
 {
     t_tabsend *x = (t_tabsend *)(w[1]);
     t_sample *in = (t_sample *)(w[2]);
-    t_int n = w[3];
+    int n = (int)w[3];
     t_word *dest = x->x_vec;
     int i = x->x_graphcount;
     if (!x->x_vec) goto bad;
@@ -792,7 +792,7 @@ static t_int *tabreceive_perform(t_int *w)
 {
     t_tabreceive *x = (t_tabreceive *)(w[1]);
     t_sample *out = (t_sample *)(w[2]);
-    t_int n = w[3];
+    int n = (int)w[3];
     t_word *from = x->x_vec;
     if (from)
     {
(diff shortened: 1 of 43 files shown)


danomatika:
With this branch, libpd compiles without warnings in Xcode and Pd compels without warnings on macOS.

src/d_array.c
```diff
@@ -708,7 +708,7 @@ static t_int *tabsend_perform(t_int *w)
 {
     t_tabsend *x = (t_tabsend *)(w[1]);
     t_sample *in = (t_sample *)(w[2]);
-    int n = w[3];
+    t_int n = w[3];
```
umlaeute: i think this is wrong.
`t_int` is practically never used for *int* values; it's main purpose seems to be to a way to provide "void-pointer arithmetic" in the dsp-graph.

(i think the use of `t_int` for `atom_getint` resp `atom_getintarg` is accidental and should be fixed)

umlaeute: i think this is wrong.
`t_int` is practically never used for *int* values; it's main purpose seems to be to a way to provide "void-pointer arithmetic" in the dsp-graph.

(i think the use of `t_int` as return-type for `atom_getint` resp `atom_getintarg` is accidental and should be fixed)

danomatika: Gotcha. t_int sounded like a generic integer type, so I was preferring that in places where I thought it made sense.

danomatika: >(i think the use of t_int as return-type for atom_getint resp atom_getintarg is accidental and should be fixed)

Should I rectify this? It would solve a good number of warnings.

umlaeute: yes, it should read `int n = (int)w[3];`

danomatika:
Updated to prefer int over t_int. Also, `atom_getint()` and `atom_getintarg()` now return an int which negates a large number of explicit casts.

PR #225, opened 2017-09-22
reimplemented "allow negative canvas positions in Tk 8.5+ for multi monitor support"

umlaeute:
this is a re-implementation of PR #224, which:
- keeps the window-positions on X11
- is more optimized (check for hacks is only done on boot time (and therefore drops the `TCL_PATCHLEVEL` variable)

i've tested this on Debian but only with a standard dual-screen setup (main monitor to the left of right-monitor)

### original PR text

As reported on the pd-list, negative canvas window positions are clipped to 0 and/or reflected on the opposite screen border.

This PR fixes that by allowing negative positions to be sent to the window manager in Tk 8.5+ and keeps the clipping for Tk 8.4 as a fallback. Also, versions of Tk < 8.5.13 on macOS keep the clipping due to the buggy nature of the TK Cocoa backend up until that version.

This will need to be tested on all platforms but, from my research, it should be fine as negative positions were added to 8.5 and back ported to 8.4 at some point. I decided to use 8.5 as a hard limit to be conservative.

This also removes the apparently unused TCL_*_VERSION variables

Tested so far on macOS 10.12.6 with Tk 8.4, Tk 8.5.9, Tk 8.7.0a.

You can try a build for newer macs (10.9+) with Tk 8.7a: http://docs.danomatika.com/pdbuilds/Pd-0.48-1-multimonitor.app.zip

Here is a minimal patch which opens sub patches in all 4 directions. Expected behavior: all open on screen, either in expected location or clipped to screen edge if the position is off screen.

~~~
#N canvas 100 300 450 300 10;
#N canvas 400 -800 450 300 top 1;
#X restore 150 50 pd top;
#N canvas 400 1800 450 300 bottom 1;
#X restore 150 250 pd bottom;
#N canvas -1800 400 450 300 left 1;
#X restore 50 150 pd left;
#N canvas 1800 400 450 300 right 1;
#X restore 250 150 pd right;
~~~

### DIFF
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -29,31 +29,54 @@ namespace eval ::pdtk_canvas:: {
 #winfo rooty . returns contentsTop
 #winfo rootx . returns contentsLeftEdge
 
+if {$::tcl_version < 8.5 || \
+        ($::tcl_version == 8.5 && \
+             [tk windowingsystem] eq "aqua" && \
+             [lindex [split [info patchlevel] "."] 2] < 13) } {
+    # fit the geometry onto screen for Tk 8.4,
+    #  also check for Tk Cocoa backend on macOS which is only stable in 8.5.13+;
+    # newer versions of Tk can handle multiple monitors so allow negative pos
+    proc pdtk_canvas_wrap_window {x y w h} {
+        set width [lindex [wm maxsize .] 0]
+        set height [lindex [wm maxsize .] 1]
+
+        if {$w > $width} {
+            set w $width
+            set x 0
+        }
+        if {$h > $height} {
+            # 30 for window framing
+            set h [expr $height - $::menubarsize - $::windowframey]
+            set y $::menubarsize
+        }
+
+        set x [ expr $x % $width]
+        set y [ expr $y % $height]
+        if {$x < 0} {set x 0}
+        if {$y < 0} {set y 0}
+
+        return [list ${x} ${y} ${w} ${h}]
+    }
+} {
+    proc pdtk_canvas_wrap_window {x y w h} {
+        return [list ${x} ${y} ${w} ${h}]
+    }
+}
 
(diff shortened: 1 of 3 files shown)


tcl/pdtk_canvas.tcl
```diff
@@ -29,31 +29,54 @@ namespace eval ::pdtk_canvas:: {
 #winfo rooty . returns contentsTop
 #winfo rootx . returns contentsLeftEdge
 
+if {$::tcl_version < 8.5 || \
+        ($::tcl_version == 8.5 && \
+             [tk windowingsystem] eq "aqua" && \
+             [lindex [split [info patchlevel] "."] 2] < 13) } {
```
danomatika: One issue is that the patch level string can contain alpha numerical values ala "8.7.a1". scan "%d.%d.%d" catches this and skips setting the value, so setting the value to 0 ahead of time yields a default of 0 even when number can't be parsed.

danomatika:
I can confirm this is working on macOS 10.12.6.

tcl/pdtk_canvas.tcl
danomatika: I guess what I mean to say is I'm not sure how a resulting "string: < 13 comparison will play out, possibly yielding a bug where test versions of Tk will suddenly wrap the screen coords while release versions will not. It's an edge case, sure, but *could* happen ... ?

danomatika: Doh. Actually this is fine. The version of Tk I am testing with yields "8.7.a0" and the screen wrapping is not being called.

umlaeute: i think, tcl is not really performing a numeric comparision with the `<` operator, but rather a string comparision.
e.g. `"abc" < "xyz"` is valid (and returns `true`). so we are safe on the syntactic side.

whether `a19 < 19` should be true or not (on the semantic side), i cannot say.

PR #226, opened 2017-09-22

Ant1r:
This is only a GUI change, and doesn't modify any variable name for e.g, so it shouldn't impact compatibility, excepting patches or plugins or externals directly calling pdtk_path_dialog or pdtk_startup_dialog (but are they numerous?).


Ant1r:
I understand. I reopen it to master on #249.

PR #227, opened 2017-09-23
Updated metrics and object sizing, etc

danomatika:
This is a bug fix PR meant to alleviate font metrics and object sizing issues in Pd 0.48.

As many have pointed out on the pd-list, the sizing of objects is a bit different between 0.46 and 0.48. This seems to be largely due to how the font metrics are calculated with zooming and the introduction of Deja Vu Sans Mono on macOS (and probably other factors).

Pd-extended managed to have relatively consistent sizing across platforms, so some of this work comes directly from extended as well as some experimentation.

Highlights:

* updated font metrics to pd-extended values
* fixed bug where metric detection was not taking into account the desired font weight
* ~increased vertical margins slightly to offset the tighter font sizing~ *doesn't seem to be needed after testing*
* added IOHEIGHT for consistent inlet & outlet heights, io now zooms
* ~float and symbol atom heights now match object box heights and the text is thus centered better~ *decided against this*
* added in code from extended which:
  * ensures patch cords are not drawn behind objects when they are created
  * improves atom and message box right indent sizing (calculates based on size), this makes them a bit more obviously different from rectangular object boxes
* updating iem guis for full zoom compatibility (form and function):
  * [x] bang
  * [x] toggle
  * [x] number2
  * [x] vslider
  * [x] hslider
  * [x] vradio
  * [x] hradio
  * [x] vu
  * [x] mycanvas
* [x] updating arrays for zooming

Note: This includes some work first introduced in #162 but I decided that it would be a better approach to focus on fixing sizing first, then look into styling, etc later on.

This has been tested on macOS 10.12.6. It should be checked on other platforms.

Here is a test build for newer Macs (10.9+): http://docs.danomatika.com/pdbuilds/Pd-0.48-1-metrics-sizing.zip

And some screenshots:

**0.47-1**

screen shot 2017-09-23 at 10 41 41 pm

screen shot 2017-09-23 at 10 41 16 pm

**0.48-0 (before)**

screen shot 2017-09-23 at 10 29 24 pm

screen shot 2017-09-23 at 10 33 26 pm

**0.48-1 (after)**

screen shot 2017-09-23 at 10 29 09 pm

screen shot 2017-09-23 at 10 33 45 pm


porres:
How nice! 

for more comparison, I'm pasting coll's documentation that I've used many times to point this issue. Left is 047-1, right is this PR's binary. They both seem to have the exact same sizing, so this really fixes all the issues I was having. I can add that the right one looks nicer, probably because of rendering and the new font.

Thanks so much.

cheers.

screen shot 2017-09-23 at 18 01 11

porres:
though I can see that outlets are bigger, and I guess I saw a discussion about it. Looks cool, the problem is when using objects that have letters like "p", "y", it doesn't look good as it cuts it a little.


screen shot 2017-09-23 at 18 13 21

danomatika:
Yeah, I know. The font sizing does not seem to take into account descenders and that is probably why the outlets are drawn smaller. Increasing the bottom margin with fatter inlets leads to the objects being too large as well, so maybe stilling to smelling inlets & outlets works better.

porres:
with a closer look to the figure above, now I can see that this new PR has some offset that makes the text maybe a pixel lower, and that contributes to the collision of "p" and the outlet

danomatika:
I tweaked the margins so the text should be centered better and the test build has been updated.

Also, I'm not 100% sold on the atoms being the same height as objects as the text in float atoms now seems to sit too high, however it is nice that objects, floats, and symbols are all the same height now. Maybe this  will break people's layouts?

millerpuckette:
The good thing abut 0.48-0 was that it got objects to be sized a lot more consistently between platforms.  This seems to have meant widening the 10-point font on Mac from 6 to 7 pixels, as it had already been on linux.  As a result, now the "coll" help window on linux looks at least somewhat like it does on Mac.  But other stuff that was designed in linux has now changed.  The only thing that outright broke for me was a 24-point symbol box in a GOP that grew (and hence went invisible).

Maybe a compromise would be to adjust the "standard" font sizes downward (but not upward) so that they agree with what actually appeared on Macontosh?

danomatika:
>But other stuff that was designed in linux has now changed. The only thing that outright broke for me was a 24-point symbol box in a GOP that grew (and hence went invisible).

What kind of things?

There has been some cross platform testing with the font metrics previously: https://github.com/pure-data/pure-data/wiki/Crossplatform-font-metrics-&-comparisons

This harkens back to Pd-extended, so I've fallen back to largely pulling in details from there. One issue we should have fixed is different font sizes between platforms by standardizing on one font (except for Windows so far...).

danomatika:
And in checking some screenshots, one of the height differences comes from Monaco versus Deja Vu Sans Mono where Monaco is slightly taller, at least as rendered on macOS. There was a manual adjustment for this to the vertical margins (TMARGIN & BMARGIN) in Pd-extended, but I opted for adding 1 pixel to each as suggested on the pd-list. That should help with consistency, hopefully.

danomatika:
One thing to note is there are probably 2 ideas of "pixel-perfect" layouts:

* previous vanilla
* Pd-extended

Moving from one layout to the other will result in some adjustment. I remember moving from extended to vanilla necessitated some sizing changes as some things rendered larger on extended than on vanilla.

danomatika:
I think what is really needed is a canonical test patch with objects, some GOP, etc we can use as a baseline for comparison.

millerpuckette:
I don't think any particular size is the "correct" one - my original choice of size limits like (6, 10)
(for "8-point") was that there were fonts available at the time that were close to those values, so that moving back and forth between OSes didn't change sizes too much.  The change from 0.47 to 0.48 seems to have harmonized teh sizes at least between MACOS and linux, but did so at the expense of growing the fonts slightly on MACOS.  But now it seems that allowing anything to grow makes things collide that used not to.  So the best path forward I can see is to find out what the actual pixel sizes on Mac were in 0.47, and where necessary reducing something in the declared maximum sizes so that nothing grows - but also trying to limit how much things are thereby made to shrink in the other OSes.

Lucarda:
This is for w32:


For this test I pasted the files from:
https://ci.appveyor.com/project/umlaeute/pure-data/build/job/e4o2av8gdp8dp90d/artifacts

to the stable 0.48.0 and downloaded the last “pd-gui.tcl ” from :
https://github.com/pure-data/pure-data/pull/227/files

0.48.0


0.48.1-testing
048 1test

danomatika:
After doing some more comparisons, the latest commit walks back the atom sizing to match objects and corrects some details.

I've updated the Mac test build on the main post link.

IMO Pd-extended has basically set the precedent for not only many users patches but also the sizing used by the various forks at this point. I'm erring slightly on matching some of the details it established. I'm trying to find a way for us to have & eat cake between things.

For comparison, here is the same patch rendering the same horizontally in 3 versions of Pd. I'd argue if the extended sizing is working to this degree, then I think we are doing pretty well considering the amount of effort that already went into testing font sizing using DVSM years ago now.

** 0.47-1**

screen shot 2017-09-24 at 2 12 34 am

** Pd-extended**

screen shot 2017-09-24 at 2 21 30 am

** 0.48-1 (this branch)**

screen shot 2017-09-24 at 2 36 59 am

One other small note, I generally check how the included Pd manual patches look when fiddling with rendering. I know things are borked if those at 12 point font break.

danomatika:
And here is the famous coll-help.pd with the latest commit:

screen shot 2017-09-24 at 2 48 39 am

porres:
> IMO Pd-extended has basically set the precedent for not only many users patches but also the sizing used by the various forks at this point.

Agreed. For instance, Purr Data, which follows the Extended styling, and I'm hoping the cyclone documentation looks the same on both ;)

porres:
> I tweaked the margins so the text should be centered better and the test build has been updated.

Looks better now, just like 0.48, though I could say I think moving it up one pixel higher would make it better centered, I think.

danomatika:
Moving forward into related territory: vslider and hslider now fully support zooming, use the overall io sizing, and a couple bugs in their implementation were fixed as well. These are probably the most complicated of the IEM guis, so the rest should be a bit quicker hopefully.

Comparison screenshots between regular and zoomed:

screen shot 2017-09-25 at 1 13 58 am
screen shot 2017-09-25 at 1 14 07 am
screen shot 2017-09-25 at 1 14 13 am
screen shot 2017-09-25 at 1 14 18 am

Also, I made a quick hack which resizes the patch canvas when zooming which has been handy as you no longer need to scroll and/or manually resize as much. Zooming back in simply divides the size by 2 so it goes back to the same size. Let me know if that is something that would be useful to add.

danomatika:
>Looks better now, just like 0.48, though moving it up one pixel higher could make it better centered, I think.

@porres It becomes too smaller, especially when using font size 8.

Lucarda:
Up-to-date with this branch.
Using build from AppVeyor for Ms-Windows.
Everything seems fine:

slider


sliderzoom

This is not from this branch, is 0.48.0 (zoomed). The knob is different and also the label, look in the arrow:

sliderzoom0 48

danomatika:
I decided to walk back the taller vertical object margins to the original 2 & 2. The object aspect ratio just seems to tall over time.

porres:
Hi, I'm aware it needs space for letters like "p", and I find the space bellow fine. I did ask to me the text better centered by moving it up just a bit because of that, and now that you did, I also felt like it could go up yet another pixel, that's what I meant. Not sure if that was clear.

Now, I think I see what you're pointing out, that it makes a difference depending on the font size. Yeah, this thing I'm mentioning is clearer when you have a bigger font. The bigger it gets, the better you can see that the object's text is not well centered, it's lower than higher

danomatika:
macOS & Linux object comparison from Roman Haefeli on the pd-list:

attachment-0001

danomatika:
>So the best path forward I can see is to find out what the actual pixel sizes on Mac were in 0.47, and where necessary reducing something in the declared maximum sizes so that nothing grows - but also trying to limit how much things are thereby made to shrink in the other OSes.

@millerpuckette by using the following at the end of pd-gui.tcl `fit_font_into_metrics` proc:

~~~
   ::pdwindow::verbose 0 "Measured font metrics:\n"
    foreach {size width height} $::font_measured {
        ::pdwindow::verbose 0 "$size $width $height\n"
    }
    ::pdwindow::verbose 0 "Measured zoom2 font metrics:\n"
    foreach {size width height} $::font_zoom2_measured {
        ::pdwindow::verbose 0 "$size $width $height\n"
    }
~~~

**0.47-1 macOS**

~~~
Measured font metrics:
8 5 10
10 6 13
13 8 16
17 10 21
20 12 25
36 22 45
Measured zoom2 font metrics:
16 10 20
21 13 26
26 16 32
34 20 42
40 24 50
72 43 90
~~~

**this branch macOS**

~~~
Measured font metrics:
9 5 10
10 6 11
12 7 14
16 10 19
23 14 26
36 22 41
Measured zoom2 font metrics:
16 10 19
19 12 22
23 14 26
32 20 38
46 28 54
73 44 85
~~~

millerpuckette:
Would it be appropriate for me to test & merge this now?

danomatika:
It depends. I still have the rest of the item guis to update for zooming. One option would be to split the zoom-related stuff into a separate branch.

millerpuckette:
Nah... I can just wait a few days, no problem.

danomatika:
Ok. It should probably be more into the weekend.

I am a little skeptical about the amount of multiplication that I'm adding for the zooming, so I may go back and try to see if I put some zoom level checks in.

Lucarda:
screenshots for Windows:

l1


Zoomed:

l2

danomatika:
So far, I'm updating the iem guis to use the same IOHEIGHT value that regular objects use for the inlet/outlet height. We can change that if we want as it seems a size of 3 leads to the radio button touching the io rectangles, so perhaps 2 for iem guis and 3 for regular objects or 2 for everything? I chose 3 as it's the minimum needed to draw the io boxes with fill & outline in order to allow for future theming and/or Pd-extended style signal/control style differences.

Lucarda:
It could be “**2 for iem guis and 3 for regular objects**” as iem guis will not accept audio signals. Just a thought, 3 for all is also ok.

danomatika:
IEM guis are more or less ready for testing. I tried to test the edit messages when zoomed/non-zoomed and when saving while zoomed, but I might have missed a detail or two.

Next up: arrays

One note so far, in my testing, the corners of objects drawn with a border don't complete overlap when zoomed. See the upper left corner of this zoomed float atom:

screen shot 2017-10-02 at 3 32 16 pm

screen shot 2017-10-02 at 3 32 16 pm

I wondering if this is something I'm seeing on mac only or if it's happening on other systems. I have a fix but don't want to commit to it unless this is a cross-platform issue. It seems to be an incomplete overlap between the first and last points in the line.

Ant1r:
same here on ubuntu 12.04 (and seems the same on Lucarda's windows screenshot above).
but quite a minor issue...
screenshot from 2017-10-02 15 59 33

danomatika:
Ok, so it's how Tk is drawing the lines. I'll add `-capstyle projecting` to fix this.

danomatika:
>but quite a minor issue...

And an easy fix it turns out.

src/g_text.c
```diff
@@ -1008,11 +1016,16 @@ static void text_getrect(t_gobj *z, t_glist *glist,
...
+        width = (x->te_width > 0 ? x->te_width : 6) * glist_fontwidth(glist);
+        height = glist_fontheight(glist);
+        if (glist_getzoom(glist) > 1)
+        {
+            /* zoom margins */
+            width += ATOM_RMARGIN * glist_getzoom(glist);
+            height += ATOM_BMARGIN * glist_getzoom(glist);
+        }
+        else
+            width += ATOM_RMARGIN; height += ATOM_BMARGIN;
```
umlaeute: is this intentional? it seems that `height` is incremented regardless of the result of `glist_getzoom()` (but `width` is only incremented if not zoomed)

danomatika: ? It's incremented in the line above. The perhaps rather dumb construction here is to avoid the zoom multiplication unless we're actually zooming.

Lucarda:
Updated to this branch. (on ms-windows)
Same little issue but with [number2], here zoomed with Pd & 300% with windows magnifier.

l3

danomatika:
>Same little issue but with [number2]

Strange. It's workign for me. Perhaps you have an older build.

screen shot 2017-10-02 at 10 04 30 pm

danomatika:
Here's the progress so far.

My test patch no-zoom:

screen shot 2017-10-02 at 10 09 57 pm

My test patch zoomed:

screen shot 2017-10-02 at 10 07 42 pm

Ant1r:
Things seem not completely fixed here, on Ubuntu16.04 (this branch):

not zoomed:
capture du 2017-10-02 23-16-14
zoomed:
capture du 2017-10-02 23-19-26

Lucarda:
Is very strange that [route 1 2 3] has 5 outlets when zommed!

Lucarda:
Couldn't replicate on Ubuntu 16.04:

screenshot from 2017-10-02 19-39-23

screenshot from 2017-10-02 19-39-46

umlaeute:
On 2017-10-02 18:11, Dan Wilcox wrote:
> danomatika commented on this pull request.
> 
> 
> 
>> @@ -1008,11 +1016,16 @@ static void text_getrect(t_gobj *z, t_glist *glist,
>  
>      if (x->te_type == T_ATOM && x->te_width > 0)
>      {
> -        int font = glist_getfont(glist);
> -        int fontwidth = glist_fontwidth(glist),
> -            fontheight = glist_fontheight(glist);
> -        width = (x->te_width > 0 ? x->te_width : 6) * fontwidth + 2;
> -        height = fontheight + 1; /* borrowed from TMARGIN, etc, in g_rtext.c */
> +        width = (x->te_width > 0 ? x->te_width : 6) * glist_fontwidth(glist);
> +        height = glist_fontheight(glist);
> +        if (glist_getzoom(glist) > 1)
> +        {
> +            /* zoom margins */
> +            width += ATOM_RMARGIN * glist_getzoom(glist);
> +            height += ATOM_BMARGIN * glist_getzoom(glist);
> +        }
> +        else
> +            width += ATOM_RMARGIN; height += ATOM_BMARGIN;
> 
> ? It's incremented in the line above. The perhaps rather dumb construction here is to avoid the zoom multiplication unless we're actually zooming.
> 

the problem is that height gets doubly incremented.
if you properly indent the code it looks like:

~~~
if (glist_getzoom(glist) > 1)
{
    /* zoom margins */
    width += ATOM_RMARGIN * glist_getzoom(glist);
    height += ATOM_BMARGIN * glist_getzoom(glist);
}
else
    width += ATOM_RMARGIN;
height += ATOM_BMARGIN;
~~~
so if we are zooming, then height gets incremented by
 "(ATOM_BMARGIN + 1) * glist_getzoom(glist)"
which seems not to be wanted.
if it *is* wanted, i'd suggest changing the code to:
~~~
if (glist_getzoom(glist) > 1)
{
    /* zoom margins */
    width += ATOM_RMARGIN * glist_getzoom(glist);
    height += (ATOM_BMARGIN + 1) * glist_getzoom(glist);
}
else
{
    width += ATOM_RMARGIN;
    height += ATOM_BMARGIN;
}
~~~

Ant1r:
> Is very strange that [route 1 2 3] has 5 outlets

Sorry my issue wasn't related to [route] (which was actually [route 1 2 3 4] so had 5 outlets), but was more about difference of heights behaviour (overlapping of objects was to highlight comparison of heights):
- on the no-zoom version, only 2 different heights exist: number/number2 and object/toggle/bang
- on the zoom one, 4 different heights : number2 < toggle/bang < number < object

Also on Linux the '0' of the not-zoomed number touches the inlet, which is not the case for mac.
And inlets/outlets seem significantly bigger on Linux than on mac (which may be related to the fact globally the font size on Linux seems a bit too small).

src/g_text.c
umlaeute: the problem is that height gets doubly incremented.
if you properly indent the code it looks like:

~~~
if (glist_getzoom(glist) > 1)
{
    /* zoom margins */
    width += ATOM_RMARGIN * glist_getzoom(glist);
    height += ATOM_BMARGIN * glist_getzoom(glist);
}
else
    width += ATOM_RMARGIN;
height += ATOM_BMARGIN;
~~~
so if we are zooming, then height gets incremented by
 "(ATOM_BMARGIN + 1) * glist_getzoom(glist)"
which seems not to be wanted.
if it *is* wanted, i'd suggest changing the code to:
~~~
if (glist_getzoom(glist) > 1)
{
    /* zoom margins */
    width += ATOM_RMARGIN * glist_getzoom(glist);
    height += (ATOM_BMARGIN + 1) * glist_getzoom(glist);
}
else
{
    width += ATOM_RMARGIN;
    height += ATOM_BMARGIN;
}
~~~

danomatika: Ah yeah, I see what you mean. Fixed in both g_text.c and g_rtext.c. That was the cause of the vertical sizing being too large when zoomed.

danomatika:
Vertical sizing when zoomed now fixed in latest commit.

danomatika:
>And inlets/outlets seem significantly bigger on Linux than on mac (which may be related to the fact globally the font size on Linux seems a bit too small).

The inlets are 1 pixel bigger to match the outlets. How does the font/object sizing compare to Pd-extended as the font metrics are borrowed from it?

Lucarda:
Yesterday Ubuntu:
screenshot from 2017-10-02 19-39-46

today;
screenshot from 2017-10-03 14-36-34

today windows:

win1

Lucarda:
These are the slight differences from pd-extended on ms-windows.
I've done Photoshop for layers.
The red ghost is this branch:



zoom 200 with photoshop:



zoom 400:

Ant1r:
I've the same as Lucarda; we now have on the zoomed view 3 different heights: 
number2 < (toggle15 == bang15 == number) < object
while on non-zoomed still:
(number == number2) < (object == toggle15 == bang15)

> How does the font/object sizing compare to Pd-extended?

I'm sorry I don't have Pd-extended anymore installed here... But let's compare the same patch rendered with this branch vs 0.47-1 (not zoomed):

now:
capture du 2017-10-03 22-16-46

0.47-1:
capture du 2017-10-03 22-26-15

this shows that the overall look is slightly less "airy". But of course the spacing is much closer from with older versions, and the harmonization between OSes is indeed a very good thing.

Ant1r:
I've noticed that zoomed inlets are 4 pixels high, while outlets are only 3 (I've added pixel-staircases to highlight that). 
Also, one pixel is still missing at the upper left corner.

number

danomatika:
nbx corner should fixed. As for the io sizing, I'm trying to facilitate the option for Pd-extended style rectangles. The placement is not perfect yet but it's getting closer. Here's the io with a white fill:

non-zoomed

screen shot 2017-10-04 at 10 54 11 am

zoomed

screen shot 2017-10-04 at 10 54 17 am

The issue so far is that the border is not being correctly overlapped when zooming yet.

danomatika:
>I've done Photoshop for layers

The vertical margins in Pd-extended are 2 pixels taller than vanilla. Is that worth looking into again. I thought it was a bit too tall, but I suppose it facilitates the font not touching the IO better. Perhaps I've gotten used to the tighter vanilla boxes. OTOH I don't want to potentially break people's vanilla patches that rely on vertical positioning...

Lucarda:
> OTOH I don't want to potentially break people's vanilla patches that rely on vertical positioning...

Testing with 0.47 gives that [objects] are == to Pd-extended so there won't be “vertical positioning” trouble with older vanilla patches.


Here 
- blue ghost for 0.47-1
- red ghost this branch
- grey for Pd-extended

blue047

Lucarda:
Sorry for my **WRONG** report up here.

The BLUE ghost for 0.47 was measured with a tweaked pdgui.tcl that i was using an year ago.

umlaeute:
@lucarda: you can remove your posts in the issue by clicking on the `x` sign of your posts header. i think this is better than to keep misleading information and additional "please ignore" posts which might easily be overlooked (or misunderstood to which post they actually refer to)

Lucarda:
@umlaeute 
Thanx.

umlaeute:
@Lucarda :-) but please also delete your "thanks" post.

danomatika:
After using this a bit, I decided to revert the IEM gui io height back to 2 and use the Pd-extended vertical margins which seem to adjust the font positioning better to where the object's text contents do not touch the inlets. Descender characters like 'p' do touch the outlets but I see that as less of a visual/readability problem and Pd-extended exhibited this behavior already.

For comparison, here is my test patch on Pd-extended and this branch:

diff

Lucarda:
Troubles with 65e152a **(Pd-extended vertical margins)** on Windows & Ubuntu:

Windows:

with-65e152a-windows


Ubuntu:

with-65e152a-ubuntu

danomatika:
Ok, I reverted the vertical margins again. I see now the tweaks were probably based on how Monaco is rendered on Mac and courier is rendered on win/linux. I think its better, overall to have consistent margins between all platforms.

For comparison, here's a gif showing my test patch in Pd-extended, this branch with Monaco (like extended), and this branch with Deja Vu Sans Mono. The Monaco metrics are clearly taller whereas DVSM seems to be a bit tighter vertically:

diff2

Lucarda:
Good:

Windows:
wgood

Ubuntu:
ugood

danomatika:
Latest commit adds my test patch: `docs/7.stuff/tools/sizingtest.pd`

Ant1r:
Looks good on Ununtu16.04! And it zooms well too.
capture du 2017-10-12 21-34-34

there's a pixel missing at the right of vslider's knob, and at the bottom of hslider's one:
capture du 2017-10-12 21-30-51

When zoomed, the gap is both sides of the knob:
capture du 2017-10-12 21-31-05

danomatika:
>there's a pixel missing at the right of vslider's knob

Yeah, I did that on purpose as the original knob overlapped the right/button border.

danomatika:
>Looks good on Ununtu16.04!

One issue I see is that the "0" in the number boxes seems to touch the inlet on Ubuntu but not on Mac or Windows. I wonder how it looks with the smallest font size. I *really* don't want to (re)introduce platform margin hacks, so maybe we should revisit a some extra vertical padding? I dunno. Maybe I'm staring at pixels too much at this point...

Ant1r:
> "0" in the number boxes seems to touch the inlet on Ubuntu

it actually touches it with font size 8 or 10; from 12 there's at least one pixel space. 

> I really don't want to (re)introduce platform margin hacks

don't!

 

> maybe we should revisit a some extra vertical padding?

maybe just one more pixel above the text? only one...

Ant1r:
About sliders knob, I' ve found that changing (here for vslider_draw_update):
```
        sys_vgui(".x%lx.c coords %lxKNOB %d %d %d %d\n",
                 glist_getcanvas(glist), x, xpos + IEMGUI_ZOOM(x), r,
                 xpos + x->x_gui.x_w - IEMGUI_ZOOM(x), r);
``` 
to:
```
        sys_vgui(".x%lx.c coords %lxKNOB %d %d %d %d\n",
                 glist_getcanvas(glist), x, xpos + 1, r,
                 xpos + x->x_gui.x_w - IEMGUI_ZOOM(x) + 1, r);
```
was doing the trick (no overlap, no gap, either zoom). Is this working on your system(s)?

danomatika:
Your fix has the right edge overlapping into the border. I find the following to work the best (more or less):

~~~
sys_vgui(".x%lx.c coords %lxKNOB %d %d %d %d\n",
                 glist_getcanvas(glist), x, xpos + 1, r,
                 xpos + x->x_gui.x_w - IEMGUI_ZOOM(x), r);
~~~

danomatika:
>maybe just one more pixel above the text? only one...

@Ant1r How does it look into win/linux if you set TMARGIN to 3 in g_rtext.c?

Ant1r:
> Your fix has the right edge overlapping into the border for me. I find the following to work the best (more or less)

so definitively TclTk doesn't render the same on different OSes (sorry if I seem to discover this now...).

your version:
vsl1_2 vsl1_1 
mine was:
vsl0_2 vsl0_1
Maybe we could try to _raise_ the outline, so knob overlapping would turn to "underlapping" (which wouldn't be visible).

--------
About number box:
> TMARGIN to 3 in g_rtext.c

see for fontsize 8-10-12:
nbr_tmargin3_fontsize8 nbr_tmargin3_fontsize10 nbr_tmargin3_fontsize12

now the "0" touches the outlet. Changing BMARGIN seems having no effect on number box.

Lucarda:
sizingtest.pd on Windows, all good:

sizingtest-win

Ant1r:
> raise the outline

obviously this doesn't work, i.e. creating the BASE after just totally hides the KNOB...

danomatika:
Yeah. I opted initially for "close enough" as it's really a detail.

danomatika:
For trying things out:

* object sizing margins are at the top of g_rtext.c
* atom margin is at the top of g_text.c

I'm using a build right now that has top and bottom margins of 3 which provides a little more space between the text and the io. Also keeps the boxes from seeming too slim and they are easier to read. Of course, I wonder how dependent upon the font this (subjective) view is.

Ant1r:
Ok, thanks for clarifying.
For me the best setting is:

g_rtext.c
#define TMARGIN 3
#define BMARGIN 2

g_text.c:
#define ATOM_BMARGIN 5

Then objects, messages and atom have exactly the same height, and there is always at least one pixel space above and below the text, whatever the font size.

danomatika:
Ok, I added 1 to the vertical margins. I think we should keep the slight difference in the atom sizing as it helps make them more distinct from the object boxes.

Ant1r:
Yeah, great.
In font size 8 the 'p' in a symbol atom is a bit too much eaten by the outlet, but it's not that serious.
capture du 2017-10-14 11-56-11

porres:
hi, would this be ready to go?

danomatika:
>In font size 8 the 'p' in a symbol atom is a bit too much eaten by the outlet, but it's not that serious.

I'm inclined to let that go for now as I think many of us are used to the float and symbol atoms being slimmer.

>hi, would this be ready to go?

It pretty much is. I've been using a build of 0.48 with this branch for a while now.

One issue we're missing is array zoom sizing, but maybe that can be another branch in the future?

porres:
> One issue we're missing is array zoom sizing, but maybe that can be another branch in the future?

sounds good to me.

danomatika:
Array zooming is more or less working already, actually. I've now added thicker lines when zoomed.

A couple of issues I'm seeing:

* Adding an array to a canvas and then zooming works fine, but for some reason the arrays in C08.analog.sequencer.pd don't zoom.
* Opening an array into it's own canvas, zooming, then closing resizes the GOP in the parent patch regardless of the parent's zoom level. IMO the sub window should allow zooming (or open using the zoom level of the parent) but keep the zoom level of the parent when closing.

danomatika:
A possibly related bug to array GOPS:

1. open a patch with a GOP, zoom, then open the GOP -> sub window is zoomed
2. unzoom sub window
3. close subwindow -> GOP now resized to it's unzoomed size while GUI contents are not

This is the result of doing the above with sizingtest.pd:

screen shot 2017-11-04 at 2 48 29 am

danomatika:
Latest commit fixes GOP zoom level not matching parent's after closing sub window.

danomatika:
Nevermind, GOP array objects are *not* sizing correctly when zoomed yet.

danomatika:
Ok, GOP array zooming is now working.

There was a comment about not sending zoom messages to "old style GOPs" but this was what was keeping the arrays from zooming, so I've removed that filter. Will this be a problem? I've not noticed anything detrimental in my testing so far.

danomatika:
Note: Data structures currently do not zoom and I'm not willing to tackle that now. I've gone ahead and made sure the zooming in g_template.c is only used for graph objects aka GOP arrays.

porres:
awesome, can't wait for this merge

Lucarda:
>Ok, GOP array zooming is now working.

Super!
I know you know this was not working yesterday. Screen on w32:

arrayzoom

Ant1r:
Really nice. Even array inside GOP inside GOP works!

PR #233, opened 2017-10-10
Nicer to read Pd Manual with "Sans Serif" font

Lucarda:
Pd html manual with "Sans Serif" font

Also changed to html5 and new CSS.

### DIFF
--- a/doc/1.manual/index.htm
+++ b/doc/1.manual/index.htm
@@ -1,4 +1,4 @@
-<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01//EN">
+<!DOCTYPE html>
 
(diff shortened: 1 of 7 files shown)


millerpuckette:
Are 232 and 233 (this one) the same, or are they alternatives, or are they independent?  I don't know what's going on here.

Lucarda:
> Are 232 and 233 (this one) the same, or are they alternatives, or are they independent?

The main difference is just the Font used. They both share the same update to HTML5 (all modern browsers support it since long ago).

You can choose any you like most about the font  (I made 2 PRs just to be able to preview them). Later it will be very easy to change the font-family, font-sizes and link-colors and other cosmetics.

PR #241, opened 2017-10-29
value right inlet

danomatika:
This PR adds a right inlet to the [value] object as [suggested on the pd-list](https://lists.puredata.info/pipermail/pd-list/2017-10/120875.html).

Here's a test patch:

screen shot 2017-10-31 at 2 14 16 pm

~~~
#N canvas 169 142 450 300 10;
#X obj 306 159 value b;
#X obj 86 132 bng 15 250 50 0 empty empty empty 17 7 0 10 -262144 -1
-1;
#X obj 86 132 bng 15 250 50 0 empty empty empty 17 7 0 10 -262144 -1
-1;
#X obj 306 130 bng 15 250 50 0 empty empty empty 17 7 0 10 -262144
-1 -1;
#X obj 306 130 bng 15 250 50 0 empty empty empty 17 7 0 10 -262144
-1 -1;
#X floatatom 86 190 5 0 0 0 - - -, f 5;
#X floatatom 306 189 5 0 0 0 - - -, f 5;
#X floatatom 185 39 5 0 0 0 - - -, f 5;
#X floatatom 185 137 5 0 0 0 - - -, f 5;
#X obj 159 57 bng 15 250 50 0 empty empty empty 17 7 0 10 -262144 -1
-1;
#X obj 159 57 bng 15 250 50 0 empty empty empty 17 7 0 10 -262144 -1
-1;
#X obj 86 160 value a;
#X obj 185 105 value;
#X msg 232 37 symbol a;
#X msg 238 59 symbol b;
#X msg 245 81 symbol c;
#X connect 0 0 6 0;
#X connect 1 0 11 0;
#X connect 3 0 0 0;
#X connect 7 0 12 0;
#X connect 9 0 12 0;
#X connect 11 0 5 0;
#X connect 12 0 8 0;
#X connect 13 0 12 1;
#X connect 14 0 12 1;
#X connect 15 0 12 1;
~~~

### DIFF
--- a/src/x_connective.c
+++ b/src/x_connective.c
@@ -1567,8 +1567,7 @@ void value_release(t_symbol *s)
  * value_getfloat -- obtain the float value of a "value" object
  *                  return 0 on success, 1 otherwise
  */
-int
-value_getfloat(t_symbol *s, t_float *f)
+int value_getfloat(t_symbol *s, t_float *f)
 {
     t_vcommon *c = (t_vcommon *)pd_findbyclass(s, vcommon_class);
     if (!c)
@@ -1581,8 +1580,7 @@ value_getfloat(t_symbol *s, t_float *f)
  * value_setfloat -- set the float value of a "value" object
  *                  return 0 on success, 1 otherwise
  */
-int
-value_setfloat(t_symbol *s, t_float f)
+int value_setfloat(t_symbol *s, t_float f)
 {
     t_vcommon *c = (t_vcommon *)pd_findbyclass(s, vcommon_class);
     if (!c)
@@ -1615,6 +1613,13 @@ static void value_float(t_value *x, t_float f)
     *x->x_floatstar = f;
 }
 
+static void value_set(t_value *x, t_symbol *s)
+{
+    value_release(x->x_sym);
+    x->x_sym = s;
+    x->x_floatstar = value_get(s);
+}
+
 static void value_ff(t_value *x)
 {
     value_release(x->x_sym);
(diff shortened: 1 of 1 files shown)


porres:
awesome :)

umlaeute:
what's the advantage of using a `set` message over a second inlet?

(i think a `set` message is kind of problematic if we ever want to extend `[value]` to be able to hold other values than floats)

Spacechild1:
I second IOhannes on that. also, a second inlet would mirror the behaviour of objects like [send] and the interface of [array] and [text].

porres:
> (i think a set message is kind of problematic if we ever want to extend [value] to be able to hold other values than floats)

now that we're at it, I ALWAYS HOPED THAT IT'D HOLD OTHER VALUES THAN FLOATS :)

umlaeute:
although i don't think that capitalisation will make it happen any second sooner...

danomatika:
It depends on your expectations. There are objects which take a set message and others which provide an inlet.

IMO an object called "value" only works with numbers. If there would be a similar object for symbols, I'd imagine it would have a (slightly) different name.

In any case, either method is  it that much code as you can see in the first commit.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Ok. The set message has now been changed to an optional right inlet.

Note: As far as I understand the object messaging system, sending into the second inlet essential sends a message to the object which can also be sent via the main left inlet (if you know the message name). For that reason, I changed the message from "set" to "symbol2" to partially hide it as well as follow the naming conventions for similar inlet function class in other core objects.

porres:
> IMO an object called "value" only works with numbers. If there would be a similar object for symbols, I'd imagine it would have a (slightly) different name.

[var]?

Spacechild1:
@danomatika  a trick I've seen is to use an empty symbol ("") as the selector to make sure the method is only called via the right inlet and not via a message to the left inlet.

umlaeute:
fwiw: while it is complicated to create a selector that is an empty symbol (""), it's not impossible (however, i've been using this excessively :-))

Issue #243, opened 2017-11-02
Multi instance support for [text define]


pierreguillot:
When I use libPd with PDINSTANCE and PDTHREADS and I use pd_instance pdinstance_new(), if I load a patch with [text define], it crashes in [text-define_new()](https://github.com/pure-data/pure-data/blob/master/src/x_text.c#L280) because x->x_scalar is NULL (because in [scalar_new()](https://github.com/pure-data/pure-data/blob/master/src/g_scalar.c#L113), the function template_findbyname(templatesym) returns NULL). So it seems that the function can't find a template linked with the symbol "pd-text". 

After several investigation, I found that at initialization, pd_init() calls [conf_init()](https://github.com/pure-data/pure-data/blob/master/src/m_pd.c#L311)  that calls [x_qlist_setup()](https://github.com/pure-data/pure-data/blob/master/src/m_conf.c#L91) that calls [text_template_init()](https://github.com/pure-data/pure-data/blob/master/src/x_text.c#L1970) etc. and that ends by calling [template_new()](https://github.com/pure-data/pure-data/blob/master/src/g_template.c#L154) that binds a template to the the symbol "pd-text". 

This binding is only valid at initialization and is not duplicated when using new pdinstances that why it crashes. And I guess there is the same problem with the symbols "pd-float" and "pd-float-array" with garray_init().

Here is the discussion (with myself :) ) on [pd-dev list](https://lists.puredata.info/pipermail/pd-dev/2017-11/021313.html). And the initial [issue](https://github.com/libpd/libpd/issues/196) on the libpd repository.

I create a PR to correct the bug but as I'm not sure that this is the best solution, I also prefer to create this issue.

danomatika:
This would be good to have in 0.48-1 and libpd needs it as well.

pierreguillot:
PR Merged

PR #245, opened 2017-11-02
Fix multi instance support for [text define] & [array define]

pierreguillot:
The correction solves this [issue](https://github.com/pure-data/pure-data/issues/243) and is pretty straightforward but I'm not sure that's the best approach. If @millerpuckette or another Pd master could check it. (and sorry for the noise)

### DIFF
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -136,8 +136,6 @@ and float-arrays. */
 void garray_init( void)
 {
     t_binbuf *b;
-    if (garray_arraytemplatecanvas)
-        return;
     b = binbuf_new();
 
     glob_setfilename(0, gensym("_float_template"), gensym("."));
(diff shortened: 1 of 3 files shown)


pierreguillot:
In fact I'm not sure that I totally solved the problem

umlaeute:
@pierreguillot could you elaborate on that?

pierreguillot:
Sorry, I just wanted first to note that the problem persisted. 

So, here is the error message:
```
array: couldn't find template pd-float-array
consistency check failed: array_define_yrange
array: couldn't find named array '1001-env'
...
```
It's caused by the 2nd lines of the function `text_template_init()` and `garray_init()` that returns if the function has already been called (but we need to call the function for each pdinstance). So I have to remove these lines (the functions are only called at initialization and now with this PR with `pdinstance_new()` so it won't be a problem I guess).

PR #249, opened 2017-11-07

Ant1r:
(was previously #226, now redirected to master)
This is only a GUI change, and doesn't modify any variable name for e.g, so it shouldn't impact compatibility, excepting patches or plugins or externals directly calling pdtk_path_dialog or pdtk_startup_dialog (but are they numerous?).

### DIFF
--- a/src/s_path.c
+++ b/src/s_path.c
@@ -632,7 +632,7 @@ void glob_start_path_dialog(t_pd *dummy)
      char buf[MAXPDSTRING];
 
     sys_set_searchpath();
-    sprintf(buf, "pdtk_path_dialog %%s %d %d\n", sys_usestdpath, sys_verbose);
+    sprintf(buf, "pdtk_path_dialog %%s %d\n", sys_usestdpath);
     gfxstub_new(&glob_pdobject, (void *)glob_start_path_dialog, buf);
 }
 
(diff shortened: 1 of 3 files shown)


danomatika:
Since this touches the FUDI messaging between core and GUI, I think this is less likely to be accepted. We don't know what might be relying on that value being with the patch dialog.

To a certain perspective, having the verbose button where it is makes sense in that most of the prints you see involve printing the search paths...

Ant1r:
> We don't know what might be relying on that value

I've no problem if you want to close this PR, but just for my understanding, what could use this value? GUI plugins for instance? Could some externals rely to this too?

danomatika:
>I've no problem if you want to close this PR, but just for my understanding, what could use this value? GUI plugins for instance? Could some externals rely to this too?

I don't know. libpd, for instance, can open a connection to the gui. It's relatively likely that a app using an older version of libpd may try to talk to a newer Pd gui. Even though libpd doesn't really use the GUI's settings, I'm not sure what would happen if you opened and closed the dialog.

Ant1r:
Ok thanks, let's close it. Really not a big deal.

danomatika:
This is something I wanted to fix with a united preferences dialog, but it will have to wait until the next version of Pd.

PR #250, opened 2017-11-07
update french translation

Ant1r:
(was previously #221, simplified and redirected to master)
Fixes, typos and complements.


Ant1r:
I don't feel "Propriétés de la boîte Atom" very natural, but you're right, we cannot keep "Nombre". 
Why not just "Propriétés"?

pierreguillot:
I don't think that **atom** is more "natural" in english and it's coherent with the Pd documentation while just **Propriétés** is not informative. Otherwise what about **Propriétés de la boîte Nombre/Symbole** ? But I'm not very good at naming things (I called by my Pd plugin "Camomile", you can imagine my degree of efficiency in such domain :) ).

PR #256, opened 2017-11-23
use more 'const' qualifiers in functions (and t_symbol)

umlaeute:
this PR adds `const` to a lot of exported functions, allowing the compiler to perform better optimisations and the userdev to write safer code.

it also declares the `s_name` member of t_symbol to be `const` (which it really is; i've seen crashing bugs in the last 15 years, because people would start to modify the string they get from a symbol)


danomatika:
Sounds good. Eliminate failure sources *before* runtime...

danomatika:
When this is further along, I should test it against libpd.

pierreguillot:
Perhaps I'm wrong but if you change the prototype of the functions in *m_pd.h*, the externals that use these functions won't be compatible anymore. Am I right or wrong?

danomatika:
The won't build only if they are trying to do things the API really shouldn't support, such as modifying const things.

pierreguillot:
I have supposed that the loading results "unresolved external symbol" because the prototypes of the functions has changed so what you tell is that const doesn't make any difference for this.

umlaeute:
@pierreguillot  the problem would only exist if the declaration where exported in a C++-context (where the names get mangled to hold a signature of the actually passed types in order to support polymorphism).
Since Pd's API is only C (and therefore all its symbols are explicitely *not* mangled for polymorphism), there shouldn't  be a problem.

pierreguillot:
Thanks for the clarification :)

danomatika:
Can we bring this up to date for the 0.49 release cycle?

umlaeute:
it would probably be simpler to restart from scratch, but i might want to give it a try.

the problem with this patch is that it is quite evasive and thus fragile.

millerpuckette:
merged.  This touched a lot of lines so will probably cause merge conflicts with other PRs, but if I delayed merging this it would be the other way around I think,

umlaeute:
thanks for merging. indeed the touching of so many lines was the main problem with keeping this PR mergable in the first place.

PR #265, opened 2017-12-04
updated the make app target to build Tcl/Tk 8.5 frameworks

danomatika:
This PR updates the `make app` autotools make target to download and build the Tcl/TK 8.5 frameworks to match that of the current Pd release builds.

This makes building an equivalent build to the releases require only a few steps, namely:

    ./autogen.sh
    ./configure
    make
    make app

The ARCH_CFLAGS used when configuring are passed to the `mac/tcltk-wish.sh` script to ensure Wish is built with the same architecture that Pd was configured for.

An existing Wish is checked for before building to ensure a prebuilt Wish is re-used. One thing to note is if someone is scripting different architecture builds, the Wish app should be re(moved)/renamed before running `make app` to make sure the wrong Wish architecture is not used.

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -228,9 +228,9 @@ To build the Pd macOS application, simply run:
 
     make app
 
-This builds Pd-#.#.#.app in the Pd source directory which can be then be
-double-clicked and/or copied to /Applications. For more info &
-options regarding the Pd .app bundle, see mac/README.txt
+This builds a Pd-#.#.#.app with embedded Tcl/Tk frameworks in the Pd source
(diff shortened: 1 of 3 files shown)


danomatika:
One possible problem with this is that the default "make app" target now relies on an internet connection since we download and build Tk 8.5.

danomatika:
@millerpuckette I think I'm going to stow this for now as requiring an internet connection when doing the `make app` build target is not really ideal.

I think the best solution is to update the stuff.tgz with the new version of the Wish.app that is used with the current releases. This way people build Pd on their own know they have the same thing as the release for testing, etc.

PR #266, opened 2017-12-04
updated miditester to print channel from channel objects

danomatika:
A small PR which updates the miditester patch to print from the channel objects (notein, ctlin, etc). Also, simplifies the print headers.


danomatika:
This was missed in latest bug fix branch round.

PR #268, opened 2017-12-05
bugfix/helpfiles

umlaeute:
this contains a few fixes to help-patches (mostly to make the "test audio and midi" patch loadable multiple times without red glaring errors).

it also has the .po files regenerated.


umlaeute:
this has either been applied independently or can be found in the update/0.48-2 PR.

PR #272, opened 2017-12-17
Update expr-help.pd

porres:
This is just a minor update that is mostly aesthetical as it now aligns to the new font size from 0.48-1


porres:
cool, I'm now just waiting to see if Shahrokh provided a new version (which was promised before the end of 2017), so I can send an updated help file

cheers

PR #275, opened 2017-12-21
[WIP] automake settings for MinGW (building pd.dll without libtool)

Spacechild1:
alternative version of https://github.com/pure-data/pure-data/pull/262 without libtool (because it didn't allow `-static-libstc++` and `-static-libgcc` for building pd.dll)

bonus: significantly faster build times. 

works fine on Windows (with ASIO). can someone test on Linux and OSX?

EDIT: 
also adds a windows specific install target (`install-windows`) to create a standalone package from pdprototype.tgz.


Lucarda:
On Msys1.0 builds fine with no ASIO.

The dependancies are these 2:

- pthreadGC-3.dll
- libgcc_s_dw2-1.dll

With ASIO here's the error log file:
[msys1-asio-error.txt](https://github.com/pure-data/pure-data/files/1577704/msys1-asio-error.txt)

Spacechild1:
thanks. it's WINVER again, I'll fix it.
BTW: msys1 is actually outdated, it's better to use msys2.

Lucarda:
Ok it build fine with Msys 1 & 2

Thanks for the “install-windows”!!!

These are dependancywalker results running the builds on a clean machine:

Msys2 (everything fine)

msys2



For Msys1.0 **pthreadGC-3.dll** is needed which calls **libgcc_s_dw2-1.dll**, this one seems to have nothing to do with how **pd.dll** was compiled, it seems to be inseparable from pthreadGC-3.dll

msys1


**Notes:**
On both builds the menu media
- standard (MMIO)
- ASIO (via portaudio)
had disappear.

menu1


But we can select audio driver via audio settings...

menu2

Spacechild1:
Did you configure with both --enable-asio and --enable-mmio ?

Lucarda:
Oops, nope, just `./configure --enable-asio`
: - )

Spacechild1:
NOTE: travis CI fails at configure stage for some reason. Makefile.in is not build although I added the /msw/Makefile to configure.ac. It works fine when I run ./autogen.sh ./configure etc. on my machine, though...

Spacechild1:
General opinion of mine: I'm not sure we should try to fix dependencies for old/outdated MinGW distributions. The branch seems to work fine for recent distributions like Msys2 and I think that's good enough.

Lucarda:
I agree.
We can update or create documentation on puredata.info on how to build with Msys2 and also mention the especial needs when building with Msys1.0

Also include this info on INSTALL.txt

umlaeute:
the build is failing on travis; could you have a look?

msw/Makefile.am
```diff
@@ -0,0 +1,22 @@
...
+##### Targets #####
+
+.PHONY: install-windows
+
+if WINDOWS
+
+# make a standalone distro on Windows
+install-windows:
+	make -C $(top_builddir) install prefix=$(prefix)/pd-$(VERSION)
+	$(top_builddir)/msw/msw-install.sh $(DESTDIR)$(prefix)/pd-$(VERSION)
```
umlaeute: this should most likey have `$(top_srcdir)`

Spacechild1:
@umlaeute as I wrote above:

> NOTE: travis CI fails at configure stage for some reason (Makefile.in is not build although I added the /msw/Makefile to configure.ac). It works fine when I run ./autogen.sh ./configure etc. on my machine, though...

do you know why?

umlaeute:
it seems that travis is not building the latest commit but  b6c27ac instead (i have no idea why)

umlaeute:
i've a (couple of) fix(es ) - where can i push them?

Spacechild1:
great!
you can make a PR to my branch https://github.com/Spacechild1/pure-data/tree/mingw-autotools2
or push it somewhere on your git so I can pull it back in. whatever you consider best practice...

Spacechild1:
I see a slight problem for exec_prefix=/pd/ and later renaming to pd-$VERSION. let's assume DESTDIR already contains a folder named /pd (maybe some earlier install or totally unrelated). the user might expect that `make install-windows` would install into DESTDIR/pd-0.48.1 while not touching the DESTDIR/pd folder.
two suggestions:
a) use something less generic for exec_prefix and later rename to either /pd oder /pd-$VERSION
b) warn the user if DESTDIR/pd already exists

Spacechild1:
c) set ecex_prefix already to the final folder name (less elegant but safest option IMHO)

Lucarda:
On INSTALL.txt

How about adding:

To install toolchains for msys2 do:
`pacman -S gcc make autoconf automake libtool`

Spacechild1:
@danomatika  could you also have a look when you have time? works pretty fine already.

INSTALL.txt
```diff
@@ -273,7 +273,29 @@ aliases to provide the pd command.
 
 ### Windows
 
-*Windows-specific info goes here*
+Pd on Windows is built with either MinGW or Cygwin.
+
+available audio APIs which can be enabled/disabled via ./configure are:
```
danomatika: Captital "A" here.

INSTALL.txt
```diff
@@ -273,7 +273,29 @@ aliases to provide the pd command.
...
+* JACK			--enable-jack --disable-jack
+
+Note: because of license restrictions, Pd can't ship the necessary ASIO files. 
+If you want to use ASIO, have a look at /asio/README.txt for further instructions.
+
+The standard 'make install' won't work outside your Cygwin/Msys2 environment (if at all).
+There is, however, a special target called 'make install-windows' for a self-contained 
+Pd package which should run out of the box. After 'make', just do:
+
+make install-windows DESTDIR="/path/to/pd"
```
danomatika: Use 4 leading spaces here as this is terminal command. The document is basically written in Markdown format and I used indentation for terminal commands.

INSTALL.txt
```diff
@@ -273,7 +273,29 @@ aliases to provide the pd command.
...
+
+available audio APIs which can be enabled/disabled via ./configure are:
+* MMIO (slow!) 	--enable-mmio --disable-mmio
+* ASIO			--enable-asio --disable-asio
+* JACK			--enable-jack --disable-jack
+
+Note: because of license restrictions, Pd can't ship the necessary ASIO files. 
+If you want to use ASIO, have a look at /asio/README.txt for further instructions.
+
+The standard 'make install' won't work outside your Cygwin/Msys2 environment (if at all).
```
danomatika: Can we use double quote around "make install" etc to match the rest of the document?

danomatika:
Initial review:

* I noted some small formatting issues in the INSTALL.txt
* We could add a little more detail in the steps/packages needed to get an Msys2 + MinGW build environment set up. The idea being to provide interested beginners with enough info as well as provide a relatively canonical set of instructions. Here are my own notes when I setup my Windows test system in the summer which could be added/edited to match the overall macOS & Linux sections:

~~~
### MinGW

Using x86_64 Mysys2 installer: http://www.msys2.org/

Note: Msys2 development seems to change frequently, so some of the package names
      may have changed after this document was written,

* Install to default location (C:\msys, C:\msys64, etc)
* Follow the setup/update info on the Msys2 webpage
* Run msys & update package database & core runtimes:

    pacman -Syu

* Close the terminal window, reopen, & run the command again to update the system.

_Following instructions are for 64-bit. For 32-bit, change all instances of "x86_64" to "i686" in the package names._

Install the developer toolchain:

    pacman -S mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang

Install make and autotools:

    pacman -S make pkgconfig autoconf automake libtool

Install gettext for translations:

    pacman -S mingw-w64-x86_64-gettext 

Install git (when clonding the Pd sources from githib):

    pacman -S git

Install nsis installer tool (for building the Pd installer):

  pacman -S mingw-w64-x86_64-nsis
~~~

* I see `-DWINVER=0x0502 -D_WIN32_WINNT=0x0502` is being set in multiple Makefile.ams. I think it would be better to move that into configure.ac and perhaps exported as a Makevariable that is then used conditional where it's needed. This way it only needs to be changed once in the future.

Spacechild1:
thanks! you're welcome to edit INSTALL.txt if you want, you have write access to this branch.

danomatika:
@umlaeute At some point, probaly not in this PR, we should consider changing the portmidi and portaudio convenience libs to build *without* libtool. They are static only, so there isn;t really a need to use libtool and it would speed up the windows build.

danomatika:
More feedback:

Build
------

Build is definitely faster!

Seeing this automake warning:

~~~
src/Makefile.am:11: warning: variable 'pd_core_LDFLAGS' is defined but no program or
src/Makefile.am:11: library has 'pd_core' as canonical name (possible typo)
~~~

This is because it's expecting the "pd_core" naming to match a given bin or lib. I think it just needs a dummy libpd_core to be defined a with the first mingw-autools branch. I'll look into this.

App/Install-Windows
-----------------------

The msw-install.sh should use a default location so `make install-windows` doesn't fail, but generates a pd/$VERSION folder in the top build directory. This is similar to how `make app` works on macOS and removes a requirement beginners may (and probably will) miss. I'm also a believer of setting defaults wherever possible while providing overrides.

I feel like the `make install-windows` target could be renamed `make app` as it basically does the same action as `make app` on macOS: copy/install Pd into an application subfolder as needed by the OS -> macOS: "Pd-0.48.1.app" & Windows: "pd-0.48.1"

The missing $DESTDIR detection in msw-install.sh didn't work for me as running `make install-windows` gave me the following:
~~~
C:\msys64\mingw64\bin\strip.exe: '0.48.1/pd/bin/pd.exe': No such file
C:\msys64\mingw64\bin\strip.exe: '0.48.1/pd/bin/pd.com': No such file
C:\msys64\mingw64\bin\strip.exe: '0.48.1/pd/bin/pd.dll': No such file
mv: cannot stat '0.48.1/pd/lib/pd/*': No such file or directory
cp: cannot stat '../msw/../extra/*/*.c': No such file or directory
make[1]: Leaving directory '/home/wilcox/src/pure-data/msw'
~~~

Running
---------

Crash on startup when running pd.exe.

I can start the gui fine using wish:

    test/pd-0.48.1/bin/wish85.exe test/pd-0.48.1/tcl/pd-gui.tcl

but Pd crashes when turning on DSP.

Spacechild1:
> Crash on startup when running pd.exe.

can you try to run pd.exe in the console and see what errors you get?
how did you configure? which Windows verson? just so that I can try to reproduce this.

> but Pd crashes when turning on DSP

which audio API?

Spacechild1:
> The msw-install.sh should use a default location so make install-windows doesn't fail, but generates a pd/$VERSION folder in the top build directory.

you're right, I'll have a look.

danomatika:
>>The msw-install.sh should use a default location so make install-windows doesn't fail, but generates a >>pd/$VERSION folder in the top build directory.
>
>you're right, I'll have a look.

I handled this when building the macOS app by using the `mac` directory as a temp directory and them moving the finished directory to the top build directory in the end. That ended up making some of the path stuff more complicated though, but it works.

danomatika:
I have the fix for the automake warning: simply use "core" as the suffix so automake doesn't think any of the variables associated with libs or binaries. I'll do a push soon with this.

Spacechild1:
cool! I'll fix the install script in the evening. in the meantime you could post the crash report with your system info.
@Lucarda do you get a crash? on which system did you test?

danomatika:
@umlaeute I decided to try switching the local portaudio & portmidi to static (non-libtool) convenience builds and it works fine. Much faster. I think I'll go ahead and add this change to the PR as it should be fine on the other platforms since it's static only.

Now building just the extra folder takes almost 2x as long as building Pd itself thanks to libtool. Oh well, at least it's not 30+ minutes!

Good job @Spacechild1 and @Lucarda . I'm glad y'all have stepped up to help on Windows as splitting the work makes us all more effective.

Spacechild1:
> Now building just the extra folder takes almost 2x as long as building Pd itself thanks to libtool.

exactly :-D

just to be sure: do you still get a crash when running Pd?

danomatika:
>just to be sure: do you still get a crash when running Pd?

Double-click pd.exe: There is no real crash report, just the error dialog saying the program needs to terminate, weirdly in German for my English system.

Running Pd from the command line works fine, so it may be a path issue...

Spacechild1:
very strange... which system are you on?

danomatika:
>very strange... which system are you on?

Windows 10 Pro 64bit, Version 1709, Build 16299.125

danomatika:
Latest commits do the following:

* build portaudio, portmidi, & asio as static libs without libtool
* enable mmio and asio by default on Windows, this matches Miller's release
* asio header check is now a warning
* "ASIO" is now added to the audio api list in the configure output

danomatika:
Note @millerpuckette: We've updated the ASIO sdk source folder naming to "ASIOSDK" from the "ASIOSDK*.*" naming from Steinberger so version updates are less of a hassle. I've updated `src/makefile.mingw` and `msw/mingw-compile.sh` to reflect this.

danomatika:
I have an adapted msw-install.sh working. I'll throw it up soon with a `make app` target.

danomatika:
Ok. I added a msw-app.sh script adapted from msw-install.sh & osx-app.sh which matches the behavior of the latter. There is now a simple make target which builds the app folder using the current version via:

    make app

umlaeute:
regarding `AS_IF` on the long run we probably should; it's not very high on my priority list, so i usually just fix it as i pass by.

regarding build time: your numbers are amazing; when i do an ordinary cross-build (mingw64 on Debian/amd64; 16GB RAM + SSD disk; 4* Intel i5 @ 2.5GHz) with `make -j4`, it takes:
- 21 seconds (using 'makefile.mingw'
- 38 seconds (using autotools on 'master')
- 16 seconds (using autotools on this branch)
all measurements are without ASIO, without disk warmup and only taking a single run.
anyhow: while the speedup is considerable, I never considered a 40 seconds build time to be problematic (unlike 30 minutes)

Lucarda:
> @Lucarda do you get a crash? on which system did you test?

No, 

Built on win8.1 Msys1 & Msys2.
Both tested on another Win8.1 machine.

I will build this branch in a couple of hours.

Lucarda:
Built with Msys2:

```
./autogen.sh
./configure --enable-mmio –enable-asio
make install-windows prefix=c:/auto-enero1
```

Everything compiles but there is nothing in "auto-enero1" (install-windows):

```
../msw/msw-install.sh  0.48.1
installing to: 0.48.1
C:\msys64\mingw32\bin\strip.exe: '0.48.1/pd/bin/pd.exe': No such file
C:\msys64\mingw32\bin\strip.exe: '0.48.1/pd/bin/pd.com': No such file
C:\msys64\mingw32\bin\strip.exe: '0.48.1/pd/bin/pd.dll': No such file
mv: cannot stat '0.48.1/pd/lib/pd/*': No such file or directory
cp: cannot stat '../msw/../extra/*/*.c': No such file or directory
make[1]: Leaving directory '/c/Users/Lucarda/Downloads/pure-data-mingw-autotools2/msw'
```


Then i did:
`make app`

I got a **nice** pd-0.48.1 folder.

Runs fine with win10 & win8.1. (no crashes)

danomatika:
>make install-windows prefix=c:/auto-enero1

That should be: `make install-windows DESTDIR=c:/auto-enero1 VERSION=0.48.1`

Whoops, I think I removed the install-windows instructions from INSTALL.txt.

Lucarda:
@danomatika 

About your crashes:

Are you sure you are compiling with the Msys2 MinGW 32 bit shell?
And all your mingw-pkgs are 32 bit?

danomatika:
>Are you sure you are compiling with the Msys2 MinGW 32 bit shell?
>And all your mingw-pkgs are 32 bit?

I'm compiling 64 bit with a MinGW 64 bit shell. Packages are all 64 bit. Since the GUI is separate from the core, a 32 bit GUI works fine with a 64 bit core. Running it works fine from the command line, it's just double-clicking that fails for some reason...

Lucarda:
I think you can not load any external with a 64bit Pd.
Pd on windows is 32bit only.

danomatika:
If the externals are built 64 bit, it would work fine. It's not that Pd *has* to be 32 bit still... but that's another issue. I'm only testing Pd and it's core externals. Are the dlls distributed with the pdprototype.tgz 32 bit? However then why can I run it fine for the command line? That leads me to believe the dlls aren't the issue.

Lucarda:
>Are the dlls distributed with the pdprototype.tgz 32 bit? 

There has never been a 64-bit Pd yet.  : )
All externals are 32bit only.

umlaeute:
On 2018-01-10 01:28, Dan Wilcox wrote:
> Are the dlls distributed with the pdprototype.tgz 32 bit? However then why can I run it fine for the command line? 

because they are not needed?
the dlls in pdprototypes' bin/ come in three groups:
- those required for the Pd-GI (tcl*, tk*, wish*)
- those required for Pd-core ("libwinpthread-1.dll")
- those required for externals (the rest)

practically the 3rd group evolved from the 2nd: Pd (being build by MSVC)
would depend on (e.g.) msvcrt.dll and thus ship it; externals (being
build with MSVC) would also depend on msvcrt.dll but not ship it because
they were working anyhow (since Pd already comes with that dll).
now that Pd is being build with mingw for the official binaries,
msvcrt.dll has become obsolete (from Pd's POV); but because many 3rd
party externals would stop working if it was removed, it is still included.

now in this branch, Pd links statically against libc, obsoleting even
more libraries (probably all of them).

this, and the fact that you probably have any still required dlls (if
any!) in your PATH, allows you to run a 64bit Pd from the cmdline.

umlaeute:
On 2018-01-10 01:31, Lucas Cordiviola wrote:
> There has never been a 64-bit Pd yet.  : )
> All externals are 32bit only.

that's not technically true.
e.g. appveyor builds a 64bit binary of Pd for W32 for some time.

what is true is, that there hasn't been an official 64bit build for W32
available for download from anywhere.
and (presumably for that very reason) nobody has uploaded 64bit W32
externals to deken (or made publicly available otherwise).

danomatika:
Overall, I think what we need is to have it working both 32 and 64 bit. To that end, I'm thinking of adding a script which builds Tcl/Tk and creates the "prototype" folder. This way, people could optionally experiment with newer versions and, as with the macOS build, it would suddenly become much easier to transition the release builds to newer Tcl/TK and/or 64 bit.

In any case, I'm installing Msys2 32 bit and will update the INSTALL.txt to reflect both 32 and 64. I'll add a note that the current release builds are 32 bit, so it's recommended to build with 32.

danomatika:
@umlaeute Looks like pd.res is not being cleaned properly. Looking into it.

umlaeute:
@Spacechild1 should we remove the `install-windows` target and the `msw-install.sh` script in favour of `app` resp. `msw-app.sh`? they both serve essentially the same purpopse

umlaeute:
@danomatika weird, i do:

~~~
$ git describe --always
f1a864d
$ make
[...]
$ find . -name pd.res
./src/pd.res
$ make clean
[...]
$ find . -name pd.res
$
~~~

which looks pretty good to me.

danomatika:
I'm just not seeing it being removed when running `make clean` in the src dir:
~~~
$ make clean
 rm -f pd.exe pdsend.exe pdreceive.exe
 rm -f pd pdsend pdreceive
rm -rf .libs _libs
rm -f pd.dll pd.lib pd.def pd.com libpd.a
rm -rf ../bin
rm -f *.o
rm -f *.lo
~~

umlaeute:
hmm, here it looks like:

~~~
$ make clean
 rm -f pd.exe pdsend.exe pdreceive.exe
 rm -f pd pdsend pdreceive
test -z "pd.dll pd.lib pd.def pd.com libpd.a pd.res" || rm -f pd.dll pd.lib pd.def pd.com libpd.a pd.res
rm -rf .libs _libs
rm -rf ../bin
rm -f *.o
rm -f *.lo
~~~

which version of automake/autoconf do you use?
which revision of Pd?

danomatika:
I'm on a newer commit:
~~~
$ find . -name pd.res
./src/pd.res
$ make clean
[...]
$ find . -name pd.res
./src/pd.res
~~~

umlaeute:
well, it also works with `1e4ad10`

the relevant line in the src/Makefile looks like:

~~~make
clean-generic:
        -test -z "$(CLEANFILES)" || rm -f $(CLEANFILES)
~~~

fwiw, this is all on a linux system, with a mingw32 cross-build toolchain.

danomatika:
Ok, I distcleaned, ran autogen & configure, and now it's working:

~~~
$ make clean
 rm -f pd.exe pdsend.exe pdreceive.exe
 rm -f pd pdsend pdreceive
test -z "pd.dll pd.lib pd.def pd.com libpd.a pd.res" || rm -f pd.dll pd.lib pd.def pd.com libpd.a pd.res
rm -rf .libs _libs
rm -rf ../bin
rm -f *.o
rm -f *.lo
~~~

danomatika:
I've now built Pd with Msys2/MinGW 32bit and double-clicking pd.exe works fine. I'll update the INSTALL.txt.

danomatika:
INstall.txt updated. @umlaeute Maybe you can include cross-compilation info in the msw/README.txt? I will update it to explain the Win application directory layout.

umlaeute:
@danomatika ithink your problem with 64bit Pd is #7 (you have to manually set the `PD_LONGINTTYPE`)

danomatika:
Why is the Codacy build failing? I can't see the log unless I make an account :P

umlaeute:
oh my, i wasn't aware that you need to log in to see *anything* with codacy. it probably wasn't such a good idea then.

in any case: the issue they detected was that the `strip` variable was unused.

btw, since i've implemented the `strip` part, the 32bit Windows "app" came down from 27MB to 19MB

danomatika:
>@danomatika ithink your problem with 64bit Pd is #7 (you have to manually set the PD_LONGINTTYPE)

I just tried that on the 64 bit build and same crash when double clicking to run.

danomatika:
Update: I have building and using custom versions of Tcl/Tk working with the app directory. Tk 8.5.19 working nicely.

pd-win-8519

Spacechild1:
nice work, Dan! I think we can delete the `install-windows` target + msw-install.sh.
now with `make app`, sources are not being installed. IOhannes included them in msw-install.sh and I think it's nice to have them. what do you think?

umlaeute:
i've taken the liberty to remove `install-windows` and friends and add the sources to `app`

danomatika:
How do the sources help without including the entire build tree and build files? There isn't currently an easy way to "build in place" like the makefile and I'd argue you don't need to since you can easily generate an entire app directory now.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Also, the script takes a -s/--sources flag to enable copying the source files in. It's just jot currently set in the Makefile.am app target.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
As a follow up, I'm not against "sourcefull" releases in principal, the issue is more that this is more or less , I think, against the autotools phiolsophy which makes it more difficult to force things to work the same way as the makefile a.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
@danomatika a few points:

- you are (of course) right that the full sources are of limited use without a build-system
- but the header files (traditionally in the `src/` folder) are of utmost importance for anybody who wants to build an external
  - the full sources are an overkill; otoh they are a good reference point for writing objects
- actually the `-s/--sources` flags were not implemented (never parsed for that option) - until recently
- since recently - the `app` target uses the `-s` flag...

danomatika:
The script already installed the headers, or at least that was my intent. The sources option would then install the c files as well.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
yes, i became aware of that while writing my comment

danomatika:
Ahhh! I always thought CPPFLAGS was for C++. That makes more sense.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
I've updated the INSTALL.txt to reflect how Msys2 actually works, and not me earlier broken understanding, ie. you can build for 32 bit with the Msys2 64 bit install and vice versa. Can someone double check what Ive written?

Lucarda:
I think all the beginning should change to:

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

### Windows

Pd on Windows is built with either MinGW or Cygwin which provide the required
open source compiler chain & build environment.

It is recommended to use Msys2 distribution which provides both a Unix command
shell and MinGW. 

Download the 64bit installer unless you are on a 32bit Windows.

http://www.msys2.org/

Then install to the default location (C:\msys32 or C:\msys64) and follow the
setup/update info on the Msys2 webpage.

Msys2 64bit also provides a 32bit shell.

Currently, Pd is built as 32 bit for Windows but you can experiment with a 64 bit Windows.
--enable-universal build option has no effect in MinGW.


Note: Msys2 development seems to change frequently, so some of the package names
      below may have changed after this document was written.

Next open....

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

umlaeute:
On 01/11/2018 10:04 PM, Lucas Cordiviola wrote:
> Pd on Windows is built with either MinGW or Cygwin which provide the required
> open source compiler chain & build environment.

i think this is either factually wrong ("Pd on W32 is *not* built with
either MinGW or Cygwin"; at least not for the official binaries which
currently use MinGW *and* MSVC, both running on linux) or irrelevant
("open source" compiler).

> Currently, Pd is built as 32 bit for Windows but you can experiment with a 64 bit Windows.

i don't see the point here.
we could add a note, that "64 bit Pd on Windows is still considered
experimental", if that is what you actually mean.

Lucarda:
@umlaeute 
Thanks for clarification.
Anyway those things were not written by me. Those are on the .txt.

Spacechild1:
> i think this is either factually wrong ("Pd on W32 is *not* built with either MinGW or Cygwin"; at least not for the official binaries which currently use MinGW *and* MSVC, both running on linux) or irrelevant ("open source" compiler).

this is a Windows specific subsection about the autotools build system (not about how Miller builds Pd) and the info that you can use either MinGW or Cygwin is not irrelevant in my eyes. it says we support both, but we recommend MinGW via Msys2 (mainly because we didn't test on Cygwin :-p - or did we?)

Lucarda:
Changes as @umlaeute suggested:

```
### Windows


It is recommended to use Msys2 distribution which provides both a Unix command
shell and MinGW. 

Download the 64bit installer unless you are on a 32bit Windows.

http://www.msys2.org/

Then install to the default location (C:\msys32 or C:\msys64) and follow the
setup/update info on the Msys2 webpage.

Msys2 64bit also provides a 32bit shell.

Currently, Pd is built as 32 bit for Windows. 

--enable-universal build option has no effect in MinGW.

Also 64 bit Pd on Windows is still considered experimental.


Note: Msys2 development seems to change frequently, so some of the package names
      below may have changed after this document was written.

Next open....
```

Spacechild1:
I would keep the first line about MinGW and Cygwin. The buildsystem theoretically also supports Cygwin, so why not mention that?

Lucarda:
```
### Windows

Pd on Windows is built with either MinGW or Cygwin which provide the required
open source compiler chain & build environment.

It is recommended to use Msys2 distribution which provides both a Unix command
shell and MinGW. 

Download the 64bit installer unless you are on a 32bit Windows.

http://www.msys2.org/

Then install to the default location (C:\msys32 or C:\msys64) and follow the
setup/update info on the Msys2 webpage.

Msys2 64bit also provides a 32bit shell.

Currently, Pd is built as 32 bit for Windows. 

--enable-universal build option has no effect in MinGW.

Also 64 bit Pd on Windows is still considered experimental.


Note: Msys2 development seems to change frequently, so some of the package names
      below may have changed after this document was written.

Next open....
```

umlaeute:
On 01/11/2018 10:57 PM, Lucas Cordiviola wrote:
> ```
> ### Windows
> 
> Pd on Windows is built with either MinGW or Cygwin which provide the required
> open source compiler chain & build environment.

how about "can be built with"?
my main critique is the factual "is built", not about whether we should
mention mingw and/or cygwin.

> 
> Currently, Pd is built as 32 bit for Windows. 
> 

again: i would avoid the factual "is".


these are installation instructions not a bible. they are mean to tell
people what they *can* do not what they *should* do.
the wording should be accordingly.

gf,msard
IOhannes

umlaeute:
> Pd on Windows is built with either MinGW or Cygwin which provide the required
> open source compiler chain & build environment.

how about "*can* be built with"?
my main critique is the factual "is built", not about mentioning mingw and/or cygwin.

> 
> Currently, Pd is built as 32 bit for Windows. 
> 

again: i would avoid the factual "is".


these are installation instructions not a bible. they are mean to tell
people what they *can* do not what they *should* do.
the wording should be accordingly.

Spacechild1:
Sure, i have no problem with that!

danomatika:
I've incorporated the feedback. Please check the INSTALL.txt again.

danomatika:
In this line:

>Pd on Windows can be built with either MinGW or Cygwin which provide the
required compiler chain & build environment.

Rge "rewuired" part refers to the core build requirements at the top of the document, not that only MinGW or Cygwin is required (ie. implying MSVC isn't possible). Should that be reworded.

danomatika:
I changed the opening line in the section to:

>Pd on Windows can be built with either MinGW or Cygwin which provide the
core build requirements: a compiler chain & shell environment.

Lucarda:
I know this is technically correct:

```
Be aware of the differences in some of the Msys2 software package names, where
"i686" denotes 32 bit and "x86_64" 64 bit. Due to how MinGW is designed, you
cannot build a 64 bit Pd with a 32 bit MinGW and vice versa. This also means the
Pd configure --enable-universal build option has no effect in MinGW.
```

But I feel starting like this is less confusing for beginners:

`Msys2 uses two MinGW compilers, one for 32bit and another for 64bit. Be aware of the differences in some of the Msys2 software package names....`

danomatika:
Does the paragraph before state that already?

>Msys2 provides both 32 and 64 bit command shells for MinGW.

danomatika:
Ok, I condensed some of that text together and removed the "be aware" part as it's not really needed and implied in the next section with the package naming.

Lucarda:
Good.

take alook at https://github.com/Spacechild1/pure-data/blob/mingw-autotools2/INSTALL.txt
there's new very long line.

danomatika:
I've now committed the tcltk-dir.sh script which you can use to build Tcl/Tk on Windows. Both scripts now work together in a similar way to those in the mac directory and I've updated msw/README to explain how to use them.

Here's a screenshot of the current master running using 32 bit Tk 8.6.8:
pd-win-868

danomatika:
I've added two Win32 builds to <http://docs.danomatika.com/pdbuilds>, one with the default Tk 8.5 and the other with Tk 8.6.8. I'm curious if these builds from Windows 10 will work on earlier versiosn of Windows such as 7 & 8.

danomatika:
Note: The custom Tk build seems to be larger. Perhaps there are some more files we can remove when installing the custom Tcl/Tk to the Pd app directory.

I'm now looking into converting the Tcl/Tk .a files into .libs so existing externals can link to them.

Lucarda:
>  I'm curious if these builds from Windows 10 will work on earlier versiosn of Windows such as 7 & 8.

I will test on win8.1 ASAP.

---------

Other experiment:

Yesterday I built an w64 Pd with:

`./configure --enable-mmio --enable-asio CPPFLAGS=-DPD_LONGINTTYPE=__int64`

The app don’t work and seeing it with depwalker I get a lot of dependancies:




Is there a way to tell configure to prepend the flag `-static` ?
at least to try to get rid of all those dlls

danomatika:
I think it looks that way with 32 bit as well, not sure why. I had a similar list of dependencies with a *working* 32 bit build...

danomatika:
Also, you don't need:

    ./configure --enable-mmio --enable-asio

As that is now the default behavior.

Lucarda:
Tested pd-0.48-1_win32_tk86 on win8.1.
Working nicely:

d1

Lucarda:
> I think it looks that way with 32 bit as well, not sure why. I had a similar list of dependencies with a working 32 bit build...

Yes there are false positives. Many more on win10 (depwalker is up to win8 and it is known to have some issues on win10).

Is there a way I can try passing the flag `-static` ?

danomatika:
>Is there a way I can try passing the flag -static ?

You could try:

    ./configure CFLAGS="-static"

but I'm not sure if that will have an effect.

Lucarda:
I will try:

`./configure CFLAGS="-static" CPPFLAGS=-DPD_LONGINTTYPE=__int64`

Do I also have to put cuotes to the longinttype?

danomatika:
>Do I also have to put quotes to the longinttype?

Safer to do so on the commandline as you don't want any weird character evaluation.

danomatika:
I'm done for today. From what I can tell, Tcl/Tk gets built with the TDBC library which is why the overall directory size is bigger. We should find some way to avoid building/installing it as Pd doesn't need it.

Also, I noticed a note in the Tcl readme that basically says Tcl 9.0 will *not* support Win32, so it's good timing to get a 64 bit Pd. (EDIT: Of course Tcl 9.0 may be *years* away, but still.)

danomatika:
Nevermind, I found it. We can simply remove the tdbc and sqlite libs from the Tcl install. That cuts the directory size down by half.

danomatika:
>So I simply change that one with the one that comes in the C:\msys64\mingw64\bin.

The script could be updated to check if the dll exists at `/bin/libwinpdthread.dll` and install it. The dlls and Tcl/Tk in the pdprototype.tgz are 32 bit.

Lucarda:
> The dlls and Tcl/Tk in the pdprototype.tgz are 32 bit.

Sure, on the "task manager" they show up as 32-bit, but not Pd.

Also I have successfully compiled "Cyclone" for 64-bit and is working. 32-bit externals don't work (of course).

danomatika:
>>The dlls and Tcl/Tk in the pdprototype.tgz are 32 bit.
>
>Sure, on the "task manager" they show up as 32-bit, but not Pd.

One thing to note is that the Pd gui and the Pd core can run with different architectures, ie. 64 bit core and 32 bit core gui, since they are separate processes. THis can be confusing, however, so it's better to build them both with the same settings if possible. THis is also why IOhannes added a [deken] check where the GUI asks the core what architecture it is in case the GUI is different.

danomatika:
@Lucarda msw-app.sh now copies the pthread dll from your MinGW if you use a custom Tcl/Tk, so you shouldn;t need to do it manually now.

Lucarda:
I will test.
I will try to get a Tck/tk 64bit.

msw/msw-app.sh
```diff
@@ -206,8 +206,16 @@ if [ "x$prototype_tk" = xfalse ] ; then
...
-    # remove some Tcl libraries Pd doesn't need
-    rm -rf $APP/lib/sqlite* $APP/lib/tdbc*
+    # remove bundled Tcl packges Pd doesn't need
+    rm -rf $APP/lib/itcl* $APP/lib/sqlite* $APP/lib/tdbc*
+
+    # install pthread from MinGW
+    if [ -v MINGW_PREFIX ] ; then
+        # MINGW_PREFIX should be either "/mingw32" or "/mingw64"
+        if [ -e $MINGW_PREFIX/bin/libwinpthread-1.dll ] ; then
+            cp -v $MINGW_PREFIX/bin/libwinpthread-1.dll $APP/bin
```
umlaeute: this is not going to fly on cross-compilation builds. (e.g. on my Debian system, the file is `/usr/i686-w64-mingw32/lib/libwinpthread-1.dll` resp. `/usr/x86_64-w64-mingw32/lib/libwinpthread-1.dll`, so the `MINGW_PREFIX` on my system would arguably be something like `/usr/x86_64-w64-mingw32`

msw/msw-app.sh
```diff
@@ -206,8 +206,16 @@ if [ "x$prototype_tk" = xfalse ] ; then
     cp -R $TK/bin $APP/
     cp -R $TK/lib $APP/
 
-    # remove some Tcl libraries Pd doesn't need
-    rm -rf $APP/lib/sqlite* $APP/lib/tdbc*
+    # remove bundled Tcl packges Pd doesn't need
+    rm -rf $APP/lib/itcl* $APP/lib/sqlite* $APP/lib/tdbc*
+
+    # install pthread from MinGW
+    if [ -v MINGW_PREFIX ] ; then
```
umlaeute: hmm, so compiling with `MINGW_PREFIX=""` will evaluate to *true*. i think a simple check for non-emptiness is probably more what we want.

msw/msw-app.sh
umlaeute: i really don't know, but maybe we could use something calculated from the output of `$CC -print-libgcc-file-name`

danomatika: This is more of a convenience on Windows than a "works everywhere" solution. I figure advanced users doing a cross compile might be able to copy the dll themselves if they have to...

danomatika:
>there is a bug in the help for msw-app.sh

Strange because I was using `--tk` just fine earlier today.

msw/msw-app.sh
danomatika: Simple solution would be to try $MINGW_PREFIX/lib first, then $MINGW_PREFIX/bin second.

Lucarda:
Strange, I took the --t from the readme : https://github.com/Spacechild1/pure-data/tree/mingw-autotools2/msw

.............................
Once your custom Tcl/Tk is built, you can use it as the Tk directory source for
msw-app.sh with the -t/--tk option:

    build Pd with a custom Tcl/Tk 8.6.8 directory
    msw/msw-app.sh -t tcltk-8.6.8

.....................................


I checked the -help and there was no -t so I used:

msw/msw-app.sh -tk tcltk-8.6.8 0.48-1-w64-tcltk-w64 

And I got a pd--tk folder with wish85 from the pdprototype.tgz. 

NOTE: now I see that I used the -tk instead of --tk.

But with -t I got what I wanted.

danomatika:
>But with -t I got what I wanted.

"-t" is the short version of "--tk". They both do the same thing.

Lucarda:
That explains it. : )

Lucarda:
Btw, Thanx!!!
Great work!!!
Super.

msw/msw-app.sh
```diff
@@ -29,7 +29,7 @@ BUILD=..
 #----------------------------------------------------------
 help() {
 cat <<EOF
-Usage: msw-app.sh [OPTIONS] [VERSION]
+Usage: $0 [OPTIONS] [VERSION]
```
danomatika: I'm not a fan of this as it will print the whole path when called from another directory and screw up the formatting.

msw/msw-app.sh
danomatika: Fixed.

msw/msw-app.sh
danomatika: Fixed.

danomatika:
FYI: This branch builds fine on macOS.

Spacechild1:
@danomatika, @umlaeute:
could someone add this branch to the pure-data repo? 
probably replacing https://github.com/pure-data/pure-data/tree/mingw-autotools

umlaeute:
@Spacechild1 i almost did the other day, but was distracted.
note, that the entire conversation in *this* PR will be "lost" (that is: not taken over)

Spacechild1:
> note, that the entire conversation in this PR will be "lost" (that is: not taken over)

OK for me.

danomatika:
@millerpuckette I think you only needed to merge #299 which was a cleanup of this PR.

millerpuckette:
I think that's what I did (299, not 275)

danomatika:
Ah right. I'm seeing the automated merge messaging since both PRs contain basically the same commit hashes.

umlaeute:
#299 was only putting the PR into the "pure-data/pure-data" repository (moving it away from @Spacechild1 's fork - on their request)

PR #278, opened 2017-12-28
check sys_libdir exists before reading it

Ant1r:
This fixes libpd/libpd#214: libpd crashing when patch contains [declare -stdpath]

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -1366,7 +1366,7 @@ static void canvas_completepath(char *from, char *to, int bufsize)
     {
         to[0] = '\0';
     }
-    else
+    else if(sys_libdir)
     {   // if not absolute path, append Pd lib dir
         strncpy(to, sys_libdir->s_name, bufsize-10);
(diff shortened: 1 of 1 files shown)


danomatika:
@millerpuckette Is it too late for this to make 0.48-1? It would be good to have in the net libpd release.

umlaeute:
given that Pd-0.48-1 was released yesterday, I'd assume it *is* too late (please let's not get into adding bugfixes after the release).
time for Pd-0.48-2 then!

danomatika:
Hah yeah I know. I just wasn't able to spend time looking into this earlier and it's a libpd crasher.

umlaeute:
i've started a new [updates/0.48-2](https://github.com/pure-data/pure-data/tree/update/0.48-2) bugfix branch, that already includes this patch.

umlaeute:
i've started a new [updates/0.48-2](https://github.com/pure-data/pure-data/tree/update/0.48-2) bugfix branch, that already includes this patch.

PR #287, opened 2018-01-10
Fix Avg function from expr 0.56

porres:
The way Avg is doing still accounts for an extra index and ruins the output if you exceed the index range. The maximum index limit should be "size - 1", not just "size". This fixes it.


umlaeute:
included in update/0.48-2

PR #288, opened 2018-01-11
Fix thread concurrency issues and multi-instance issues

pierreguillot:
- 0c38ada & c7ae002: set **PERTHREAD** (TLS) the variables **stackcount** and **outlet_eventno**  because if these variables are increased or decreased in a concurrency context the behavior is undefined and results mostly to a stack overflow for the **stackcount**. **outlet_eventno** doesn't seems to have any influence on the program but an ounce of prevention is worth a pound of cure. (and sorry for TSL instead of TLS in the commits names...)
- 8f3eec1: The objects **key**, **keyup** and **keyname** are bound to **static** symbols - that breaks the multi-instance support. To fix this the symbols are generated in the **new** and **free** methods of the objects with **gensym()** - ensure that the symbols correspond to the current pd instance. 
- cb2e44f: set **PERTHREAD** (TLS) the FFT static variables from d_fft_fftsg.c to support FFT objects in a multithread context. We can also add [this](https://github.com/pure-data/pure-data/pull/288) if you want to fix size the problem of the FFT. 

Instead of using **PERTHREAD**, another solution would be to put the variable in the pd instance structure. I started to do it for the FFT (see [this](https://github.com/pierreguillot/pure-data/commit/b8671dfc8bbb1f3328feced7133939dee0d4a3a9)) but it can also be done for  **stackcount**. anyway, the **PERTHREAD** works fine.

- ddb167e & 1f0f6a2: add multi-instance support for MSVC (by @eldruin)

### DIFF
--- a/src/d_fft_fftsg.c
+++ b/src/d_fft_fftsg.c
@@ -36,10 +36,10 @@ void rdft(int, int, FFTFLT *, int *, FFTFLT *);
 
 int ilog2(int n);
 
-static int ooura_maxn;
-static int *ooura_bitrev;
-static int ooura_bitrevsize;
-static FFTFLT *ooura_costab;
+static PERTHREAD int ooura_maxn;
+static PERTHREAD int *ooura_bitrev;
+static PERTHREAD int ooura_bitrevsize;
+static PERTHREAD FFTFLT *ooura_costab;
 
 static int ooura_init( int n)
 {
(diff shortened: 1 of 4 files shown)


millerpuckette:
about the EXTERN_NOAPI commit - can't we just not do that and say
extern PERTHREAD t_pdinstance *pd_this;

?  That would be simpler.
The rest of this looks good to me.

pierreguillot:
> about the EXTERN_NOAPI commit - can't we just not do that and say
> extern PERTHREAD t_pdinstance *pd_this

Yes, indeed! My last commit fix this.

pierreguillot:
merged

eldruin:
Glad to see that the MSVC fixes could be merged :)

PR #290, opened 2018-01-12
Update README.txt from the extra folder

porres:
Several fixes for this text, which is outdated, like removing "expr" from the list, adding new objects, fixing some typos and other stuff...

Note that the text still mentions the paf~ object, which has been removed for a while, I'm not doing anything about it cause I just hope it gets back into the extras as there are no more patent issues these days.

### DIFF
--- a/extra/README.txt
+++ b/extra/README.txt
@@ -9,7 +9,8 @@ You can additionally compile fiddle~, bonk~, and paf~ for Max/MSP.
 contents:
 
 externs:
-fiddle~ -- pitch tracker
+fiddle~ - pitch tracker
+sigmund~ - sinusoidal analysis and pitch tracking
 bob~ - moog analog resonant filter simulation
 bonk~ - percussion detector
 choice - find the "best fit" of incoming list


porres:
so, on a second commit, I also added missing new objects to the README, like bob~

extra/README.txt
```diff
@@ -12,7 +12,7 @@ externs:
 fiddle~ -- pitch tracker
 bob~ - moog analog resonant filter simulation
 bonk~ - percussion detector
-choose - find the "best fit" of incoming vector with stored profiles
+choice - find the "best fit" of incoming list
```
umlaeute: what's wrong with the original description?

extra/README.txt
```diff
@@ -25,7 +25,7 @@ hilbert~ - Hilbert transform for SSB modulation
 complex-mod~ - ring modulation for complex (real+imaginary) audio signals
 rev1~, etc. - reverberators
 
-These objects are part of the regular Pd distribution as of Pd version
-0.30.  Macintosh versions of fiddle~, bonk~, and paf~ are available
-from http://www.crca.ucsd.edu/~tapel
+The extra library is part of the regular Pd distribution as of Pd version
```
umlaeute: why do you consider `extra` a single library?

umlaeute:
probably just upload it to deken?

porres:
cool, closing then

PR #297, opened 2018-01-24
consistent dllextents across platforms (and easier to add new ones)

umlaeute:
this PR uses a variable-length array of dllextensions rather than two strings (which obviously can only hold 2 different dll-extensions).

this allows us to easily have only 1 dllextent on certain platforms and 4 on others (without quadrupling the same dll-extension in the first case, which in turn leads to searching for the same file 4 times)

it also simplifies calculating the arch-specific part by putting it into one codeblock that is used by all platforms.
this automatically adds support for arch-specific dll-extensions for (e.g.)  Windows-on-amd64 and Linux-on-PPC.

finally, it also adds support for the `arm64` architecture.

it's somewhat related to https://github.com/pure-data/pd-lib-builder/issues/40 (which is mostly about externals for 64bit MSW)

### DIFF
--- a/src/s_loader.c
+++ b/src/s_loader.c
@@ -37,27 +37,35 @@ objects.  The specific name is the letter b, l, d, or m for  BSD, linux,
 darwin, or microsoft, followed by a more specific string, either "fat" for
 a fat binary or an indication of the instruction set. */
 
+#if defined(__x86_64__)
+# define ARCHEXT "amd64"
+#elif defined(__i386__) || defined(_M_IX86)
+# define ARCHEXT "i386"
+#elif defined(__arm__)
+# define ARCHEXT "arm"
+#elif defined(__ppc__)
+# define ARCHEXT "ppc"
+#endif
+
+#ifdef ARCHEXT
+#define ARCHDLLEXT(prefix) prefix ARCHEXT ,
+#else
+#define ARCHDLLEXT(prefix)
+#endif
+
+
 static const char*sys_dllextent[] = {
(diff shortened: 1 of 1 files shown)


umlaeute:
seems like this fails on VisualStudio for now. i'll try to fix this by tomorrow and in the meantime i've marked this PR as `WIP`

umlaeute:
it turned out to be a git-commit error on my side (leaving a necessary line out of the commit). fixed.

afaict, this is ready to be merged :-)

umlaeute:
personally i'd remove that compat extension ASAP.
i see that you are one of the users that actually uploaded packages to deken with the *bad* extension, but i also see that you regularily uploading.

so: do you really have such a large user-base that will update Pd as fast as possible but who can not be bothered to install a new version of your libraries?

Ant1r:
Actually my request was not really motivated by my Pof library, but more by the kind of project needing (for easier installation process) to distribute some externals in multiple binary versions (one for each targeted OS); these binaries are then located into an common "extra" subdirectory, which is [declared] as path. In this case you cannot use the .pd_linux extension. 

But: I've just successfully tried another way: .pd_linux externals for 64bit can be located into different directory than 32bit ones, both paths [declare]d, and Pd can load the right one. Thus a project tarball can be made compatible with both Pd versions, using .pd_linux extensions.

So... please go ahead! Thanks for all.

umlaeute:
for the record: it's pretty easy to also add an additional `l_ia64` legacy extension for amd64/linux systems, if you really really want that (though i hope that i have made my preference clear by now ;-))

Issue #185, opened 2017-08-15
deken now marks wrong architectures as compatible


umlaeute:
since the `platform(bits)` field is no longer taken into account when calculating the compatibility of deken-installed externals (316f6553), wrong architectures are displayed as compatible.

chr15m:
@umlaeute yep as mentioned on the other comment this whole thing became very complicated when I started implementing it and digging down into different CPU/OS/arch combinations. Basically I guess our obvious end-goal should be showing/highlighting those packages in Deken that match with what the user can actually run on their system.

umlaeute:
this was merged into master with f7dc020 and released as 0.48-1

PR #299, opened 2018-01-26
autotools updates for fully functional W32-builds

### LINKED ISSUE #25: [WIP] usable mingw32 builds
currently, Pd-for-windows is built with **both** `msvc` (to create a usable `pd.lib`) and `mingw32` (for the distributed binary).
it seems that the *pd.lib* as created by *mingw32* is not really usable, yet.

this should of course be fixed.

here's a few tests if the created binaries are usable (*new* means Pd-binaries/libs/externals created solely by mingw; *old* refers to the traditional double-compilation)

- [ ] the new `pd` loads
- [ ] the new `pd` can load *new* externals
- [ ] the new `pd` can load *old* externals
- [ ] an *old* `pd` can load *new* externals
- [ ] an existing external can be built (linked) against the *new* `pd.lib` using mingw (without modifications of the build-system)
- [ ] an existing external can be built (linked) against the *new* `pd.lib` using MSVC (without modifications of the build-system)

umlaeute:
this PR fixes the autotools buildsystem for W32 builds.
it is the same as #275 (but the branch has been moved into the pure-data repository on @Spacechild1's request). it also obsoletes #262.
it closes issue #25 

(afaict) the binaries produced are fully compatible with the ones currently shipped in the W32 precompiled packages.
This means that with this PR applied, it is no longer necessary to use a VisualStudio running under Wine in order to create a functional `pd.lib` (that can be used to create externals with VS).

Building Pd on W32 (MinGW):

    ./configure && make

Cross-compiling Pd on Linux (provided that the cross-compiler toolchain is installed):

    ./configure --host=i686-w64-mingw32 && make

Creating a ready to use application bundle (after the above steps):

    make app

there's also a number of goodies, like buliding/including Tcl/tk and creating a 64bit Wndows application (see INSTALL.txt and msw/README.txt)


Spacechild1:
> it is no longer necessary to use a VisualStudio running under Wine in order to create a functional pd.lib (that can be used to create externals with VS).

actually, it was never necessary :-)

umlaeute:
@Spacechild1 hush!

millerpuckette:
almost worked right out of the box... here's what I did to cross-compile a w64 app in linux:

./autogen.sh 
./configure  --host=x86_64-w64-mingw32 CPPFLAGS=-DPD_LONGINTTYPE=__int64
vi libtool  ((edit libtool to remove 'msvcrt' mention))
make app
cd pd-0.48.1/bin
cp /usr/x86_64-w64-mingw32/sys-root/mingw/bin/libwinpthread-1.dll .
wine ./pd.exe  ((works!!!))

For some reason, the generated 'libtool' file had an unnecessary ref to a 'msvcrt' library; if this is not suppressed there's warning every time you laungh pd or pdsend, etc.   Also, I had to replace the 32-bit version of libwinpthread-1.dll with a 64-bit one (without this pd didn't run at all). 

I'm happy to go ahead and merge this as is, or alternatively can wait to see if any of you want to clean those two minor issues up first...

Meanwhile I think I should put out a test version compiled on my site.

millerpuckette:
Here's a belated question though... should we be making the 64-bit windows version use 64-bit floats?  I often wonder if I should have done that for MacOS when it would have been easy; but it would be easy to do now in windows I think.

Lucarda:
@millerpuckette 

We found that w64 Pd core can run with **w32 tk/tcl** but is better to avoid it and use an **w64 tk/tcl**.

There are instructions on how to do that in INSTALL.txt. I did it and it worked out of the box. 

Patrice informed to the list that he couldn't use 64bit ASIO drivers with a **w32 tk/tcl**. Also everything on Deken is better with an **w64 tk/tcl**.

@danomatika 

For some reason the APP script places a pthread.h on the SRC dir. This pthread.h messes up compiling w64 [externals] with static link. I think is better to have it out of the pd/src. All compilers have their own pthread.h.

millerpuckette:
Thanks for the warning about W32 TCL... I'll attempt to get the 6t4-bit one running tonight when I have another block of time (I hope).  Also I should chase down a couple of compiler warnings and think again about the best float size to be using...

danomatika:
@Lucarda 

>For some reason the APP script places a pthread.h on the SRC dir. This pthread.h messes up compiling w64 [externals] with static link. I think is better to have it out of the pd/src. All compilers have their own pthread.h.

I don't know where that's coming from, maybe it's in the pdprototype.tgz. I can add a rule to explicitly remove it.

@millerpuckette 

>almost worked right out of the box...

Sweet. Glad things are paying off now. I designed the msw-app.sh & tcltk-dir.sh scripts to have the same behavior as the similar scripts in the mac directory. They should be straightforward to use for automating things and have their own built in help.

Lucarda:
We should think of a **pdprototype64.tgz**. Any thoughts against it?

Both pdprototypes should include our new and beloved https://github.com/pure-data/pdfontloader 

I naively think there's no need to create a new "win64" along "win32" on pd-gui.tcl

Lucarda:
@millerpuckette 
>Here's a belated question though... should we be making the 64-bit windows version use 64-bit floats? I often wonder if I should have done that for MacOS when it would have been easy; but it would be easy to do now in windows I think.

I don't know if this is impossible but given the fact that there's no external compiled yet, could be the case that w64 Pd can also be "Double-precision floating-point"?

I will love the [+1] counter going beyond 16777216

:-)

If I understand you correctly is that what you mean?

If yes please, please, please make it "Double"

porres:
I get the feeling that's not what it meant... cause.. if so... why not making it double everywhere? we all kinda want it, don't we? :) I'd love to see table indexes over 16777216

Lucarda:
>  we all kinda want it, don't we? :) I'd love to see table indexes over 16777216

That why i think this is comming. It has other goodies also.
Also I think it will come for osx and linux. But all the Deken repository will have to add double-precision [externals].

umlaeute:
everybody would love that.

here's a few catches:
- we have 64bit but single-precision binaries on all platforms except Windows for years now (millers website offers 64bit binaries for OSX for almost 5 years now; on linux we have 64bit binaries for about 15 years).
if we switch the w64 binaries to double precision, we need to find a migration path for the other platforms as well.
- there's still a number of externals who silently assume that the numbers are 32bit floating point (and just use "float" as the internal data representation). at best these will work incorrectly; mostly they will just crash.
- even those externals that consistently use t_float are not guaranteed to work correctly (see also my [paper on PdCon16](http://phaidra.kug.ac.at/o:38661)). these will not crash, but produce incorrect results.
- running a double-precision external within a single-precision pd (or vice versa) will crash the system. i've [proposed a solution](https://lists.puredata.info/pipermail/pd-dev/2015-02/020073.html) for that years ago (but nobody every picked it up, including myself)
- deken is not quite ready for double-precision right now. that is: neither will the developer oriented cmdline tool properly detect double-precision binaries, nor will the user oriented plugin properly filter out the externals with the wrong precision. (in theory we do have a re-appropriated indicator of the precision size in the meta-data; in practice this is currently ignored, as there was confusion about the actual meaning of this value; see #185, https://github.com/pure-data/deken/issues/161 and simliar).

as said previously, i am *all* for switching to double precision rather sooner than later.
maybe even at the cost of broken externals.

however, i would rather have this done in a coordinated effort (and not by sneaking it in by means of support for an OS that few developers seem to be actively using).

i expect most externals to work fine on W64 (after all, 64bit Pd has been used and tested for ages).
however, i do expect quite a lot of externals to have problems in a double-precisision environment - and somebody will need to have these externals fixed. who? and how?
i probably should just patch-import the externals in the https://github.com/pd-externals/ repository, so people have a targeted place to report the issues and (other) people can fix them.

umlaeute:
one more thing about deken: each and every 64bit external uploaded for Linux is currently *incorrectly* tagged as `64bit precision` (due to the change in the semantics of the bitwidth field).
also there are already a few W64 externals uploaded with a bitwidth of `64`, but it is unclear how that field is used (i didn't have time to test)

danomatika:
Personally, I'd argue for one step at a time:

1. 64 bit windows with single precision
2. update deken and existing packages to correctly handle the precision/bitwidth field so there is a good base of single-precision externals (as we already have)
3. double-check pdlibbuilder double precision support works on all platforms
4. announce an overall shift to double precision with 64 bit builds at some point in the near future (this summer, 1-2 years?)
5. encourage external maintainers to upgrade to pdlibbuilder and provide double precision 64 bit builds
6. encourage motivated users to make their own Pd & external double precision builds for testing
7. start rolling out 64 bit double precision builds at the appointed time for ALL platforms

danomatika:
And for those that "gotta have it now", we make sure that building Pd with double precision is easy by adding a configure option for it. This way motivated people can build Pd themselves when needed.

millerpuckette:
Cool, you've sold me (also partly the fact that I just tried to get a 64 bit
float size compile for windows and don't see teh expected effects yet, so
I have to do more debugging.  So it seems wisest for me to put out a 
32-bit-float version that's much more likely to work well.

cheers
Miller

millerpuckette:
All good, staying with 32 bit floats for now.  I've replaced tcltk with a
64-bit one and uploaded a new compiled version.

I'm hitting some compiler warnings about pointer size conversions that look
like they might cause trouble, so will do some cleaning.  Also need to test
compiling with the PR in place on other platforms (should be no problem), then
will merge PR.

cheers and thanks to all your help...
M

Spacechild1:
`msw/0.*` is missing in .gitignore

millerpuckette:
I went ahead and merged this although I still have some problems but will report them on pd-dev.

PR #282, opened 2018-01-09
Bugfixes to go into 0.48-2

umlaeute:
this PR contains a collection of bug-fixes for 0.48-2

PR #301, opened 2018-02-02
ensure iemgui labels are on top of controls and io

danomatika:
This PR updates the iemguis to ensure that labels are *always* on top. Previous work for 0.48-1 raised controls above labels but this was pointed out as a bug on the pd-list by Roman Haefeli:

>Up until 0.48, the label text of any gui widgets (sliders, etc.) appeared above the front. Since 0.48, the front covers the label text, changing the appearance of patches that have overlapping front and label.
>
>See here:
>front_above_label

This PR reverts to labels above controls and fixes a bug where label draw order would be lost after configuring send/receive names.


umlaeute:
i've included that in the #282 `update/0.48-2` PR.

PR #302, opened 2018-02-09
Make code used in libpd compile using MSVC

eldruin:
Libpd includes some sources from the pure-data repository.
These need a couple adaptions in order for them to build correctly when using Microsoft Visual C compiler.
The commit messages should be descriptive enough.
Please let me know about any further improvements I could try under MSVC.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -32,11 +32,14 @@ extern int pd_compatibilitylevel;   /* e.g., 43 for pd 0.43 compatibility */
 #ifdef _WIN32
 #ifdef PD_INTERNAL
 #define EXTERN __declspec(dllexport) extern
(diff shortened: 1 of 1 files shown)


danomatika:
Cool to see VS actually working. The last time I tried about 5 years ago, it was so many errors due to unsupported C version it seemed hopeless.

pierreguillot:
Since your fix is for multi-instance support I pull your branch in this PR #288. And I suggest you to use the commits that I made if you want to avoid some crashes.

millerpuckette:
About the second commit (Fix compiler error C2492 on MSVC) - could I just say
extern PERTHREAD ...
and not make the new macro?

pierreguillot:
It's merged into master, you can close the issue.

PR #307, opened 2018-02-20
various bugfixes for [text sequence] and [text search]

Spacechild1:
this PR fixes a couple of bugs and adds some minor improvements to [text search] and [text sequence].
summary:

1) there is a bug in the text client argument parsing, skipping valid flags/arguments in some cases.
2) allow (or explicitly disallow) flags or other predefined symbol as first creation argument in [text search] and [text sequence]. while it's possible to do [text search 0], [text search > 0] currently doesn't work as expected since the ">" is treated as the text buffer name. similar thing would happen with [text sequence -g]. I wrote a function to check the first argument against predefined flags/symbols. however, allowing this usage would introduce a regression in case someone used a text buffer named like one of those symbols (which is quite unlikely, though), so I added a switch to explicitly disallow it (posting a helpful error message).
3) fix segfaults with -t flag in [text sequence]
4) better error messages for wrong flag usages

1) + 2) fix issue https://sourceforge.net/p/pure-data/bugs/1308/

### DIFF
--- a/src/x_text.c
+++ b/src/x_text.c
@@ -1318,6 +1318,7 @@ static void *text_sequence_new(t_symbol *s, int argc, t_atom *argv)
     x->x_eaten = 0;
     x->x_loop = 0;
     x->x_lastto = 0;
+    x->x_clock = clock_new(x, (t_method)text_sequence_tick);
     while (argc && argv->a_type == A_SYMBOL &&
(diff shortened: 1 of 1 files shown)


millerpuckette:
cherry-picked commit 6f99757be84350bd35265f77e0fff5a6c2d938f5

PR #309, opened 2018-02-23
Fixed documentation 16.more.array.pd

pierreguillot:
- There were two arrays (array98 & array99) that were strangely merged together (perhaps due to an incompatiblity with the patch format of previous versions?). I removed the array 98 and the only message associated with it. 
- I changed the size of the array99 to 6, this way it is coherent with the GOP bounds associated with the array.

I tried to change the minimum of things but I'm not sure it's the best fix. But I guess that the patch must be updated in a way or another.

Using the patch revealed one bug:
- Renaming an array doesn't rename it graphically (you have to update the window in a way or another).

And I have two suggestions:
- Using the message bounds should also resize the array
- Using a list to change the elements of an array should also resize the array if needed or at least notifying if the elements are out of bounds.

### DIFF
--- a/doc/2.control.examples/16.more.arrays.pd
+++ b/doc/2.control.examples/16.more.arrays.pd
@@ -1,8 +1,8 @@
 #N canvas 19 83 830 601 12;
-#X graph graph1 0 -1 5 1 569 304 769 154;
-#X array array99 5 float;
-#X array array98 7 float;
-#X pop;
+#N canvas 0 22 450 278 (subpatch) 0;
+#X array array99 7 float 0;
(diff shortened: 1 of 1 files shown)


millerpuckette:
I made some somewhat different edits... the array98 and 99 were intentional but misleading - I tried to clarify somewhat.  And yes it's time for me to fix teh bug about the name not updating... I think though
that sending lists shouldn't resize the arrays because, at least in classical computer usage, resizing arrays isn't a real-time-safe thing to do.

pierreguillot:
Perfect! I'll try to have of look at the bug about the name. Thanks!

Issue #242, opened 2017-10-31
loadbang in clone is loaded later


porres:
maybe there's no way around it and this is known, but I'm here to point it out and discuss anyway.

<img width="643" alt="screen shot 2017-10-31 at 13 44 31" src="https://user-images.githubusercontent.com/5360142/32233483-a4dc764c-be41-11e7-9451-29215b599155.png">

here you see that the loadbang from an abstraction is "superceeded" by the loadbang from the parent patch. But if the abstraction is loaded inside clone, it is fired later than the parent patch. 

This is problematic if you have an abstraction that loads arguments or default values that need to be superceeded by incomming messages.

I was able to solve this by using [initbang] inside the abstraction instead of [loadbang]. Seems like a solution if nothing else can be done, but this would mean that vanilla really needed something like [initbang]

cheers

umlaeute:
no need for `[initbang]`.

simply make sure that whatever should get loadbanged will also get triggered the first time something is received in the *inlet* (filtering out duplicates).

note: i'm not saying that the current behaviour is not a bug; but i don't think it warrants the introduction of `[initbang]` (this is a separate issue)

porres:
> simply make sure that whatever should get loadbanged will also get triggered the first time something is received in the inlet (filtering out duplicates).

sorry, I don't undesrtand

danomatika:
>>simply make sure that whatever should get loadbanged will also get triggered the first time something is received in the inlet (filtering out duplicates).
>
>sorry, I don't understand

I think he is suggesting to use something like a spigot which bangs the [f $1] on the first message the object receives and then closes afterward, something like:

![screen shot 2017-12-01 at 3 22 13 pm](https://user-images.githubusercontent.com/480637/33486853-b47e4ea4-d6ab-11e7-9416-37d7e2cc3cf8.png)

Also, you could create a control inlet to send the value. I mainly use [clone] this way and without the object arguments.

porres:
> I think he is suggesting to use something like a spigot which bangs the [f $1] on the first message the object receives and then closes afterward, something like:

But that would mean changing the whole patch/abstraction. And I don't wanna do that. Say it is from an external library and I don't wanna mess with it. 

And anyway, I'm not looking for hacks or ways around this. I'm already dealing with this my own way... The only point is that I think this is a bug, and this is just a bug report then. 

Now, is there any opposition to the idea this is a bug? 

cheers

danomatika:
[...]

PR #310, opened 2018-02-23
[WIP] [clone] improvements/bug fixes

Spacechild1:
this PR basically fixes two bugs with [clone]:

1) prevent cloned abstraction from closing after being saved. I implemented a clone_reload() method so we don't always have to remake the whole clone object in canvas_doreload(). 
I check if inlets/outlets are still the same number and type (signal/non-signal). 
if yes, I reload all instances except for the one being saved, otherwise I let canvas_doreload() remake it (later we can implement a better way to handle this).
I chose to make some existing code into functions to avoid further repetitions and copypaste errors.
this fixes https://sourceforge.net/p/pure-data/bugs/1279/ and https://github.com/pure-data/pure-data/issues/495

~~2) send "loadbang" to clone objects at the same time as abstractions to ensure the correct order of intitialization. this fixes https://github.com/pure-data/pure-data/issues/242#~~

... more stuff ...


src/g_canvas.c
```diff
@@ -914,14 +920,17 @@ void canvas_loadbangsubpatches(t_canvas *x)
     t_symbol *s = gensym("loadbang");
     for (y = x->gl_list; y; y = y->g_next)
         if (pd_class(&y->g_pd) == canvas_class)
-    {
-        if (!canvas_isabstraction((t_canvas *)y))
-            canvas_loadbangsubpatches((t_canvas *)y);
-    }
+        {
+            if (!canvas_isabstraction((t_canvas *)y))
```
umlaeute: it seems that there are whitespace issues here.
do you mind replacing all the TABS with 4 SPACES?

Spacechild1: purpose because in Allman/BSD indentation style (which is used throughout Pd) brackets should be on the same level as the control structure - which in this case is the if-statement and not the for-loop. IMHO, putting the brackets on the same level as the for-loop is a bit weird and misleading. I can switch back to the original indentation if you think it's better style.

umlaeute: no that's fine.
i was confused by the github diff display (I didn't check out the actual files; and now see that my comment was void).
please disregard my comment :-)

umlaeute:
sure. done. (although i don't know about the impact of that action)

millerpuckette:
waiting for 0.49 for this one.  I think it should be possible to fix the problem of teh disappearing
clone copy (on save) more succinctly.  Also, I tried the "loadbang" fix, but at least as I applied it,
I ended up seeing each loadbang going off twice on reload.

Spacechild1:
thanks for having a look!
hmmm, I checked again and it works fine for me. here's a minimal testpatch:
[clone-loadbang.zip](https://github.com/pure-data/pure-data/files/2259363/clone-loadbang.zip)

when I load this patch, the loadbangs fire in the correct order and only once: 1) clone subpatch, 2) clone canvas, 3) parent subpatch, 4) parent canvas. 

can you try with this patch and/or can you send me your test patch? did you test with all commits or only certain commits applied?

millerpuckette:
Aha - trying it with all the commits removes the problem - I must simply not have figured out
how to apply only the loadbang-handling without the (more involved) re-loading fix.  I'd rather
save teh reloading thing for 0.49 but would like to get loadbanging fixed for 0.48-2.

Spacechild1:
weird, the re-loading fix shouldn't affect the loadbangs IIRC. I'll have a look.

Spacechild1:
do you have an example which triggers the crash? the fix looked rather straight forward and worked for me. after all,` scalar_new()` in `text_define_new()` doesn't really add the scalar to the glist (where it would be freed automatically), so we have to free it manually somewhere...

Spacechild1:
while we're at it: you might also consider https://github.com/pure-data/pure-data/pull/307/commits/6f99757be84350bd35265f77e0fff5a6c2d938f5 which fixes a hard crash for the -t flag in [text sequence]

Spacechild1:
Ha, I think I found the problem! Here's a fix: https://github.com/pure-data/pure-data/pull/382
both `textbuf_free()` and `word_free()` (called by `scalar_free()`) attempted to free the same binbuf. unfortunately, double deletion is undefined behaviour, this might explain why it didn't crash for me... sorry!
does it work now for you?

millerpuckette:
I believe the loadbang-order part of this one is already merged, and I want to think harder about
the stay-visible trick and hopefully do it more simply.

Spacechild1:
ok!

danomatika:
Is this PR still relevant/alive? (I'm trying to close *shadow* merged stuff.)

Spacechild1:
I'm closing it for now

Spacechild1:
@millerpuckette Hi Miller, starting from April I want to pick up working on [clone],as there is at least one new bug to fix (https://github.com/pure-data/pure-data/issues/410) and I want to propose some features. 

I would like to continue work from here, so my question is: are you still sceptical about my solution for 1) (_prevent cloned abstraction from closing after being saved_)? I really would like to see this fixed, as I'm using [clone] a lot, so do you think you could accept this as a temporal fix? (I think it can be changed later without any problems, in case you've found a more elegant solution). 

otherwise, I make a new branch/PR.

millerpuckette:
I think it's better to keep them separate for now at least.  I remember the stay-visible changes looked complicated and fragile to me.

Spacechild1:
ok, I'll start with a new PR then.

PR #313, opened 2018-02-25
remove *edit* suffix when displaying canvas menuclose dialog

danomatika:
The new " \*edit\*" suffix is sent added to the filename which is set as the canvas window title. When closing a dirty canvas window, the dialog text is

    Do you want to save the changes you made in "SomePatch *edit*"?

when it should probably be the filename without the suffix

    Do you want to save the changes you made in "SomePatch"?

This PR simply removes the suffix using regsub if the filename is long enough to contain it. As far as I can tell, this is the only place in the gui where the window title is used to grab the filename. If not, this could probably be rolled into a new canvas procedure and reused ala

    proc ::pdtk_canvas::pdtk_canvas_cleanfilename {mytoplevel}

### DIFF
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -12,6 +12,10 @@ namespace eval ::pdtk_canvas:: {
     namespace export pdtk_canvas_menuclose
 }
 
+# store the filename associated with this window,
+# so we can use it during menuclose
+array set ::pdtk_canvas::::window_fullname {}
+
 # One thing that is tricky to understand is the difference between a Tk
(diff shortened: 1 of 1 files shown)


umlaeute:
i took the liberty to rebase the PR on current master (so it can be merged again) and to rewrite the code, to not rely on regexes but instead storing the full path (without any state decorations) in a dictionary that is associated with the current window.
very similar to what is already done in the `::windowname` array (which i have left untouched, as it is used in the window-browser and the search results and having the full path might often confuse people)

umlaeute:
included in the #394 PR.

PR #315, opened 2018-02-26
pointing to github externals howto

x37v:
more up-to-date


umlaeute:
closing, as the link in the documentation already redirects to github (and I prefer having the canonical URL not be owned by e-corp)

x37v:
Redirect is good, I can't remember but I think when I made this change it wasn't a redirect, it was just a stale howto.

umlaeute:
yes sure.

iirc, i implemented the redirect (back then) as a reaction to that issue.

x37v:
great!

PR #318, opened 2018-03-04
fix a case of bad message formatting

jpcima:


umlaeute:
fixed with pure-data/deken@61ce0f16c03ef8bb77a4bd181b96e04220fb2e83 (included in `0.48-2`)

Issue #184, opened 2017-08-15
suggestion: make [declare -path] also search user added paths


porres:
This is an issue that relates to other Pull requests and parallel discussions, like: https://github.com/pure-data/pure-data/pull/183

In short, some of us were used to relying on "additional standard paths" for managing externals, but @millerpuckette  points out (and I quote) "_the way forward is to somehow get the "path" mechanism to do what people are asking for._" see full message at: https://lists.puredata.info/pipermail/pd-list/2017-07/119839.html

Regarding that, there is a user case related to [declare] if you use "standard paths" and the -stdlib and -stdpath flags that I hope we can achieve with the -lib and -path flags as well. 

Say you have externals elsewhere than Standard Paths (which should only be the "extra" folder, according to Pd's manual). The idea is that you have to add them to pd's Path, but if you wanna use [declare], you can only use relative or absolute paths, which I think are much more complicated to manage. So we could make the "-path" flag also search and find subdirectories in user added paths.

Let me illustrate. If I have cyclone in a standard path, I can keep the search path empty and just use declare to call it. This seems to be what Miller suggests it's a good practice in https://github.com/pure-data/pure-data/pull/152#issuecomment-321332197 - and I agree it's a nice one, but we need an enhancement to [declare] so it really works.

See the picture below: search path is empty, and declare is being used to load cyclone with the "-stdpath" flag.

<img width="717" alt="29137682-85ada536-7d17-11e7-9bda-447dca5e092a" src="https://user-images.githubusercontent.com/5360142/29331594-637ca2a8-81d3-11e7-9b0b-6749991ad94f.png">

So, now we do not want to mess with standard paths no more, and we can put it in a different folder, such as ~/Documents/Pd/externals - if I have cyclone in there, I cannot use [declare] in a simple way like this: 

<img width="880" alt="29137866-2446d852-7d18-11e7-97a2-21c394a31b66" src="https://user-images.githubusercontent.com/5360142/29331646-8b915f0e-81d3-11e7-8dac-a85348521935.png">

Unless I do an absolute path, and that is much more complicated than before with standard paths. But if [declare -path] also searches in the added search paths, this would happen.

cheers

Spacechild1:
[declare -path] is explicitly meant for patch relative searching - for example to load a specific version of a library bundled in a project. if [declare -path] would look also in user-paths, this could create many problems and defeats some of the purpose of -path.
[...]

PR #320, opened 2018-03-08
Bumped to deken/v0.3.0 which supports the new .dek package format.

umlaeute:
- new fileformat
- support for double precision Pd (clarification of the "bits" field)
- uninstall-before-install
- bells
- deken-specific preferences
- proxy/https support (if the relevant system libraries are installed)
- whistles and hoops

### DIFF
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -963,9 +963,9 @@ static void sys_init_deken()
         "arm64"
 #elif defined (__ARM_ARCH)
         "armv" stringify(__ARM_ARCH)
-# if defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__)
-#  if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
-        "l"
+# if defined(__BYTE_ORDER__) && defined(__ORDER_BIG_ENDIAN__)
+#  if __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+        "b"
 #  endif
 # endif
 #else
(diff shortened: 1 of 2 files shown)


umlaeute:
@danomatika it would be nice to have the deken-preferences in Pd's common preferences menu...

porres:
Hi, I wrote this to the list, but maybe here is better. About one of the "bells" -  _**set installation path (suggesting Pd's default search paths, and offering to create those that are missing)**_ - I see a problem. In short, see: this: https://github.com/pure-data/pure-data/pull/139 and how we should consider only the "extra" folder as **the** standard folder. And also how the new external folder since 0.48-0 already fixes this issue and makes this kinda pointless now. Given that, I'm even suggesting a Pull Request to clear this up for good, it's here: https://github.com/pure-data/pure-data/pull/183

See also: https://github.com/pure-data/pure-data/issues/184 which discusses the behaviour of [declare] and how we could just use "-path" and forget about "-stdpath" 

Now, I don't know how the plugin works, but I may assume it gets these paths from Pd, so maybe if my Pull Request gets accepted, no change needs to be made in the Deken plugin and it'll just stop showing them... cause as it is, this is adding unnecessary complexity and confusion.

porres:
> Now, I don't know how the plugin works, but I may assume it gets these paths from Pd, so maybe if my Pull Request gets accepted, no change needs to be made in the Deken plugin and it'll just stop showing them... 

So, I tested it, and it is true. And I also closed that Pull Request and opened a new one here https://github.com/pure-data/pure-data/pull/321

umlaeute:
included in `update/0.48-2` #336

PR #322, opened 2018-03-09
updating the help file of [declare] to reflect the changes in this branch

porres:


porres:
this got it, but I guess I'm not happy with it yet, seems I'm making a mistake. I will check it out with care and propose an update if I find necessary, thanks

PR #323, opened 2018-03-10
update/fix list-store-help.pd

Spacechild1:
* replace old 'range' method name with new 'get'.
this is especially important for the serializer example which currently causes an infinite loop.
* remove 'non-list message convert to list' example since this is not true
(BTW: this is actually *a good thing* because it prevents collisions if we add more methods to [list store] in the future!)


umlaeute:
the `range` bug has already been fixed in f83c88ba7ac68368248797ef0d0febc3526432bb

the 'non-list message convert to list' has been included in #394

PR #324, opened 2018-03-10
document -yrange and -pix flag for [array define]

Spacechild1:


umlaeute:
included in `update/0.48-2`

PR #325, opened 2018-03-10
document "click" and "close" methods in [text define]

Spacechild1:


porres:
I can't believe I just opened an issue suggesting to include such features and found out they already exist and weren't documented, what are the odds? :)

porres:
oh, can you update your help files to sat "changed for Pd 0.48-2" and stuff?

Spacechild1:
it already says "updated for Pd 0.48", no need to include the bugfix version number. [soundfiler], [noise~] etc. don't either.

porres:
> no need to include the bugfix version number. [soundfiler], [noise~] etc. don't either.

well, I think they actually do or should, as I've seen many updated help files with bugfix version number, and I've also been explicitly suggesting and doing this with no resistance.  See the help of [midiin], [expr], for instance.

Spacechild1:
OK, but I guess Miller will update the version number whenever he merges this.

porres:
I wouldn't say that's guaranteed to happen, and probably easier and more guaranteed if you do it yourself. 

Now, I can't also be the one say what's the criteria for changing number versions on the help update. Is it supposed to mean the object was updated only? I figure and guess it could say either the documentation has (significantly) changed, the object, or both. But that's something for @millerpuckette  to answer

umlaeute:
included in `update/0.48-2`

PR #327, opened 2018-03-10
Conform array, text and scalar interface for gpointer access

Spacechild1:
both [text] and [array] have their own way to get a pointer to the scalar. I thought it might be a good idea to add the "send" method to [text define] and the bang method (+ outlet) to [array define] to make the interface more uniform.


millerpuckette:
merged

PR #328, opened 2018-03-16
separate inlet & outlet height defines

danomatika:
As requested on the pd-list, make inlet and outlet height separate.


umlaeute:
merged into `update/0.48-2`, so closing this.

PR #335, opened 2018-03-23
Remove unused variables

eldruin:
I enabled all warnings and saw lots of obvious cases of unused variables that were just declared but never used.
I have removed some of the obvious cases, although I did not have time to fix all of them, also paying attention to the variables not being used in commented-out or deactivated code.
I put the changes in a commit per file to ease partial change acceptance.


umlaeute:
merged

PR #338, opened 2018-03-29
Fix implicit functions and definitions for msvc

pierreguillot:
b90f5cc: s_utf8.h is necessary for `u8_utf8toucs2()` [L. 1393](https://github.com/pierreguillot/pure-data/blob/8c40ab48edf8e2f6899cdc7ad68645a652d649cd/src/g_canvas.c#L1393) of g_canvas.c

8c40ab4: io.h is necessary for `_waccess()` [L. 1394](https://github.com/pierreguillot/pure-data/blob/8c40ab48edf8e2f6899cdc7ad68645a652d649cd/src/g_canvas.c#L1394) of g_canvas.c

c738415 & 4eaf62b: `_atoldbl` takes a `char*` for the 2nd arg and not `char**` the macro fix this.
 
b29dcc5:  `mayer_realfft()` is already declared in m_pd.h.

fcdabfe: `rint()` is already declared with math.h

### DIFF
--- a/extra/sigmund~/sigmund~.c
+++ b/extra/sigmund~/sigmund~.c
@@ -26,11 +26,6 @@ typedef double t_floatarg;
 and usable in other contexts.  The one external requirement is a real
 single-precision FFT, invoked as in the Mayer one: */
 
-#ifdef _MSC_VER /* this is only needed with Microsoft's compiler */
-__declspec(dllimport) extern
-#endif
-void mayer_realfft(int npoints, t_sample *buf);
-
 /* this routine is passed a buffer of npoints values, and returns the
 N/2+1 real parts of the DFT (frequency zero through Nyquist), followed
 by the N/2-1 imaginary points, in order of decreasing frequency.  Pd 0.41,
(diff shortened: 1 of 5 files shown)


extra/sigmund~/sigmund~.c
```diff
@@ -26,11 +26,6 @@ typedef double t_floatarg;
 and usable in other contexts.  The one external requirement is a real
 single-precision FFT, invoked as in the Mayer one: */
 
-#ifdef _MSC_VER /* this is only needed with Microsoft's compiler */
-__declspec(dllimport) extern
-#endif
-void mayer_realfft(int npoints, t_sample *buf);
-
```
umlaeute: could this code-snippet be related to (and requierd for) compiling sigmung~ on Max/MSP?

pierreguillot: I admit I don't know... we should ask Miller. If so a good solution would be to include this code between `_MSC_VER` and `MSP`:
```
#ifdef MSP
#ifdef _MSC_VER /* this is only needed with Microsoft's compiler */
__declspec(dllimport) extern
#endif /* _MSC_VER */
void mayer_realfft(int npoints, t_sample *buf);
#endif /* MSP */
```

umlaeute:
included as 0cb0b3d6aa43e2464594b8e07339fdffc1d8623e...f958e98c930e60ac5b224d7e03a5934fe6af3188

PR #340, opened 2018-03-30
Fix inconsistent variable type

pierreguillot:
fix the type of `x_pathsize` of `t_oscformat` from `int` to `size_t`


umlaeute:
cherry-picked into `update/0.48-2` with 0f6b6631bfa7f74209617335e3d7a684ac1e38b2

PR #346, opened 2018-04-05
expand $0 to canvas id in messages

danomatika:
Back by popular demand....

~~~
[loadbang]
|
[ $0 <
|
[print]

-----------
print: 1003
~~~

Pulled from Pd-l2ork, so has been tested in the wild. This also includes a compatibility check to be safe, although the only issue would be for people "expecting" a 0 from $0 within a message which I doubt there are any. (Philosophical $ selector design issues aside...)

EDIT: I also updated the message help patch:

screen shot 2018-04-05 at 9 26 23 pm

### DIFF
--- a/doc/5.reference/message-help.pd
+++ b/doc/5.reference/message-help.pd
@@ -1,33 +1,32 @@
-#N canvas 70 162 648 546 12;
-#X msg 67 10 message boxes;
-#X text 34 33 Message boxes hold one or more message. Anytime the message
+#N canvas 141 36 453 655 12;
+#X msg 47 10 message boxes;
+#X text 14 37 Message boxes hold one or more message. Anytime the message
 box receives any message at all \, the messages in the box are all
 sent to their destinations.;
-#X obj 120 295 print;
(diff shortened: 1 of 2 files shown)


Spacechild1:
here's an alternative implementation: https://github.com/pure-data/pure-data/pull/347

danomatika:
Closing this for now in favor of #347.

umlaeute:
@danomatika should we delete this branch?

danomatika:
Here's the relevant commit from the Purr Data implementation: https://git.purrdata.net/jwilkes/purr-data/merge_requests/200

I wonder if there is some advantage over this approach to that of Purr Data?

Spacechild1:
personally, I'm not a big fan of Jonathan's solution (canvas_dollarzero taking a t_pd * argument and checking if it's messresponder_class), I rather took the chance to get rid of some global state in binbuf_realizedollarsym and binbuf_eval. Performance wise there shouldn't be any significant difference (I haven't measured, though!), so I think it's rather a matter of taste.

PR #347, opened 2018-04-06
allow $0 expansion in message boxes

Spacechild1:
this is an alternative implementation to https://github.com/pure-data/pure-data/pull/346 which uses no hidden dynamic memory allocations but is more invasive: 

binbuf_eval and binbuf_realizedollsym now take the $0 as a parameter instead of getting it inside the routine from the current canvas. this allows message boxes to pass the $0 value of their parent canvas.

the two functions are renamed to binbuf_doeval and binbuf_dorealizedollsym, the original function names are kept as thin wrappers around the new ones.


danomatika:
This looks good and implements it as suggested by the commenting. I would also grab the message-help.pd file from my PR for this.

Spacechild1:
resolved conflicts with master

porres:
> resolved conflicts with master

there seems to be new ones :/

porres:
would this also expand "$0" inside [text] and other similar objects (qlist/textfile)?

Spacechild1:
It seems like dollar (symbols) are not expanded in `[text]` at all, i.e. not only $0. This should be a seperate issue/PR.

porres:
> It seems like dollar (symbols) are not expanded in [text] at all, i.e. not only $0. This should be a seperate issue/PR.

please open it, as I'm not sure what you mean.

porres:
> please open it, as I'm not sure what you mean.

things seem to work fine for me

Spacechild1:
> things seem to work fine for me

Are you saying that a line like "$1 $2 $3" in `[text define]` expands to the patch's creation arguments? This doesn't work for me... Or did you mean something different?

millerpuckette:
No, but the text_sequence object expands them to values supplied using the "args" message to [text sequence]..

Spacechild1:
Ha! That's a feature I've never noticed! So it's clear to me now that "$1" etc. is not supposed to expand to the patch's creation arguments. This still leaves the possibility of expanding $0, so I see what @porres meant.

BTW, reading the source code, I discovered an undocumented list method, which is like a "bang", but also temporarily sets the arguments. Will open an issue to add this to the help file.

porres:
yeah, and I'm using "$1" inside [text sequence] to get the value of "$0" and it's just kind of a drag, so I wonder if this PR would take care of that as well ;)

Issue #251, opened 2017-11-11
pd~ on windows


chikashimiyama:
I'm trying to run a very heavy patch using pd~ on windows 10. The version of Pd is 0.48-0.
I opened simply pd~ help patch with DSP on, clicked "pd~ start pd~subprocess.pd" message box, 
then the second pd window opens and the application freezes. (all windows do not react to my mouse clicks)

Lucarda:
Strange!
I couldn't replicate the issue. I opened [pd~] help patch, start the sub-process and everything is working as expected.

Win10
Pd 0.48.0

Spacechild1:
hi, what's the output from the console? in which directory does your Pd live? pd~ doesn't like directory paths with whitespace. try to move your Pd folder to a path without whitespace and see if it solves the problem.

Spacechild1:
@chikashimiyama is this still a problem with Pd 0.50 or can we close this?

Spacechild1:
Closing because can't reproduce and OP doesn't respond. This has likely been fixed by one of my pd~ bugfixes.

PR #349, opened 2018-04-07
[WIP] bugfixes/improvements for pd~

Spacechild1:
* fix errors with path names containing whitespace. On Windows, Pd is usually installed in C:\Program Files (x86) and such a path would be treated as several arguments, messing up the argument parsing in the subprocess. This fixes https://github.com/pure-data/pure-data/issues/251

* return from pd_tilde_readmessage if no binary message could be received instead of going into an infinite loop. I also added a return code to the function so we can call pd_tilde_close on failure. this gracefully handles the case where the subprocess is closed from within, preventing the parent process from freezing.

* ignore audio API in sys_loadpreferences when an external scheduler is used. the idea is to prevent unnecessary scanning of audio devices (which can take some time on some systems for various reasons) so that the subprocess is opened instantly. the "-schedlib" flag will set the audio api to 0 in sys_argparse anyway.

EDIT:

* the parent process keeps reading from the pipe in a child thread, preventing possible deadlocks.

### DIFF
--- a/extra/pd~/GNUmakefile.am
+++ b/extra/pd~/GNUmakefile.am
@@ -22,7 +22,7 @@ AUTOMAKE_OPTIONS = foreign
 AM_CFLAGS = @EXTERNAL_CFLAGS@
 AM_CPPFLAGS	+= -I$(top_srcdir)/src -DPD
 AM_LIBS = $(LIBM)
-AM_LDFLAGS = -module -avoid-version -shared @EXTERNAL_LDFLAGS@ \
+AM_LDFLAGS = -lpthread -module -avoid-version -shared @EXTERNAL_LDFLAGS@ \
     -shrext .@EXTERNAL_EXTENSION@ -L$(top_builddir)/src
 
 externaldir = $(pkglibdir)/extra/$(NAME)
--- a/extra/pd~/pd~.c
+++ b/extra/pd~/pd~.c
@@ -584,6 +584,13 @@ static void pd_tilde_donew(t_pd_tilde *x, char *pddir, char *schedlibdir,
         execv(cmdbuf, execargv);
         _exit(1);
     }
+    do {
+      unsigned int i;
+      for(i=FIXEDARG; execargv[i]; i++) {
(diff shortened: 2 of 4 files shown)


Spacechild1:
OK, this is definitely a deadlock of the two pipes, probably when the pipe buffer size is exceeded (as suggested it this article: https://blogs.msdn.microsoft.com/oldnewthing/20110707-00/?p=10223).
A large -fifo size initially pumps a lot of messages into the pipe. But worse:  I can also trigger the deadlock by sending lots of messages into or out of the subprocesses. 
This means that pd~ in its current implementation is a bit unsafe to use... 
any ideas? the article suggests using different threads for reading and writing.

Spacechild1:
OK, I got the reading thread to work and the processes are not deadlocking anymore. 
I also wrote some error handling for vector sizes != DEFDACBLKSIZE which the original code is missing.

For now, I only implemented it for Pd binary mode but I'll also do it for ASCII mode and Max if this solution turns out to be the way to go.

@umlaeute can you have a quick look when you have time?

here's some crude test patch:
[pd~-test.zip](https://github.com/pure-data/pure-data/files/1920828/pd.-test.zip)

Spacechild1:
actually, the fifo is only useful up to certain length because the subprocess can only block so long before the parent process will completely fill the pipe buffer and block as well. the buffer size is 65536 bytes which corresponds to 128 blocks of two channel audio. 
one solution could be to have the subprocess constantly drain the pipe in a separate reading thread, just like the parent process. of course this would mean adding more synchronization, possibly degrading performance, but I have to measure it. maybe it could be an option which can be enabled via the schedlib args.

millerpuckette:
holding off on the big change (commit 7) but the others are straightforward and look safe

danomatika:
Is this more/less merged? Should we close the PR or are there changes @Spacechild1 can make?

Spacechild1:
the first 6 commits are merged. I need some time to finish the deadlock fix, so I'm closing it for now.

Issue #355, opened 2018-05-06
suggestion: add argument inlet to [log]


porres:
this is so it would work consistently in the same way as [log~]

porres:
this was done! so I'm closing it

PR #356, opened 2018-05-06
update math functions ([log], [pow]/[pow~])

porres:
The [log] update  makes it consistent and equivalent to [log~]

I had opened an issue earlier about it - see https://github.com/pure-data/pure-data/issues/355 - but decided to implement it myself :) 

This is totally backwards compatible, of course, and I just copied/adapted the same code from [log~]. I find it quite inconsistent and annoying that [log] and [log~] are not the same, and I really think it is useful to be able to set the base of the logarithm to other than "e" in [log].

It all compiles and runs/works fine! I also updated the help file so it's really good to go.

cheers


porres:
This is totally backwards compatible, of course, and I just copied/adapted the same code from [log~]. I find it quite inconsistent and annoying that [log] and [log~] were not the same, and I really think it is useful to be able to set the base of the logarithm to other than "e". 

I tested this, it all compiles and runs/works fine! I also updated the help file. Anyway, it's really good to go.

cheers

porres:
Hi, I added a second commit for allowing [pow] and [pow~] to handle negative numbers.

It seems Pd avoids generating 'nan' and 'inf', one good example is how both [log] and [log~] will output "-1000" if you give it a <= 0 input.

Perhaps the motivation to not allow negative inputs here had a similar motivation, but I think it's very problematic to not be able to calculate the power of negative numbers and I don't think we should rely on [expr] for that. Most people also consider [expr] to be more expensive and it's just inconsistent.

But if the concern is to filter out inf/nan and turn them to zero, what I did was expand the check to allow negative numbers but still give a "0" output if it'd generate an inf (if left input is 0 and right input is < 0) or nan (if left input is negative and right input is not an integer).

umlaeute:
Closes https://github.com/pure-data/pure-data/issues/355

btw, adding a `Closes https://github.com/pure-data/pure-data/issues/355` message to your commit message (somewhere in the body) would automatically close the issue once the PR gets accepted)

millerpuckette:
I don't understand why - I can't seem to download this PR...  the URL

https://github.com/pure-data/pure-data/pull/356.patch

is just returning "Sorry, this diff is unavailable".

porres:
I'm still figuring out how to use git, but now I guess I know how to create branches :) - I'll redo this one

porres:
here's the new PR https://github.com/pure-data/pure-data/pull/427

PR #357, opened 2018-05-07
update all (#1)

cuinjune:
* noise~ inlet now sets random seed, updated noise~-help.pd to reflect this

* Update midi-help.pd

* dlopen failure considered an error, not a debuggig message

* fix non-portable pointer comversions in DSP routines

* changed 'float' seed message to noise~ to 'seed' to make it compatible
with 'random' object

* fix compiler warning

* fix "example 3" in list help which misused list store object


umlaeute:
most of your fixes seem to already be in the `update/0.48-2` branch.

to make life easier for @millerpuckette, i guess we should not have too many branches with the same features. so, would you mind redoing your PR like:
- create a PR against `update/0.48-2` (rather than `master`)
- use atomic commits (that is: one commit per feature/bugfix, rather than all in one)
- drop all the commits that have already been merged into `update/0.48-2`

umlaeute:
closing this due to it's duplicate nature

PR #359, opened 2018-05-15
update help files

porres:
- expr (forgot to include "%")
- float (added "see also [int]")
- int (added value example and "see also [float]")


doc/5.reference/expr-help.pd
```diff
@@ -1,4 +1,4 @@
-#N canvas 275 90 942 546 12;
```
umlaeute: the changes to expr-help are much more than just adding the forgotten "%".
since the help-patch is rather deeply nested, it's hard to spot all the changes. what else did you change?

umlaeute: the changes for float-help look good enough (but i don't understand why you changed `[value x]` to `[value f]`)

umlaeute: the changes to *int-help* have removed important examples, namely the truncating/rounding towards zero.

i think this is a much more important property than being able to send to a `[receive]` or `[value]`, so you shouldn't remove this information.

porres: that's the only information I changed, the other changes are just reorganizing the patch, shifting a bit to the right and left, window size, things like that. Note I authored all the big recent changes in this help file patch, I wouldn't really wanna change anything.

porres: trying to track back my thought process, I suspect I first copied [value x] to the help file of [int], and then thought it was a good idea to have different variable names. Then it seems I ended up changing both

porres: here's a new patch https://github.com/pure-data/pure-data/compare/master...porres:patch-1

porres: By the way, here's another simple and new update, related to this one https://github.com/pure-data/pure-data/pull/395

millerpuckette:
I updated "int" and "float" help files - haven't touched "expr" though, as I wasn't able to follow the changes.  (Turns out git is terrible at showing changes in patches).

danomatika:
Is this PR still relevant? Has this been updated in other PRs and/or commits by now?

porres:
most of it yeah, not the epxr help file yet, so I dunno, should I make a new PR for that?

danomatika:
If you either commit the expr-help.pd update here or send it to me somehow, I can just include it in the running bug fix branch. Otherwise, make a new PR from master and submit to the `update/0.49` branch.

porres:
I tried my best to send this to the right place and I don't really believe it worked, what do you say? 

anyway, here's the new PR https://github.com/pure-data/pure-data/pull/448

PR #360, opened 2018-05-15
initialize Inlet~ with float argument/inlet

porres:
adding a control inlet and float argument to [inlet~], this is crucial for creating abstractions that behave like externals, where you can initialize a default value other than zero when there's no signal connections, such as when giving it a default argument value. 

For instance, an oscillator abstraction that can initialize a frequency value, but also receive frequency input from a signal connection.

The inlet allows you to change that default value or specify the argument via some more complicated logic. Such as, if there are no arguments, load "x" value, but if there's one argument value "y", load that value instead. Or even, an optional first argument can be a symbol, then you need to get the value from the 2nd argument into inlet.


porres:
I agree with this report, https://app.codacy.com/app/PureData/pure-data/pullRequest?prid=1653132 where it says the code quality could be better.

I'm just copying some of the code/idea from dinlet~ https://github.com/MetaluNet/moonlib/blob/externals/moonlib/dinlet~.c

And by the way, I don't think one should solve this with an external, better to have it as a built in feature. One problem with dinlet~, for instance, is that it always calls the help file of inlet~, and doesn't have all of the features from inlet~, like up/downsampling.

I also noted dinlet~ crashes on windows, so we should test if this works on windows here.

Anyway, about the bad coding, copying the inlet union doesn't look like a good thing to do, and I suspect there are other better ways of doing this, quite likely not involving the need to duplicate the code.

Unfortunately, I'm not that great of a programmer, so this is what I could do...

cheers

porres:
also, this was supposed to be only the last commit, as the others have been made part of different pull requests here

porres:
Oh, I surely tested this on MacOS and it works fine!!!

Lucarda:
I have tested this PR on Windows and it works fine.

inlet



Also I found no compatibility troubles with older patches.

:)

porres:
Awesome, thanks a lot, don't know why, but the dinlet~ external, which is basically this, never seemed to work on windows.

umlaeute:
@porres please create a new PR that* only* includes this feature (without all the other commits).

porres:
done: https://github.com/pure-data/pure-data/pull/363

PR #361, opened 2018-05-15
re-including paf~ as an extra

porres:
paf~ is a complex and useful external, and it is still referred to in the audio examples  (F13 example)

but it really seems it needs a revision, and I can't really do that :/


umlaeute:
this was closed as it has been superseded by the clean PR https://github.com/pure-data/pure-data/pull/364

PR #363, opened 2018-05-16
inlet~ enhancements

porres:
The first feature adds a control inlet and float argument to [inlet~], this is crucial for creating abstractions that behave like externals, where you can initialize a default value other than zero when there's no signal connection (in short,  giving it a default argument value).

For instance, an oscillator abstraction that can initialize a frequency value, but also receive frequency input from a signal connection.

The inlet in [inlet~] allows you to change that default value or specify the argument via some more complicated logic. Such as: if there are no arguments, load "x" value, but if there's one argument value "y", load that value instead. Or even: an optional first argument can be a symbol, then you need to get the value from the 2nd argument into the inlet.

a screenshot with [inlet~] loading default values: 

screen shot 2018-05-16 at 15 21 21


porres:
I agree with the codacy report where it says the code quality could be better :)

I'm just copying some of the code/idea from dinlet~ https://github.com/MetaluNet/moonlib/blob/externals/moonlib/dinlet~.c

And by the way, I don't think one should solve this with an external, better to have it as a built in feature. One problem with dinlet~, for instance, is that it always calls the help file of inlet~, and doesn't have all of the features from inlet~, like up/downsampling. Most importantly, for some reason, dinlet~ crashes on windows, but a windows test here, so farm, works quite well!  I also tested this on MacOS and it works great!!! 

Anyway, about the bad coding, copying the inlet union doesn't look like a good thing to do, and I suspect there are other better ways of doing this, quite likely not involving the need to duplicate the code.

Of course, there are no backwards compatibility issues here ;)

Unfortunately, I'm not that great of a programmer, so this is what I could do...

cheers

porres:
A second commit allows data through [inlet~], see: 
screen shot 2018-05-16 at 15 15 40

porres:
further changes would need to be made to clone so it could also make use of inlet~'s new features

porres:
> although i don't fully understand the reasoning behind the extra argument if you can do the same with the inlet.

yeah, I see, you can achieve the same thing with the inlet than with the argument, but not the other way around. If I were to cut one of them, I'd stick with the inlet only, for sure.

I don't have strong feelings for keeping the extra argument, so I wouldn't fight for it, it's merely easier and more convenient if you just need to insert the value right there in the object rather than putting more objects outside, like [float] and [loadbang], connect them, etc. So you'd use the inlet only if you needed a more versatile and complex scenario.

porres:
BTW, I got the 2nd commit from this branch - https://github.com/x37v/pure-data/commit/0ec9854adf18364cc93252d8a951d351175a8f71 - which I thought had been made into a PR, but not. I'm actually not fond of all of those features, and even if I were, I also think it's best to propose the least intrusive and more essential changes first. you know, go "step by step"

umlaeute:
if you take other people's commits, you should `cherry-pick` them, leaving their authorship and credits intact.

since you also changed the permissions of `src/g_graph.c` and `src/m_obj.c` to make them executable :man_facepalming: I'd rather you revert your commit and take the others (doing a force-push)

umlaeute:
i just did what i suggested in the last post, so you can safely ignore it :-)

porres:
thanks, I wouldn't really know what to do exactly anyway :)

Spacechild1:
> i just did what i suggested in the last post, so you can safely ignore it :-)

thanks :-) BTW, the patch for [clone] is just the next commit:
https://github.com/pure-data/pure-data/commit/e8ae8d66c2b33e718d57ada10e10c21f599bd2f5

like Alex, I'm still unsure what's better: adding an extra outlet for messages or using an object like [route~]. I'm just afraid this might follow the strange path of [initbang], where we have API support but no (internal) object.

porres:
> BTW, the patch for [clone] is just the next commit

weird thing is that I tried that one and, for some reason, it didn't happen for me in my tests, so I didn't use it.

did you try it?

Spacechild1:
I think so... but let me check again

porres:
and we'd also need to do something to make the inlet and argument work in clone as well, but I wouldn't know what :/

porres:
> I'm just afraid this might follow the strange path of [initbang], where we have API support but no (internal) object.

I hear you, though I've seen Miller considering including initbang in vanilla.

Well, I also think it'd be important for vanilla to have more stuff, but first things first, if we take a bigger bite than we can chew, I'm more afraid if we get nothing instead ;) 

In the end, I'd even be ok with being able to code a crazy [inlet~] the way I want, but this isn't possible and, more importantly, my point is that I think at least this changes I'm proposing need to be incorporated in Vanilla's inlet~.

Spacechild1:
can you merge this commit? it fixes some missing bound checking.
https://github.com/pure-data/pure-data/commit/22775e199befa602719472e2773534295c597784

Spacechild1:
I checked again, [route~] works fine inside [clone]. here's a crude test patch:
[vinlet~-test.zip](https://github.com/pure-data/pure-data/files/2011186/vinlet.-test.zip)

it's indeed tricky to get the default float value inside a cloned abstraction because the abstraction inlet is always connected to the [clone] inlet. I don't see a simple way to do it. if we implement the connectivity checking we could dynamically connect/disconnect the two inlets within clone_dsp. this might do the trick.

porres:
ok, recompiled and tested. Yeah, it works now, so I don't know what was wrong before.

> it's indeed tricky to get the default float value inside a cloned abstraction because the abstraction inlet is always connected to the [clone] inlet. I don't see a simple way to do it.

Well, maybe something for later then. I can live without it for now. Honestly, I might not even need it myself. I can see maybe others doing it, but I'm really happier using clone as part of a patch that is loaded as an abstraction. Maybe if I post it, it will be clearer.

screen shot 2018-05-16 at 21 23 47

you see, I actually needed this inlet functionality for this [oscbank~] object, to set a default frequency value, and also route all messages via one single [inlet~] objects. But I didn't need it inside clone, as I still need to load clone inside the abstraction.

porres:
hmm, hadn't thought of that, so another object, like an external, connected to [inlet~] could sense if it has connections and also count the number of connections, huh?

porres:
there's actually good reasoning for keeping audio and signal into a same outlet/cord, cause pd objects can respond to that. Here, you can use it to feed a signal and a message to [fexpr~], for example

screen shot 2018-05-17 at 00 49 49

so if you need to split it, then use [route~], if the idea is to have something like [route~] in vanilla, maybe include [route~] as a vanilla, which looks pretty much like a vanilla object, quite simple and straightforward.

I like this design when things are simpler, such as with [inlet~] having only one outlet, then if you want to do something different, include an object for that.

porres:
yeah, sure, I had thought of that, but it looked kinda weird to me. But I'd also be totally fine if that's where we're heading, no real problem for me, just to be clear. 

I'm just saying what we have here is already good enough and doable, I'm satisfied, but I wouldn't really oppose if you  give it 2/3 extra outlets, get all sorts of bells and whistles into it etc... ;) y'all can also send me pull requests and I can include them here on this PR, no problem

Spacechild1:
> I like this design when things are simpler, such as with [inlet~] having only one outlet, then if you want to do something more complicated, put another object in your patch. 

I agree with this. also I have the feeling that putting additional outlets to all [inlet~]s would look clumsy. 
after all, _traditionally_ you would only use this one the first inlet of an abstraction, so putting a single [route~] seems more elegant to me.

Spacechild1:
> hmm, hadn't thought of that, so another object, like an external, connected to [inlet~] could sense if it has connections and also count the number of connections, huh?

I guess I had a blackout  :-) I edited my list accordingly.

Spacechild1:
BTW, I think the connectivity sensing (if it we choose to add it) definitly shouldn't be part of the traditional [inlet~] interface because - as I understood it - it's mainly useful for live patching (or for some edge cases, like automatically normalizing all signal input). so I think this would be better enabled as needed with a flag or an alternative constructor. just my two cents

umlaeute:
@Spacechild1 wrote:
> I agree with this. also I have the feeling that putting additional outlets to all [inlet~]s would look clumsy.

*look* clumsy? I fail to see (no pun intended) why the visual representation should hinder the required functionality.

> after all, traditionally you would only use this on the first inlet of an abstraction, so putting a single [route~] seems more elegant to me.

*traditionally* the `[inlet~]` wouldn't forward messages. if it starts forwarding messages to the already connected objects, you would suddenly get a change of behaviour without touching the patch (and I pray that @millerpuckette doesn't need to add another compatibility flag for disabling the forwarding) - rather than getting an error message, you are suddenly controlling a nested object.

Spacechild1:
> why the visual representation should hinder the required functionality.

of course it doesn't. just my (probably not so relevant) aesthetical preference.

>  rather than getting an error message, you are suddenly controlling a nested object.

good point!!

but even if we go for 2 seperate outlets, I'm still thinking about doing it with an alternative constructor (something like [vinlet~]). I guess we'd mostly use this feature to emulate CLASSMAINSIGNALIN and therefore all the other [inlet~]s in an abstraction usually wouldn't need the extra outlet. making a visual separation could be conceptually clearer. but I'm just thinking out loud.

Spacechild1:
now that I think of it, the missing error messages are a good reason *not* do add this feature to all signal inlets by default. by using an alternative constructor or a flag we explicitly decide which signal inlets accept messages and which don't (and therefore report an error to the user).

Spacechild1:
what about something like this? https://github.com/pure-data/pure-data/compare/master...Spacechild1:inlet-enhancements

Spacechild1:
vinlet
test patch:
[vinlet-test-new.zip](https://github.com/pure-data/pure-data/files/2013484/vinlet-test-new.zip)

porres:
and we have a different PR with more features now :) - https://github.com/pure-data/pure-data/pull/365 

maybe we can keep this as a "minimal changes" proposal, and the other one for the combo with extra outlets and stuff

Spacechild1:
shall I make a PR to switch to 2 outlets since we all agree it's safer?

porres:
yeah, ok, so 2 outlets, and then you don't need another object, and who cares if it looks "clumsy" :) 

for later: a flag that routes float inputs to the 2nd outlet

porres:
ok, 2nd inlet working fine, and also routing inside [clone] without issues. 

further work: make argument/inlet work inside [clone] / think about a flag to also route float from signals 

screen shot 2018-05-17 at 16 58 59

umlaeute:
i'm pretty convinced that flags are a bogus way to change the API.
having a flag (or a different objectclass) just adds too much brainload to the user - and i don't buy that it is confusing to have a 3rd outlet all the time. i guess that people will get used to it pretty quickly. (even if they never ever need it).

i believe the more options there are, the more confusing it gets.
ideally, there is a minimum number of possibilities to achieve any given goal.

- the default value for a signal is redundant if we know whether a signal is connected. just use `[sig~]` for that (and get rid of the extra argument and the inlet of `[inlet~]`; we really don't need three different ways to implement that)
- passing the float to the message outlet *all the time* allows one to have both float-to-signal (ala CLASS_MAINSIGNALIN) and distinct signal/float handling.
- prefix the signal-count with a routable selector (e.g. `connected`) - this allows for easy extension to additional information without adding more brainload.

Spacechild1:
hah, that's true! distinct signal/float handling is a bit tricky but indeed possible (you have to put a [*~] after the inlet and set it to 0 if no signal is connected to "disable" float to signal promotion)

porres:
> the default value for a signal is redundant if we know whether a signal is connected. just use [sig~] for that (and get rid of the extra argument and the inlet of [inlet~])

how exactly?

> passing the float to the message outlet all the time allows one to have both float-to-signal (ala CLASS_MAINSIGNALIN) and distinct signal/float handling.

A distinct signal/float implies passing the floats to the message outlet instead of promoting to signals, but that is a breaking change. If you had an abstraction that could take floats or signals, now they can only take signals. This is why I think you need a flag or something to say "don't promote this to signal and pass floats to the message outlet instead"

Spacechild1:
the key is to keep float to signal promotion *and* pass the float. see my answer above how to "disable" float to signal promotion in your abstraction.

porres:
yeah, I can see that if you keep float/signal promotion.

I can't see how to achieve the inlet/argument functionality yet though

Spacechild1:
inlet

we should probably put such examples in the help patch. would you do that?

porres:
why do you have [route signals]? what's coming out of the 3rd outlet?

in A) you'd also need to have the mid outlet of inlet~ into sig~, right? Is it really less confusing being left only with this rather complicated patch with 4 extra objects than just an extra float argument? Anyway, as long as I have the functionality, I'm still happy... but seriously, I gotta say, it's just a better option to have that functionality in the object. Seems the only argument against it is that, well, we found a hacky/nastier way to do it if you don't have it, so let's not do it :)

in B), can you have a signal value of, say, 10 going to unsig~ while floats, like 67, are sent to the number box? Where can I test your new super thing?

Spacechild1:
inlet -korr

here's the corrected version.

porres:
well, to me, the need to create an abstraction with 11 objects (well, actually you might not need some of these, but still...) is really an even more drastic example of a "hacky" way, like I was saying, instead of just having the argument do all that for you... documenting this and explaining the logic behind this to people is also a lot more complicated than just using an argument. Would you really rather do this for the sake of simplicity? This is just not elegant, and less efficient as well.

I see the point that maybe too many options is confusing, but this won't even be considered an option at all if we just have the argument. This only comes up as a workaround to how to do it all on your own if you don't have it... ok, I can see a proof of concept, but how is this an argument to not implement something that would be simpler?

Spacechild1:
yeah, your default value feature might be used quite often, so I'm also not totally convinced that we should let users build this themselves... maybe ship an abstraction? 
nevertheless, you could do a PR with your feature to @x37v's branch, with the connectivity sensing API support I can later implement it for [clone].
OTOH, separating floats from signals is rather a rare need (no built-in object does this and only few externals do) and quite easy to build, so I think it's OK not to introduce a flag.

porres:
> maybe ship an abstraction?

if you need to facilitate by giving something, why not the inlet/argument functionality? What's so wrong about it? It's the one common feature for objects. Maybe if there was already an easy alternative, but no, so it's really pushing it, as I've been pointing, to give a rather complicated solution as an idea that this is already sorted. 

3 ways of doing it is too much? Well, first, two of them are what you always find in Pd. You have an argument and an inlet to [f]. Why two? Why both an inlet and an argument? And the third one is something no one will ever care to use and would rather just use this that literally takes one second.

So yeah, I'm not taking this down from my Pull Request. It's not up to me in the end, but this is my proposal, it's what makes sense to me.

> you could do a PR with your feature to @x37v's branch

yeah, that'd be great, we can join forces and make things simpler into one Pull Request with everything. Maybe not all functionalities will end up to be taken, but some of them can be "cherry picked". And then I could close this one.

But then, also, you've figured it out a better way to code my proposal. You know, I'm the least skilled programmer here. So, if you want, you can just go ahead and include it yourself too. I can better help by testing and working on the documentation.

cheers

x37v:
@Spacechild1 @porres I think the inlet argument for the default value makes sense, I'd definitely take a PR into my branch for that. I don't think separating floats back out from signals makes sense, I like the initial promotion that we merged in from @Spacechild1 yesterday. Unfortunately I don't have time until next week to look into this myself but if one of you want to put a PR together for the inlet argument default value I'd love to review it.

Spacechild1:
I can do that next week or so. the default value is a bit tricky to implement for [clone] but it's totally possible. I would keep sending the floats, it doesn't hurt and enables users to treat them differently from signals if they want.

porres:
>  I don't think separating floats back out from signals makes sense, I like the initial promotion that we merged in from @Spacechild1 yesterday

yeah, the idea is to keep promoting floats to signals, but also sending them to the control messages outlet. This would allow distinct float/signal handling, which could be cool and useful in some cases. I could compromise and meet halfway to the idea that this won't be a widely used feature, so maybe not bothering coming up with a "flag" right now.

If it weren't for this, I'd still go for a separate object like [route~] to manage this scenario, because I'm also all for simplicity and less brain load. But it'd be impossible as any other object down the chain only receive the promoted to signal result, thus unable to route it. So control outlet it is ;)

I wouldn't have much use for the 3rd outlet, and I don't think many would, but with it you could at least work to get separate float/signal handling. So it actually balances things

porres:
actually I would like to keep this as an alternative suggestion, a lighter version without the third outlet, in the events the other one looks like "too much" - not that I'm against it

Spacechild1:
I think #365 is easier to follow. in the end, we can always do something with the third outlet. so I would also suggest to move the whole discussion to the other PR to avoid confusion.

porres:
I'm now convinced a third outlet could be fun, so whatever, I'm closing this one and let's just deal with #365

PR #365, opened 2018-05-17
inlet~ features

x37v:
@Spacechild1, @porres and I have added 2 new outlets to `inlet~`.

* outlet 1 - acts as it always did, giving the signal patched into the abstraction/subpatch
* outlet 2 - gives control information given to the `[inlet~]`
  * this way abstractions can act more like built ins that take both signals and messages, like `[dac~]`
  * also, now `[clone]`'s abstractions can always be controlled by its first inlet instead of only its first control inlet.
* outlet 3 - gives a count of signals connected to the `[inlet~]`
  * this can be used to make no-click connections. I patch my abstraction to a summing node, then I connect a signal to my abstraction which detects that there is a connection and fades in an envelope.
  * unfortunately the disconnection information isn't received in time to use this to make no-click disconnects, though maybe @millerpuckette has an update to allow for that.

I think both outlets are useful but could make a separate PR for outlet 2 as it is definitely the most important if you ask me.


porres:
ok, this has similar features to https://github.com/pure-data/pure-data/pull/363 and could even replace it, but it'd need to consider the float argument and inlet to inlet~, have you seen that? What do you think? Would you include it?

Spacechild1:
cool, but maybe consider not making this the default behaviour of [inlet~] but rather use a flag or alternative constructor (something like this: https://github.com/pure-data/pure-data/compare/master...Spacechild1:inlet-enhancements).
reasons are:
a) backwards compatibility
b) just like with objects/externals, not all signal inlets of an abstraction are supposed to accept messages (traditionally it's only the first one). to communicate this to the user, "ordinary" signal inlets should keep posting an error.

x37v:
@Spacechild1 
> a) backwards compatibility

This PR is fully backward compatible, it only adds new features and doesn't change existing features.

RE *b*, I get that but I don't see any reason not to make all `[inlet~]` more capable.. say you have a mixer and you want to have enveloped mutes. This message paired with signal lets a message to "mute" a inlet be sent to the inlet its muting. Why add extra complexity to object creation just to lose capability?

x37v:
> This PR is fully backward compatible, it only adds new features and doesn't change existing features.

or, @Spacechild1, maybe you mean making abstractions that are backward compatible with previous versions of PD?

x37v:
> ok, this has similar features to #363 and could even replace it, but it'd need to consider the float argument and inlet to inlet~, have you seen that? What do you think? Would you include it?

@porres I'm not sure what I think about having floats pass through the `[inlet~]`

Don't a lot of other signal objects promote a float to a signal? Do you have an example of a built in object that segregates float and signal on the same inlet?

Spacechild1:
> Don't a lot of other signal objects promote a float to a signal? Do you have an example of a built in object that segregates float and signal on the same inlet?

Alex' inlet to [inlet~] doesn't prevent float to signal promotion.

x37v:
@Spacechild1 

>Alex' inlet to [inlet~] doesn't prevent float to signal promotion.

Ahh, maybe I misread it then, it simply sets the signal value to a specified value if you give it an argument? oh huh, that seems kinda nice actually.
The explanation of the inlet for `[inlet~]` is sort of confusing but I'm guessing that it just lets you override the value set as the argument? like most other objects?

Spacechild1:
exactly!

x37v:
@Spacechild1 ahh, that makes sense. I wonder if it should be an arg or just a newly named object which exposes these 2 outlets and @porres 's default argument changes.

I was initially inspired to have control outlets on `[inlet~]` because of `[clone]` and the fact that the control interface for its cloned instances isn't consistently accessible via its first inlet. Making a newly named object or an object that requires an argument doesn't resolve this issue without making you modify all your abstractions, but I do get the trade-off with error messages.

Spacechild1:
>  without making you modify all your abstractions

don't you have to modify them anyway? old abstractions have been written under the premise that signal inlets don't forward messages

porres:
you can code an object to have a special method for float instead of promoting it to signal

my inlet doesn't let you do that, floats are always promoted to signals, I just thought, well, for abstractions, it could be like that. Whatever other way to deal with this felt too complicated for me.

> now you don't get error messages when you send a message to a signal inlet which isn't supposed to accept messages, instead you might accidentally trigger something funny inside the abstraction.

i don't think that's relevant at all, you had something that didnt accept messages, so you shouldn't be using messages, and now why would you send it messages? And if it doesn't work, so what? Isn't these things supposed to be documented and tell you how to use them, like, what kind of messages it takes and stuff?

but whatever ;)

x37v:
@Spacechild1 
> don't you have to modify them anyway? old abstractions have been written under the premise that signal inlets don't forward messages

In order to use the extra control outlet you do, but to have a unified interface to `[clone]` instance control you don't. I guess these are kind of 2 separate issues but they end up being resolved together.

Spacechild1:
I still don't understand :-(... can you give an example?

Spacechild1:
@porres yes, it's rather unlikely, but you know, Murphy's Law... :-) the point is rather that the user should receive an error message if he/she does something wrong. I don't like it at all if software is silent about my mistakes, I'd rather prefer it to shout at me.

x37v:
@Spacechild1 
before this patch, if i have an abstraction, *my-abstraction-name.pd* with only one inlet: `[inlet~]` I have no way to view the abstraction via messages to `[clone my-abstraction-name]`
for example: `|vis 1 1(` doesn't work.
If it has 2 inlets, the first is audio and the second is a signal, I can view the abstraction as well as send messages to it, but now via the 2nd inlet of `[clone my-abstraction-name]` and not the first.

porres:
Ok, so, a 2nd outlet will now not care whatever you send this, right? if before it gave you a warning, now it won't anymore, and there's no problem with that, right? Right? 

So ok, 2nd outlet it is, I'm sold, it has the advantage of not needing an extra object to route it, and it can't have whatever concerns for compatibility that would need another (let's pray to god not) compatibility flag.

Moreover, we can think of a flag that separates the signal input and the float input, so the float can be routed to the 2nd outlet, then you can make it behave exactly like any other coded external.

The argument and control inlet to [inlet~] keep just promoting floats to signals

how about that?

porres:
>  I don't like it at all if software is silent about my mistakes, I'd rather prefer it to shout at me.

there's no mistake in sending whatever to [inlet~], when you couldn't before, the error is supposed to happen where this connection goes, say it goes to a [fexpr~] object and you send a message "banana", [fexpr~] will then tell you "no method for banana" - there's no silencing going on...

Spacechild1:
@x37v now I see what you mean! you're totally right.
I've just figured out that signal inlets for [clone] can always be generic and forward messages. I just updated my branch: https://github.com/pure-data/pure-data/compare/master...Spacechild1:inlet-enhancements
sending [vis 1 1( to a signal inlet of [clone] now always works but instance messages like [1 foo bar( give an error if the abstraction itself has an ordinary signal inlet.

Spacechild1:
> Ok, so, a 2nd outlet will now not care whatever you send this, right? if before it gave you a warning, now it won't anymore, and there's no problem with that, right? Right?

you're right, the two outlet solution is actually safe. for the missing errors: I can probably live with that.

> Moreover, we can think of a flag that separates the signal input and the float input

this is possible, we would only need another flag in the t_inlet struct. we can easily pack it inside the unsigned char of i_generic.

we have to think about the three outlets, though. don't you think it's overkill to have the third outlet (number of connections) all the time? maybe only enable it with a flag? or another constructor? after all, this is a feature which is not needed that often, I suppose...

porres:
> we have to think about the three outlets, though. don't you think it's overkill to have the third outlet (number of connections) all the time? maybe only enable it with a flag? or another constructor? after all, this is a feature which is not needed that often, I suppose...

sure, I'm with you all the way. It's very specific, I don't really get or am convinced why it is needed at all :)

Spacechild1:
from what I understood, it could be useful in dynamic patching, e.g. to automatically fade when you connect/disconnect a signal. or to automatically normalize the input (dividing the signal by the number of connections), but I think we all agree these are rather edge cases. so we probably shouldn't put the third outlet by default.
OTOH, API support for getting the number of connected signals could be useful, like for implementing your floatsignalvalue for cloned abstractions.

porres:
so, how about that  inlet argument in here as well? ;)

Spacechild1:
I've been terribly busy, but I'll do it soon.

porres:
hi, still busy? :) cause we may have a shot for this to be included in 0.49 soon

Spacechild1:
I'm terribly busy again :-) but I actually did the PR in June: https://github.com/x37v/pure-data/pull/2
@x37v  do you want to merge it?

x37v:
done!

Spacechild1:
you only approved it but not actually merged it ;-)

x37v:
ha, oops, now it has been merged!

porres:
Hi, I edited the help file to include the newly added functionalities, please consider it and merge it

https://github.com/x37v/pure-data/pull/3

x37v:
hey alex,
I'll have to look at this a little later.. I forget if the branch in my repo has
the default setting for [inlet~] that you detail in there, i think not..
the rest looks good though, i'd definitely get in this or part of it :)

x37v:
@porres oops, look like your PR was into the wrong branch but i copied it over to this one. Thanks!

Spacechild1:
personally, I wouldn't clutch all [inlets~] with the third outlet but rather enable it with a flag or with a special constructor (e.g. [vinlet~]). maybe @millerpuckette can tell us his general thoughts about this?

x37v:
@millerpuckette i think the 2nd outlet is very useful but so far the 3rd outlet isn't as useful as I'd like it to be.
My hope was that we could use the signal connectivity info to allow for clickless connectivity for abstractions (ramp output from zero on input connected, ramp to zero on disconnected and manage the input disconnection with either a block size recording played backwards or just an input delay so that ramps happen after/before the discontinuity..)
It has been some time since I've visited this so maybe I'll try to implement the clickless abstraction and see if that actually works.. but either way, I'm willing to make the 3rd outlet optional via an argument or take it out entirely if you think its just noise.

x37v:
[clickless.tar.gz](https://github.com/pure-data/pure-data/files/3907387/clickless.tar.gz)

Here is an example of a 'clickless' connectivity abstraction. This simply does a delay and ramp approach, but I could see it happening without a delay if we can make the abstraction itself manage the disconnection from its parent after a ramp down. It would still manage a ramp up on initial connection.

Nothing we can do if an additional signal is added to the first connection though.

porres:
> Here is an example of a 'clickless'

do you have a screenshot?

millerpuckette:
I think the second outlet is well wirth it, and so would be taking "float" messages to set the state - the rest of this seems to be adding complexity to something that's already too complicated.

porres:
Yeah, from the start I felt like a 3rd outlet might be a bit too much, this is why I kept a similar PR with just a 2nd outlet. So I wouldn't mind seeing it go...

Other than that, I can't wait for the 2nd outlet feature!!! It would really fix several abstractions of mine.

x37v:
> > Here is an example of a 'clickless'
> 
> do you have a screenshot?

clickless

x37v:
> I think the second outlet is well wirth it, and so would be taking "float" messages to set the state - the rest of this seems to be adding complexity to something that's already too complicated.

@millerpuckette being able to write abstractions that allow users to quickly patch dsp and avoid discontinuities seems really useful to me but if that feature is what will hold back the other feature I'll remove it. I can put together an augmented PR tomorrow with:

1. the 2nd outlet from `[inlet~]` that passes messages sent to that inlet of the abstraction.
2. the inlet to `[inlet~]` that sets the default output of `[inlet~]` if nothing is attached.

@millerpuckette is that what you're thinking?

millerpuckette:
That sounds right to me.  I think the clickless-hot-patching idea is cool but I think it's an unusual situation for which it would be OK to require a "modded" Pd in order to keep the base version of Pd as clean as possible.

x37v:
@millerpuckette I believe this is all in now.. largely work by @Spacechild1 

the only thing that I've seen that doesn't seem to work is that the default value set via `[inlet~ float]` or via sending a float to `[inlet~]` doesn't do anything inside `[clone]`

1. `[inlet~]` can now pass messages through to abstractions just like internals/externals
2. `[inlet~]` can have a default signal value set via an argument or its inlet
3. `[clone]` consistently uses the left most inlet for control messages like `vis`

help file updated by @porres with slight edits from me.

millerpuckette:
I'm going to look for a way to do this without having to extensively mess with m_obj.c - I'm trying to keep that as uncomplicated as I can.

porres:
> I'm trying to keep that as uncomplicated as I can.

while you're at it, maybe try to fix the default value (via argument/inlet)inside [clone]? :)

millerpuckette:
> while you're at it, maybe try to fix the default value (via argument/inlet)inside [clone]? :)

I can't remember what needs fixing there.

millerpuckette:
OK, I think master has a good solution (but getting notified about edits is on back burner for now).

porres:
It seems everything works fine if you give it a default value as an argument or via the inlet. But as Alex Normal ( @x37v ) mentioned, this feature isn't working yet when you have inlet~ in an abstraction loaded in [clone]. I can try it out, but not now.

x37v:
with master commit `2fa003d03d2deb6792659ce6cbb81cda9df38581` this PR is no longer needed.

I verified that the `[clone absname~ 1]` outputs 0 if absname~ is simply an inlet~ with a non-zero default value set via a float, hooked to an outlet~
but this was a problem with this PR as well and maybe kind of an edge case?

Thanks @millerpuckette !

millerpuckette:
oops, no, that's important.  (and checking on it revealed an unrelated serious problem too!)  I'll get back on this.

porres:
yeah, I tested and couldn't get the default value from inlet~ inside clone.

Anyway, thanks a lot for this :) I know I personally requested for this one so I feel considered :)

Spacechild1:
Thanks a lot Miller for including this! Now we can build abstractions which feel just like externals :-)

porres:
Not sure if up to https://github.com/pure-data/pure-data/commit/dac518a85429db980648ff0f29fac926fb7d1412 we should have this fixed, but I tested and I still can't have an abstraction inside clone getting a default float into its inlet

And also, I found another bug :/ https://github.com/pure-data/pure-data/issues/809

Spacechild1:
> but I tested and I still can't have an abstraction inside clone getting a default float value into inlet~

I think Miller said he's working on it.

millerpuckette:
It turns out it's impossible without really messing with things, since there's always a signal connected to the inlet~ in the abstraction.  Clone itself makes a separate inlet, and that one promotes to signal.  So even if it were possible, sending "float" to one included object's inlet~ would end up affecting all of them - probably not a desirable result.

porres:
> So even if it were possible, sending "float" to one included object's inlet~ would end up affecting all of them - probably not a desirable result.

hey, I never really thought about it, and I'm not sure if I follow it, but it seems is something that is just not pertinent? Meaning, we're just gonna leave it at that? Or are you considering messing with things to allow this to happen?

Coming to think of it, at first I can't really think of a use case where I'd need an inlet~ in clone having a default argument if no signal is connected.

x37v:
> > So even if it were possible, sending "float" to one included object's inlet~ would end up affecting all of them - probably not a desirable result.
> 
> hey, I never really thought about it, and I'm not sure if I follow it, but it seems is something that is just not pertinent? Meaning, we're just gonna leave it at that? Or are you considering messing with things to allow this to happen?
> 
> Coming to think of it, at first I can't really think of a use case where I'd need an inlet~ in clone having a default argument if no signal is connected.

I could see it being useful if you're using signals for control and want to have defaults if nothing is connected.. for instance, a signal controlled frequency for an oscillator bank, could be nice to have an audible frequency if nothing is connected.
but either way, i think having the `fwd` feature and the default setting in non cloned abstractions is the most important part of this.

porres:
In this case, you want the [clone] itself to behave as an external. I guess for that there are workarounds. All you need is an abstraction wrapper over your abstraction loaded via clone, and this is what I was already doing for my abstractions in the ELSE library before all this, cause for different reasons, it's just better if you do that. Now, I'm starting to update them, and for my oscillator bank, I'm just doing this.

Sure, maybe it'd be good to have it working in clone, but I'm just saying it is not a show stopper, and I don't see a specific use case that we cannot deal with via a workaround. Of course I'm taking into account that we'd need to undertake some drastic overhaul in Pd's code to achieve this. And I also wouldn't like this to not be part of the next release because we can't figure this out now.

So, worst case scenario, just mention in the documentation of inlet~ and clone that you need to be aware of this.

Screen Shot 2019-12-05 at 17 08 23

x37v:
btw, with the default value settings, I was able to rework my `clickless` abstraction to work without the 3rd outlet. I'm guessing it is block size dependent:  
newclickless
[clickless.tar.gz](https://github.com/pure-data/pure-data/files/3935378/clickless.tar.gz)

PR #366, opened 2018-05-26
[soundfiler] - remove tiny default value of "maxsize" and prevent signed integer overflow

### LINKED ISSUE #380: -maxsize no bigger than 2.14748e+09 in [soundfiler] on windows
In windows, `[soundfiler]` won't accept a `-maxsize` flag bigger than *2.14748e+09* or something (I got that with trial and error) for some reason.

the console shows a typical syntax error for `[soundfiler]`:

```
usage: read [flags] filename tablename...
flags: -skip <n> -resize -maxsize <n> ...
-raw <headerbytes> <channels> <bytespersamp> <endian (b, l, or n)>.
```

Spacechild1:
1)
the default value for maxsize has been 4000000 samples which IMHO is not necessary. I basically end up always doing [read -resize -maxsize 1e007 ...( when loading soundfiles... 
I think "maxsize" should be opt-in so I've set the default value to LONG_MAX. 

EDIT: I could've also used a magic number like -1 instead of LONG_MAX and test against it to avoid including "limits.h"

2)
users would often do something like "-maxsize HUGENUMBER" to practically circumvent the default value. on Windows, however, `long `is always 32 bit, and something like "1e010" will lead to integer overflow, most likely resulting in a negative number and making soundfiler_read fail. I added a simple check to prevent this. this fixes #380

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -1441,8 +1441,7 @@ long soundfiler_dowrite(void *obj, t_canvas *canvas,
     int argc, t_atom *argv, t_soundfile_info *info)
 {
     int endianness, swap, filetype, normalize, i;
-    long onset, nframes, itemsleft,
-        maxsize = DEFMAXSIZE, itemswritten = 0, j;
+    long onset, nframes, itemsleft, itemswritten = 0, j;
     t_garray *garrays[MAXSFCHANS];
     t_word *vectors[MAXSFCHANS];
     char sampbuf[SAMPBUFSIZE];
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -23,6 +23,7 @@ objects use Posix-like threads.  */
 #include <string.h>
 #include <errno.h>
 #include <math.h>
+#include <limits.h>
 
 #include "m_pd.h"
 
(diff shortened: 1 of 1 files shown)


umlaeute:
given that this can easily lock up your computer (esp. on 64bit systems with virtual memory) i don't really think that `LONG_MAX` is such a good idea.

Lucarda:
How will this behave when we have the double-precision Pd. One of the goals is being able to load tables larger than 16777216 points.

umlaeute:
@Lucarda double-precision is not at all about being able to load tables with more that 16777216 points. (which you can do with current Pd).

apart from that, I don't see how the patch would be effected at-all by double-precision Pd (or vice-versa)

Spacechild1:
> given that this can easily lock up your computer

this might only happen if a user tries to load a ridiculously large audio file - which would be the user's fault - and I wouldn't call that *easily*. Theoretically, this can already happen when resizing arrays and I think it's even more likely to accidentally type some huge number than accidentally loading an extremely large audio file.

if you really think this might be a problem, what about 0x7fffffff then? this would be 13h of mono audio which I guess is enough 99.9% of the time.

millerpuckette:
I grabbed teh overflow-protection commit from here but not the others (hoping to fix bugs in
0.48 without introducing new problems).  I think a reasonable default maximum might be 0x7fffffff
(the 4 million one was chosen back in days when memories were much smaller; today it seems much
less likely you'd hose your computer by accident by loading a soundfile.)

Spacechild1:
ok, thanks!

Spacechild1:
@millerpuckette are you still planning to get rid of the 4000000 limit?

PR #368, opened 2018-05-28
Help fixes

reduzent:
- pointer-help.pd: fix arguments to second pointer example (pointer takes struct names as arguments, not subpatch names)
- element-help: fix typo (mane -> name)


umlaeute:
included in update/0.48-2

PR #374, opened 2018-06-04
feature: intelligent patching

umlaeute:
this PR includes a number of improvements on handling connections (and patching in general) from within Pd (this list is getting updated as new features are added)

- connect two objects, by selecting them and pressing <kbd>$modifier</kbd>-<kbd>k</kbd>
- extend connections to the right by selecting a connection and pressing `Duplicate` (<kbd>$modifier</kbd>-<kbd>d</kbd>) (that is: if the selected chord connects `foo:1` with `bar:0`, `Duplicate` will create a connection `foo:2 bar:1`, then `foo:3 bar:2`, ...
- connect one outlet to multiple inlets by hovering the dangling connection over an outlet and pressing <kbd>Shift</kbd> (like in [Max](https://youtu.be/K1OMDaaHujo))
- connecting multiple objects to a single inlet
  - select a group of (source) objects and connect one of the objects to an inlet (of an unselected object) - all selected objects will connect to the same inlet
- connecting multiple objects to a the inlets of an object
  - select a group of (source) objects together with a *sink* object and connect one of the sources to an inlet of the sink - the source objects will connect to the inlets of the sink from left-to-right.
- connecting a single outlet to multiple objects
  - connect a (signal) source with one of a group of selected objects, and the source will be connected to ll selected objects 

- connecting the outlets of an object to multiple objects
  - select a source object together with a group of *sink* objects and connect one of the source's outlet to an inlet of any of the sink objects - the source object's outlets will connect to the inlets of the sinks from left-to-right.

- insert an object into a connection
  - select a (disconnected) object and a single connection and press <kbd>$modifier</kbd>+<kbd>k</kbd> - the object will be inserted into the connection
  - OR select two connected objects and a third disconnected object and press <kbd>$modifier</kbd>+<kbd>k</kbd> - the object will be inserted into the connection

- remove an object from a connection ("bypass")
  - select three chained up objects (`[objA] -> [objB] -> [objC]` and press <kbd>$modifier</kbd>+<kbd>k</kbd> - the middle object `[objB]` will be bypassed by disconnecting it (and establishing a direct connection between the other two)

- remove all connections from/to an object
  - select a single object and press <kbd>$modifier</kbd>+<kbd>k</kbd> - all connections are gone

- swap connections
  - if you have mistakenly connected the wrong inlets (or outlets) of two objects, you can easily swap connections by selecting one connection and then <kbd>Shift</kbd>-clicking another connection (the two connections must share at least one object) Closes: https://github.com/pure-data/pure-data/issues/387

- pasting without overlapping objects
  - when pasting some objects into the canvas, Pd will check whether it there are already objects on the same position (as the first object in the copybuffer). if so, Pd will slightly offset the pasted object (similar to `duplicate`)

- Paste Replace
  - you can replace all selected objects with one of your liking by copying a single object (or msgbox,...) and then "Paste Replace" the selection (via the menu)

- continue to autopatch in subpatches
  - when patching, Pd will automatically connect the last object to the newly created one (that's old news). When the new object is a subpatch, Pd will now automatically create an `[inlet]` (or `[inlet~]` depending on what the source object wants to send you) within the subpatch, thus automatically establishing a connection into your new subpatch, where you can continue to autopatch) Closes: https://github.com/pure-data/pure-data/issues/390


Ant1r:
nice... it will soon become indispensable!

umlaeute:
more features:
- connect one of a group of selected objects to a sink, and all selected objects will be conntected to the sink (provided they have enough outlets and the iolets are compatible)
  - e.g. selecting `[unpack]`, `[tabplay~ foo]` and `[timer]` and connecting the 2nd outlet of `[unpack]` with `[print]`, will connect `[tabplay~]`'s second outlet as well, but not `[timer]` (as there is no 2nd outlet)

- connect a (signal) source with one of a group of selected objects, and the source will be connected to ll selected objects

umlaeute:
with the recent commits, connection handling is almost on par with purr-data's *intelligent patching* facility:

> * If you select exactly two objects A and B, say, and then connect one of the outlets from A to one of the inlets of B, then starting from the initial outlet-inlet pair the remaining outlets of A will be connected to the corresponding inlets of B.
> 
> * If you select two (or more) objects B and C, say, and then connect an outlet of a third, unselected object A to an inlet of B, then the corresponding connection from A to C will be done automatically. Conversely, you can also connect an outlet of B to an inlet of A to have the corresponding C-A connection completed for you.
> 
> * If you select three (or more) objects A, B and C, say, where A has two outlets or more, and then connect an outlet of A to an inlet of B, then the next outlet of A will be connected to the corresponding inlet of C. Conversely, you can also connect an outlet of B to an inlet of A, and have the corresponding outlet of C connected to the next inlet of A. This works for an arbitrary number of source or target objects, considering the “other” objects in left-to-right, top-to-bottom order.1
> 
> * Finally, pressing the shift key while doing connections will let you do multiple connections from the same outlet in one go.

(I say "almost", because §3 (connecting multiple objects with a single multi-io-object) is sometimes ambiguous - as in Purr Data; however PurrData allows the user to select an alternate connection algorithm by pressing <kbd>Control</kbd> during connecting in the case of ambiguity...something i haven't implemented yet)

porres:
If it's not too much to ask, here's another feasture. Upcoming max 8 will have it and I like it even better than what's here already. So, holding shift and dragging objects in and out of cords magically connect/disconnect them.

It's the 1st feature in here: https://www.youtube.com/watch?v=Nk6fv4hXaGg&feature=youtu.be

LGoodacre:
Are you still taking suggestions for this? I've always thought it would be great if you could autopatch a subpatch with an inlet already given. Ie. if you have an object selected, press Ctrl 1 and type [pd something], the connection remains intact and you get a subpatch with a single inlet at the top. (Better yet, the inlet should be selected, so you can continue the autopatching chain within the subpatch).

umlaeute:
@LGoodacre i'm doing something like this in my live-coding performances by means of the iemgut's `autoabstraction` feature (which will create new abstractions from a "template" for any unknown object the users tries to create) - but only for abstractions not for subpatches.

so i hear you.

one problem i'm facing there is that it's undecided whether the inlet should be a signal inlet or a message inlet.

LGoodacre:
I see what you mean. Is it possible to scan the selected object to see if it contains a '~' character?

You can run into problems with signal vs. control inlets with regular autopatching anyway, (ie. [osc~]--[f]). So it seems to me that the confusion with inlets in a new auto-subpatch feature would be no worse than the current situation.

umlaeute:
@LGoodacre i've implemented a draft (based on this branch) in my personal clone of this repo: https://github.com/umlaeute/pure-data/tree/feature/connections (it took surprisingly much code to implement that feature, so i'm hesitant to add it to the PR)

danomatika:
This might be better as different issues, but since some of the logic in this PR is related. here goes.

A Max-using friend suggested also adding Max's "Paste replace" feature. Essentially, it allows you to select some a object, copy, then select some destination objects, and pasting overwrites the selected objects with instances of the source object, keeping the existing connections, where possible. I believe it's activated by hold Alt while pasting.

Another thing which I know is a different issue on here but just throwing it in again: copy paste into the same patch creates a new copy ON TOP of the existing objects. This is obviously a pain point and I noticed that Max's default paste creates the copies offset like Pd's "Duplicate." I think the best solution is to simple add the offset if the copy source canvas and paste dest canvas are the same. If you copy from one canvas and paste into a completely different canvas, then keep the same location as we have now. That would be the simplest/smartest solution IMO.

umlaeute:
i cannot possibly imagine how the "paste replace" would actually work (without being able to try it).
but anyhow, @LGoodacre, @porres, @danomatika and whoever, please create separate tickets for each feature you'd like to see. it's getting really hard to track requests in a long discussion thread.

if you don't want to soil Pd's issue tracker with feature requests that are unlikely to get implemented, you could use the [tracker of my clone](/umlaeute/pure-data/issues), although I don't see why.

Lucarda:
+1

I've tested this PR on windows and it is running fine.

:)

danomatika:
I'll test in macOS soon.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

fdch:
Hi there, I can't seem to get the <kbd>shift</kbd> key to work with cords in macos. It seems to be disabled: once I hit <kbd>shift</kbd> I can no longer connect cords until I release it. I compiled from here, and tried with both Wish 8.6.6 and 8.6.7 (using `osx-app.sh`).

umlaeute:
updated to merge cleanly again...

umlaeute:
brought up-to-date with current `master`, so it applies cleanly again.

Issue #372, opened 2018-05-31
Abstractions named "anything"


HenriAugusto:
If you try creating an abstraction named "anything" it will not be instantiated. Also after trying to create an [anything] PD won't instantiate other valid abstractions!

It seems that after the failed [anything] PD **will** instantiate previously used abstractions but **will not** instantiate others.

**edit:** I could not find the reason in the source code so i don't know if this happens by design. If its not by design then it's a bug. If it is by design there should be some kind of file documenting that reserved name.

porres:
seems you cannot have a compiled object called anything as well...


<img width="805" alt="screen shot 2018-05-31 at 13 00 37" src="https://user-images.githubusercontent.com/5360142/40793434-bf11c872-64d2-11e8-938b-8efe94db54fd.png">

Spacechild1:
that is actually a pretty funny bug! here's what's going on:

when you type something in an object box you're actually sending a message to pd_objectmaker with the selector being bound to a "new method" (for externals) or "do_create_abstraction" (for abstractions). 
generally, if a message can't be found in a class (for pd_objectmaker this means object/abstraction hasn't been loaded yet), pd_typedmess falls back to the anything method - in the case of pd_objectmaker that's "new_anything" (see m_class.pd), which is responsible for loading unknown objects.
now when you do [anything ...], you basically override pd_objectmaker's anything method, preventing it from loading any new abstractions/externals.

for why the [anything] abstraction itself doesn't load, I'm not quite sure yet...

anyway, the fix is easy: add a check in "new_anything" that `s != &s_anything`. 
I'll put it my list and make a PR.

porres:
awesome, cause now I wanna make an object called [anything] :)

Spacechild1:
can be closed

HenriAugusto:
okay

porres:
have you tested this? cause I still can't create an abstraction called anything, or is this not in 0.48-2? just in the next release?

Spacechild1:
> cause I still can't create an abstraction called anything

yes, but don't you get a warning: "object name anything not allowed"? creating such an object breaks Pd by design, so we explicitly disallow it.

porres:
oh... i didn't realize the idea was that you now can;t do it, haha

PR #381, opened 2018-06-20
disallow creating an object with the name "anything"

Spacechild1:
...because this would eventually override pd_objectmakers's anything method, essentially breaking the abstraction/external loading mechanism. 
fixes https://github.com/pure-data/pure-data/issues/372

### DIFF
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -778,6 +778,10 @@ void new_anything(void *dummy, t_symbol *s, int argc, t_atom *argv)
       error("maximum object loading depth %d reached", MAXOBJDEPTH);
       return;
     }
+    if (s == &s_anything){
+      error("object name \"%s\" not allowed", s->s_name);
+      return;
+    }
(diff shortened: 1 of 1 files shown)


millerpuckette:
This is an interesting way an extern could bring Pd to its knees - but it will always be possible
for unfortunatelt written externs to crash Pd so I think this doesn't need to be enshrined in code -
perhaps it's worth a warning in the externs howto instead.

Spacechild1:
but it's quite easy for users to name one of their abstractions "anything", as this user did: https://github.com/pure-data/pure-data/issues/372
without knowing the internals of Pd, there's no way of knowing that this is not allowed...
Won't it be better to have this fixed?

millerpuckette:
Ah, right, I hadn't caught on that even making an 'anything' abstraction would add a
new method.  I wonder what other side effects we can expect from that... anyhow, I'm convinced
and have adopted this.

PR #382, opened 2018-06-21
don't increment gl_valid when deleting non-scalar objects

Spacechild1:
currently, gl_valid is incremented whenever you delete any gobject. this means that pointers to scalars become stale even when you add/delete non-scalar objects to the same canvas (adding an object means deleting the temporary object). 

I run into this problem in a large personal Pd project which combines data structures and abstractions on the same canvas.

this PR enables adding/deleting non-scalar objects to the same canvas without affecting pointers.
it also fixes a memory leak where text_define_free wouldn't delete the scalar.

note that [text define] has to be handled specially because it owns the scalar and uses the parent canvas only as a stub, so deleting a [text define] wouldn't automatically free the scalar and increment gl_valid - this has to be done manually in text_define_free.

here's a test patch:

[gl_valid_test.zip](https://github.com/pure-data/pure-data/files/2123458/gl_valid_test.zip)

### DIFF
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -129,6 +129,8 @@ void glist_delete(t_glist *x, t_gobj *y)
         g->g_next = y->g_next;
         break;
     }
+    if (y->g_pd == scalar_class)
+        x->gl_valid = ++glist_valid;
     pd_free(&y->g_pd);
     if (rtext)
         rtext_free(rtext);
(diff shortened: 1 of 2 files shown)


millerpuckette:
merged

millerpuckette:
OK, I believe I've got this patched and running, and no problems yet.

PR #383, opened 2018-06-21
add "delete" method to [pointer]

Spacechild1:
the [delete( message deletes the current scalar and outputs the next one. in case it was the last one in the list, a "bang" is sent to the rightmost outlet instead, just like with [next(.
we ensure that the gpointer stays valid so you can, for example, iterate over a glist and conditionally delete scalars.

here's a test patch:
[pointer_delete_test.zip](https://github.com/pure-data/pure-data/files/2125500/pointer_delete_test.zip)


Spacechild1:
ok, now it should be safe

Spacechild1:
I've implemented @millerpuckette 's suggestions of caching the previous scalar. glist_delete now takes a third parameter which acts as a hint*) where to start searching for the previous gobj. I can image this could be handy in other situations.

here's an updated test-patch: 
[pointer_delete_test.zip](https://github.com/pure-data/pure-data/files/2348812/pointer_delete_test.zip)

*) it is a hint because gpointer only caches the last valid *scalar*. there might be other gobjs in between. we can't simply cache the last gobj because it might go stale without notice (since 0.48.2 deleting non-scalar objects won't increment gl_valid anymore).

Spacechild1:
:-D that's why I like doubly linked list. you can just merge up to and including https://github.com/pure-data/pure-data/pull/383/commits/dfb04160cdf23325315f1d8bb52b1055d49442cf

PR #396, opened 2018-06-29
Fix font values from sys_fakefromgui()

pierreguillot:
There is small error: all the font values are shifted so the width is used as the size, the height as the width and the size as the width. So if you use this, most of the objects will have a wrong size. 
The commit also "reorganizes" the `defaultfontshit[]` array with spaces, this way the values are aligned and, I think, more readable but I can change back if you prefer.

### DIFF
--- a/src/s_main.c
+++ b/src/s_main.c
@@ -286,8 +286,8 @@ void glob_initfromgui(void *dummy, t_symbol *s, int argc, t_atom *argv)
 
 // font char metric triples: pointsize width(pixels) height(pixels)
 static int defaultfontshit[] = {
- 8,  5, 11, 10,  6, 13, 12,  7, 16, 16, 10, 19, 24, 14, 29, 36, 22, 44,
-16, 10, 22, 20, 12, 26, 24, 14, 32, 32, 20, 38, 48, 28, 58, 72, 44, 88
+ 8,  5,  11,  10, 6,  13,  12, 7,  16,  16, 10, 19,  24, 14, 29,  36, 22, 44,
+ 16, 10, 22,  20, 12, 26,  24, 14, 32,  32, 20, 38,  48, 28, 58,  72, 44, 88
 }; // normal & zoomed (2x)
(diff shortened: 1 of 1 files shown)


umlaeute:
how can i trigger a problem with the (current, buggy) behaviour?

pierreguillot:
You can set the [line 256](https://github.com/pierreguillot/pure-data/blob/4c720ed007bd396a55d6ad3ed9874559d1c3dc61/src/s_main.c#L256) to 1 to print the font values. Then using the option `-nogui` should call the function `sys_fakefromgui()` and you'll see that the values are not coherent with the ones defined [line 288](https://github.com/pierreguillot/pure-data/blob/4c720ed007bd396a55d6ad3ed9874559d1c3dc61/src/s_main.c#L288). The result should be:
```
font (8  5  11)
font (10 6 13)
...
```

umlaeute:
so is there an actual bug your are fixing or just an inconsistency (that might never trigger a *real* problem)?

(both deserve fixing)

pierreguillot:
I would say an inconsistency... as even if `defaultfontshit[]` is not well used I don't think there is any consequence for a "standard" Pd user that use the flag `-nogui`. But as I recreate the GUI using these values (with libpd and JUCE) for me there is a problem. But I will check tomorrow with different font sizes if there is no crash using `-nogui` (I only tried with 10 or 12 not with 8, 16, 24 or 36).

pierreguillot:
So I tried and  it seems at first sight that opening patches with various font sizes in the nogui mode doesn't crash. Nevertheless using the nogui mode generates the  message `verbose(4): ignoring invalid font-metrics from GUI` (from the [line 249](https://github.com/pierreguillot/pure-data/blob/4c720ed007bd396a55d6ad3ed9874559d1c3dc61/src/s_main.c#L249) (so you don't even have to set the [line 256](https://github.com/pierreguillot/pure-data/blob/4c720ed007bd396a55d6ad3ed9874559d1c3dc61/src/s_main.c#L256) to 1 to see that there is an inconsistency).

umlaeute:
included in #394 (feel free to close this PR)

PR #400, opened 2018-07-10
Update noise~-help.pd

porres:
just correctly saying this is being updated in 0.48-2, not in 0.48


millerpuckette:
I think it's better just to give the major version for help updates... because that's the
grain of compatibility (i.e., minor version numbers are just updates, not, hopefully,
functional changes)

PR #404, opened 2018-07-31
Fix build on OpenBSD

thodg:
These two commits fix building PureData on OpenBSD 6.3

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -55,7 +55,7 @@ DIST_SUBDIRS = asio doc extra font mac man msw po portmidi portaudio src tcl
 # optionaly build localizations
 if MSGFMT
 locales:
-	make -C po all
+	${MAKE} -C po all
 else
 locales:
 endif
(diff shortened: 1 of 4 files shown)


millerpuckette:
merged

Makefile.am
```diff
@@ -55,20 +55,20 @@ DIST_SUBDIRS = asio doc extra font mac man msw po portmidi portaudio src tcl
...
 # forward target to mac/Makefile
 app:
-	make -C mac app
+	${MAKE} -C mac app
 endif
 if WINDOWS
 # forward target to msw/Makefile
 app:
-	make -C msw app
+	${MAKE} -C msw app
```
umlaeute: shouldn't that read `$(MAKE)` rather than `${MAKE0` (parentheses, rather than curly braces)?

umlaeute:
merged as 055805d6114795602c1af04b257efe6a3013ca87...c81676628d14f260b6795dafd4aa8a1af088bbd9

PR #409, opened 2018-08-07
in clone-help.pd call an abstraction abstraction, not subpatch

mxa:
renamed the abstraction and to clone-abstraction.pd and call it an abstraction in the clone helpfile.

### DIFF
--- a/doc/5.reference/clone-abstraction.pd
+++ b/doc/5.reference/clone-abstraction.pd
@@ -1,4 +1,4 @@
-#N canvas 647 320 470 387 10;
+#N canvas 737 330 470 387 10;
 #X obj 112 74 inlet;
 #X obj 112 171 osc~;
 #X obj 112 224 outlet~;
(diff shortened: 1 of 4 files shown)


mxa:
Wanted to make a clean pull request for e5ca483 but didn't know how to do it.

umlaeute:
@mxa 
- all files **must** be listed in the relevant `Makefile.am` 
   - if you rename files, you must also update the `Makefile.am`
- you shouldn't create PRs from your `master` branch. instead create a separate feature-branch
- TravisCI currently fails because of some pull error (probably because i pushed to your `master`; i don't really know); that's supposed to be unrelated to the actual code which i think is now fine...at least a local `make distcheck` succeeds.

umlaeute:
ah no; the TravisCI failures are because i was confused becase of your `master`, therefore pushed to the wrong branch, deleted that branch and now TravisCI cannot find it any more...

millerpuckette:
Should I just apply these changes by hand?  Especially as it involved renaming a file, which merging ought to do correctly but I've been bitten by that one.

mxa:
@millerpuckette probably wise to do this by hand.

danomatika:
I went ahead and applied this manually to #438 via commit 79d58eaf3564e695305a75c68d443ac844bfa9c9.

mxa:
Thank you Dan and Miller.

Issue #407, opened 2018-08-05
TCL Invalid Command Name


shreeswifty:
Linux 64 Ubuntu 
Linux shree-linux2 4.4.0-116-lowlatency #140-Ubuntu SMP PREEMPT Mon Feb 12 22:41:05 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

when opening midi alsa window  

is it supposed to be "raise" -- not sure but,,,

(Tcl) INVALID COMMAND NAME: invalid command name "rase"
    while executing
"rase "$id""
    (procedure "pdtk_alsa_midi_dialog" line 67)
    invoked from within
"pdtk_alsa_midi_dialog .gfxstub212e2c0 1 1 1 0 0 0 0 0 1 1"
    ("uplevel" body line 5)
    invoked from within
"uplevel #0 $docmds"opened alsa client 128 in:1 out:1
opened alsa client 128 in:1 out:1

porres:
this is with pd 0.48-2test1, right?

PR #413, opened 2018-08-16
running bugfixes branch for 0.48-x

umlaeute:
this is a curated branch that only contains trivial bug fixes, translation updates and documentation updates.

rules for inclusion in the PR:

- only bugfixes
- the patch is trivial
- the patch is "unanimous" (no protests, "buts" or second thoughts have been (ever) voiced)

## included fixes

- [x] [expr~ crash due to un initialiized value](https://sourceforge.net/p/pure-data/bugs/1315/)
- [x] typo (`rase` instead of `raise` for raising a dialog window) resulting in backtraces in the Pd-console #407 
- [x] regression that would not properly `restore` subpatches when saving a patch (thus leading to all kind of weirdnesses, starting from empty objectboxes...)
- [x] a couple of potential buffer-overflows (detected by codacy)
- [x] handling of failed `realloc()` (detected by codacy)
- [x] adding hashbangs to the build-scripts so the terminal knows which interpreter to choose (rather than relying on miller's login-shell)
- [x] removed double entry for `Preferences` from the `Edit`-menu (it's already in the `File`-menu)
- [x] fixed some warning when doing assignments within conditionals
- [x] re-generated translations
- [x] improved `mac/tcltk-wish.sh` to not enforce (re)downloading of TclTk sources

### DIFF
--- a/extra/stdout/stdout.c
+++ b/extra/stdout/stdout.c
@@ -76,7 +76,7 @@ static void stdout_binary(t_stdout *x, int argc, t_atom *argv)
         argc = BUFSIZE;
     for (i=0; i<argc; i++)
         ((unsigned char *)buf)[i] = atom_getfloatarg(i, argc, argv);
-    buf[i>BUFSIZE?BUFSIZE:i] = 0;
+    buf[i>=BUFSIZE?(BUFSIZE-1):i] = 0;
     fwrite(buf, 1, argc, stdout);
 
     if (x->x_flush || !argc)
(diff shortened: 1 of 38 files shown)


danomatika:
Looks good so far. That "raise" typo was probably mine.

danomatika:
The last commit is not really a bug fix, but more something I added to me local copy an forgot to commit a while ago.

umlaeute:
rebased into master 3de09d4c20523bcb40736d2efce1280c44fcad15...80b3891537bb3e30e18228afb54858c62c35e9e5

PR #420, opened 2018-09-03
offset patch cord y pos to center within io

danomatika:
This fixes a display "bug" where signal patch cord corners can be seen at certain patch cord angles between objects. The fix basically introduces a y offset based on the inlet/outlet height and zoom setting.

The only downside I can see is that the current PR introduces some extra math when updating the canvas io. This could probably be avoided by precomputing the offset and storing it somewhere.

Quick examples at zoom 2. Notice the corner of the [osc~] outlet cord showing beyond the object border.

screen shot 2018-09-03 at 2 04 50 pm

screen shot 2018-09-03 at 2 06 18 pm

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -801,13 +801,14 @@ static void canvas_drawlines(t_canvas *x)
 {
     t_linetraverser t;
     t_outconnect *oc;
+    int iyoffset = (IHEIGHT/2)*x->gl_zoom, oyoffset = (OHEIGHT/2)*x->gl_zoom;
     {
         linetraverser_start(&t, x);
         while ((oc = linetraverser_next(&t)))
             sys_vgui(
         ".x%lx.c create line %d %d %d %d -width %d -tags [list l%lx cord]\n",
                 glist_getcanvas(x),
-                t.tr_lx1, t.tr_ly1, t.tr_lx2, t.tr_ly2,
+                t.tr_lx1, t.tr_ly1-iyoffset, t.tr_lx2, t.tr_ly2+oyoffset,
                 (outlet_getsymbol(t.tr_outlet) == &s_signal ? 2:1) * x->gl_zoom,
                 oc);
     }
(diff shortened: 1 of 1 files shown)


src/g_canvas.c
```diff
@@ -801,13 +801,14 @@ static void canvas_drawlines(t_canvas *x)
 {
     t_linetraverser t;
     t_outconnect *oc;
+    int iyoffset = (IHEIGHT/2)*x->gl_zoom, oyoffset = (OHEIGHT/2)*x->gl_zoom;
```
umlaeute: this seems to do integer arithmetic, potentially getting an unintended 1-pixel offset.
what about the following (untested)?

~~~C
int iyoffset = IHEIGHT*0.5*x->gl_zoom, oyoffset = OHEIGHT*0.5*x->gl_zoom;
~~~

umlaeute:
i'm not entirely sure i understand the problem you are trying to solve.
is it, that the cord seems to originate from the bottom of the outlet (rather than it's center), but having rectangular caps - and the corners of those rectangles might not lie inside the object-box (or outlet-box).

personally, i don't care very much about such details, but for others it might be important.

so to review the patch: looks straight enough (apart from the integer-arithmentic)

danomatika:
>but having rectangular caps - and the corners of those rectangles might not lie inside the object-box

That's the issue exactly. Its not a functional problem but it looks "ugly" when the signal cords show their corners at more extreme angles. I tried using a projecting cap style, but that didn't help.

The integer math was on purpose as I didn't want an accidental round upwards. It just needs to be a *little* offset but also handle the situation where people would use a different IHEIGHT/OHEIGHT.

danomatika:
Here's another screenshot with 0.48-2 and more exaggerated angles:

screen shot 2018-09-03 at 4 02 47 pm

danomatika:
After some more testing, a "quick and dirty" solution to this would be to simply set the offset to gl_zoom. That gives enough to hide the corners without the extra math at the loss of vertically centering the cords based on the IHEIGHT/OHEIGHT;

danomatika:
Here's a similar screenshot using just gl_zoom:

screen shot 2018-09-03 at 4 09 15 pm

danomatika:
I've gone ahead and changed this PR to use gl_zoom. It works just as well and is simpler.

danomatika:

umlaeute:
wouldn't using non-square line-endings (e.g. rounded line endings) fix the problem as well?

danomatika:
Line projections would work, but I didn’t find that as a drawing option i Tk. We can look again.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

PR #423, opened 2018-09-04
Fix audio issues related to portaudio

Spacechild1:
* prevent Pd from hanging when you choose an audio device blocksize larger than 64 and accidentally set a delay which is shorter than your block size
* check for "-h" and "-help" early on, before opening audio devices because this might take some seconds when you have Jack on Windows (and the server isn't started). this can be a bit unnerving when you only want to see the help
* prevent Pd from hanging when a device gets disconnected. we set a timeout for polling/waiting after which Pd prints an error and tries to reopen the audio device. if that fails, Pd closes the device and the user can later reopen it from the menu. this works for FAKEBLOCKING with and without THREADSIGNAL. I also added the error handling to the blocking mode (which didn't really hang Pd)

### DIFF
--- a/src/s_audio.c
+++ b/src/s_audio.c
@@ -435,11 +435,13 @@ void sys_reopen_audio( void)
     if (sys_audioapi == API_PORTAUDIO)
     {
         int blksize = (audio_blocksize ? audio_blocksize : 64);
+        int nbufs = sys_advance_samples / blksize;
+        if (nbufs < 1) nbufs = 1;
         if (sys_verbose)
-            fprintf(stderr, "blksize %d, advance %d\n", blksize, sys_advance_samples/blksize);
+            fprintf(stderr, "blksize %d, advance %d\n", blksize, nbufs);
         outcome = pa_open_audio((naudioindev > 0 ? chindev[0] : 0),
         (naudiooutdev > 0 ? choutdev[0] : 0), rate, STUFF->st_soundin,
-            STUFF->st_soundout, blksize, sys_advance_samples/blksize,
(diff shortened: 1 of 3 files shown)


umlaeute:
speaking of audio bugs: could you also provide a patch to run `pd -h` or `pd -help` (or both) to get the usage without initializing the audio engine? (i remember this to be a problem with your setup (related to jack-on-windows) but don't have the complete picture of the problem)

Spacechild1:
ah yes, thanks for reminding me. there's actually a simple workaround: "pd -noprefs -h", but I want to fix it nevertheless.

there are more things I wanna do, like prevening Pd from crashing when an audio device gets unplugged while DSP is running and certain latency problems which I think happen with audio devices which operate on non-power-of-2 block sizes, but those will need some time...

Spacechild1:
OK, I stopped Pd from hanging when the audio device gets disconnected. I leave the polling loop after a certain amount of sleeps, print an error message and Pd try to reopen the audio device. if that fails, the user can check what's wrong and simply reopen the device from the menu. the most important part ist that Pd will *not* freeze. this can be disasterous when you have, say, a short power outage on stage (like I've experienced more than once) and you have to force quit Pd. with this patch, Pd will try to recover on its own. I've tested this successfully by quickly plugging/unplugging the USB cable :-) 

of course I could also check for Pa_IsStreamActive within the polling loop, so it will quit immediately, but I'm not sure if this is really an advantage because chances of automatic recovery will be significantly smaller...

this patch only concerns the standard configuration: FAKEBLOCKING and *no* THREADSIGNAL. I haven't tried the other possible configurations and I don't know if I should care to?

PR #424, opened 2018-09-05
removed debian/ folder to conform with Debian's packaging policy

umlaeute:
this is with my Debian hat on (being responsible for the official Debian packages of puredata):

Debian packaging should stay an independent process from upstream development, and not include their own `debian/` folder.

Now, we already exclude the `debian/` folder from the tarballs that are automatically generated via `git archive` (such as the release tarballs from github) as well as from the `make dist` tarballs.

Nevertheless, this is bitrotting code that was last used and updated about 8 years ago (in 2010).

It's time to make it go away.


umlaeute:
just for the record: the bitrot is not theoretical, its factual, as it fails to build a Debian package.

PR #195, opened 2017-08-18
icon update

danomatika:
(This is a re-uploaded PR of #194 to keep history clean from intermediate icon files.)

Updated icons for all platforms:

* macOS: pd.icns & pd-file.icns (much higher resolution)
* Windows: pd.ico (now with multiple sizes)
* Linux: pd.gif (512x512) & puredata.xpm (hand-tuned to match others)

This will give Pd a consistent icon between systems.

On Linux, pd-gui.tcl now checks if the gif is available in the sys_guidir and loads it as the application icon. I added pd.gif to the tcl/Makefile.am but I'm not sure what is the best way to hand things on Linux since the desktop environment might load icons in a different manner, ie. Debian uses the xpm. Maybe there should be a configure flag to disable installing the gif? This is a question for @umlaeute.

Also, tcl/Makefile.am now only installs the relevant icon file based on the platform (.ico for Windows and .gif for Linux).

Removed `src/pd.ico` as it's already loaded from the tcl location in `pd-gui.tcl`. The `src/makefile.mingw` was updated to reflect this.

The design is based on the Pd-extended icon proportions while maintaining the minimal look of the vanilla icon. It was generated from an actual patch using the sag export plugin, tweaked in Inkscape, and saved as an SVG. The font is DejaVu Sans Mono bold for better readability and the line weights are thicker than normal. The icon repo is [pd-icon](https://github.com/pure-data/pd-icon).

Pd icon

~~~
                                
                                
   .........                    
   .##...##.                    
   .#.###.#.                    
   ..#####..                    
   ..#####..                    
   ..#####..                    
   ..#####..                    
   .#.###.#.                    
   .##...##.                    
   .........                    
    .                           
    .                           
   ..........................   
   ....$$$$$$$$$$$$$$$$$$....   
   .$$$$$$$$$$$$$$$$$$$$$$$$.   
   .$$$$$$$$$$$$$$$$$$$$$$$$.   
   .$.....$$$$$$$..$$$$$$$$$.   
   .$..$$..$$$$$$..$$$$$$$$$.   
   .$..$$..$$$$$$..$$$$$$$$$.   
   .$..$$..$$$.....$$$$$$$$$.   
   .$.....$$$..$$..$$$$$$$$$.   
   .$..$$$$$$..$$..$$$$$$$$$.   
   .$..$$$$$$..$$..$$$$$$$$$.   
   .$..$$$$$$$.....$$$$$$$$$.   
[...]

PR #428, opened 2018-09-10
re-adding icons and changes from update/icon2 branch

danomatika:
This is a PR to address changes missing from the manual merge of #195.


danomatika:
@umlaeute Since we removed the `debian` dir, I put the pd.xpm into the `tcl` dir. Also, is the conditional loading of the pd.gif in `tcl/pd-gui.tcl` needed on Linux? I figured it couldn't hurt, although is there a change this would work against the window manager?

PR #430, opened 2018-09-10
textwindow edit bindings

danomatika:
This PR adds select all, undo, & redo key bindings to text windows (ie. those used with qlist, text define, etc).

One issue is that the window bindings currently conflict with those of the parent patch: mod+z seems to fire the redo event on both the text window & the patch window. This may be a question for @umlaeute.


tcl/pdtk_textwindow.tcl
```diff
@@ -23,13 +23,17 @@ proc pdtk_textwindow_open {name geometry title font} {
...
         text $name.text -relief raised -highlightthickness 0 -bd 2 \
-            -font [get_font_for_size $font] \
+            -font [get_font_for_size $font] -exportselection 1 -undo 1 \
             -yscrollcommand "$name.scroll set" -background white
         scrollbar $name.scroll -command "$name.text yview"
         pack $name.scroll -side right -fill y
         pack $name.text -side left -fill both -expand 1
         bind $name.text <$::modifier-Key-s> "pdtk_textwindow_send $name"
         bind $name.text <$::modifier-Key-w> "pdtk_textwindow_close $name 1"
+        bind $name.text <$::modifier-Key-a> "pdtk_textwindow_selectall $name; break"
```
umlaeute: it seems that there are default bindings here (Linux tcltk8.6) of <key>Ctrl</key>-<key>A</key> and <key>Ctrl</key>-<key>E</key> which move the cursor to the beginning resp. end of the current line. while i quite like the default behaviour, i have to admit that i only discovered it while testing this PR :-)

danomatika: We use MOD+A already for select all, so I was following that precedent. It's also what I'm used to and I think it would be helpful to be able to do it *somehow*.

danomatika: Annoyingly, those also work on macOS, except they are not mapped to the Command key.

millerpuckette:
I tried this on linux; ctrl-a did nothing and ctrl-z undid on the parent there (causing my text object to vanish :)

danomatika:
Yeah, this was really just and idea and I didn’t go much farther with it. It’s not terribly important.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Similarly, I need to go through my other PRs to see which are still relevant/finished.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
i was unable to trigger any problems with Pd-0.51-4.

Anyhow, i've included this PR in #1353 and am therefore closing this in favour of the cleaned up one.

PR #431, opened 2018-09-10
add flag to [drawpolygon] to only disable changing vertices

Spacechild1:
while "-xr" disables *all* mouse actions in run mode, "-xv" only disables changing vertices while still reporting mouse clicks to the struct.

example use case: build a GUI with data structures but prevent the user from accidentally change the shape of the widgets.


millerpuckette:
Not sure about this yet... will look at it again after compiling test version but want to figure it out.

Spacechild1:
ok! it's not really urgent and can be done later anytime.

PR #435, opened 2018-09-10
add object to test gpointers for equality

Spacechild1:
output 1 if pointer to left inlet and stored pointer (right inlet) point to the same thing.
otherwise output 0.

alternative implementation: https://github.com/pure-data/pure-data/pull/436

EDIT: we could also create this object via a special constructor argument to [pointer], e.g. [pointer ==]. the thing is that this would break patches which have a struct called "==". maybe it is so unlikely that we can ignore it. after all, Miller reserved "text" for [text], accepting that it *might* cause a regression in someone's patch.

### DIFF
--- a/src/g_traversal.c
+++ b/src/g_traversal.c
@@ -520,6 +520,49 @@ static void ptrobj_setup(void)
     class_addbang(ptrobj_class, ptrobj_bang);
 }
 
+/*----------------------- compare ------------------------- */
+
+static t_class *compare_class;
+
+typedef struct _compare
+{
+    t_object x_obj;
+    t_gpointer x_gp;
+} t_compare;
+
+static void *compare_new(t_symbol *classname, int argc, t_atom *argv)
+{
+    t_compare *x = (t_compare *)pd_new(compare_class);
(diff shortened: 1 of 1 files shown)


danomatika:
I think #436 is a better version.

Spacechild1:
NOTE: we could also create this object via a special constructor of [pointer], e.g. [pointer ==]. the thing is that this would break patches which have a struct called "==". maybe it is so unlikely that we can ignore it. after all, Miller reserved "text" for [text], accepting that it *might* cause a regression in someone's patch.

umlaeute:
i'd also prefer if the pointer-related objects lived in their own namespace. `[equal]`, `[compare]` (and `[gem]` and `[set]`) are just way too generic names.

Spacechild1:
me too! the problem is that [pointer] already takes symbols as creation arguments. do you think [pointer ==] would be acceptable?
what are your thoughts about #436 (message version)?

Spacechild1:
closing because alternative PR https://github.com/pure-data/pure-data/pull/436 got merged

PR #436, opened 2018-09-11
add message to [pointer] to test gpointers for equality

Spacechild1:
if the gpointer of the message points to the same thing (scalar or word) as the stored pointer,
output the pointer.
if they are not equal, send a bang through the last outlet.

EDIT: the result could also be sent as float (0 or 1) to the last outlet. I'm starting to like this better because it wouldn't interfere with the rest of the interface (the user can use [route bang float] on the last outlet)

alternative implementation: https://github.com/pure-data/pure-data/pull/435

### DIFF
--- a/src/g_traversal.c
+++ b/src/g_traversal.c
@@ -396,6 +396,34 @@ static void ptrobj_delete(t_ptrobj *x)
     }
 }
 
+static void ptrobj_equal(t_ptrobj *x, t_gpointer *gp)
+{
+    t_symbol *templatesym;
+    int n, which;
+    t_typedout *to;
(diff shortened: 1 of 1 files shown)


Spacechild1:
seems we have the same preference after all since [compare] is of the same spirit as [pointer equals]. note, however, that reserving symbols as pointer constructor arguments can be problematic (see my comment in #435)
this PR adds a *method* to [pointer], which I don't like so much because it complicates the interface (the right most outlet now can have two meanings: "end of glist" and "not equal"). 
there are several other ways to deal with the message, e.g. sending 1 or 0 to the last outlet and the user can to [route bang float] to distinguish.

PR #437, opened 2018-09-11
also trigger savestate located in subpatches

Ant1r:


danomatika:
This is a good point! [savestate] would not be as useful if it didn't work within a [pd] nesting.

Ant1r:
Thanks!
I think it would be even better if a [savestate] in an abstraction could refer to the containing patch (i.e `[savestate 1]` ), allowing to call stuff like `[autosavedfloat myFloat]`, but it might be a little harder to do...

millerpuckette:
merged

PR #440, opened 2018-09-11
proposal for declare: default to '-path'+'-lib' when no flag is specified

Ant1r:
this allows to write:
`[declare myLib]`
whether myLib is a single binary file, multiple binaries, or abstractions set.

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -1658,27 +1658,38 @@ void canvas_declare(t_canvas *x, t_symbol *s, int argc, t_atom *argv)
     for (i = 0; i < argc; i++)
     {
         const char *flag = atom_getsymbolarg(i, argc, argv)->s_name;
-        if ((argc > i+1) && !strcmp(flag, "-path"))
+        if (!strcmp(flag, "-path"))
         {
-            canvas_path(x, e, atom_getsymbolarg(i+1, argc, argv)->s_name);
-            i++;
+            if(argc > i+1) {
+                canvas_path(x, e, atom_getsymbolarg(i+1, argc, argv)->s_name);
+                i++;
+            }
         }
-        else if ((argc > i+1) && !strcmp(flag, "-stdpath"))
+        else if (!strcmp(flag, "-stdpath"))
         {
-            canvas_stdpath(e, atom_getsymbolarg(i+1, argc, argv)->s_name);
-            i++;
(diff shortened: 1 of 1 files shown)


umlaeute:
hmm, but what if you want to declare a library that is called `-path`?

probably adding another flag that adds both *path* and *lib* is a better idea.

Spacechild1:
tbh, a library called "-path" seems just bizarre to me. if I see such a library I would probably file a bug report to the author :-)

Ant1r:
@millerpuckette don't you think it could be good to have this in 0.49? So that we can update the [declare] man-page once and for all, instead of being in an intermediate situation again.

Ant1r:
moreover, this should go in conjunction with #463.

millerpuckette:
I need to rethink the whole "declare" mess from the beginning.  I simply don't understand how search paths work anymore and am afraid the whole thing will become too confusing for humans to use.

porres:
Up to now I was using the "add-to-path" message to Pd in the library binary to add the library's path to the preferences path... I thought this would be basically the same as doing [declare -lib libname -path libname], but it turns out it doesn't as I discussed in https://github.com/pure-data/pure-data/issues/1179

so, just here to say that I like the idea that I also hoped there could be a simpler way to use declare, and that the fact that sometimes we need to use both -lib and -path or know if one or the other is confusing. So I highly praise this new simpler and more effective design :)

marcodsad:
Hey, Marco Donnarumma here to upvote the simplification of the system as it is being proposed here.
I also would like to point out that this doesn't help only new users, but also old and faithful ones! :P
From a user's standpoint, simplifying this system is an urgent necessity, imho. Users should not have to spent time understanding this kind of system.

A -path library is just a bad idea :) I wouldn't think this is a scenario realistic enough to impede the simplified implementation described here. my 2 cent!

porres:
hi, why is this still a draft?

Ant1r:
> why is this still a draft?

because:
- Miller seemed to think that the [declare] should be revised because incomprehensible; actually my own feeling differs a bit, but that means that this PR won't be easily accepted.
- actually I'm no longer sure that this PR has no drawbacks. The overhead could be noticeable in some cases, like when an abstraction with a [declare] is cloned many times; so I think performance should be measured somehow before removing the Draft flag.

Issue #352, opened 2018-04-22
functions keys not working on pd-0.48


porres:
This is with pd 0.48 and 0.48-1 4 bits on macOS 10.12.6, with a macbook Pro using the new touchbar thingy.

But it was working with Pd < 0.48. Here's a screenshot of 0.48-1 on the left and 0.47-1 (also 64 bits) on the right when trying to get "F4", but other functions keys (F1, F2, etc) have a similar issue. This also affects an external like [cyclone/comment] which uses "F4".

<img width="1100" alt="screen shot 2018-04-22 at 18 43 55" src="https://user-images.githubusercontent.com/5360142/39100266-b124c09a-465d-11e8-9200-019bcf2130a2.png">

I also find it funny how [key] and [keyup] 0.48 spit out a big number when hitting a function and 0.47 doesn't

umlaeute:
so is the big-number unique? (e.g. it differs between <kbd>F4</kbd> and  <kbd>F5</kbd>)

also, which tcl/tk version are the two binaries using?
if you cannot tell, you might want to install the `tclprompt-plugin` and run the following command in the new `Tcl` entry just below the Pd-console (after re-starting Pd):

~~~tcl
::pdwindow::post "Tcl:${tcl_version} Tk:${tk_version}\n"
~~~

porres:
> so is the big-number unique? (e.g. it differs between F4 and F5)

It differs, F4 = 63239 / F5 = 63240

Log 4 in 0.48-1 gives me Tk 8.5.19, but I can't get it like that in 0.47, but I know from the Pd list it was using an earlier 8.4 version

porres:
yeah, this seems like a mac only issue

danomatika:
I see this too. Looking into it today.

danomatika:
>if you cannot tell, you might want to install the tclprompt-plugin and run the following command in >the new Tcl entry just below the Pd-console (after re-starting Pd):

@umlaeute You don't need to do this with 0.48+ as I added the TK version as a verbose print.

danomatika:
Actually, the modifier keys do not give good values either (command, control, option, shift).

[...]

PR #442, opened 2018-09-11
catch macOS F1-F12 key values from TK 8.5+

danomatika:
This is a simple fix for #352 by adding additional checks for these key values as they are not given string names by Tk for some reason. This might be some that could be fixed more in Tk, but this is an easy workaround for Pd.


millerpuckette:
merged

Issue #443, opened 2018-09-11
pd 0.49test3 printing annoying errors on console (macOS)


porres:
this never happen, I have no idea

![screen shot 2018-09-11 at 13 33 50](https://user-images.githubusercontent.com/5360142/45374106-8936f600-b5c7-11e8-8e2f-ad4a65feabb1.png)

Spacechild1:
I'm the culprit. does it print it all the time? what are your audio settings?

porres:
Nothing, that's why it's weird...

Spacechild1:
Ok, and with audio?

danomatika:
I'm not seeing this with the current master. I start Pd, choose a USB sound interface for input & output, start the test tone patch, unplug, and Pd doesn't crash. No prints.

The only time I see any prints/hangs are when I try to re-select the device in the audio dialog but, of course, the portaudio device list is now stale. I'd suggest printing some sort of one-time error line that would at least suggest restarting Pd when this happens using Portaudio since it cannot handle dynamic hot plugging.

porres:
I'm just using the default settings, with mac's bultin audio card, I never went into the audio settings, I don't know what's causing this. I'm just using Pd and suddenly this starts popping

Spacechild1:
> since it cannot handle dynamic hot plugging.

well, on Windows it works. but since it seems to cause weirdness on OSX, I better make this fix Windows only for now and later (when I have more time and access to a Mac) apply it to OSX.

umlaeute:
is this issue resolved with accepting #445

Spacechild1:
waiting for Alex to confirm

porres:
confirmed

porres:
still happening with test3, so I changed the issue name, but it does occur much less

porres:
I'm actually getting something else, time for a new report

PR #445, opened 2018-09-11
revert the portaudio polling limit for non-Windows systems

Spacechild1:
because it seems to cause weirdness as reported by https://github.com/pure-data/pure-data/issues/443#issuecomment-420425508
I'll work on it when I have access to a Mac. In the mean, keep it Windows only.

### DIFF
--- a/src/s_audio_pa.c
+++ b/src/s_audio_pa.c
@@ -92,12 +92,9 @@ static pthread_cond_t pa_sem;
 #endif
 /* maximum number of ms sleeps before we stop polling and try to reopen the device */
 #ifndef MAX_NUM_POLLS
-#ifdef _WIN32
-#define MAX_NUM_POLLS 1000
-#else
-#define MAX_NUM_POLLS 0x7fffffff /* polling limit doesn't work on non-Windows systems yet */
-#endif /* WIN32 */
-#endif /* MAX_NUM_POLLS */
+#define MAX_NUM_POLLS 1000 /* maximum number of unsuccessful polls before giving up */
+PaError PaUtil_ValidateStreamPointer(PaStream * stream);
+#endif
 #endif /* THREADSIGNAL */
 #endif  /* FAKEBLOCKING */
 
@@ -544,7 +541,9 @@ int pa_send_dacs(void)
                 break;
             }
 #else
-            if (!--counter)
+            if (PaUtil_ValidateStreamPointer(&pa_stream) < 0)
+                counter--;
+            if (!counter)
             {
                 locked = 1;
                 break;
@@ -587,7 +586,9 @@ int pa_send_dacs(void)
                 break;
             }
 #else
-            if (!--counter)
+            if (PaUtil_ValidateStreamPointer(&pa_stream) < 0)
+                counter--;
+            if (!counter)
             {
                 locked = 1;
                 break;
--- a/src/s_audio_pa.c
+++ b/src/s_audio_pa.c
@@ -92,8 +92,12 @@ static pthread_cond_t pa_sem;
 #endif
 /* maximum number of ms sleeps before we stop polling and try to reopen the device */
 #ifndef MAX_NUM_POLLS
+#ifdef _WIN32
 #define MAX_NUM_POLLS 1000
-#endif
+#else
+#define MAX_NUM_POLLS 0x7fffffff /* polling limit doesn't work on non-Windows systems yet */
+#endif /* WIN32 */
+#endif /* MAX_NUM_POLLS */
 #endif /* THREADSIGNAL */
 #endif  /* FAKEBLOCKING */
 


Spacechild1:
@porres can you do me a favour and quickly build this branch and see if the errors are gone?

Spacechild1:
this means that usleep(1000) isn't guaranteed to sleep at least for 1ms (unlike Sleep()). of course this throws everything off... I'll go ahead and check the device state after the sleep. I tried to avoid that but it's obviously safer.

millerpuckette:
Actually on thinking about it I have trouble believing usleep() isn't sleeping... I don't know what's
going on at all in fact.

millerpuckette:
It's better... Pd doesn't hang at least.  On Mac, one still has to restart Pd to get the audio device back up after reconnecting.  This seems to be a limitation of portaudio on Mac - the device list also doesn't change when attaching/detaching devices until I restart Pd.  Later I want to see if re-initializing portaudio from scratch every time audio is started helps or makes it even worse.

millerpuckette:
Anyhow, merged it and will close this.

Spacechild1:
thanks! in october I'm doing some crossplatform work where I have access to a Mac and I want to investigate this further.

> Later I want to see if re-initializing portaudio from scratch every time audio is started helps

this might be bad in some circumstances, initialization sometimes takes some seconds (e.g. with Jack on Windows), and this would be annoying to happen only when toggling DSP. I could, however, imagine a "rescan devices" button in the audio settings which reinitializes portaudio. (something similar could be very handy with portmidi, so we can plug in MIDI devices after Pd has started).
some apps, especially REAPER, handle audio devices excecptionally well, maybe I will ask Justin what he's doing (I guess something entirely homegrown). OTOH, I'm currently working on an audio app using RtAudio and it appears to have fewer limitations than portaudio. haven't tried it on a Mac yet, though...

danomatika:
I've contemplated writing a CoreAudio backend for macOS which would be bale to handle this as well as dynamic hot plugging. I may need this for another project anyway, so we'll see.

Spacechild1:
that would be great, so I could further mess with portaudio without fearing to break stuff on OSX :-)

danomatika:
Well, I wouldn't replace the Portaudio implementation. It would be a new option and, possibly, the default.

porres:
Is this in test2? cause it seems it ain't gone yet

millerpuckette:
(porres) - by that, do you mean that you still have to restart Pd to get the device to work if you unplug it and reconnect it?  I don't think that's fixable right now... or is there a different problem still lurking and if so how can I reproduce it?

Spacechild1:
@millerpuckette i think he is referring to https://github.com/pure-data/pure-data/issues/443
@porres can you try with the latest master?

it should only print this error if polling gets 1000x a bad stream pointer and I don't know why on earth this should happen for a connected device... just for the record, @porres was using the internal soundcard. @millerpuckette which device did you test with?

Spacechild1:
btw @porres: do you get those errors continuously while DSP is running or just a couple of them. do you get this only the first time you start DSP or everytime you (re)start DSP?

millerpuckette:
I tested with an external USB device (so I could hot-unplug it and enjoy the devastation), and also,
less intensively, with an internal audio device - with the internal device I had no errors, and with the external one, nothing bad until I unplugged it.

Spacechild1:
ok, this sounds alright. maybe you didn't include the polling fix (only increment counter when PaUtil_ValidateStreamPointer < 0) in test2? anyway, let's wait for Alex to test with the latest master. I think his original issue was caused by the bad usleep behaviour which you also experienced. now it *should* work.

danomatika:
I'll just chime in and say it would be nice if unplugging a device didn't cause Pd to stall, but at least let you save before restarting. I've been bitten by that before.

Spacechild1:
with the current master hotplugging should not stall Pd on OSX (while the reconnecting doesn't work). Dan, can you confirm this?

millerpuckette:
Test2 already had this compiled in:
            if (PaUtil_ValidateStreamPointer(&pa_stream) < 0)
                counter--;
            if (!counter)
            {
                locked = 1;
                break;
            }

danomatika:
I can confirm that unplugging now does not hang Pd. However, there is no error message printed. I imagine the first response would be to go to the audio dialog and try to select a different device, in which case *then* Pd hangs. I think this is 1 step forward and possibly one step back as it also leads to confusion after the expectation of "I can just reselect a different device." There needs to be an error messages to effect of "whoops, portaudio device lost, save your work and restart pd"

Spacechild1:
sure, I just don't understand why there's no error message printed. here's the actual loop (without all the ifdefs):

> while (sys_ringbuf_getwriteavailable(&pa_outring) <
>             (long)(STUFF->st_outchannels * DEFDACBLKSIZE * sizeof(float)))
>         {
>             rtnval = SENDDACS_SLEPT;
>             if (PaUtil_ValidateStreamPointer(&pa_stream) < 0)
>                 counter--;
>             if (!counter)
>             {
>                 locked = 1;
>                 break;
>             }
>             usleep(1000);
>         }
>

danomatika:
Shouldn't `THREADSIGNAL` be defined...?

Spacechild1:
THREADSIGNAL is *not* defined by default in the source... we rather sleep. or does it get THREADSIGNAL somehow defined in the build process? I don't think so. actually, THREADSIGNAL has been broken anyway because the mutex and condition variable haven't been intialized.

Spacechild1:
what happened in old Pd was that while loop would go on forever if there was no device (and thus no available samples to read). now my idea was to check the device state everytime before we sleep. if it returns a negative value this indicates an error. there's no other way to break from the loop than either a) reading samples b) accumulating 1000x device errors... and once we break because of an device error, `locked` is set to 1 and we therefore *must* get some error message.

Spacechild1:
so basically my question is: if hotplugging in also did hang Pd on OSX, it means it couldn't get out of the polling loop. now, I added a way to break out of the loop, but this sets `locked` to 1. so how can it get out of the loop without printing an error afterwards?

danomatika:
I added a print to stderr to the close/reopen section but it's not printed. What does appear is an assertion within portaudio which is likely borking everything after that:

~~~
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
Assertion failed: (!err), function AudioIOProc, file portaudio/src/hostapi/coreaudio/pa_mac_core.c, line 2467.
Pd: signal 6
gui socket 6 - closing audio...
~~~

danomatika:
It's basically an unfinished part of the portaudio Coreaduio implementation. Commenting the assertion allows me to see the Pd error messages as expected:
~~~
error -9988: Invalid stream pointer
trying to reopen audio device
Error opening audio: Internal PortAudio error
audio device not responding - closing audio.
~~~

Port audio prints a number of errors as it unwinds:
~~~
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
||PaMacCore (AUHAL)|| Error on line 2466: err='-50', msg=Unknown Error
KAaaaaaa~
||PaMacCore (AUHAL)|| Warning on line 525: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 525: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 445: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 445: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Error on line 1301: err='-10851', msg=Audio Unit: Invalid Property Value
||PaMacCore (AUHAL)|| OpenStream @ 44100 returned: -9986: Internal PortAudio error
||PaMacCore (AUHAL)|| Warning on line 525: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 525: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 445: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 445: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Error on line 1301: err='-10851', msg=Audio Unit: Invalid Property Value
||PaMacCore (AUHAL)|| OpenStream @ 44100 returned: -9986: Internal PortAudio error
~~~

Spacechild1:
ha, nice! PaError are actually supposed to start at -10000. I have no idea how Pa_IsStreamActive can return -50...

Spacechild1:
can you now reconnect afterwards from the menu or by toggling DSP?

danomatika:
I can select the built in device and start sound, but I cannot re-select the external device as I get the following Pd error:

    Error opening audio: Internal PortAudio error

And portaudio spits out:

~~~
||PaMacCore (AUHAL)|| Warning on line 525: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 525: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 445: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Warning on line 445: err=''!dev'', msg=Audio Hardware: Bad Device
||PaMacCore (AUHAL)|| Error on line 1301: err='-10851', msg=Audio Unit: Invalid Property Value
||PaMacCore (AUHAL)|| OpenStream @ 44100 returned: -9986: Internal PortAudio error
~~~

danomatika:
Error -50 is: `Mac OS error -50 (paramErr): error in user parameter list`, so not particularly useful.

Spacechild1:
OK, but at least Pd doesn't hang when trying to reconnect? that's good news.

porres:
yeah, errors from #443 are gone, so it seems, but with the current build, not test2

thanks!

Spacechild1:
I will quickly make a fix for the assertion problem.

Spacechild1:
commenting out the asserts fixes the local portaudio would only fix the local portaudio, though... BTW: how old is the portaudio shipped with Pd?

danomatika:
I've made local changes and forwarded them upstream before. I believe I updated portaudio for the last release when I fixed a bug in it's CoreMIDI handling.

Spacechild1:
shall we add a "check your hardware connection and reopen the device from the menu" for _WIN32 and "please restart Pd" for OSX for now?

danomatika:
I would add something like that with a TODO comment explaining the current situation.

Spacechild1:
I'm so glad we've finally sorted this out! I assume you'll make a PR with your changes.

danomatika:
How is this simple error message?

"audio device interruption detected, save your work and restart pd"

danomatika:
Note: I'm not seeing the print now on unplug and I'm not sure why. I will have to investigate tomorrow.

danomatika:
Ok I found it. I was working with a branch that didn't have the revert applied yet (aka 46fbcba0b8b3ae90df9ea0b4ba29ecdc3cb64015). Using `Pa_IsStreamActive()` never triggers the lock mechanism as portaudio doesn't seem to return the error code, even though it prints it. Using `PaUtil_ValidateStreamPointer()` as before does, so the error is noticed by Pd and the prints happen.

Spacechild1:
https://github.com/pure-data/pure-data/commit/46fbcba0b8b3ae90df9ea0b4ba29ecdc3cb64015 added the device state check via PaUtil_ValidateStreamPointer() instead of relying on just relying on usleep.
then IOhannes opened the following issue:
https://github.com/pure-data/pure-data/issues/450
in the current master, PaUtil_ValidateStreamPointer() is replaced by Pa_IsStreamActive - which internally calls PaUtil_ValidateStreamPointer - so the behaviour should be the same.
I think it's best if you work from the current master. or are you saying the current master isn't working for you when you apply your fixes to the CoreAudio implementation?

danomatika:
`PaUtil_ValidateStreamPointer()` works and `Pa_IsStreamActive` doesn't: portaudio just keeps printing errors and the stream is never closes.

Spacechild1:
PaError Pa_IsStreamActive( PaStream *stream )
{
    PaError result = PaUtil_ValidateStreamPointer( stream );

    PA_LOGAPI_ENTER_PARAMS( "Pa_IsStreamActive" );
    PA_LOGAPI(("\tPaStream* stream: 0x%p\n", stream ));

    if( result == paNoError )
        result = PA_STREAM_INTERFACE(stream)->IsActive( stream );


    PA_LOGAPI_EXIT_PAERROR( "Pa_IsStreamActive", result );

    return result;
}

danomatika:
I've been looking in the portaudio source for a while now. Also, a better way to refer to sources is to link to the specific line, ala https://github.com/pure-data/pure-data/blob/master/portaudio/portaudio/src/common/pa_front.c#L1484

Spacechild1:
oh, thanks!

in 46fbcba the loop breaks if PaUtil_ValidateStreamPointer() < 0 which implies result != paNoError so the CoreAudio specific part shouldn't be called anyway. maybe something happens in the logging calls?

Spacechild1:
as a last ressort: since on MacOS most people use the internal portaudio (I guess), we could make the validator function a macro so that only MacOS uses the private function for now.

danomatika:
I got it. I added handling for the -50 error in portaudio so it will abort and close the stream. This is the right way to do it anyway.

Spacechild1:
so this means we can use Pa_IsStreamActive?

danomatika:
>so this means we can use Pa_IsStreamActive?

Yes.

One stylistic note: make sure code you paste into markdown will render as code, otherwise it can be harder to read. For longer stuff, use triple tildes or backticks:

~~~
```
some code
```
~~~

becomes

```
some code
```

Spacechild1:
ah, the triple tildes do the trick! thanks

PR #447, opened 2018-09-12
more running commits for 0.49-test1

danomatika:
Another PR for smaller commits and fixes during 0.49-test1 development. Paging @umlaeute.

Current important stuff:
* [x] mac/tctk-wish.sh: added support for applying patches to the tcl and/or tk source trees when building Wish for macOS
* [x] added Miller's Tk 8.5.19 back port key fix as a patch
* [ ] key repeat bug on macOS (?)
* [ ] ~hang handling when unplugging audio device not working on macOS (?)~
* [ ] pdfontloader Tcl extension for loading DejaVu Sans Mono on Windows


Spacechild1:
hang handling on OSX has been fixed: https://github.com/pure-data/pure-data/pull/445

Lucarda:
@danomatika  can you refactor the pdfontloader thing to be shipped on the "Font" folder.

The current implementation wouldn't work on 64bit Pd. We can fix that later (I already have a compiled version for 64bit)

Also **DejaVuSansMono-Bold.ttf** is not shipped on the font folder and thats the one we need.

porres:
can you include this one here? https://github.com/pure-data/pure-data/pull/448

only you could do that, right?

millerpuckette:
Do you want pdprototype.tgz to include ._pdfontloader.dll (some kind of Apple file I think)?  May I
just delete that?

millerpuckette:
merged

Lucarda:
@danomatika 

I think the pdfontoader won't work because in the "make app" script the "Font" folder is not copied so the .ttf is not present.

It will be a good to update the "make app" script and also to include DejaVuSansMono-Bold.ttf on the the Font folder.

Lucarda:
Also update pd-gui.tcl to load both **DejaVuSansMono-Bold.ttf**  and **DejaVuSansMono.ttf** .

If not using **DejaVuSansMono-Bold.ttf** it renders horrible as Pd defaults to use bold.

Lucarda:
Yeah.

I think all the pdfonloader thing can be done just in the Font folder.

pdfontloader.dll for **32bit tk85** and pdfontloader64.dll for **64bit tk86** should live there. Also DejaVuSansMono-Bold.ttf

Lucarda:
I already have a compiled pdfontloader64.dll for tk86 64bit.

millerpuckette:
question: is it better that we include pdfontloader64.dll -- or alternatively, should I just be releasing 64-bit Pd with 32-bit Tcl/Tk?  I _think_ the pdfontloader64.dll idea is the better one but not sure.
Also, is there a canonical place one gets DejaVuSansMono-Bold.ttf from?

umlaeute:
https://github.com/dejavu-fonts/dejavu-fonts/

danomatika:
I'll add the font stuff.

As for pdfontloader, what's in pdprototype.tgz is 32 bit and is built for the Tk 8.5.10 that's there. This is the baseline. If you want to have pdfontloader included in the build process for Windows as well, I can add that. I was just unsure of pushing that now at the, more or less, last minute.

Lucarda:
> Also, is there a canonical place one gets DejaVuSansMono-Bold.ttf from?

https://dejavu-fonts.github.io/

> question: is it better that we include pdfontloader64.dll -- or alternatively, should I just be releasing 64-bit Pd with 32-bit Tcl/Tk? I think the pdfontloader64.dll idea is the better one but not sure.

I'm not sure either. It seems there is no problems at all with the 32bit tcl/tk.

Lucarda:
I like the idea of using the 64bit tcl/tk and start finding solutions for the coexisting archs

Lucarda:
> @Lucarda It should. "make app" calls msw/msw-app.sh which copies the font folder: https://github.com/pure-data/pure-data/blob/master/msw/msw-app.sh#L230 Is that not working?


Ok, I'll try to compile. My remark came because the Font dir is not present in Miller's 32bit distributions.

millerpuckette:
OK, so I'm making the distribution wrong on 2 counts: I need to add the 'font' subdir, and I've been
pointing the msw-app.sh script to a 64-bit version of Tcl/Tk.  (That might also explain a couple of hiccups I ran into building 64-bit Pd with the script - I had to replace a dll or 2 before I could run it).  Will attempt to compile a clean 32-bit windows tcl/tk now :)

danomatika:
I have the font added now. Opening a new PR or you can see the commit in the `update/0.49` branch.

danomatika:
I'll integrate pdfontloader into the windows build scripts now in a separate branch/PR.

Lucarda:
@danomatika 

Right, the msw/msw-app.sh copies the font folder.

pd-gui.tcl should also load DejaVuSansMono-Bold.ttf

@millerpuckette 

I think is safe to use tcl/tk 8.5.10 that is on pdprototype.tgz for a 64bit Pd.

I'll try to compile and report results.

:)

danomatika:
>pd-gui.tcl should also load DejaVuSansMono-Bold.ttf

This is done in 5024e8b040684974a24ab9358eb8600bdee448ef

millerpuckette:
Actually since I was using a 64-bit 8.5.19 (wihch I had compiled with your script having assumed that I needed to feed it the 64-bit flag) I can easily just switch to a script-generated 32-bit 8.5.19.  This will have the advantage of being the same tcl/tk as is used on the Mac.  I'll try to put this all together tonight :)

danomatika:
@millerpuckette Quick question: do you want pdfontloader in the main pd repo or is it fine to stay in it's current location? I'm thinking the quickest solution would be to download and build it along with Tcl/Tk in the `msw/tcltk-dir.sh` script as it needs to be told which version of Tcl it's for.

Lucarda:
@millerpuckette 

FYI:

I compiled a 64bit Pd with just:

```
./autogen.sh
./configure
make
make app
```

and everything is running fine with the tcl/tk 8.5.10 coming from pdprototype.tgz.

danomatika:
@millerpuckette FYI: pdfontloader is at https://github.com/pure-data/pdfontloader. The makefile should make it relatively easy to build against the tcltk dir that `msw/tcltk-dir.sh` generates.

Issue #450, opened 2018-09-12
s_audio_pa.c: uses private portaudio headers


umlaeute:
the portaudio backend now uses a function like `PaUtil_ValidateStreamPointer` (at least), which appears to be a *private* function of the portaudio library.
since Pd includes a full copy of portaudio (for system that lack proper dependency handling), this is technically possible.

however, these private symbols are not exported by `libportaudio` (e.g. as found in the official Debian package of portaudio), so one cannot link against such  systemwide portaudio.

this is a serious regression (esp. for the Debian world). please allow the usage of a dyamically linked  portaudio.

umlaeute:
mention @Spacechild1

Spacechild1:
sorry, I forgot that there's the dynamic linking option... I'll swap PaUtil_ValidateStreamPointer for the public Pa_IsStreamActive.

danomatika:
We're no longer using `PaUtil_ValidateStreamPointer`.

PR #451, opened 2018-09-12
fix regression in portaudio

Spacechild1:
replace the private PaUtil_ValidateStreamPointer with the public Pa_IsStreamActive. 

solves https://github.com/pure-data/pure-data/issues/450#issuecomment-420761829

lesson learned: stay away from private headers


millerpuckette:
merged

PR #459, opened 2018-09-13
better error messages for portaudio hotplug fix

Spacechild1:
make error message platform specific

* on Windows we can plug the device back in and reopen it from the menu
* on OSX we have to restart Pd

### DIFF
--- a/src/s_audio_pa.c
+++ b/src/s_audio_pa.c
@@ -667,6 +667,12 @@ int pa_send_dacs(void)
             error("successfully reopened audio device");
         else
             error("audio device not responding - closing audio.");
+        #ifdef _WIN32
+            error("check your hardware connection and reopen the device from the menu");
+        #else
+            /* TODO: find out why we can't reopen a hotplugged device on OSX */
+            error("please restart Pd");
+        #endif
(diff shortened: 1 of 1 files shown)


Spacechild1:
you're right, we can actually skip trying to reopen on OSX! can you add this line or something similar to the last Windows error message:

`error("check your hardware connection and reopen the device from the menu")`

since you'll have to make a PR with your portaudio changes it makes sense that you also include the updated error messages. I'm closing this.

src/s_audio_pa.c
```diff
@@ -667,6 +667,12 @@ int pa_send_dacs(void)
             error("successfully reopened audio device");
         else
             error("audio device not responding - closing audio.");
+        #ifdef _WIN32
+            error("check your hardware connection and reopen the device from the menu");
+        #else
+            /* TODO: find out why we can't reopen a hotplugged device on OSX */
```
danomatika: "macOS" not "OSX" as of over a year ago now...

Spacechild1: :-D now you can tell how much I know about Apple

PR #454, opened 2018-09-12
add missing  [drawtext] / [drawsymbol] objects to documentation

porres:
added references to [drawtext] and [drawsymbol] into help-intro.pd

added [drawtext] in drawtext-help.pd file as it wasn't mentioned anywhere in any help file... and seems like it could be into drawtext-help.pd if any :) - but I didn't add any information on what it does or what's different about, because I just have no idea!

PR #460, opened 2018-09-13
more updates for 0.49

danomatika:
* [x] added DejaVu Sans Mono bold
* [x] detecting audio device disconnect works on macOS (required portaudio fixes which were forwarded upstream)
* [x] added unplug error message suggestions based on platform, ie. on mac "you may need to save and restart pd"

and a few other small things


porres:
I didn't include this in the 0.49 branch, but maybe it could be included as it's so simple? https://github.com/pure-data/pure-data/pull/454

PR #461, opened 2018-09-13
consistent post, error, & bug message styling

danomatika:
This is a detail PR which attempts to make the pd window post, error, and bug messages consistent from what I can tell is Pd's overall message style. The main approach was:

* prefer initial lower case, ie. "Couldn't do X" -> "couldn't do X"
* error response to an action starts with "... " ie.
   ~~~
   trying to open X...
   ... failed
   ~~~
* use "MIDI" instead of "midi"
* fix any typos
* update translations

I didn't change the content of the message unless it didn't make sense or did not match the context of the code, ie. a couple copy/pastes without editing the string. ~~~Also, I didn't change anything in [deken].~~~

Additionally:

* commented out the GUI apple_events.tcl verbose prints as they aren't really needed for normal use and clutter the verbose log when debugging


danomatika:
@umlaeute I didn't want to touch the deken messages? Should we consider that?

umlaeute:
@danomatika as long as there is a single commit that changes the `pd_deken.tcl` file (so the change can easily be backported to deken upstream), i don't have a problem with it.

danomatika:
Should the messages in the *deken* log also follow a similar format?

danomatika:
Ok, deken updated. I also moved the "[deken]: " header string out of the translation strings.

danomatika:
This branch is now up to date (should merge cleanly).

umlaeute:
iirc i kept the `[deken]: ` header string in the translation to make it distinguishable from similar strings (sometimes the same phrase must be translated differently based on the context), and to help the translators find the context of the translation.

it's not a big deal though (and i'm happy with removing it)

danomatika:
This is now up to date.

millerpuckette:
blindly resolved conflicts in de.po by taking the patch's version of everything

danomatika:
No need. The POs are updated by gettext automatically if you have it installed and run makenin the app directory.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Sorry, what I meant to finish saying was:

With get text installed and Pd configured to use it, running make or make in the /po directory will automatically update the translation files. That's the whole point, you shouldn't have to worry about changing the localized able text in the source files, only if the translations themselves are touched.

PR #463, opened 2018-09-13
remove lib "already loaded" error printing

Ant1r:
to avoid cluttering the Pd console, e.g when many abstractions are declaring the
same library.

I have chosen to simply comment the error out, for memory.

### DIFF
--- a/src/s_loader.c
+++ b/src/s_loader.c
@@ -329,7 +329,7 @@ int sys_load_lib(t_canvas *canvas, const char *classname)
 
     if (sys_onloadlist(classname))
     {
-        error("%s: already loaded", classname);
+        /*error("%s: already loaded", classname);*/
         return (1);
     }
         /* if classname is absolute, try this first */
--- a/src/s_loader.c
+++ b/src/s_loader.c
@@ -327,11 +327,8 @@ int sys_load_lib(t_canvas *canvas, const char *classname)
     data.canvas = canvas;
     data.ok = 0;
 
-    if (sys_onloadlist(classname))
-    {
-        /*error("%s: already loaded", classname);*/
-        return (1);
-    }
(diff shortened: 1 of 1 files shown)


src/s_loader.c
```diff
@@ -329,7 +329,7 @@ int sys_load_lib(t_canvas *canvas, const char *classname)
 
     if (sys_onloadlist(classname))
     {
-        error("%s: already loaded", classname);
+        /*error("%s: already loaded", classname);*/
```
umlaeute: i'm not a big fan of disabling code by commenting it.
why don't you just delete the line?

danomatika:
As a note, if we *really* need to notify the user without being too annoying, we'd need to keep track of which classnames we're already made a point for, maybe with a string set. IMHO, simply removing it is the better solution for now.

src/s_loader.c
danomatika: Yeah. I say remove it altogether.

Ant1r: you're right it's not nice.
I was embarrassed because actually Miller changed a "verbose(1, ...)" for an "error" some months ago (fee700b3b35caddab2ab4a48a726a6cdbf78e0ce). 
OK i'm removing it for good.

porres:
how about telling it didn't find/load?

danomatika:
>how about telling it didn't find/load?

...? That's what it was already doing.

AFAICT the problem is that it would tell you *every time* an abstraction had a [declare] to load the same library. If you have a clone with 100 abstractions using [declare], you would then get this messages printed 99 times when the parent patch is opened. Since there is not real issue if you are reloading the same classname, maybe it's better to not print?

millerpuckette:
merged (after tweaking code style)

danomatika:
This has been merged.

PR #466, opened 2018-09-13
Update release notes for 0.49

porres:
- I changed the order and put two sections that mentioned declare and search paths closer
- I added information on how you can use soundfiler to only read samples, not load them
- I added changes to log, pow, and pow~


millerpuckette:
merged (and some verbiage improved afterward)

PR #467, opened 2018-09-14
fix obscure bug with [donecanvasdialog(

Spacechild1:
Pd 0.49 introduced a weird bug with the [donecanvasdialog( message.

if you send this message to a canvas and this canvas has GOP enabled, deleting the canvas will
a) show a Tcl error if its a subpatch
b) segfault if its an abstraction

here's the backtrace:

~~~
Thread 1 received signal SIGSEGV, Segmentation fault.
0x675a6a09 in atom_string (a=0xfeeefeee, buf=0x28f012 "", bufsize=500)
    at m_atom.c:70
70          switch(a->a_type)
(gdb) bt
#0  0x675a6a09 in atom_string (a=0xfeeefeee, buf=0x28f012 "", bufsize=500)
    at m_atom.c:70
#1  0x67562a1b in canvas_reflecttitle (x=0x37896c8) at g_canvas.c:607
#2  0x67573d0a in canvas_editmode (x=0x37896c8, state=1) at g_editor.c:4603
#3  0x6756a506 in canvas_undo_canvas_apply (x=0x37896c8, z=0x377e1a0,
    action=0) at g_editor.c:1377
#4  0x6759cbb5 in canvas_undo_doit (x=0x37896c8, udo=0x377e1f0, action=0,
    funname=0x676378c8 <__FUNCTION__.4674> "canvas_undo_free") at g_undo.c:146
#5  0x6759d197 in canvas_undo_free (x=0x37896c8) at g_undo.c:326
#6  0x675631ab in canvas_free (x=0x37896c8) at g_canvas.c:787
#7  0x675afd87 in pd_free (x=0x37896c8) at m_pd.c:31
#8  0x67576bfb in glist_delete (x=0x3778078, y=0x37896c8) at g_graph.c:134
#9  0x6757118f in canvas_doclear (x=0x3778078) at g_editor.c:3698
#10 0x6756f4e6 in canvas_key (x=0x3778078, s=0x50bdd0, ac=3, av=0x28f8b4)
    at g_editor.c:3122
#11 0x675ad1f2 in pd_typedmess (x=0x3778078, s=0x50bdd0, argc=3,
    argv=0x28f8b4) at m_class.c:949
#12 0x6757a0c5 in guiconnect_anything (x=0x37783c0, s=0x50bdd0, ac=3,
    av=0x28f8b4) at g_guiconnect.c:73
#13 0x675ad5d9 in pd_typedmess (x=0x37783c0, s=0x50bdd0, argc=3,
[...]

PR #468, opened 2018-09-14
updates for 0.49 test3

### LINKED ISSUE #379: "format s, 42"--[oscparse] crash
https://github.com/pure-data/pure-data/blob/f83c88ba7ac68368248797ef0d0febc3526432bb/src/x_misc.c#L555

           if (*sp)
                typecode = *sp++;
            else if (argv[j].a_type == A_SYMBOL)
                typecode = 's';
            else typecode = 'f';
            if (typecode == 's')
                msgindex += ROUNDUPTO4(strlen(argv[j].a_w.w_symbol->s_name) + 1);

`oscformat` trusts the type code given in the format specifier `sp`, before checking the type of the atom.  When the type code is `'s'` and the atom is not a symbol, the `w_symbol` field is bogus and an invalid memory access crash results.

danomatika:
Some more running updates:

* catch malformed messages to [oscformat], fixes #379
* fixed spelling error in posted message
* fixed undo-related crashes with `donecanvasdialog`, fixes https://github.com/pure-data/pure-data/pull/467

### DIFF
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -564,7 +564,7 @@ static void oscformat_list(t_oscformat *x, t_symbol *s, int argc, t_atom *argv)
                 msgindex += ROUNDUPTO4(strlen(argv[j].a_w.w_symbol->s_name) + 1);
             else
             {
-                pd_error(x, "oscformat: expected symbol but got float");
+                pd_error(x, "oscformat: expected symbol for argument %d", j+1);
                 return;
             }
         }
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -559,7 +559,15 @@ static void oscformat_list(t_oscformat *x, t_symbol *s, int argc, t_atom *argv)
             typecode = 's';
         else typecode = 'f';
         if (typecode == 's')
-            msgindex += ROUNDUPTO4(strlen(argv[j].a_w.w_symbol->s_name) + 1);
+        {
+            if (argv[j].a_type == A_SYMBOL)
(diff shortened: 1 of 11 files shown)


src/x_misc.c
```diff
@@ -559,7 +559,15 @@ static void oscformat_list(t_oscformat *x, t_symbol *s, int argc, t_atom *argv)
             typecode = 's';
         else typecode = 'f';
         if (typecode == 's')
-            msgindex += ROUNDUPTO4(strlen(argv[j].a_w.w_symbol->s_name) + 1);
+        {
+            if (argv[j].a_type == A_SYMBOL)
+                msgindex += ROUNDUPTO4(strlen(argv[j].a_w.w_symbol->s_name) + 1);
+            else
+            {
+                pd_error(x, "oscformat: expected symbol but got float");
```
umlaeute: iirc this error message is consistent with similar errors (even though the "float" might also be a "gpointer").
probably a better wording would be `"oscformat: expected symbol as argument #%d" % (j,))`

danomatika: I thought of that, but the types that oscformat takes are only floats and symbols anyway. The rest get coerced into a float.

danomatika: I would use "oscformat: expected symbol for argument %d" but should the index print be 0 or 1-indexed? ... :)

danomatika: Done. 1-indexed makes sense in this context.

millerpuckette:
merged

PR #474, opened 2018-09-17
better portaudio device unplug handling for macOS

danomatika:
This is a follow up to the assert & error fix/hack so portaudio won't hang when a device is unplugged. This PR now adds *proper* error handling by using a CoreAudio device listener which receives a notification when a device is no longer alive and can abort the portaudio stream.

This should hopefully be a fix for #470 (and tangentially #443).

A 64bit test build with the fix for macOS is here: 
http://docs.danomatika.com/pdbuilds/Pd-0.49-0test3-unplugfix.zip

1. plug in external sound device
2. start Pd & choose the device\*
3. turn on DSP
4. unplug device
5. Pd should not hang and bring an error message

\* _Please repeat 3 times: device as both input/output, only input, & only output_

Mainly, we *should not* see any error messages until the unplug happens and Pd should not hang when this happens (as it did in previous versions).

The error message should look like this:

screen shot 2018-09-17 at 12 16 04 pm

paging @Spacechild1 @porres

### DIFF
--- a/portaudio/portaudio/src/hostapi/coreaudio/pa_mac_core.c
+++ b/portaudio/portaudio/src/hostapi/coreaudio/pa_mac_core.c
@@ -1102,6 +1102,34 @@ static OSStatus AudioDevicePropertyGenericListenerProc( AudioDeviceID inDevice, 
     return osErr;
 }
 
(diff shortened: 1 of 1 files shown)


danomatika:
Although #470 is still open, I think this PR is an improvement.

PR #476, opened 2018-09-18
[WIP]  32 and 64 bit installer

Lucarda:
This allows to have the 32bit and 64bit Pd installed on the same system without clashing.

It does:

- Individual start menu entries.
- Individual desktop icons.
- Individual default installation paths.

Tested and both installations can coexist with no trouble but more work on `src/s_file.c` and `tcl/pd_guiprefs.tcl` needs to be done so that each Pd has its own 'windows registry' settings.


umlaeute:
also i don't understand why you changed the install dir to `Pd-32bit` and `Pd-64bit`.
shouldn't the location (`%ProgramFiles%` vs `%ProgramFiles(x86)%`) be enough to separate them (don't tr yto optimize for the obscure case, where somebody wants to install both 32bit and 64bit versions side-by-side into the same directory).

Lucarda:
@danomatika and @umlaeute

These two files needs Pure-Data-32 / Pure-Data-64 ifs.

- https://github.com/pure-data/pure-data/blob/master/src/s_file.c#L284

- https://github.com/pure-data/pure-data/blob/master/tcl/pd_guiprefs.tcl#L153

I can't handle these C / Tcl.

Help needed!

:)

Lucarda:
> also i don't understand why you changed the install dir to Pd-32bit and Pd-64bit

This can be changed but it was on pourpuse to make it totally clear.

umlaeute:
> These two files needs Pure-Data-32 / Pure-Data-64 ifs.
>     https://github.com/pure-data/pure-data/blob/master/src/s_file.c#L284
>     https://github.com/pure-data/pure-data/blob/master/tcl/pd_guiprefs.tcl#L153
> I can't handle these C / Tcl.

on the C-side, the fix is trivial.
the question is more, how to do that properly (for all platforms, not just Win).

in the Tcl/Tk side, i'm not entirely sure, whether we should change it at all.
there is no proper way, to detect (from the GUI-side) which architecture the Pd-core is running (at this stage in the startup phase).
a better idea would be to be able to set the registry base key (or base path or whatever is used on the other systems) by a single configuration file that resides in the tcl/ folder (and which could be generated by the installer - either online or offline)




>>    also i don't understand why you changed the install dir to Pd-32bit and Pd-64bit
>
> This can be changed but it was on purpose to make it totally clear.

why?
i mean, it's probably fine to add such info to the start-menu entry, but there's little point in adding it to the installation path (there's so much information we might want to make totally clear; think `C:\Archivos de programa (x86)\Pd-0.49-1 for Windows (64bit) doubleprecision by Miller Puckette - a bit like MaxMSP but BSD3 licensed\` ;-) more seriously, it adds redundancy which can easily be made self-contradictory (in the given example: why is a 64bit binary installed in the 32bit wow path?))

Spacechild1:
> shouldn't the location (%ProgramFiles% vs %ProgramFiles(x86)%) be enough to separate them

I rather agree with @umlaeute. although some programs do this (e.g. REAPER (x64)), I don't think it's necessary.

Lucarda:
> shouldn't the location (%ProgramFiles% vs %ProgramFiles(x86)%) be enough to separate them

Ok, no trouble at all with that.

:)

Lucarda:
> in the Tcl/Tk side, i'm not entirely sure, whether we should change it at all.
there is no proper way, to detect (from the GUI-side) which architecture the Pd-core is running (at this stage in the startup phase).
a better idea would be to be able to set the registry base key (or base path or whatever is used on the other systems) by a single configuration file that resides in the tcl/ folder (and which could be generated by the installer - either online or offline)

This is Ok if possible with the installer but what about the Zip app?

You already did the magic detection with Deken, I guess this detection is late?

Idea: Now that we have `pref file` couldn't we not use at all the 'windows registry' exept for file association?

umlaeute:
> This is Ok if possible with the installer but what about the Zip app?

just use a different default (in the ZIP-file).
- one registry root for the 32bit installation
- another registry root for the 64bit installation
- yet another registry root for all ZIP "installations".

if the user want to keep their ZIP-installations in different config
spaces, they will have to to tweak manually (just like the users who
want to have Pd-0.43 and Pd-0.49 co-exist but with different preferences).

> You already did the magic detection with Deken, I guess this detection is late?

yes.

> Idea: Now that we have `pref file` couldn't we not use at all the 'windows registry' exept for file association?

heresy. (if you don't want windows, why do you use it).

apart from that: what will this buy us? you are basically exchanging a
file on a virtual filesystem with a file on a real filesystem.

the only advantage would be, that the preference file could be stored
alongside the pd.exe (e.g. in the same directory).
however, not every user of %ProgramFiles%\Pd\bin\pd.exe can be assumed
to have write access to %ProgramFiles%\Pd\.
which prevents these users from saving their preferences.

finally: i think it is entirely possible to create a bat/cmd/.ps1 script
that starts Pd with `-noprefs -prefsfile my64bit.prefs`

Lucarda:
> 
> - one registry root for the 32bit installation
> - another registry root for the 64bit installation
> - yet another registry root for all ZIP "installations".

How you do the "registry root for all ZIP "installations"?
The installer is made from the zip version.
 what did you meant with:

> by a single configuration file that resides in the tcl/ folder (and which could be generated by the installer - either online or offline)

?

umlaeute:
> How you do the "registry root for all ZIP "installations"?
> The installer is made from the zip version.

use the default registry root for ZIP "installations". 
add an override of the default for NSI-installations (while building the NSI installer).

> what did you meant with:
> >   by a single configuration file that resides in the tcl/ folder (and which could be generated by the installer - either online or offline)

"put a single file containing the registry root into the tcl/-folder. this file could be generated by the installer-script; or missing."
this sounds an awful lot like what I wrote in my incomprehensible style, so I guess it is not much better. which part should i clarify?

Lucarda:
> "put a single file containing the registry root into the tcl/-folder. this file could be generated by the installer-script; or missing."
this sounds an awful lot like what I wrote in my incomprehensible style, so I guess it is not much better. which part should i clarify?

I think I got it. Some txt with just:

`"HKEY_CURRENT_USER\\Software\\Pure-Data-32"`

or

`"HKEY_CURRENT_USER\\Software\\Pure-Data-64"`

?

I like it. :)

Lucarda:
I think the Txt thing resolves everything.

:)

umlaeute:
yep, something like this. i'd like to hear the opinion of others on this though (@danomatika  :-))

Lucarda:
Ok, I'll try to make a nicer nsi script

:).

Lucarda:
@umlaeute I made the changes that you requested, can you review?

The thing now goes with `makensis -DARCHI=32 pd.nsi`

Lucarda:
Something went wrong when trying to do some little fix.

I'll do another PR because I cant fix this one any more.

But this one works with `./build-nsi.sh <path-to-pd-dir> <pd-version-number> <arch (32 or 64)>`

msw/pd.nsi
```diff
@@ -7,7 +38,12 @@
 !define PRODUCT_VERSION "PDVERSION"
 !define PRODUCT_PUBLISHER "Miller Puckette"
 !define PRODUCT_WEB_SITE "http://www.puredata.info"
-!define PRODUCT_DIR_REGKEY "Software\Microsoft\Windows\CurrentVersion\App Paths\pd.exe"
+;
+; removed as of 0.49 to prevent 32/64 installer clashing. No negative side effects were observed.
+;
+;			!define PRODUCT_DIR_REGKEY "Software\Microsoft\Windows\CurrentVersion\App Paths\pd.exe"
```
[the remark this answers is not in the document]
reduzent: Most likely it there was no particular reason, but it got there because it was in the original template that I edited

Lucarda:
> get rid of the `...\Pd-${ARCHI}bit\` directory.

there's no such dir.
Is the 'start menu' / 'desktons icons' thing.

msw/pd.nsi
```diff
@@ -65,8 +101,15 @@ FunctionEnd
...
 OutFile "/tmp/pd-${PRODUCT_VERSION}.windows-installer.exe"
-InstallDir "$PROGRAMFILES\Pd"
-InstallDirRegKey HKLM "${PRODUCT_DIR_REGKEY}" ""
+InstallDir "$PROGRAMFILES${ARCHI}\Pd"
+
+;
+; removed as of 0.49 to prevent 32/64 installer clashing. No negative side effects were observed.
+;
+;
+;			InstallDirRegKey HKLM "${PRODUCT_DIR_REGKEY}" ""
```
umlaeute: looking up the [NSIS Documentation](http://nsis.sourceforge.net/Reference/InstallDirRegKey), I think the usage of this is to remember the installation directory across re-installs. As such, it seems clear that their are no *immediate* side-effects visible when doing a single installation. The change will only become apparent in consecutive use of the installer.
As such, I guess it should stay in (probably with a explanatory comment why it should stay)

umlaeute:
>  Is the 'start menu' / 'desktons icons' thing.

ah indeed. static code analysis only gets you so far, if you don't know the eco-system well.
sorry for the noise

msw/pd.nsi
Lucarda: If that is present you get the wrong default installation dir for the parallel 2nd installation something like: c:\program files\pd\bin. I tested it.

umlaeute: i guess, that's because the `InstallDirRegKey` must be different for 32bit and 64bit installations

Lucarda:
Just a GIT question:

I resolved some conflicts from the web and now I can no longer push, I get:

```
Pushing to https://github.com/Lucarda/pure-data.git
To https://github.com/Lucarda/pure-data.git
 ! [rejected]          windows32and64installer -> windows32and64installer (non-fast-forward)
error: failed to push some refs to 'https://github.com/Lucarda/pure-data.git'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.
```
Any hints to not make a new PR?

umlaeute:
just pull the changes from github, resolve any conflicts locally and push again.

btw, this is basically what the error message says:
> hint: Updates were rejected because the tip of your current branch is behind
> hint: its remote counterpart. Integrate the remote changes (e.g.
> hint: 'git pull ...') before pushing again.

(interfacing git via the github-webgui mostly sucks (unless for trivial cases))

Lucarda:
resolve any conflicts locally?

umlaeute:
yes.

Lucarda:
```
$ git pull --all
error: Pulling is not possible because you have unmerged files.
hint: Fix them up in the work tree, and then use 'git add/rm <file>'
hint: as appropriate to mark resolution and make a commit.
fatal: Exiting because of an unresolved conflict.
```

Lucarda:
Ok, sory for the noise. I have to have some sleep anyway.

I'll do a new PR.

:)

umlaeute:
sure, get some sleep.
when you awake, make sure to read the *hints* you get when doing a pull:
> error: Pulling is not possible because you have unmerged files.
> hint: Fix them up in the work tree, and then use 'git add/rm <file>'
> hint: as appropriate to mark resolution and make a commit.

so do a `git add + git commit` on the (local) files you want to get in. after that do the `git pull`.
it will probably want to merge your changes with the changes it pulled from github. do the merge and resolve conflicts (alternatively - if you want a more linear history - you could also try to rebase your branch on the github branch or do a cherry-pick of the local commit).
once you've resolved everything, push.

i think it's better to once learn to use git (there's no shame here; i think mastering git has a steep learning curve for everybody), instead of doing new PRs whenever you run into conflicts.

oh, and since this is totally unrelated to the PR at hand, I've removed all this chat-conversation about `git` (and will remove this comment as well, once I'm sure you've read it)

Spacechild1:
here's how to actually resolve merge conflicts: https://git-scm.com/book/en/v2/Git-Branching-Basic-Branching-and-Merging -> Basic Merge Conflicts
I found the git-book to be an extremely useful resource for learning how to use git myself (which is not so long ago, so I feel you :-).

(@umlaeute shall delete this message as well)

Lucarda:
After reading both post above I power my dev machine and when started windows git there was lots of files changed.

I resolved conflicts and now is working.

Thanx!

Lucarda:
Ok, I have tested and the only known bug is that if you install first the 32 and then the 64, if 64 gets uninstalled the 32 loses the file association.

Not much a big problem as you can fix it with "open with" from the explorer.

@umlaeute can you review?

This only fixes the 32/64 installer. Work must be done on the C/tcl side so that the 32 app don't share the registry with the 64 app.

### The installer is ready.

Lucarda:
@millerpuckette 

This allows to have 32 and 64 bit Pd installed on the same system.
It defaults installation to:

- 64bit-OS 64bit-Pd: `"%Program Files%\Pd"`
- 64bit-OS 32bit-Pd: `"%Program Files (x86)%\Pd"`
- 32bit-OS 32bit-Pd: `"%Program Files%\Pd"`

It creates different:
- Start-menu entries
- Desktop icons

Some work must be done outside of the installer so that both apps don't share the same 'registry' but all that can be done with the installer's scripts is done in this PR.


usage:

`./build-nsi.sh <path-to-pd-dir> <pd-version-number> <arch (32 or 64)>`


:)

danomatika:
Looks good to me.

umlaeute:
i've updated the `build-nsi.sh` script to automatically detect the `arch` value if it is missing (please test :-))

Lucarda:
fdf39d7 is running nicely.

Lucarda:
Why did you erase the commits? they were almost going. :)

I got that part but I'm starting to fell asleep.

will hang out 1 hour more or less :)

umlaeute:
i erased them because they are in the wrong PR.
i moved them over to #481

Lucarda:
Oops. Can you undo this merge.

Is was included on #481

Lucarda:
Sorry for the noise `master` seems Ok.

Issue #483, opened 2018-09-21
Pd-GUI no longer honors `-path`


umlaeute:
up till `Pd-0.48-2` I could start Pd with:

~~~
pd -path /path/to/my/pdguiplugins/
~~~

and the Pd-GUI would automatically load any GUI-plugins found in `/path/to/my/pdguiplugins/`.
It seems this no longer works with `Pd-0.49-0test3`.

It *does* work if I add `/path/to/my/pdguiplugins/` to my searchpaths in the preferences and restart Pd.

Since GUI-plugins are only loaded at startup, this makes it a bit inconvenient when using GUI-plugins which you only want to enable ad-hoc.

I guess this is related to separating searchpaths given with `-path` and via the path-preferences.

danomatika:
The temp paths probably need to be added when loading. I thought it was already doing that, but I'll check.

millerpuckette:
I looked at this and can confirm that 0.48 and 0.49 act differently, although I can't see that any of the relevant code has changed.. mysterious.

danomatika:
Might be related to the changes to [declare]?

danomatika:
I think I see it. st_temppath is only searched in `g_canvas.c` canvas_open() but not used by canvas_path() or canvas_lib(). I'll add and test.

Spacechild1:
btw, here's another regression with [declare]: https://github.com/pure-data/pure-data/issues/410

danomatika:
This is a different issue dealing with the GUI not knowing about the st_temppath. [declare] seems to be working fine with `-path` but GUI plugins do not.

PR #485, opened 2018-09-21
send st_temppath to GUI

danomatika:
This PR adds sending of the `st_temppath` core paths from the commandline `-path` option to GUI so it they be used for plugins, helpbrowser, and deken. This is accomplished by adding a new GUI variable ` $::sys_temppath`. This should fix the regression of `-path` not be used for loading GUI plugins.

This is a fix for #483 which was introduced when separating the `-path` temporary paths from the savable `st_searchpaths`.

### DIFF
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -142,6 +142,7 @@ struct _instanceinter
 extern int sys_guisetportnumber;
 extern int sys_addhist(int phase);
 void sys_set_searchpath(void);
+void sys_set_temppath(void);
 void sys_set_extrapath(void);
 void sys_set_startup(void);
 void sys_stopgui( void);
@@ -1285,6 +1286,7 @@ static int sys_do_startgui(const char *libdir)
     sys_get_audio_apis(apibuf);
     sys_get_midi_apis(apibuf2);
     sys_set_searchpath();     /* tell GUI about path and startup flags */
+    sys_set_temppath();
     sys_set_extrapath();
     sys_set_startup();
                        /* ... and about font, medio APIS, etc */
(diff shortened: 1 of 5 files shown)


millerpuckette:
that fixed it for me.

umlaeute:
my guts say that they don't *really* belong into the deken-preferences.
but then, you can add any user-defined path (as in: not coming from any Pd search path variable at all) via the deken-preferences as well, so i don't see much harm.
the point of these preferences is to present the user with an "interesting" set of paths to install things to.

umlaeute:
and on second thought, it's probably *good* to present these paths :-)

danomatika:
I removed them as they are temporary. I think it only makes sense to show the search paths that are built in (aka "standard") or saved specifically by the user (user search paths).

umlaeute:
well, the reasoning was more: what if you want to create a self-contained application for your project, where you want all libraries be installed inside the project directory.
typically, one would create a start-script that runs something like the following (this pretty much pre-dates the current `[declare]` magic)

~~~sh
pd -path abs:libs -open MAIN.pd
~~~

if the deken-preferences showed the temporary paths, this would make it very simple to install deken packages into the project-specific `libs` folder without needing to manually navigate there.

of course, deken would save the installation path (for the next invocation of Pd) which might be an undesired side-effect (but that also happens when manually selecting the installation folder).

so i think there are uses, even for temporary paths.

but the benefits are probably small enough, and i don't care very much.

danomatika:
>of course, deken would save the installation path (for the next invocation of Pd) which might be an undesired side-effect (but that also happens when manually selecting the installation folder).

That's the kind of action I was trying to *avoid* with the whole temppath approach.

PR #487, opened 2018-09-22
dump tcl source code on error

claudeha:
printing the offending tcl code on error may help developers diagnose the source


claudeha:
not sure whether this is actually useful or not - usually the line in question is enough to find relevant source code, and this can end up printing pages of irrelevant stuff....

PR #491, opened 2018-09-23
more simple bugfixes for Pd-0.49

### LINKED ISSUE #478: Windows/Pd doesn't honour directory started from
When starting Pd on Windows from the command line in my project-directory, the first "Open" (<kbd>Ctrl</kbd>-<kbd>O</kbd>) will fire up the open-dialog in `$::env(HOME)/Documents` (or in my `$::env(HOME)` if the former does not exist).

i think this is unexpected behaviour and quite confusing:
- I keep my projects wherever i like (on my USB-stick; on my 2 TB external disk)
- I started from the cmd (since i know what i'm doing)
- I `cd`ed into my project directory (because this is where all my files live)
- I call `....\Pd.com` (passing my project's `MAIN.pd` patch, which gets successfully opened!)
- I trigger an `[openpanel]` because i want to load one of my project's *soundfiles* (which happen to reside besides my `MAIN.pd` patch).
   - alternatively I just use <kbd>Ctrl</kbd>-<kbd>O</kbd> to open an additional Pd-patch
- the "file open dialog" lands me in `~/Documents/`, and I have to navigate to my project again!!!

(note: while i write "I" it is not actually me who is experiencing this; rather it's my work-colleagues who keep pestering me about it but haven't created any bug-report yet...; also they don't do these steps (apart from the last one) manually, instead they just open a `myproject.bat` file which does all the things for them.)

this has been introduced with `Pd-0.48-0`, as part of the "Pd Documents" revamp.
The problem does not exist on Linux.
I see the value of providing a default documents directory for ordinary users (kind of).
But for power-users (who manually start Pd from the cmdline), this is impractical.

i suggest to:
- keep the current behaviour  when Pd is started from the "Start Menu" resp. the "Application Link" (on the desktop). afaict, these should start Pd via the GUI (but i haven't checked the installers; if this is not true, probably the installers should be fixed)
- when starting Pd from the cmdline (directly calling `pd.exe` resp. `pd.com`) the current working directory should be honoured.

i'm not entirely sure how Pd should behave when being started by clicking on a Pd-patch.
probably:
- when Pd is launched while opening the Pd-patch, the fileopendir should be set to the directory containing the Pd-patch
- when the patch is opened into a running Pd-instance, don't change whatever fileopendir is currently set to

umlaeute:
trying to keep bugfixes even simpler than last time...

- [x] on windows, when starting Pd via pd.exe (or pd.com), use the current-directory as the starting point for fileopendir/filenewdir (as is already the case with the other OSs); Closes #478

### DIFF
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -383,9 +383,6 @@ proc init_for_platform {} {
             option add *DialogWindow*font menufont startupFile
             option add *PdWindow*font menufont startupFile
             option add *ErrorDialog*font menufont startupFile
-            # initial dir is home
-            set ::filenewdir $::env(HOME)
-            set ::fileopendir $::env(HOME)
             # set file types that open/save recognize
             set ::filetypes \
                 [list \


tcl/pd-gui.tcl
```diff
@@ -383,9 +383,6 @@ proc init_for_platform {} {
             option add *DialogWindow*font menufont startupFile
             option add *PdWindow*font menufont startupFile
             option add *ErrorDialog*font menufont startupFile
-            # initial dir is home
```
danomatika: The docsdir plugin uses this default to tell when to override the directories at https://github.com/pure-data/pure-data/blob/master/tcl/pd_docsdir.tcl#L126

Is there a default dir when the GUI is started now or will these silly be empty? This is fine, we just might need to double check things.

Maybe a simple fix for the docsdir plugin is to check if "$::filenewdir" is set? If the GUI starts without the file dir paths being set, then the docsdir uses it's path. Otherwise, the panels fall back to home either of the filedir directories do not exist (or are empty) in wheredoesthisgo.tcl, pd_menucommands.tcl, and pdtk_canvas.tcl.

umlaeute: there is always a sane default for `filenewdir` and `fileopendir` (i use the two interchangebly)

1. `$::filenewdir` is *always* set to `pwd` in  [pd-gui.tcl:174](https://github.com/pure-data/pure-data/blob/d5b786dc64070229f5b1cd59b0fe2948c0e0c065/tcl/pd-gui.tcl#L174)
1. if Pd is started via the GUI, `$::filenewdir` is always set to `${HOME}` in [pd-gui.tcl:814](https://github.com/pure-data/pure-data/blob/d5b786dc64070229f5b1cd59b0fe2948c0e0c065/tcl/pd-gui.tcl#L814)
1. *Windows* used to be the only platform that would unconditionally `set ::filenewdir $::env(HOME)` afterwards.

*1.* and *2.* establish the sane defaults which i want to see on all platforms.
*Windows* also happens to be the only platform that would ignore the pwd when being started via the Pd-core; this is #478 , and i believe is caused by *3.* and this is what i want my patch attempts to fix.

umlaeute: also note that [pd-gui.tcl:174](https://github.com/pure-data/pure-data/blob/d5b786dc64070229f5b1cd59b0fe2948c0e0c065/tcl/pd-gui.tcl#L174) makes any check whether this vars are unset useless (they will always be set).

the only situation that the defaults won't work is if `${HOME}` does not exist (in which case there's a serious problem anyhow). we could nevertheless check for that in [pd-gui.tcl:814](https://github.com/pure-data/pure-data/blob/d5b786dc64070229f5b1cd59b0fe2948c0e0c065/tcl/pd-gui.tcl#L814)

danomatika: Ok. If they are set to HOME before the plugin is loaded, then everything should be fine. The idea is that if the plugin is loaded and they are anything else, the paths should be honored and docspath is not used.

danomatika: >Windows used to be the only platform that would unconditionally set ::filenewdir $::env(HOME) afterwards.

That might be my fault. In any case, I will always applaud consistency of behavior between platforms.

danomatika:
Hah, I was a little confused until I realized we have two very similar-ly named branches right now: "update/0.49" and "update**s**/0.49."

umlaeute:
On 2018-09-24 12:37, Dan Wilcox wrote:
> Hah, I was a little confused until I realized we have two very similar-ly named branches right now: "update/0.49" and "update**s**/0.49."

indeed, sorry for the confusion (i realized my typo after i pushed).
should i just cherry-pick my single commit into the `update/0.49`
branch, create a new PR based on that and ditch the `updates/0.49`?

fgamsdr
IOhannes

danomatika:
You can add it to the other branch if you want. That might keep it simpler.

danomatika:
PR is already there: #497

umlaeute:
done (included in #497)

PR #494, opened 2018-09-23
increase portaudio max poll

### LINKED ISSUE #470: Pd 0.49test3 not recognizing audio card at 1st (macOS)
Many times when  I start Pd and try to turn DSP on I get this error

screen shot 2018-09-15 at 13 16 36

It's not always, and I have to go into audio settings and just press "ok" without changing anything, than audio comes up.

danomatika:
This PR increases the portaudio max poll to 2000 to fix #470. This detail was used in a test build but not actually committed. This is a replacement for #490 as it no longer seems necessary to patch the portaudio sources (yay!).

A test build for this approach is: 
http://docs.danomatika.com/pdbuilds/Pd-0.49-0test4-unplugfix4-nopatch.zip

There are also a few updates to the portaudio/portmidi scripts.

### DIFF
--- a/mac/README.txt
+++ b/mac/README.txt
@@ -111,7 +111,12 @@ tcltk-wish.sh --git commandline option. Oftentimes, these kinds of issues will
 appear with a newer version of macOS before they have been fixed by the open
 source community.
 
-Additionally, Pd uses an older version of Tcl/Tk for backwards compatibility on macOS. As such, small bugfixes from newer versions may need to be backported for the Pd GUI. Currently, this is handled in the tcltk-wish.sh script by applying custom patches to either the Tcl and/or Tk source trees. To skip applying patches, use the tcltk-wish.sh --no-patches commandline option. See mac/patches/README.txt for more info.
+Additionally, Pd uses an older version of Tcl/Tk for backwards compatibility on
+macOS. As such, small bugfixes from newer versions may need to be backported for
+the Pd GUI. Currently, this is handled in the tcltk-wish.sh script by applying
+custom patches to either the Tcl and/or Tk source trees. To skip applying
+patches, use the tcltk-wish.sh --no-patches commandline option. See
+mac/patches/README.txt for more info.
 
 ## Supplementary Build Scripts
 
--- a/portaudio/update.sh
+++ b/portaudio/update.sh
@@ -1,4 +1,4 @@
-##! /bin/bash
+#! /bin/bash
 #
 # this script automatically downloads and copies the portaudio library for Pd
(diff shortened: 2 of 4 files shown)


danomatika:
@millerpuckette This should be the final fix for the portaudio stuff. Hopefully @porres can confirm.

danomatika:
>So If I'm correct, this is ready to merge into master, and we can ignore PR 490.

PR #490 is no longer needed, the actual fix and relevant stuff has been moved here.

>I'm unclear as to whether the current portaudio source in the master branch is exactly pa_stable_v190600_20161030.tgz or whether any previously merged PRs have changed it - but I can verify that tonight. (If there are changes to the portaudio code I guess they need to be proposed upstream).

It's the current git master, (see [commit log here](https://app.assembla.com/spaces/portaudio/git/commits/list/?type=branch&page=3)). The only real differences since then are to apis and files we don't actually use for Pd. You can update portaudio to the latest stable release commit using the new script:

    cd portaudio
    ./update 396fe4b6699ae929d3a685b3ef8a7e97396139a4

Details are in the `portadio/README.txt` and the script header. The portmidi script works in the same manner, so it should be easy to update portaudio and for people to try new versions if they are build for themselves.

Portaudio also has a new header which contains the git revision hash: `portaudio/portaudio/src/common/pa_gitrevision.h`.

danomatika:
> (If there are changes to the portaudio code I guess they need to be proposed upstream).

Right. Luckily the changes I was using are now no longer needed, although I did forward the relevant patch upstream if they want to use parts of it.

The portmidi patch is there for now until upstream applies it to their repo. I got confirmation about this last year, but still waiting for them to push it. The main point for me is to separate any custom changes form the lib sources so we are much freer in updating the libs without losing whatever fixes we've made that aren't upstream yet.

millerpuckette:
Excellent.  I need to do a bit more work on my end to compile my PPC version (which the current portaudio code broke) but I can fix that locally.  Meanwhile everything else seems to be working.

danomatika:
In that case, you could probably use the script with an earlier commit. I just added d9d7fe664f999df0d0e49ffaf74b5c03c4bb9e07 to the `update/0.49` branch which should work for older versions of portaudio which don't have a script they added in the last release.

millerpuckette:
Maybe will try that later - ATM I'm using an ad hoc script for both PPC/Mac and for 32bit windows; haven't yet managed to migrate those to the scripts.

danomatika:
For completeness, I've tested and device unplug handling is working on macOS.

PR #497, opened 2018-09-24
0.49-0test5 updates

danomatika:
Hopefully this is the last round of cleanup for 0.49:

* [x] remove building pa_x86_plain_converters.c form portaudio, it wasn't used before and it breaks the windows cross compile (and possible the windows 32 bit build), I mistakingly re-added it with the portaudio cleanup
* [x] remove leftover portaudio configure stuff, not used since pre 0.47
* [x] portaudio update script should check if the update_gitrevision.sh exists, this should keep the script from halting with versions of portaudio earlier than the last stable release


millerpuckette:
merged

PR #499, opened 2018-09-25
changed clone error message

mxa:
there was an error message inside clone which still said it can't create a subpatch.


millerpuckette:
I think that even if some people think "subpatch" shouldn't include "abstraction", it's still more clear to say "subpatch" here than abstraction, so that users who don't know the exact terminology can still understand the messsage.

mxa:
Just checked the documentation, and indeed this seems to be in line with the definition of "subpatches": "Pd offers two mechanisms for making subpatches, called 'one-off subpatches' and 'abstractions.' " ($2.7.) Subpatch thus is the umbrella term for abstractions and subpatches of the pd-space-something kind of subpatches.

PR #507, opened 2018-10-04
'./configure --enable-jack' warns and advices if jack-dev is not found

Ant1r:
if jack development file are not found, configure then prints:
```
configure: WARNING: jack development files not found, skipping jack
    (install libjack-dev or libjack-jackd2-dev, then re-configure)

```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -339,7 +339,9 @@ AC_ARG_ENABLE([jack],
 if [ test x$jack_framework != xyes ] && [ test x$jack = xyes ] ; then
     AC_CHECK_LIB([rt], [shm_open], [LIBS="$LIBS -lrt"])
     AC_CHECK_LIB([jack], [jack_set_xrun_callback], [JACK_LIBS="-ljack" ; jack=xrun])
-    AC_CHECK_LIB([jack], [jack_set_error_function], [JACK_LIBS="-ljack" ; jack=yes], [jack=no])
+    AC_CHECK_LIB([jack], [jack_set_error_function], [JACK_LIBS="-ljack" ; jack=yes],
+        [AC_MSG_WARN([jack development files not found, skipping jack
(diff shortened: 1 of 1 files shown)


umlaeute:
these instructions are only valid of Debian-derivatives.
what about fedora, arch, gentoo,....?

Ant1r:
oh... I tried not to be too much specific (e.g i didn't mention apt-get...) but now I realize that I failed to.
What would you think better?
- not doing anything (removing this PR...)
- removing reference of actual deb packages names in the warning (generic message)
- adding instructions for some other distros ("exhaustive" message)

danomatika:
This is already in the documentation. Also, the configure confirmation print will list the audio APIs it's building for, so a missing "JACK" should be a second sign.

danomatika:
>these instructions are only valid of Debian-derivatives.
>what about fedora, arch, gentoo,....?

...or macOS. I don't think we want to get into too many platform specific error messages, if we can avoid them.

Ant1r:
> This is already in the documentation (...)

yes; the idea here is to give more help to an inattentive user trying to compile by him/herself, about how to solve jack not working (even though the user knows that libjack is installed).

danomatika:
Ok ok. I'll add this but for *all* relevant header checks.

Ant1r:
> all relevant header checks

yep I thought about it. I can try adding this (and stick to generic messages).

danomatika:
Don't worry about it. I'm doing it now using your suggestion as a starting point.

danomatika:
Ok, this is implemented by #508

Ant1r:
ok cool!

danomatika:
Also, my making a separate PR is not meant to diminish your contribution. For a smaller fix like this, it seemed easier to bring it in manually as I wanted to do some additional stuff anyway.

Ant1r:
no worry! I'm happy to help in any way I can.

PR #508, opened 2018-10-05
bugfixes & general development

danomatika:
This is a PR/branch for small, uncontested bugfixes & updates for the Pd 0.50 development cycle. 

* [x] configure: print warning message when JACK not found
* [x] configure: print error message when ALSA not found on Linux (probably requires setting alsa=no as default on other platforms)
* [x] warning fixes, as usual
* [x] spelling fixes in German translation
* [x] undo fixes/updates
* [x] documentation clarification regarding universal & fat libs
* [x] .c syntax consistency for `(void)`
* [x] fix [print -n] ":" header for floats
* [x] try matching full audio devname string before checking initial characters  
* [x] fix number2 number not consistently layered above label
* [x] change tidy up post level to debug
* [ ] more checks for memory management?
* [ ] ...?


umlaeute:
cool.

small sidenote: if those updates are small-enough and only "bugfixes", they probably could go into `0.49-x` (so `update/0.50` might be a misnomer)

configure.ac
```diff
@@ -339,7 +341,9 @@ AC_ARG_ENABLE([jack],
 if [ test x$jack_framework != xyes ] && [ test x$jack = xyes ] ; then
     AC_CHECK_LIB([rt], [shm_open], [LIBS="$LIBS -lrt"])
     AC_CHECK_LIB([jack], [jack_set_xrun_callback], [JACK_LIBS="-ljack" ; jack=xrun])
-    AC_CHECK_LIB([jack], [jack_set_error_function], [JACK_LIBS="-ljack" ; jack=yes], [jack=no])
+    AC_CHECK_LIB([jack], [jack_set_error_function], [JACK_LIBS="-ljack" ; jack=yes],
+        [AC_MSG_WARN([JACK development files not found... skipping (See INSTALL.txt)])
+        jack=no])
 fi
```
umlaeute: i'm afraid, that people (at least those, who don't read `INSTALL` files...)  will not see this information.
probably the warning should be emitted at the very end of the configure-script (*after* the summary at the end). with something like:

~~~
 AS_IF([test x$enable_jack = xyes -a x$jack != xyes],
    AC_MSG_WARN([JACK requested but not found. (See INSTALL.txt)])
)
~~~

danomatika: Well, the last example from the email thread included copy/paste from the configure output, so it comes *some* people would see it.

umlaeute: cool. so shall we remove the summary at the end of `configure` (because those who read it, can also just scan through the output of configure)?

Ant1r: > remove the summary at the end of `configure`

please don't ;-)
moving the warning(s) to the end of the summary seems sensible to me.

danomatika:
The warnings on the bottom do work, although I thought printing the summary and seeing the settings would be helpful enough. I suppose people can't scroll up to confirm? I guess having warning messages in there twice will at least be useful in the end...

Spacechild1:
@danomatika you can revert 8304bde, the portmidi code misses some #ifdefs. I have a working patch which solves Cyrille's problem: https://github.com/pure-data/pure-data/issues/581#issuecomment-494787205

I'll make a PR to properly enable/disable the rate limit per compile time flag and then we only need to discuss what should be default behavior.

millerpuckette:
I got into a twist merging this - I believe there's no more need for
portmidi/patches/mac_override_limit_rate.patch
correct?

Spacechild1:
Hi Miller, this should have been reverted with e7403e8 
The actual PR for the MIDI output fix is here: https://github.com/pure-data/pure-data/pull/641

millerpuckette:
Cool - hopefully I've got it working now.  (But I won't be surprised if I messed it up somehow.)

umlaeute:
@millerpuckette so can this branch be merged into `master`?
git should handle any previously applied commits just fine.
(I did a local `git merge update/0.50` and didn't spot any problems).

millerpuckette:
I think it's already merged.  (That would explain why you can merge it again w/o conflicts :)  OTOH, my gitcraft is highly error-prone, so let me know if I'm wrong.

millerpuckette:
Merged

Issue #509, opened 2018-10-09
[feature request] allow to programmatically set temporary search-path


umlaeute:
since Pd-0.48 we have the `add-to-path` message, that allows us to add a new search-path via a message to Pd.
this was originally intended for `deken`, but is used by external libraries as well (e.g. `cyclone`, `Gem`), to add library's path to Pd's search-path (for additional objects, like abstractions)

the message takes an option, that allows to immediately save the paths to the preference files.

However, when using from a library, it is often preferably if the dynamically added search path would not be saved to the preferences at all (to avoid duplicate adding; or multiple automatic additions of different versions).

So I'd suggest to use the "saveflag" to declare a non-saveable adding: e.g. if `saveit==-1` then the path is added to `st_temppath` instead of `st_searchpath`

danomatika:
sounds good to me

danomatika:
Is this fixed in 0.50? If not, let's move it to the 0.51 milestone.

umlaeute:
moved the milestone to the corresponding PR #510

PR #510, opened 2018-10-09
search path fixes/features

umlaeute:
### feature: extend `add-to-path` message to allow adding temporary paths
by passing "-1" as the `saveit` flag, the path will only added temporarily (using `st_temppath`), allowing libraries to dynamically add search-paths only as long as the library is loaded (during runtime of the Pd-process)

Closes: https://github.com/pure-data/pure-data/issues/509

### bugfix: search paths from the cmdline (`-path`) *before* searching paths from the preferences

ephemeral settings (startup flags) should have precedence over persistent settings (preferences).

Closes: https://github.com/pure-data/pure-data/issues/1180


danomatika:
Seems fine to me. Looks like this will work with https://github.com/pure-data/pure-data/blob/master/tcl/wheredoesthisgo.tcl#L61 as well.

PR #511, opened 2018-10-10
Windows: reset .pd files to *this* Pd

Lucarda:
This PR adds a new entry on the menu preferences (only in Windows): "Reset .pd files to this Pd".

It writes to the Registry that .pd files should open with the current Pd.

It is handy in 2 cases :

- When using a Zip version.

- When you have installed both Arch versions and you decided to change which one is the default.

It auto detects whether we are using `wish85.exe` or `wish86.exe`.

Needs permision from the user to be "Run as Administrator" to access the root of the Registry (this is where the installer had previously written).


tcl/pd_menus.tcl
```diff
@@ -575,6 +575,44 @@ proc ::pd_menus::create_preferences_menu {mymenu} {
...
+		else { set wishN "wish85" }
+		
+		package require registry 
+		
+		set detect2 $::sys_libdir
+		set dirrep [regsub -all "/" ${detect2} "\\"]
+		
+		if {[catch {registry set HKEY_CLASSES_ROOT\\.pd "" "PureData"}  errmsg ]  } {
+		::pdwindow::error "${errmsg} \n"
+		::pdwindow::post "This error is normal. To do this we need to start Pd with \"Administrator Rights\".\nSave your work and close Pd.\nThen right-click on Pd's icon and select \"Run as Administrator\".\nThen try again to \"reset .pd files\".\n"
```
umlaeute: i wouldn't call it "normal".
either it is an error, than it is an error. or it is normal, then the error message should be suppressed.
instead i'd use something like:

> "Administrator Permissions" are required to set the file associations.
> Save your work and close Pd.
> Then right-click on Pd's icon and select "Run as Administrator".
> Finally try again to "Open .pd files with this Pd."

tcl/pd_menus.tcl
```diff
@@ -575,6 +575,44 @@ proc ::pd_menus::create_preferences_menu {mymenu} {
         -command {::pd_menus::loadpreferences}
     $mymenu add command -label [_ "Forget All..."] \
         -command {::pd_menus::forgetpreferences}
+		
+	switch -- $::windowingsystem {
+       
+	   "win32" {
+	
+    $mymenu add command -label [_ "Reset .pd files to *this* Pd"] \
```
umlaeute: I'm not sure whether this sentence makes sense to the 
Probably this is better:
> Always open .pd files with *this* installation of Pd."

This is a bit long for a menu-entry.
A better place might be a general purpose perferences-window.

tcl/pd_menus.tcl
```diff
@@ -575,6 +575,44 @@ proc ::pd_menus::create_preferences_menu {mymenu} {
...
+        -command {
```
umlaeute: could you refactor the command into a `proc`?
So it could be re-used  from various places (e.g. if we want to have this option in a separate pref-window)

tcl/pd_menus.tcl
```diff
@@ -575,6 +575,44 @@ proc ::pd_menus::create_preferences_menu {mymenu} {
...
+		
+	switch -- $::windowingsystem {
+       
+	   "win32" {
+	
+    $mymenu add command -label [_ "Reset .pd files to *this* Pd"] \
+        -command { 
+		
+		if {[file exists "$::sys_libdir/bin/wish86.exe"] == 1} { set wishN "wish86" } \
+		else { set wishN "wish85" }
```
umlaeute: hmm. i think `[info nameofexecutable]` will give you the full name of the wish-executable.
no need to second-guess.

tcl/pd_menus.tcl
Lucarda: > This is a bit long for a menu-entry.
A better place might be a general purpose perferences-window.

We don't have that yet. :)

Lucarda:
I first do

```
set detect2 $::sys_libdir
set dirrep [regsub -all "/" ${detect2} "\\"]
```


I need this to replace all forward slashes.  This is also needed for calling pd-gui.tcl.

Lucarda:
The :
`set wishN [file tail [info nameofexecutable]]`

is done at first. the `tail `strips the path and leaves only the filename.

Lucarda:
> not directly related, but speaking of a "hack" here made me re-check my NSI-fixes and i noticed that I hardcoded the wish.exe to wish85.exe : urgh. does this currently break anything? is this going to break anything in the near future?

Can you do some detection for wish**.exe as you did in:

https://github.com/pure-data/pure-data/blob/master/msw/build-nsi.sh#L105

we can pass a new variable to pd-nsi.

umlaeute:
i understand the code, but i don't understand the reasoning. it seems overly complicated (it might be required for the NSIS-installer, but i think there must be a better way to do it in the Pd-GUI)

Lucarda:
Which part is over-complicated?

What you have in mind?

Putting aside the NSIS installer that is not what this PR is about.

umlaeute:
the overcomplication i see is:
- get the full path of the wish-executable with `info nameofexecutable` (although in the "wrong" format, with slashes instead of backslashes)
- extract the filename from the full path
- from somewhere else, get the path to the Pd-installation (although in the "wrong" format)
   - the Pd-installation folder is where the Pd-implementation lives (e.g. `tcl/`). On Windows it **happens** to also contain the `wish.exe`
- convert the path to the Pd-installation to the format you want (replacing slashes with backslashes)
- construct the path to the wish-executable from the Pd-installation dir and the filename f the wish-executable.

this makes an assumption, which *happens* to be true in the current implementation: the `wish.exe` resides in a well-defined place in `$::sys_libdir`.
the code breaks as soon as the assumption is no longer true (e.g. because you want to use a different wish interpreter).
i guess your answer to this is "then just fix the code", but that's just another step that could be forgotten or accidentally broken.

a much simpler way to achieve the same result, without relying on implicit knowledge of the Pd-installation layout would be:
- get the full path of the wish-executable with `info nameofexecutable` (although in the "wrong" format, with slashes instead of backslashes)
- convert the full path to the format you want

(and yes, the NSIS-installer is a totally different beast; i just keep mentioning it here as it deals with essentially the same problem, and it shouldn't be forgotten)

umlaeute:
- the full path to wish can be obtained via `[info nameofexecutable]`
- the full path to `pd-gui.tcl` is already stored in `$::pdgui::scriptname` (and could be obtained with `[info script]` within the *pd-gui.tcl* itsel.
- the path to the pd-binary is already calculated in `pd_gui.tcl:main`. it lacks the suffix (`.exe`, `.com`, ...) and is just a local variable (so it needs to be exported) that is only calculated if the GUI starts the core (so this would need to be done regardless of start-mode), but it's basically there and keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself).

Lucarda:
Cool, I tested your updates and everything is working nice.

>(and yes, the NSIS-installer is a totally different beast; i just keep mentioning it here as it deals with essentially the same problem, and it shouldn't be forgotten)

I made a PR #517 for the NSIS-installer.

PR #512, opened 2018-10-12
fixed missing letter in german translation

mxa:


po/de.po
```diff
@@ -271,7 +271,7 @@ msgid "Delay (msec):"
 msgstr "Verzögerung (msec):"
 
 msgid "Block size:"
-msgstr "Bocklänge:"
```
umlaeute: oh, that was a nice one

PR #517, opened 2018-10-20
Msw installer updates

Lucarda:
This PR finds the `wish**.exe` that we are currently using when building the Pd-installer so that `wish85.exe` does not have to be hard-coded in `pd.nsi`. 
This allows to build the installer with whatever wish**.exe we are using (`wish85.exe`, `wish86.exe`, ...).


umlaeute:
superseded by c0773945fcd9af3645799f2de94e35970503b25f

PR #521, opened 2018-11-09
changed German translations for strings where there was concensus…

mxa:
…on the Pd-list


danomatika:
Since this was integrated into PR #508, we can probably close this PR.

umlaeute:
@danomatika afaict this ticket will be automatically closed once #508 will be merged into matser.

Issue #522, opened 2018-11-10
large -npts in [sigmund~] leads to stack corruption


Spacechild1:
on my machine (Thinkpad, Win7, Pd 0.49 32-bit) this already happens at -npts > 32768 because of

`t_float *bigbuf = alloca(sizeof (t_float ) * (2*NEGBINS + 6*npts));`

`alloca` alone is dangerous but using `malloc` after a certain size isn't good either. I would suggest pre-allocating the buffer in the object and resize it when needed (e.g. on a "npts" message). will do a PR.

related to https://github.com/pure-data/pure-data/issues/68

PR #524, opened 2018-11-18
fix warnings and type inconsistencies

nicolasdanet:
In order to build the sigmund~.c file as CPP in "Spaghettis" i changed few things:

    1. Remove unused (or unreachable) values.
    2. Explicit some casts.
    3. And moreover change t_sample to t_float for buffering.

Note that in my version t_sample are single precision whereas t_float are double precision.

That changes could be incorporated in #522

### DIFF
--- a/extra/sigmund~/sigmund~.c
+++ b/extra/sigmund~/sigmund~.c
@@ -251,7 +251,7 @@ static void sigmund_getrawpeaks(int npts, t_float *insamps,
     int peakcount = 0;
     t_float *fp1, *fp2;
     t_float *rawreal, *rawimag, *maskbuf, *powbuf;
-    t_float *bigbuf = alloca(sizeof (t_float ) * (2*NEGBINS + 6*npts));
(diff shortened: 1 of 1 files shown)


extra/sigmund~/sigmund~.c
```diff
@@ -838,7 +838,7 @@ typedef struct _sigmund
     int x_npts;         /* number of points in analysis window */
     int x_npeak;        /* number of peaks to find */
     int x_loud;         /* debug level */
-    t_sample *x_inbuf;  /* input buffer */
+    t_float *x_inbuf;  /* input buffer */
```
umlaeute: this doesn't look right.
i think it should be done the other way around (more `t_sample` rather than less)

nicolasdanet: I'm not agree but it is up to you (feel free to close this PR).

umlaeute:
why did you close this?

nicolasdanet:
C++ support seems already done in #597 ; The other modifications was required only for me and i don't use that file anymore.

PR #526, opened 2018-12-12
German translation fixes added

ritsch:
Added new translation   a consens at the Pd List,

https://lists.puredata.info/pipermail/pd-list/2018-11/123843.html

and some follow ups. 
This was forked from https://github.com/chairaudio/pure-data.git with the branch from the same name
and edited and replaces his pull request.


danomatika:
Pinging @mxa, @umlaeute, & @Spacechild1. If this is good, we can merge it into a new 0.50-updates develop branch.

ritsch:
Am Mittwoch, 19. Dezember 2018, 18:24:42 CET schrieb Spacechild1:
> I like the translation! I'll fix some minor spelling issues tonight (mostly
> about missing hyphens).
Yes please and thanks for looking into it.

mfg
 winfried


-- 
Ritsch, Winfried, Ao.Univ.Prof. Dipl.-Ing.
Institut 17 Elektronische Musik und Akustik
Inffeldgasse 10/III,    8010 Graz, Austria
E-Mail:         ritsch@iem.at  - http://iem.at/~ritsch - mobil  ++436642439369

mxa:
I do believe some of the not translated terms should have proper translations, but I think in general this PR will make Pd easier for users with German (and English) knowledge.

ritsch:
> I do believe some of the not translated terms should have proper translations, but I think in general this PR will make Pd easier for users with German (and English) knowledge.

Thanks for the objection,

I see the cons, but considered also the pros: Someone without a little knowledge in English anyway will have a hard time in CM especially at the level of Pd, others which just want to edutainment can use the more beautified versions like Purr Data and others, like some composers or musicians  with little knowledge of programming and Computer and the terms within these, will be lead in a wrong direction, since not learning the Puredata as CM-language and if you change to an English version you will have some confusion.

Anyway more translation of sub-menus, warnings etc would also has to be made, which are not all included.


Another problem is if Purr Data and friends will pick up this solution or make their own translations ?
I hope I started with a rough consensus.

danomatika:
>Another problem is if Purr Data and friends will pick up this solution or make their own translations ?
I hope I started with a rough consensus.

gettext is a mature, well-supported project so I'd hope Purr Data would be using it already. These translations should be portable and the only issue would be differences and additions between the two projects.

umlaeute:
@mxa 
> I do believe some of the not translated terms should have proper translations

could you elaborate on that: which terms do you think ought to be translated that are currently not?

mxa:
@umlaeute @ritsch I think the topic has been discussed on the mailinglist sufficiently. In general I think this PR will make Pd better, so it should be merged.

That said, and since you are asking: I believe translations should make an effort to find an equivalent term in the translated language. That the rest of the program is not (yet) translated is IMO not a valid argument, neither is that computer science is dominated by English. If the user wanted to use Pd in English, she/he can still do that, and maybe eventually help files might be translated.

danomatika:
>If the user wanted to use Pd in English, she/he can still do that

I think this point reiterates the need for a GUI-only language setting. @umlaeute maybe we can roll that into a general notebook-styles settings push for 0.50?

umlaeute:
merged into `updates/0.50` with 5f2a0741dd395db6946f2aa28eaa2374f0cb0a33

PR #529, opened 2018-12-13
Memory handling improvements

eldruin:
I made some improvements with regard to memory handling


danomatika:
@umlaeute If this is accepted, we could also merge it into the update/0.50 branch.

PR #530, opened 2018-12-14
Export obj_findsignalscalar function

eldruin:
This exports the `obj_findsignalscalar` function.
This function is currently used by some plugins in [cyclone](https://github.com/porres/pd-cyclone). See [here](https://github.com/porres/pd-cyclone/search?q=obj_findsignalscalar&unscoped_q=obj_findsignalscalar).

I was not sure where to put it in the header file. Please let me know if I should put it somewhere else.

### DIFF
--- a/src/d_ugen.c
+++ b/src/d_ugen.c
@@ -17,7 +17,6 @@
 #include <stdarg.h>
 
 extern t_class *vinlet_class, *voutlet_class, *canvas_class, *text_class;
-t_float *obj_findsignalscalar(t_object *x, int m);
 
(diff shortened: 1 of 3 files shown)


umlaeute:
i don't think that `m_pd.h` (the "public public" API) is the appropriate place for this.

practically all other `obj_*` functions are exported in `m_imp.h` (the "public private" API - or was it the "private public" API?)

umlaeute:
well, cyclone already depends on the "private public" API in `m_imp.h`, e.g. `obj_starttraverseoutlet`

Issue #546, opened 2019-01-29
[feature request] -resize flag for array read message


fdch:
When you send a `[read myfile.txt(` message to an array, myfile.txt is read up to the size of the array. 

Although an abstraction can be implemented (using `[text]`, `[text size]`, `[array size]`, etc), I think it is much more efficient to have a `-resize` flag for that purpose, like when reading with `[soundfiler]`

A more extreme case would be just to consider resizing as a default (like `[text]` does), which would (perhaps) remove the need for a `-resize` flag from `[soundfiler]`.

fdch:
An example of an abstraction that resizes arrays can be found [here](https://github.com/fdch/fd_lib/tree/master/fd_read-resize)

fdch:
Hi! 
I have a working solution for this request, but I am not really sure where to make the PR. 

Basically, I followed the wiki, but I am not sure now if I have to make the PR directly to master or to some other branch that might be out there. The screenshot shows where I stopped:
![screen shot 2019-02-18 at 03 16 05](https://user-images.githubusercontent.com/13335792/52924184-e29c1180-332b-11e9-8696-0815372f241c.png)

It is officially the first time *ever* I am attempting to add some actual code to Pd. So, please, let me know if I missed some stuff!

The branch on my fork is here:
https://github.com/fdch/pure-data/tree/array-resize

Lucarda:
Hi,
The Pic is Ok, you should be able to merge to pure-data/master. 

:)

fdch:
Thanks! Will do!! here's the test patch I'm using: 
[arrayresize-test.zip](https://github.com/pure-data/pure-data/files/2873862/arrayresize-test.zip)

fdch:
This issue was addressed here https://github.com/pure-data/pure-data/pull/561

PR #561, opened 2019-02-18
'-resize' flag for reading into arrays

fdch:
Hi, 

This adds a flag to `garray_read()` message so that you can load and resize an array. The default behavior (reading up to the size of the array) is still available without the flag. With the flag you can adjust the array size to the size of the input,  assuming input files are in the same format as Pd's `garray_write()`

This is a working solution for https://github.com/pure-data/pure-data/issues/546 (there is also a test patch [over there](https://github.com/pure-data/pure-data/files/2873862/arrayresize-test.zip))

Please, let me know if you see anything that I missed (it is my first PR :).

### DIFF
--- a/doc/5.reference/array-object-help.pd
+++ b/doc/5.reference/array-object-help.pd
@@ -3,53 +3,56 @@
 #N struct array-help-struct2 float x float y array z element-struct2
 ;
 #N struct element-struct2 float x float y float w;
-#N canvas 703 157 694 542 12;
+#N canvas 586 157 694 542 12;
 #X obj 208 496 list;
 #X text 34 496 see also:;
 #X text 485 171 (click for details:), f 11;
-#N canvas 490 70 900 759 define 0;
+#N canvas 380 19 900 759 define 0;
 #X text 332 562 creation arguments:;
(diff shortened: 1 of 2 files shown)


Lucarda:
I've tested this PR on Windows (64bit Pd) as of ebf2097.
Is working nice.

:)

Lucarda:
Documentation?

Can you include it on this PR?

fdch:
@Lucarda there goes the documentation in 2b1a13c thanks for testing :)

Lucarda:
Cool.

:)

millerpuckette:
I looked at this, and realized that it made better sense to do this with the soundfiler object - so I went ahead and did that.

The PR looked OK - I had formatting beefs (lines ending in empty space; non-pd-like indenting) but the code itself looked fine.  If I were to rewrite array_read now, though, I'd use a "binbuf" instead of "scanf".  (the original routine dates to before 2000 - yikes.)

fdch:
Thanks for going over this and for implementing the request! I am glad this request made it through even if it went another way. 

I guess I should close this pr, then! Looking forward for the new release!!

PR #563, opened 2019-02-23
enable zooming for datastructures

jyg:
Hi, 
1)  I modified _g_scalar.c_ and _g_template.c_ in order to enable datastructures to be zoomed.
I tested it on the datastructures tutorial of puredata help. Seems to work fine.
2) Line 2361 in _g_template.c_, I changed CURSOR_RUNMODE_CLICKME in  CURSOR_RUNMODE_ADDPOINT in order to see the "mouse_over effect" on my computer (linux).

3) Furthermore, I would like to discuss about lines 274 & 357 in _g_scalar.c_, that seem buggy to me. the output list must have 3 elements : a _gpointer_ and 2 coordinates. Instead of this, for the click event notification, there is only the pointer + the [y] coordinate , because the [x] coordinate has been overwritten by a [gpointer] in the _template_notifyforscalar_ function. I don't know if it is better to correct the "click" message syntax or to add a new message ("clicked") with the new syntax.

### DIFF
--- a/src/g_scalar.c
+++ b/src/g_scalar.c
@@ -262,16 +262,16 @@ static void scalar_displace(t_gobj *z, t_glist *glist, int dx, int dy)
         goty = 0;
     if (gotx)
         *(t_float *)(((char *)(x->sc_vec)) + xonset) +=
-            dx * (glist_pixelstox(glist, 1) - glist_pixelstox(glist, 0));
+             dx * (glist_pixelstox(glist, 1) - glist_pixelstox(glist, 0))/ glist->gl_zoom ;
     if (goty)
         *(t_float *)(((char *)(x->sc_vec)) + yonset) +=
-            dy * (glist_pixelstoy(glist, 1) - glist_pixelstoy(glist, 0));
+            dy * (glist_pixelstoy(glist, 1) - glist_pixelstoy(glist, 0)) / glist->gl_zoom ;
     gpointer_init(&gp);
     gpointer_setglist(&gp, glist, x);
     SETPOINTER(&at[0], &gp);
     SETFLOAT(&at[1], (t_float)dx);
     SETFLOAT(&at[2], (t_float)dy);
-    template_notify(template, gensym("displace"), 2, at);
+    template_notify(template, gensym("displace"), 3, at);  // '3' instead of '2'
     scalar_redraw(x, glist);
 }
 
@@ -348,7 +348,7 @@ int scalar_doclick(t_word *data, t_template *template, t_scalar *sc,
     int hit = 0;
     t_canvas *templatecanvas = template_findcanvas(template);
     t_gobj *y;
-    t_atom at[2];
+    t_atom at[3];
     t_float basex = template_getfloat(template, gensym("x"), data, 0);
     t_float basey = template_getfloat(template, gensym("y"), data, 0);
     SETFLOAT(at, basex + xloc);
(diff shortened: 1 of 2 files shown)


Spacechild1:
> 3\. Instead of this, for the click event notification, there is only the pointer + the [y] coordinate

actually, I've always been wondering what that single number after the gpointer was but never cared to look in the source :-) I think you can just fix this bug and insert the missing x-ccordinate. I don't see how adding a third float to the list could break existing patches since, apparantly, no one has been using the float after the gpointer anyway. thanks for looking into this!

millerpuckette:
It's going to take some time for me to figure out what to do for zooming.  In testing I noticed that data structs inside GOPs already zoom correctly (no idea why) but misbehave in other ways.

Meanwhile I'm baffled why I ever thought one should zoom "text"s by changing their te_xpix and te_ypix values... there must have been some reason for that, but I now think the whole zoom thing is better handled inside glist_xtopix(), etc. instead of having direct references to gl_zoom all over the place.  (And somehow that is easier done if the te_xpix field didn't change with zooming.)

jyg:
I tried some modifications changing only g_graph.c (xtopixels, ytopixels, pixelstox, pixelstoy fonctions) and g_template.c (lines 2034 and 2624), and using only glist_getzoom() function. There is no need to modify g_scalar.c as I did before.
Zooming with datastructures seems to work fine, except for graphs inside old, proportional-style GOP.  
When I create an array and open the graph, zooming does not work properly (I get an y-offset I cannot figure why). 
But In this use-case, I would say that zooming does not have so much importance, because one can zoom by enlarging the window. Or am I wrong ?
[zoomingDS.zip](https://github.com/pure-data/pure-data/files/3452564/zoomingDS.zip)

millerpuckette:
I also tried to fix this... see "new-zoom-test" branch.  Working OK so far for me.  I do want to address the mouseover and third argument problems you mentioned above.

jyg:
Hi, I tried "new-zoom-test" branch a few days ago, but I cannot see it now. Did you suppress it ? Last time I cloned it, I noticed some issues, but I don't know if you fixed them, neither if it's time or not for giving you feedback about this.

millerpuckette:
I ended up merging it -- just grab 'master' now.  If you can find any problems
(and I bet you can) I'd be grateful to hear it.

cheers
Miller

jyg:
I just grabbed from 'master' branch, but I cannot find any of modifications  you did regarding zooming with D.S.

millerpuckette:
Hmm... somehow I lost all the changes I made!  I'll try to figure out what
I did.

M

millerpuckette:
I seem to have orphaned the changes somehow (probably had them out on a
branch that I later deleted) - "git reflog --all" turned them up.  I've
pushed them (to master) and now should go check to see if I fumbled anything
else along with them :)

M

jyg:
I get an issue when I open a gop patch. I get a red plain oval instead or red rectangle.
g_canvas.c, line 655, needs another argument (x->gl_zoom) for red gop rectangle drawing : 
`glist_getcanvas(x), x1, y1, x1, y2, x2, y2, x2, y1, x1, y1, x->gl_zoom);`

porres:
weird, all seems fine here (macOS 10.14.5)

jyg:
There is another issue I can see for which I can propose a fix :
In /doc/4.data.structures/11.array.controls.pd, when you drag a point of data11 window with zoomx2, the motion along x-axis is twice faster than pointer motion. The culprit is line 2635 in g_template.c, where we have to add a `TEMPLATE->array_motion_xperpix`factor.
so new value is ` TEMPLATE->array_motion_template, thisword, xwas + dx *  TEMPLATE->array_motion_xperpix, 1);`

porres:
so, has this been merged? can we close this?

jyg:
Hi Miller, 
I don't know if this is the good place to continue my comments. May be better on the pd-list ?
However, when I create a new patch, with zoom = x2, when I add a new object / comment / cursor etc... in this patch, this object is not created at the right place, but at something like [x*2, y*2].
May I suggest some changes in g_text.c , in the function `static void canvas_howputnew` ?
Please find attached my suggestion (signaled with /////zoom_fix).
[g_text.c.zip](https://github.com/pure-data/pure-data/files/3497661/g_text.c.zip)
regards,
J.Y.G.

jyg:
I closed this pull request and created a new one with some updated fix improvements : https://github.com/pure-data/pure-data/pull/733

PR #570, opened 2019-03-16
key, keyup, & keyname ignore events in edit mode

danomatika:
This PR adds an optional creation argument to [key], [keyup], & [keyname] to ignore events if canvas is in edit mode. Basically, this allows you to edit without triggering what you are editing.

This is creation argument is just a boolean using the first float:
- 0 : allow key events in both run & edit mode (default, current behavior)
- 1 : allow events in run mode, ignore in edit mode

### DIFF
--- a/doc/5.reference/key-help.pd
+++ b/doc/5.reference/key-help.pd
@@ -1,24 +1,30 @@
-#N canvas 146 45 546 288 12;
+#N canvas 146 45 509 365 12;
 #X obj 21 10 key;
-#X obj 54 9 keyup;
-#X obj 105 9 keyname;
+#X obj 54 10 keyup;
+#X obj 105 10 keyname;
 #X text 173 8 -- grab keyboard;
 #X obj 38 67 key;
 #X floatatom 38 95 3 0 0 0 - - -;
-#X floatatom 77 93 3 0 0 0 - - -;
+#X floatatom 77 95 3 0 0 0 - - -;
 #X obj 77 67 keyup;
-#X floatatom 128 93 3 0 0 0 - - -;
+#X floatatom 128 95 3 0 0 0 - - -;
 #X obj 128 67 keyname;
-#X symbolatom 172 94 10 0 0 0 - - -;
-#X text 280 262 updated for Pd version 0.32.;
+#X symbolatom 174 95 10 0 0 0 - - -;
 #X text 26 133 Key and keyup report the (system dependent) numbers
 of "printing" keys of the keyboard. Keyname gives the symbolic name
 of the key \, with a 1 or 0 if it's up or down \, and works with non-printing
 keys like shift or "F1".;
-#X text 18 200 Caveat -- this only works if Pd actually gets the key
+#X text 26 200 Caveat -- this only works if Pd actually gets the key
 events which can depend on the stacking order of windows and/or the
 pointer location \, depending on the system.;
+#X obj 38 262 key 1;
+#X floatatom 38 292 3 0 0 0 - - -;
+#X text 93 263 As of Pd 0.5 \, the key objects take an optional creation
+argument to ignore events when the patch is in edit mode. Set to 1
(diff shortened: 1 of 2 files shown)


porres:
that's awesome, thanks!

doc/5.reference/key-help.pd
```diff
@@ -1,24 +1,30 @@
...
+#X text 26 200 Caveat -- this only works if Pd actually gets the key
 events which can depend on the stacking order of windows and/or the
 pointer location \, depending on the system.;
+#X obj 38 262 key 1;
+#X floatatom 38 292 3 0 0 0 - - -;
+#X text 93 263 As of Pd 0.5 \, the key objects take an optional creation
+argument to ignore events when the patch is in edit mode. Set to 1
+to ignore and 0 to accept all events (default).;
+#X text 280 325 updated for Pd version 0.50.;
 #X connect 4 0 5 0;
```
umlaeute: i suggest to change to wording to give the flag a positive meaning. e.g. rather than ~~ignore events in edit mode~~ use `only output events in run mode` or similar.

also, afict the behaviour is a bit weird, as it will give you *all* events in the entire Pd instance, but only if the canvas containing the `[key]` object is not in edit mode (even if the events are created in a different patch that currently *is* in edit-mode).

i think it would be better to be able to first limit the receiving of key-events to the canvas where `keyup` lives in, and on top of that limit to run-mode.
unfortunately this requires more sever changes.

danomatika: > think it would be better to be able to first limit the receiving of key-events to the canvas where keyup lives in,

Yeah, that was my next step, probably related to the [mouse] discussion.

porres:
Hmm, how about if you just know if you're on edit mode or not?

I actually need this in an object and wrote an external for it, but it seems a useful functionality to be included in [pdcontrol]. Hence you could do something like this.

Screen Shot 2019-12-07 at 14 12 32

porres:
I could not compile and test this by the way

doc/5.reference/key-help.pd
porres: > think it would be better to be able to first limit the receiving of key-events to the canvas where keyup lives in,

Not sure I get it, but if that means thet [key] will only be operational if I'm working on its current patch window, then it seems bad as it'd break things. I have [key] objects in subpatches and I want them to be responsive if I'm not interacting with its window.

It seems like my proposal of handling this in something like [pdcontrol] is better, cause you could specify which canvas window to query for edit mode or not, so you can control and filter out [key] outputs if you have it in an abstraction and the parent patch is in edit mode.

Plus, it can also be useful for many other things, and even other externals, like filtering mouse events from cyclone/mousestate for instance.

porres:
Ok, it's done, I made a PR including updates to pdcontrol and put such functionality in it, see https://github.com/pure-data/pure-data/pull/820/commits/022303e8785d9a27639739f5c19f54839ae56791 

here's a documentation snapshot.

Screen Shot 2019-12-07 at 18 45 46

porres:
> Ok, it's done, I made a PR including updates to pdcontrol and put such functionality in it, see 022303e

Actually, see  https://github.com/pure-data/pure-data/pull/828/commits/f4816e9a31261c0efcfbf6ec62feb4be132e34f8 , which is from a new PR

danomatika:
Would this be better implemented using a flag instead of the simple boolean arg?

umlaeute:
i tend to agree with @porres' solution to factor out the edit-mode filtering into a separate object

danomatika:
Yup. I think this PR has been superseded. I'm fine with closing it as it was an initial stab for testing this approach.

umlaeute:
shall we close it then?

danomatika:
Closing in favor of #828.

umlaeute:
should we delete this branch?

danomatika:
Yeah, the branch can be deleted.

Issue #573, opened 2019-03-17
[intelligent patching] inserting objects broken for non-first connection


ch-nry:
Using <kbd>Control</kbd>-<kbd>k</kbd> to insert an object between 2 connected object, or to disconnect the middle object works only if the connection is made with the 1st outlet of the 1 object.
i.e : 
selecting this 3 object and <kbd>Control</kbd>-<kbd>k</kbd> works as expected : 
~~~
[t b b b]
  |  
  |    [f]
  |  
  [f]
~~~

selecting this 3 object and <kbd>Control</kbd>-<kbd>k</kbd> did not work as expected : 
~~~
[t b b]
      |  
      |    [f]
      |  
      [f]
~~~

umlaeute:
sidenote: i formatted your post to use `code`-sections (in order to get monospaced fonts), but i'm not sure about the indentation in the first example.

umlaeute:
to answer your original question: yep, there seems to be a bug.

in the meantime, you can (instead of selecting all three objects) select only the **connection** (between `[t b b]` and the lower `[f]`) and the **object to be inserted**, and hit <kbd>Control</kbd>-<kbd>k</kbd>

ch-nry:
yes, selecting only the connection (and the object to insert) does work.

Issue #547, opened 2019-01-29
[feature request] `esc` key binding to de-select (and avoid an extra click :) ?


fdch:
It would be nice to have something like `esc` key to de-select. Right now, doing `Select All` twice does toggle between selection and no selection. However, when there is a range of objects selected you have to either click on the canvas or toggle edit mode.

The reason I ask for this is that given the new feature for intelligent patching, I often have a group of selected objects, and I need to deselect them before making any connections, otherwise I end up with a bunch of unwanted connections.  But it would be nice to have just one way to do that with the keyboard-hand on the fly.

umlaeute:
the solution i have in mind for your underlying problem (intelligent patching getting in the way) is more like: having a modifier key to *bypass* intelligent patching.

as in:
- select multiple objects
- connect one of them to an unselected object -> all connect
- press <kbd>Shift</kbd> and keep it pressed
- connect one of the selected objects to an unselected object -> only the directly asked for connection is being made.

i'm mostly unsure about the correct key to temporarily disable intelligent patching.

and i wonder whether (de)selections should be undoable as well (but that would be another feature request).

fdch:
Imho, I would not suggest to bypass the feature, which is altogether quite amazing. 

For example, what other situation would one need to bypass it?

I probably have not used it enough to think of any. 

However, I do find often the need to deselect stuff, and intelligent patching was an example (and my own pd_user_baggage was getting in the way ;)

umlaeute:
i already have this feature request for being able to *temporarily* disable intelligent patching (in private @tmusil !!!).

i just don't see that much difference between "clicking on the canvas" and "hitting <kbd>ESC</kbd>"

fdch:
The difference as I see it is that, after selecting:

in one combined action I can:
-  move the pointer straight to the outlet while hitting `esc` (only once)
-  then just click the already deselected outlet.

Otherwise, I have to
-  click on the canvas
-  move the pointer to the outlet
-  click the deselected outlet

umlaeute:
you're cheating a bit by putting two actions into one line in your first example.
as i see it:
- hit <kbd>Esc</kbd> **or** click on the canvas
- move the pointer straight to the outlet
- click the outlet

there's not *so much* difference here.

the option i'm suggesting is:
- move the pointer straight to the outlet
- hit <kbd>Shift</kbd>, at the same time do:
- click the outlet

again, we have the same number of steps with comparable complexity.
[...]

Issue #501, opened 2018-09-26
[intelligent patching] connect single outlet with multiple inlets of the same object


porres:
say I want to connect an [osc~] into a [dac~] with 4 outputs.

I was expecting that control+K would not only connect to the first inlet, but also to the next ones if I kept using the shortcut.

how trivial is that or not?

Issue #574, opened 2019-03-17
[intelligent patching] staged undo


ch-nry:
Intelligent patching is very useful but it is sometime hard to get used to it. 
In order to facilitated the learning phase, I propose to split the "undo" in 2 action : the action automatically created, and the manual action.
I.E : create one object, create an other one, noticed that the 2 objects are unintentionally connected, cancel the automatic connection using <kbd>Control</kbd>-<kbd>z</kbd>. a 2nd <kbd>Control</kbd>-<kbd>z</kbd> will cancel the creation of the 2nd object. Curently only 1 <kbd>Control</kbd>-<kbd>z</kbd> undo the connection and the object creation.

Same principle should be used when multiples connection are created because multiples objects are selected when creating a connection.

This will make intelligent patching feature less annoying when not done on purpose.

umlaeute:
i'd rather have intelligent patching not step on your toes in the first place (see #568)

ch-nry:
since #568 have not been solved, this look like a good alternate solution (at least for me). 
Moreover, it would also solve the auto-connect feature added few years ago (I'm still not used to it):
When adding an object with a keyboard shortcut, one have to press 2 (or 3 key on french keyboard). adding an extra one is a bit to much for such common task.

umlaeute:
well, there's a branch with a solution for #568 /(i just created pr #575)

ch-nry:
PR #575  is good for intelligent patching, not for the old auto-patch feature. 
I still believe that staged undo is a good solution for the auto-patching function.

umlaeute:
i guess the biggest problem is, that with autoconnect enabled the flow is really:

1. create an empty object
2. connect the last object with the new (empty) object
3. have the user type some characters
4. re-instantiate the empty object with the content the user just typed.

so if we had (staged) Undo, this would
1. remove the newly created object and replace it with an empty object
2. remove the connection
3. remove the empty object

which is most likely not what you want. (i think you want to undo the connection without removing the created object).

ch-nry:
Hum, I understand.
the autoconnect could be changed to : 
- create an object
- have user type some characters
- re-instanciate the object
- connect with last object
But since my problem with intelligent patching is also that the new created object position is not at the mouse position, I guess I  should get used to : control 1, control Z, control 1 to create an object.

umlaeute:
changing the order makes the implementation much more complicated.

[...]

PR #575, opened 2019-03-18
intelligent patching amendments

### LINKED ISSUE #548: [feature request] making inserted object immediately editable (intelligent patching)
Once I insert an object in between two connected objects I usually need to edit it right after.

I think it would be useful to have the inserted object box text immediately available for edit. So you can:

-   Select the cord
-   Hit  <kbd>CMD + T </kbd>
-   Start typing the object name...

### LINKED ISSUE #648: triggerize position only correct for left-most iolets
e.g. triggerizing the connection in the following patch:

~~~
[unpack]
       |
       |
       |
[pack  ]
~~~

will yield something like:

~~~
[unpack]
      /
  +--+
 /
[t a ]
 \
  +-+
     \
[pack ]
~~~

rather than the expected:

~~~
[unpack]
       |
       [t a]
       |
[pack  ]
~~~

umlaeute:
some people are a bit unhappy with intelligent patching (#568, #574,...).

this PR is attempting to fix those concerns (and probably add new ones).
the PR is based on [`updates/0.50`](#508).

### fixes and features
 (this list is being updated as new things are added)
- [x] only do multi-connections when pressing a modifier-key (<kbd>Shift</kbd>); Closes:#568
- [x] allow to "deselect all" by pressing <kbd>Esc</kbd>; Closes: #547
- [x] fix insert-object for non-first connections; Closes: #573
- [x] use iolet-positions to better calculate the position of the inserted object; Closes #648 
- [x] made inserted object to be immediately editable; Closes #548 
- [x] have *Duplicate* and *Save-Abstractions* not interfere with the Copy-Paste Buffer; Closes: #644 #643 
- [x] allow fan-out of signal-outlets when connecting objects with <kbd>Ctrl</kbd>-<kbd>k</kbd>; Closes: #501 
- [x] use <kbd>Tab</kbd> to cycle the selected object/line (and <kbd>Shift</kbd>+<kbd>Tab</kbd> for reverse cycling)
- [x] while connecting, the <kbd>Tab</kbd> key cycles the outlets of the source object, resp. the inlets of the sink object (if there *is* a sink object under the mouse; inlet-cycling doesn't work on OSX yet)
- [x] fixed re-doing triggerized [trigger] (#649)
- [x] Regression: <kbd>Tab</kbd>-cycling broken in properties

### demos

https://vimeo.com/showcase/5289665/video/340437816

### experimental builds

those who cannot (or don't want to) compile Pd themselves, can try the experimental [CI-builds](https://git.iem.at/pd/pure-data/pipelines):
- [Win/32bit](https://git.iem.at/pd/pure-data/builds/artifacts/intellipatch2/download?job=w32_snapshot)
- [Win/64bit](https://git.iem.at/pd/pure-data/builds/artifacts/intellipatch2/download?job=w64_snapshot)
- [OSX/64bit](https://git.iem.at/pd/pure-data/builds/artifacts/intellipatch2/download?job=Darwin_snapshot)

these builds include some more features, including the recent changes to the `master` and the `update/0.50` branches.

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -81,6 +81,19 @@ script settings on the commandline:
     # build Pd using a system installed PortAudio
     ./configure --without-local-portaudio
 
+An important configure option for some platforms is --enable-universal which
+allows you to specify the desired architecture(s) when building Pd. For Intel
+and AMD processors, 32 bit is called "i386" and 64 bit is "x86_64". By default,
+Pd is built for the architecture of the current system, however you may want a
+32 bit Pd to work with existing 32 bit externals on a 64 bit system. You can
+override the defaults with --enable-universal:
+
+    # build 32 bit Pd
+    ./configure --enable-universal=i386
+
+    # build 64 bit Pd
+    ./configure --with-universal=x86_64
+
 The full list of available configuration options can printed by running:
 
     ./configure --help
@@ -184,18 +197,16 @@ need. For example, to install the autotools & gettext using Homebrew:
     brew link --force gettext
 
 By default, Pd is built for the current system architecture, usually 64 bit. If
-you want to override this you can use the --enable-universal configure option 
-which allows you to specify the desired architecture(s) when building Pd. For
-Intel/AMD processors, 32 bit is called "i386" and 64 bit is "x86_64". By
-default, Pd is built for the architecture of the current system, however you may
-want a 32 bit Pd to work with existing 32 bit externals on a 64 bit system. You
-can override the defaults with --enable-universal:
+you want to override this you can use the --enable-universal configure option,
+as mentioned in the main Autotools Build section. On macOS, running this option
+without arguments will build a "fat" Pd for all architectures supported by the
+compiler:
 
-    # build 32 bit Pd
-    ./configure --enable-universal=i386
+* macOS 10.6: i386, x86_64, ppc
+* macOS 10.7+: i386, x86_64
 
-    # build 64 bit Pd
-    ./configure --with-universal=x86_64
+Note: a "fat" Pd may not work on all systems and/or be able to load both 32 or
+64 bit externals. Additonally, you can specify multiple architectures directly:
 
     # build a "fat" Pd for both 32 and 64 bit
     # may not work on all systems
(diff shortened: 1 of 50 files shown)


Ant1r:
works great for me, all points are valuable enhancements or fixes, congrats!

- point 1 (only multi-connect when shift) is a bit disturbing (since Shift is also used to select/deselect and to connect on-the-fly, the multi-select sequence is now a bit tricky...) but I see no better way to solve the original issue.
- copy-paste buffer restoration saves many bad surprises...
- Tab additions will be much useful

thanks!

umlaeute:
> point 1 (only multi-connect when shift) is a bit disturbing (since Shift is also used to select/deselect and to connect on-the-fly, the multi-select sequence is now a bit tricky...) but I see no better way to solve the original issue.

that <kbd>Shift</kbd> is used to modify an existing selection by adding/removing single items never seemed to clash with the multi-connect enabler in my tests.

it does slightly interfere with connect-on-the-fly, but only slightly:
- if you want multiple multi-connections, well this is what you get :-)
- if you only want a *single* multi-connection, then the connect-on-the-fly is not a big deal: just stop after the first connection
- if you only want multiple single-connections, then you are out of luck and must unselect first.

the last part is the uglies one, but you are already in a very similar situation with the *current* state of intelligent patching, so i think we have a net improvement.

fdch:
I am loving the new feats!

Two things I have to say so far, one might be a regression, the other wishful thinking :)

First, somehow <kbd>Tab</kbd> has stopped working inside iemgui property dialog. This happens at least on macos (tried my own compilation and the darwin snapshot on the pd-list) 

Second, I think making the inserted object immediately editable is a great thing, and I also think it calls for yet another action. So, when you
-  click a cord to select it
-  then edit the new object... and without clicking outside to fully create the object, you
-  type <kbd>Cmd</kbd>+<kbd>t</kbd> (again)

I expect another object to be created below, between the newly created one and the original target connected to it. Instead, a new (second) cord is created with an object in between the two original connected objects, which seems rather useless and redundant. I believe this is because the focus remains on the original cord, but that is now really unnecessary. I think the focus can just advance itself behind the scenes to the newly created cord so that you can continue stacking inserted objects, but I don't know how this is handled under the hood.

Anyway, I will keep on posting things if I find any. If you'd like me to make separate issues, let me know! Keep up these changes cause they are just making patching more exciting every time!

Best,

f

umlaeute:
> First, somehow Tab has stopped working inside iemgui property dialog. This happens at least on macos (tried my own compilation and the darwin snapshot on the pd-list)

it seems like <kbd>Tab</kbd> stopped working for all properties. please do create a new/separate ticket about this.

millerpuckette:
merged to master

Issue #540, opened 2019-01-15
[print] with '-n' flag prints colon when receiving list of numbers


mariobuoninfante:
Tested with Pd 0.49.1 on MacOS Sierra
When sending a 'list' of numbers to [print -n], this would print a colon ':' first then the list of numbers

mariobuoninfante:
***EDIT***
Tested with Pd 0.49.1 on MacOS Sierra

When sending a 'list' of numbers to [print -n], this would print a colon ':' first then the list of numbers

umlaeute:
btw, you can directly *edit* your post, by clicking on the <kbd>...</kbd> and selecting <kbd>Edit</kbd> in your original post.

(please do update your original post; you can then delete the followup "***EDIT***" post (or i'll do it for you))

danomatika:
I actually looked into this a while ago, but my solution ended up being overly complicated.

danomatika:
This is fixed via #580.

PR #580, opened 2019-04-08
avoid colon (and leading whitespace) when printing lists with [print -n]

umlaeute:
also, only omit the "list" selector if the list starts with an A_FLOAT
(rather than *not* starting with an A_SYMBOL)

Closes: https://github.com/pure-data/pure-data/issues/540


umlaeute:
(to keep history linear, this should be cherry-picked/rebased rather than merged)

Issue #354, opened 2018-05-03
suggestion: expand trigger's functionality as in Purr Data


porres:
this is coming from a request I've seen on the facebook group. In a user discussion, someone brought it up how [trigger] in Purr Data has a different behaviour and it'd be cool if Vanilla worked like that too.

I think they got this from max, where number arguments are just sent to the corresponding outlet, whatever comes in.

<img width="184" alt="screen shot 2018-05-03 at 11 25 35" src="https://user-images.githubusercontent.com/5360142/39582624-ba635382-4ec4-11e8-8def-f72b57b57951.png">

as having number arguments is not something documented or what people should be doing in vanilla, I don't think this breaks anything

mxa:
Not only numbers, any constants like symbols too. Max help looks like this: ![photo_2018-09-16_11-50-09](https://user-images.githubusercontent.com/1220707/45595094-b0663c80-b9a6-11e8-9872-25ba1c84386e.jpg)

umlaeute:
so sending "2" to [trigger float gloat] gives you both `2.0` and `gloat`?

sounds like a great feature to me. not.

porres:
i have to say this was coming from a personal concern that purr data and vanilla could be more compatible, but this seems to be a lost battle and i'd better just give up...

if a more flexible and powerful object is needed, I can come up with an external... and I know that this would only have a better shot for consideration if I ended up doing a Pull Request or something and I guess I'm not...

umlaeute:
i don't really like the `symbol` feature, as it clashes with the current use of arguments.
this is not a problem with numbers (and already seen in e.g. ´[pack]`).

whether it is actually a good idea to add yet another functionality to `[trigger]` ("outputting *const*-values"; in addition to the current "force execution order" and "convert between message-types"), might be discussed separately.

porres:
Anyway, with another request for this feature, I just thought I'd give it a shot so I created a Pull Request that includes the code and features from Purr Data

porres:
I've split this issue into two new ones; see https://github.com/pure-data/pure-data/issues/691 and https://github.com/pure-data/pure-data/issues/692

PR #582, opened 2019-04-15
porting trigger from Purr Data

porres:
using the code and features of Purr Data into Pd.

closes https://github.com/pure-data/pure-data/issues/354

and also closes https://github.com/pure-data/pure-data/issues/583


umlaeute:
could you expand a bit on what those features backported from Purr Data actually are? (it's much easier to read a short description here, than to download/install/run Purr Data and check the help-patch)

porres:
Actually (and funny enough) Purr Data's help file doesn't mention these features :) 

But I've updated the help file of trigger for this PR including the changes. And it's nothing but what's been pointed in both  #354 and #583

Namely it: 
1) also converts anythings.

2) Offers constant values.

Regarding "2", non constant values are the usual arguments: anything, list, bang, float, symbol, pointer or the first letter of each. Anything else, like a float or a different symbol, is treated as a  or constant value. So if something like [t f b 30 stop] receives a "5" float message, the output, from right to left, is: "symbol stop", "30", "bang", "5".

Spacechild1:
I can see the value of 2. as a shortcut for connecting bangs to individual message boxes and I think it's ok for floats, but I consider mixing symbols and reserved identifiers a bit problematic. 
just look at `[1 2 3( -> [t a b c d e f]`

porres:
> just look at [1 2 3( -> [t a b c d e f]

I don't know what you mean, but I see that this doesn't work in Purr Data, as it actually cannot load one character symbols.

porres:
In max it works fine, though you have special cases for 'b', 'f', and also 'i' (integer)

porres:
I think you did not mean that the code of Purr Data was wrong or bad... but I've fixed it anyway...

now  [t a b c d e f] will output the symbols 'c', 'd' and 'e' alright... 

of course and as stated, 'a', 'b' and 'f' are special cases and will not be output as symbols, cause it would even break old patches

porres:
well, ok, I get you think it's not perfect to mix symbols, because some are special cases. I can see that, though I personally don't mind. I also realize @umlaeute isn't very happy either with the idea.

It's not like I'm dying for this feature anyway, the motivation was coming from a personal concern that purr data and vanilla could be more compatible, but this seems to be a lost battle anyway, and I don't really care that much. 

On the other hand, I've seen people from time to time request for this feature, specially for floats, of course, so maybe it's worth it.

Hence, if what gets into Pd is the float feature alone, I'd be cool with that ;) and I could adapt the code to only allow for constant Floats if that's the case! Maybe I'll do a separate PR for that case.

Oh, I hope there's no opposition to  allowing anythings to be converted to other message types, that's an annoyance I never understood and I hope we get rid of it.

Cheers

 

if a more flexible and powerful object is needed, I can come up with an external... and I know that this would only have a better shot for consideration if I ended up doing a Pull Request or something and I guess I'm not...

rnkn:
I would love to see this merged. Anything that makes Pd more intuitive is a good thing.

porres:
how do you feel about symbols?

porres:
for the record, I already have a version here without the constant symbol feature ;)

rnkn:
> how do you feel about symbols?

I have no opinion.

porres:
so, there's a new PR https://github.com/pure-data/pure-data/pull/600 we can have, instead, with just constant floats! then, this one can be removed

fdch:
I realize I am late on this discussion, but still, my 2 cents: 

I can't see why one would need this other than loading patches from Purr Data with that specific use of `[trigger]`. And I guess it will take time to get used to using `[trigger]` in that way. But, if some find this intuitive behavior I cannot argue with that. So, I really have no arguments against this feature and I would be curious to try it.

Perhaps if `[trigger]` has constant floats or lists, it would make sense that `[unpack]` had something similar

porres:
> I can't see why one would need this other than loading patches from Purr Data with that specific use of [trigger]

I don't get it, If someone is using trigger in Purr Data like that, then it has a use case.

I can better se the usefulness of having constant floats, a symbol would be more rarely used, like if you need to sequence a "stop" or a "reset" message or whatever an object requires.

> Perhaps if [trigger] has constant floats or lists

it'd have just floats

> it would make sense that [unpack] had something similar

why? and how so?

rnkn:
"I wouldn't use this so no one should" strikes again!

fdch:
As I said: I am not against this, please read carefully. I am only skeptical about it being actually useful.

About [unpack]:
I thought of initializing arguments instead of them being type declarations. This would apply to [pack], and work like [list]'s arguments, or an abstraction's args, for that matter. It would make sense to me since it is a similar treatment of arguments by initializing them instead of only declaring their types.

fdch:
Also, perhaps a "set" method might be needed to update the values if you need to.

rnkn:
@fdch: Alexandre has already explained it's usefulness, and it is default behaviour in Max. Read above.

porres:
i'm closing this one and I'm working on a new PR instead https://github.com/pure-data/pure-data/pull/694

Issue #589, opened 2019-04-22
Graph on parent Pd crash


bobczz:
Debian Stretch 64bit
Pd Vanilla 0.49

1. ctrl+n for a new window
2. put graph
3. open graph
4. put object or comment (**only these two**)
Pd closes immediately without any error message
Note: if you put on graph other elements (bang, number, slider ect) works great.

Spacechild1:
confirmed (Windows 7)

porres:
that just happened to me, macOS mojave

umlaeute:
@bobczz what's the use-case of adding an object into a graph?

bobczz:
@umlaeute the "classic" GOP use, to make, in my case, compact GUI of effects and instruments. See snapshot. Thanks!
![snapshot_20190509](https://user-images.githubusercontent.com/49860580/57481003-7208ef00-72a1-11e9-92f2-149e9aed35ef.png)

umlaeute:
i'm not talking about a GOP in general (as in: GOPifying an abstraction or a subpatch), but rather why you would want to abuse a `graph` window to put code therein.

also, your screenshot doesn't show anything related to `graph`.
btw, ordinary GOPified abstractions/subpatches work nicely.

bobczz:
The problem, for me, is the creation of a GOP:
Please be patient:
1. menu Put
2. put graph (on Linux Shift+Crtl+G)
3. right click for open and place an object on graph window and Pd crash.
Isn't this the correct procedure for GOP?
I think we talk about different things, maybe?
What you see on screenshot are GOP make with older version than 0.49.

umlaeute:
yes, but this is not how GOPs are supposed to be created (where did you learn that weird way to create GOPs?)

the intended way to create GOPs is:
- open a new patch (<kbd>Ctrl</kbd>-<kbd>n</kbd>)
- right-click the canvas and select `Properties`
- toggle the `Graph-On-Parent` checkbox
   - eventually change the size and margin parameters
- click `Apply`

bobczz:
Frankly I don't remember where I learned this method but I've always used it from many, many years... and I suppose I'm not the only one. Anyway, thank you for the explanation.

Spacechild1:
> this is not how GOPs are supposed to be created

IMO it's a valid shortcut. after all, a "graph" is nothing but a subpatch with GOP enabled, there's nothing special about them. you can even use GOP-subpatches as "graphs" for arrays.

[...]

PR #590, opened 2019-04-22
fix crash when creating objects inside graph/array

Spacechild1:
creating an object inside a graph/array leads to a segfault because `canvas_undo_get` returns a null pointer. the reason is that the private data (undo queue) isn't initialized for graphs/arrays. here's the fix:

* create private canvas data in `glist_init `(shared by canvas and graph/array)
* initialize undo queue in both `canvas_new` (canvas) and `glist_addglist` (graph/array)

fixes https://github.com/pure-data/pure-data/issues/589
requesting review from @umlaeute


Spacechild1:
@umlaeute Gentle reminder. Is this innocent enough to include in https://github.com/pure-data/pure-data/pull/650?

Issue #557, opened 2019-02-15
Start [openpanel] / [savepanel] in the patch directory


mariobuoninfante:
Hi,

It would be great if we had something that allows to start [openpanel] / [savepanel] in the patch directory.
Using [bang(  seems to open the directory from which Pd has been launched, while using [symbol ./( (in case of MacOS Sierra) opens the OS X directory.

porres:
since the help says a bang "Starts open panel in current directory", I consider this a bug.

umlaeute:
what's "the OSX directory"?

i don't have any such thing on my Linux machine, and on your macOS machine i guess *all* directories are "OSX directories"?

porres:
from testing with the help file, I got the HD root directory in macOS 14, Pd 0.49-1

mariobuoninfante:
@umlaeute it opens a dialog here: "/" (that is basically the machintosh HD). I'm not a Mac user as well, so I could probably say something wrong here, but I suppose that's the root directory in Mac (similar to Linux). And the folder, only this, is called "OSX"

pkstone:
Alternatively, a more flexible enhancement would be an object which returned the full path of the directory of the patch it is within.

This could in turn be passed to [openpanel], if desired, but a user-supplied path could overrule it.

porres:
I think now that when the documentation says a bang "Starts open panel in current directory", it just means the last opened (hence 'current') directory.

So this could be in fact a feature request and not a bug, where we could have two options to deal with it;
- change the code and set the current patch's directory as the default/current;
- Have a new object that outputs current directory.

Only the second option does the job, but we could also have both solutions, since the new object is desired for other purposes in Pd.

pkstone:
I think your second option is the superior one, as it gives the most flexibility. For instance, I may just want the directory of the current patch as the root; I may want to add a path onto it to access a specific directory *within* my patch’s directory.


Best,

Phil

porres:
I tried to see if I could implement option 1 and I failed miserably...

I'm happy if we only have the second option as a solution, anyway, I think we need it one way or another. Pd has to have new object that deals with directories and stuff. By the way, there have been parallel discussions on adding such a new object.

see https://github.com/pure-data/pure-data/issues/273

and https://github.com/pure-data/pure-data/issues/502

porres:
and option 2 is available here :) #617

Spacechild1:
Now that we have `[pdcontrol]`, can we close this? 

[...]

PR #592, opened 2019-04-23
[feature]: add 'mode' argument to [openpanel]

Spacechild1:
this PR adds an optional 'mode' argument to [openpanel]:

0: choose a single file (the default)
1: choose a directory
2: choose multiple files (output as Pd list)

this feature is fully backwards compatible

### DIFF
--- a/tcl/wheredoesthisgo.tcl
+++ b/tcl/wheredoesthisgo.tcl
@@ -28,7 +28,7 @@ proc open_file {filename} {
 # ------------------------------------------------------------------------------
 # procs for panels (openpanel, savepanel)
 
-proc pdtk_openpanel {target localdir mode} {
+proc pdtk_openpanel {target localdir {mode 0}} {
     if {! [file isdirectory $localdir]} {
         if { ! [file isdirectory $::fileopendir]} {
             set ::fileopendir $::env(HOME)
--- a/tcl/wheredoesthisgo.tcl
+++ b/tcl/wheredoesthisgo.tcl
@@ -35,14 +35,26 @@ proc pdtk_openpanel {target localdir mode} {
         }
         set localdir $::fileopendir
     }
-    if {$mode == 1} {
-        set result [tk_chooseDirectory -initialdir $localdir]
-    } else {
-        set result [tk_getOpenFile -initialdir $localdir]
+    # 0: file, 1: directory, 2: multiple files
+    switch $mode {
+        0 { set result [tk_getOpenFile -initialdir $localdir] }
+        1 { set result [tk_chooseDirectory -initialdir $localdir] }
+        2 { set result [tk_getOpenFile -multiple 1 -initialdir $localdir] }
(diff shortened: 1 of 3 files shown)


porres:
Hi, I know this is unrelated, but since you're tackling with [openpanel], would you also check this issue out? https://github.com/pure-data/pure-data/issues/557

cheers

ps. great addition by the way ;)

tcl/wheredoesthisgo.tcl
```diff
@@ -28,17 +28,21 @@ proc open_file {filename} {
 # ------------------------------------------------------------------------------
 # procs for panels (openpanel, savepanel)
 
-proc pdtk_openpanel {target localdir} {
+proc pdtk_openpanel {target localdir mode} {
```
umlaeute: from a compatibility pov, it probably makes sense to make the `mode` argument optional and default to `0`:

~~~tcl
proc pdtk_openpanel {target localdir {mode 0}} {
~~~

Spacechild1:
> Hi, I know this is unrelated, but since you're tackling with [openpanel], would you also check this issue out? #557

I basically agree with https://github.com/pure-data/pure-data/issues/557#issuecomment-484692745 that it would be great to have an object which outputs the current patch directory. this would solve your issue but also many others.

porres:
but don't you agree the object has a bug? it's supposed to open at the patch dir by default, according to the documentationw, so it's not working as intended. I agree Pd Vanilla should have an object that outputs a directory, but that's something unrelated to this issue. I know it would fix this, but I don't think the object should keep bugged and not working as intended even if we get to have this new object as a workaround.

umlaeute:
i guess discussion about whether #557 describes a bug or not should happen in #557, not in *this* thread.

porres:
I see you changes the code to "gensym("callback"), A_GIMME, 0);" 

I'm just curious why you did that and I would like to know what is the "callback" method used for.

Spacechild1:
> 
> 
> I see you changes the code to "gensym("callback"), A_GIMME, 0);"
> 
> I'm just curious why you did that and I would like to know what is the "callback" method used for.

the callback method is indirectly called by the tcl code if the file dialog returned a result. the A_GIMME signature is necessary to receive lists instead of a single symbol or float.

porres:
yeah, I had figured it out, thanks

Spacechild1:
rebased on current master

porres:
awesome to see this in :) - well, this might be too late now, but having a good look, maybe it just feels a bit more intuitive to have mode 1 allow you to select multiple files and then mode 2 to allow you to get the directory

porres:
It isn't too late though as this hasn't been released yet, I can open a feature request and then we can see if we can change this in time.

porres:
> It isn't too late though as this hasn't been released yet, I can open a feature request and then we can see if we can change this in time.

Well, here's a PR, what do you think?

https://github.com/pure-data/pure-data/pull/821

PR #593, opened 2019-04-23
[Fix] Windows XP usability.

Lucarda:
@millerpuckette 

From the question on #577 about `winsock2.h` `ws2tcpip.h` being compatible or not with XP (tests shown that they are), I found that the current 0.49-0 (Pd-32bit) was unusable on XP because of **tk/tcl 8.5.10**.

The good news is that placing the hole **tk/tcl 8.5.19** (which is 32bit) that is shipped in Pd-64bit fixes the issues and Pd is **happily running** on XP.

I changed the files on msw/pdprototype.tgz using a Debian machine. The new files were taken from pd-0.49-0.msw.zip

With this PR both msw distributions shares the same tk/tcl 8.5.19.


millerpuckette:
I'm not sure but I think this problem is fixed now (0.50 32-bits works on Windows XP) - let me know if I'm missing anything.

Lucarda:
I have just tested 0.50 32-bit and yes it works.

0.50 is shipped with Tk 8.5.19

:)

PR #594, opened 2019-04-24
add more methods to [list store]

Spacechild1:
this PR adds new methods to the [list store] object. it also introduces new features for existing [list store] methods.

#### new methods

* `[set <index> <list>(` - set existing list items to other values, starting at index
* `[insert <index> <list>(` - insert list before index
(`[append(` and `[prepend(` are just special cases of `[insert(` and are kept for backwards compatibility
* `[delete <index> <n>(` - delete n items starting at index
* ~~`[pop <n>(` - output the last n items and delete them~~
* `[send <s>(` - send the stored list to a named receiver
* ~~`[length(` - send the list length to the second outlet as a message `[length <f>(`~~
* ~~`[iter <start> <count> <step>(` - efficient and flexible iteration (see help patch)
(probably not worth the complexity and should be removed in favor of https://github.com/pure-data/pure-data/pull/595)~~

#### new features for existing methods

* the count argument of `[get(` is optional and defaults to 1 so we can simply write `[get 2(` to get the list item at index 2. (this also applys to the new `[delete(` method.)
* a negative count argument to `[get(` is not an error anymore but will rather output all items till the end of the list, starting at the given index, e.g. `[get 2 -1( --> [list store 1 2 3 4 5]` will yield `3 4 5`. (this also applys to the new `[delete(` method.)

#### motivation

the main idea is to turn [list store] into a more powerful dynamically sized heterogenous container, so you can easily implement queues, stacks, arrays etc. containing any Pd data types (even pointers!) - which is rather clumsy with the traditional list objects and typically involves many unnecessary copies.


Spacechild1:
I'm quite unsure about the semantics and behavior of the `[iter(` method. Any comments are highly welcome!

Somehow it would be nice to also output the index, because the idea is to make iteration as painless as possible. I can see two ways:

a) `[iter(` outputs a pair `[<index> <value>(` for every list item

b) the index is sent through the second outlet - but this means `[length(` should probably output a `[length <f>(` message instead of a float, so we can use [route] to distinguish it from an index.
anyway, the right outlet would stay backwards compatible with existing patches.

Spacechild1:
> Personnaly, i prefer to use "replace" than "set" keyword for that case.

I chose "set" because its complementary to "get" and its already used in [text] and [array]. but I don't care too much about the name.

Spacechild1:
> 
> 
> Ok ; "set" is probably more consistent. In Spaghettis i added a [list iterate] object since IMHO [list store] is already over complex. Thus my POV is not really useful ;-)

that's funny. while you were writing this, I was working on a new list method `[list foreach]` which probably does the same thing :-) https://github.com/pure-data/pure-data/pull/595

I'm actually thinking of dropping the `[iter(` method in favor of `[list foreach]` (if it gets accepted). the only advantage of `[iter(` is that it doesn't involve a list copy but that's probably neglegible... dunno

nicolasdanet:
Mine send a bang at the end of the iteration. I don't have already implemented the reverse and step option ; but i'll probably do it as it is a good idea (as also the set, delete, insert, pop methods for [list store]).

Spacechild1:
had to rebase and force push to resolve a nasty merge conflict in list-help.pd

millerpuckette:
I think 'insert' and 'delete' are useful, but 'pop' etc are easy enough to do otherewise.  As to 'iter',
I prefer the 'list foreach' idea - iteration doesn't really fit into the function of 'list store'.  (Otoh, I'd
probably design 'list foreach' differently from how you propose it.)  Also from the other PR, I'm thinking about 'list unjoin' (or maybe call it 'list crack' which is easier to remember) but not yetconvinced that anything like 'list join' is needed

Spacechild1:
> but 'pop' etc are easy enough to do otherewise.

Actually, the pop operation is needlessly complicated to do with the other methods. With `[append(` + `[pop(` it's really easy to build a LIFO stack. But if you don't want it, I can remove it.

> As to 'iter', I prefer the 'list foreach' idea

Me too. Should I go ahead and remove the `[iter(` method?

> but not yetconvinced that anything like 'list join' is needed

It's not *needed*, but concatenating strings in Pd is currently pretty painful, given that it should be a rather simple task...

Spacechild1:
@millerpuckette if you're interested in merging this PR, just tell me what I should change/remove and I'll also fix the new merge conflict in the help file.

millerpuckette:
> @millerpuckette if you're interested in merging this PR, just tell me what I should change/remove and I'll also fix the new merge conflict in the help file.

I think the 'set', 'delete', 'insert' and 'send' messages are the important ones.

Spacechild1:
OK! Can we also keep the `[length`( message? ;-) There's no other way to efficiently(!) get the length of the currently stored list...

millerpuckette:
> OK! Can we also keep the `[length`( message? ;-) There's no other way to efficiently(!) get the length of the currently stored list...

I think 'send' fills the function OK (not perfectly I'll admit) - and it's confusing to use the same outlet for two quite different functions and ugly to sprout yet another outlet.

Spacechild1:
I see your concerns, but note that `[send(` will make a copy of the whole list. This kind of defeats the purpose of `[list store]`, which was specifically designed for in-place operations. I think we really need something like the `[length(` method in one way or the other for completeness...

Maybe you have any other ideas? In the meantime I'll remove `[iter(` and `[pop(` and clean up the help-patch.pd (fighting with git)

porres:
> and clean up the help-patch.pd (fighting with git)

yeah, I know what's that like :) what I've been doing is: I get the current file upstream and replace it for the one in the branch for a commit, then I edit the patch again into a new commit

Spacechild1:
@millerpuckette I've removed `[iter(`, `[pop(` and `[length(` and fixed the help patch. Later we can think about what's the best way to get the list length...

porres:
> fixed the help patch.

Hi, I checked the help patch, and you've overridden some of the work I had done realigning the patch (which got accepted and cause the conflict, sorry). I also spotted a few issues, like not properly mentioning this would be updated for 0.51, and you've removed the send method but still kept a [r foo] object. Anyway, I worked on this help patch again and dealt with it all, I'll send you as a Pull Request if you don't mind, ok?

porres:
actually, by checking the conversation, it seems you still have a send method, I'll look closely, cause I didn't see a send message - let me build this branch and check things out more properly

porres:
yeah, I see the send message still there, sorry

porres:
sorry for the noise, I deleted some of the messages here cause I got confused, I'm just revising the help patch cause some of the work I had done got overriden and, if you don't mind, I'll send you a PR

Spacechild1:
@porres sure, make a PR!

porres:
> make a PR!

I think it's up there already

porres:
just compiled and tested this, nice addition @Spacechild1 !

porres:
> As to 'iter', I prefer the 'list foreach' idea - iteration doesn't really fit into the function of 'list store'. (Otoh, I'd probably design 'list foreach' differently from how you propose it.)

What I liked about 'foreach' is that you can use it to send a list of items, where each item goes to a different instance number of clone. Like sending a list of frequencies to an oscillator bank. All you need to do is send the output of [list foreach] to a swap object and pack the output. I have an external for that, which is a vanilla abstraction by the way, based on [list]. With this that would be not necessary or quite damn stupid and simple abstraction with 3 objects.

> Also from the other PR, I'm thinking about 'list unjoin' (or maybe call it 'list crack' which is easier to remember) but not yet convinced that anything like 'list join' is needed

I also have an external which is a vanilla abstraction that works like simplified version of [list join]. My only use for it so far is to send the joined symbol to a canvas gui label for displaying purposes. I also like that this is an easy way to convert floats to symbols. But yeah, I can't think of much other uses.

porres:
>  with 3 objects

2 actually, [list foreach] and [list prepend]

PR #595, opened 2019-04-24
add [list foreach], [list join] and [list unjoin] objects

Spacechild1:
#### [list foreach]

This list method iterates over all items in a given list (or anything). Each item (or sublist) is sent through the first outlet and its index through the second outlet.

You can set the step size as a creation argument or with the second inlet. A negative step size will iterate in reverse. The default step size is 1.

Examples:

[list foreach]: output all items in sequence
[list foreach -1]: output all items in reverse
[list foreach 2]: output pairs in sequence
[list foreach -3]: output triples in reverse

My motivation is to (finally) provide a convenient and simple way to iterate over a list without manually building a for loop. Also, it is more efficient because it doesn't involve a single list copy (which only makes a difference with very large lists, of course). most modern programming languages (Python, Lua, C++11, etc.) have introduced range based loops for a reason :-)

#### [list join]

Converts a list into a symbol by joining the individual items using a given delimiter.

#### [list unjoin]

Splits a symbol into individual tokens based on a given delimiter. 

For both `[list join]` and `[list unjoin]` the delimiter can be any symbol (including the empty symbol) and it can be set as an creation argument over via the second inlet. The default delimiter is a single whitespace.

The idea for `[list join]` and `[list unjoin]` came from zexy's `[l2s]` and `[s2l]` but my implementation is different (more robust and faster, with linear complexity). I've tested the objects with both small and very large lists and all sorts of delimiters. they also gracefully handle unexpected input like bangs, floats or pointers. anythings are promoted to symbols/lists. 

I use these objects quiten often and especially `[list unjoin]` is a pain to implement as an abstraction (with arbitrary delimiters), so I think this could be part of Pd vanilla.

### DIFF
--- a/src/x_list.c
+++ b/src/x_list.c
@@ -820,6 +820,122 @@ static void list_tosymbol_setup(void)
     class_sethelpsymbol(list_tosymbol_class, &s_list);
 }
 
+/* ------------- list foreach --------------------- */
+
+t_class *list_foreach_class;
+
+typedef struct _list_foreach
+{
+    t_object x_obj;
+    t_outlet *x_out1;
+    t_outlet *x_out2;
+    t_float x_step;
+} t_list_foreach;
+
+static void *list_foreach_new(t_floatarg f)
+{
+    t_list_foreach *x = (t_list_foreach *)pd_new(list_foreach_class);
+    x->x_out1 = outlet_new(&x->x_obj, 0);
+    x->x_out2 = outlet_new(&x->x_obj, &s_float);
+    x->x_step = f;
+    floatinlet_new(&x->x_obj, &x->x_step);
+    return (x);
+}
+
+static void list_foreach_do(t_list_foreach *x,
+    int argc, t_atom *argv, int step, int onset)
+{
+    int i;
+    if (step > 0) /* forward iteration */
+    {
+        if (step == 1)
+            for (i = 0; i < argc; i++)
+            {
+                outlet_float(x->x_out2, i + onset);
+                outlet_list(x->x_out1, &s_list, 1, &argv[i]);
+            }
+        else
+            for (i = 0; i < argc; i += step)
+            {
+                int n = (i + step) > argc ? argc - i : step;
+                outlet_float(x->x_out2, i + onset);
+                outlet_list(x->x_out1, &s_list, n, &argv[i]);
+            }
+    }
+    else /* backward iteration */
+    {
+        if (step == -1)
+            for (i = argc - 1; i >= 0; i--)
+            {
+                outlet_float(x->x_out2, i + onset);
+                outlet_list(x->x_out1, &s_list, 1, &argv[i]);
+            }
+        else
+            for (i = argc; i > 0; i += step)
+            {
+                int n, index = i + step;
+                if (index < 0)
+                    index = 0, n = i;
+                else
+                    n = -step;
+                outlet_float(x->x_out2, index + onset);
+                outlet_list(x->x_out1, &s_list, n, &argv[index]);
+            }
+    }
+}
+
+static void list_foreach_list(t_list_foreach *x, t_symbol *s,
+    int argc, t_atom *argv)
+{
+        /* step can be negative but mustn't be zero! default: 1 */
+    int step = (int)x->x_step ? (int)x->x_step : 1;
+    list_foreach_do(x, argc, argv, step, 0);
+}
+
+static void list_foreach_anything(t_list_foreach *x, t_symbol *s,
+    int argc, t_atom *argv)
+{
+        /* step can be negative but mustn't be zero! default: 1 */
+    int step = (int)x->x_step ? x->x_step : 1;
+        /* use some trickery to avoid unnecessary copy of the whole list */
+    t_atom *buf;
+    int n = abs(step);
+    if (n > argc + 1)
+        n = argc + 1;
+    if (step < 0)
+    {
+            /* when iterating backwards, the buffer must contain
+            the "remaining" elements - but only if there are any! */
+        int rem = (argc + 1) % -step;
+        if (rem && n > rem)
+            n = rem;
+    }
+    ALLOCA(t_atom, buf, n, LIST_NGETBYTE);
+    SETSYMBOL(buf, s);
+    atoms_copy(n - 1, argv, buf + 1);
+    if (step > 0)
+    {
+        list_foreach_do(x, n, buf, step, 0);
+        list_foreach_do(x, argc - (n - 1), argv + (n - 1), step, n);
+    }
+    else
+    {
+        list_foreach_do(x, argc - (n - 1), argv + (n - 1), step, n);
+        list_foreach_do(x, n, buf, step, 0);
+    }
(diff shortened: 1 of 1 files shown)


porres:
what do [join] and [unjoin] do?

Spacechild1:
> 
> 
> what do [join] and [unjoin] do?

I've just finished updating the description

danomatika:
I gotta say that I'm not a fan of "unjoin" and I understand the choice since "split" is already taken. It's too bad [list split] wasn't named [list moses].

Anyway, I suggest [list divide] as the antonym for [list join].

Spacechild1:
Actually, I now think it would be nicer if the argument for `[list foreach]` was the number of items per iteration, e.g. `[1 2 3 4( -> [list foreach 2]` would output `[1 2(` and `[3 4(`. 

It is always possible to put a `[list split]` afterwards to effectively skip elements, e.g. `[1 2 3 4( -> [list foreach 2] -> [list split 1]` would output every second item.

Spacechild1:
@millerpuckette what do you think? you mentioned that you had some other ideas on how you would design such an object.

Spacechild1:
Will resolve conflicts if there is interest in merging.

Spacechild1:
Yabadabadoo!

porres:
documentation coming up soon ;)

porres:
Ok, one think I noticed and doesn't feel quite right to me is that [list unjoin] doesn't handle well a list input and simply outputs a bang! What do you say @Spacechild1 ?

Spacechild1:
@poress Good catch! This appears to be on purpose, but I don't know what I was thinking... I just pushed a fix to develop. `[list unjoin]` is now consistent with `[list fromsymbol]` in that it only accepts symbol messages.

(We might consider accepting list and anything messages by taking the first element resp. the selector, but this should be done for both objects.)

Since you're writing the docs, here are few notes from my side:
1. `[list join]` and `[list unjoin]`: the delimiter can be any string, not only a single character.
2. You can even use an empty symbol as the delimiter; with `[list join]` this just concatenates the elements; with `[list unjoin]` this effectively breaks the symbol into its individual characters.
3. The default delimiter is a single whitespace.
4. `[list unjoin]`: any substring that looks like a Pd float (and has no extra characters!) is parsed as a float atom
5. `[list unjoin]`: empty tokens are ignored, i.e. repeated delimiters to not produce empty symbol atoms. For example, this means that `[symbol foo//bar////5.3( -> [list unjoin /]` results in the list `foo bar 5.3`.

Spacechild1:
> The default delimiter is a single whitespace.

Now that I think of it, I'm not sure if that's really a good idea. It kind of made sense to me at the time, but now I tend to think that the default delimiter should rather be the empty string.

First, concatenating/separating without delimiter seems like an equally, if not more common operation. Second, you can't really type an empty string as a creation argument, so it would be handy to have it as the default.

What do you all think? @millerpuckette

ben-wes:
> you can't really type an empty string as a creation argument

that seems like a good argument to me for making it the default.

another question: do you remember the use case(s) you had in mind for the right index outlet of `[list foreach]`? i'm not against it - just trying to understand the design decision for this vs. manual index count.

Spacechild1:
>  just trying to understand the design decision for this vs. manual index count.

So that you don't have to manually count the index ;-)

In my experience, you often need the index of the current item. Just to give a few examples:
1. modify or filter the items based on their position
2. put the (modified) items in another collection at the same index
3. look up elements in another list at the same positions
4. etc.

In fact, it's so common that many languages offer looping constructs that return both the index and the item.

Python:
```python
a = [ "one", "two", "three" ]
for i, v in enumerate(a):
    print(i, v)
```
Lua:
```lua
a = {"one", "two", "three"}
for i, v in ipairs(a) do
    print(i, v)
end
```
JS:
```js
const a= [ "one", "two", "three" ];
for (const [i, v] of a.entries()) {
  console.log(`${i}: ${v}`);
}
```
etc.

ben-wes:
@Spacechild1 thanks for the detailed response and examples! while i was aware of the functionality in those other languages, i think i hadn't used that index yet in my patches (or maybe i forgot).

those use cases sure don't sound too exotic.

sorry for the noise! - for a moment, it felt a bit arbitrary to me when testing. but should have put more thought here myself!

porres:
> you can't really type an empty string as a creation argument

you actually can if you escape it with "\ "

Spacechild1:
> you actually can if you escape it with "\ "

I'm talking about an empty symbol, not a single whitespace. Since you can type a whitespace, but not the empty string, that might be even more reason to use the latter as the default separator.

Side note: if you type`\` - without a subsequent whitespace! - you indeed get an empty string because it's not a known escape sequence. However, this gets lost when saving and reopening the patch.

millerpuckette:
I don't understand what an empty string as a delimiter would mean - 
doesn't any string at all match the empty string in its zeroth character?

I'd suggest using whitespace as the delimiter, so that, for example, a 
double space would only count as one, as any reasonable parser would 
do.  But perhaps I misunderstand the purpose of this object...?

cheers
M

millerpuckette:
P.S. - shouldn't there be a list verb that just calls binbuf_text() to 
parse a string into a Pd message as if it were read from a text file?  
Again, perhaps that is a different function from this 'list unjoin' 
thingy ...

Spacechild1:
> I don't understand what an empty string as a delimiter would mean -
doesn't any string at all match the empty string in its zeroth character?

With `[list join]`, an empty string delimiter means that the list items are concatenated directly:
```
[list foo bar baz( -> [list join] -> symbol foobarbaz
```

With `[list unjoin]` the empty string delimiter is a bit of a special case as it just breaks the string into its characters:
```
[symbol foobar( -> [list unjoin] -> list f o o b a r
```

As you said, you can't really match an empty string, but I think the current behavior is well defined and useful.

> I'd suggest using whitespace as the delimiter, so that, for example, a double space would only count as one, as any reasonable parser would do. But perhaps I misunderstand the purpose of this object...?

`[list unjoin]` skips empty tokens, so repeated delimiters are effectively only counted as one. With a single space delimiter, you would get the behavior that you've described (albeit only for space, not arbitrary whitespace).

In general, `[list unjoin]` splits a string into tokens by a delimiter string. Think of it as the exact opposite of `[list join]`. Here's a little practical example:
```
// convert a time string into a list of numbers
[symbol 09:50:36( -> [list unjoin :] -> list 9 50 36
```

It's *not* meant as a general FUDI-string to Pd-list parser, there are other objects for that (see below).

> P.S. - shouldn't there be a list verb that just calls binbuf_text() to parse a string into a Pd message as if it were read from a text file? 

You can actually do that with `[list fromsymbol] -> [fudiparse]`!

@porres Would you maybe add a little example to the fudiparse help file? I think it's not entirely obvious. EDIT: we should also demonstrate the opposite, i.e. turning a Pd list into a FUDI string: `[fudiformat] -> [list tosymbol]`.

Spacechild1:
@millerpuckette Would you be fine with me changing the default delimiter from a single space to the empty string? Again, the reasoning is that these two are at least equally common, but - unlike the space - you can't type the empty string as creation argument.

PR #596, opened 2019-04-26
closebang bug fixes

Spacechild1:
* send closebangs *before any object is actually removed* and in the same order as loadbang and initbang:
1) abstractions / clone objects (recursively)
2) subpatches (recursively)
3) remaining objects on the canvas

the current ordering is messy because when an object receives a closebang, other objects might have already been destroyed. so if you want to save a table with [closebang] for example, you have to put it into the right place in the glist (the details are complicated). here's a patch which demonstrates the problem:
[closebang-issue.zip](https://github.com/pure-data/pure-data/files/3130537/closebang-issue.zip)
what I would want instead is that no object is actually deleted unless all closebang messages have been sent, so I don't have worry about where I put my [closebang] objects.

* don't send closebang twice to cloned abstractions

* fix a bug where non-canvas objects on the root canvas wouldn't receive a closebang message.

---

externals which use `glist_delete` from the private canvas API still work as expected. 
the `[clear(` message also works. 
cutting objects is ok but `canvas_closebang` is called *on every selected canvas in series* - which means the closebang order is only correct for each selected canvas. otoh, it's the same with loadbang and initbang when pasting multiple canvases, so at least its consistent. later we might want to fix the initialization order for cut&paste...

here's a test patch (needs [iemguts] with a version **before 2018**):
[closebang-test.zip](https://github.com/pure-data/pure-data/files/3121633/closebang-test.zip)

needs review from @umlaeute (see also https://git.iem.at/pd/iemguts/issues/8)

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -897,7 +897,7 @@ int glist_fontheight(t_glist *x)
     return (sys_zoomfontheight(glist_getfont(x), glist_getzoom(x), 0));
 }
 
-void glist_dodelete(t_glist *x, t_gobj *y, int close);
+void glist_dodelete(t_glist *x, t_gobj *y);
 
 void canvas_free(t_canvas *x)
 {
@@ -909,7 +909,7 @@ void canvas_free(t_canvas *x)
         canvas_whichfind = 0;
     glist_noselect(x);
     while ((y = x->gl_list))
-        glist_dodelete(x, y, 0);
+        glist_dodelete(x, y);
     if (x == glist_getcanvas(x))
         canvas_vis(x, 0);
     if (x->gl_editor)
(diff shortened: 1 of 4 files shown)


Spacechild1:
@umlaeute gentle reminder ;-)

Spacechild1:
@umlaeute if you have time, it would be great if you could have a look, since this PR is mainly aimed at [iemguts/closebang].

umlaeute (approved):
fine by me.

millerpuckette:
Would it work to change canvas_loadbang(t_canvas *x) to canvas_loadbang(t_canvas *x, t_floatarg action) where
acvtion can be LB_LOAD or LB_CLOSE?  Then canvas_free could just blithely call canvas_loadbang(x, LB_CLOSE) and we could tell anyone wanting to catch the CLOSE message that it can happen multiple times but just ignore all but the first one you get.

Spacechild1:
We can't really send the "closebang" from within `canvas_free` because it would mess up the ordering, as explained in the PR description. Most importantly, I make sure that all "closebangs" are sent before any objects are actually deleted, just like "loadbangs" are sent after all objects have been created.

PR #597, opened 2019-04-26
sigmund~: fix possible stack overflow

Spacechild1:
there are two calls to `alloca` in sigmund~ which can easily lead to a stack overflow for large numbers of points (on Windows with a typical 1 MB stack it's already happening above 32768 points). 

I've introduced a stack limit of 400 kB (= max. 16384 points) above which heap allocation is used. this kind of fixes the DSP method (it's never called recursively) but sigmund~ could still crash if someone, for whatever reason, recursively calls its list method (which analyzes an array). the second commit changes the list method to always allocate its buffer on the heap.

there are other possible approaches:
* `sigmund_getrawpeaks` could use thread local storage (similar to https://github.com/pure-data/pure-data/pull/81)
* `sigmund_list `could use a buffer in the t_sigmund object struct

fixes https://github.com/pure-data/pure-data/issues/522


nicolasdanet:
Since you are fixing sigmund~ you probably could check also:

- https://lists.puredata.info/pipermail/pd-dev/2018-11/021800.html
- https://lists.puredata.info/pipermail/pd-dev/2018-11/021801.html
- https://lists.puredata.info/pipermail/pd-dev/2018-11/021802.html

Spacechild1:
> 
> 
> Since you are fixing sigmund~ you probably could check also:
> 
>     * https://lists.puredata.info/pipermail/pd-dev/2018-11/021800.html
> 
>     * https://lists.puredata.info/pipermail/pd-dev/2018-11/021801.html
> 
>     * https://lists.puredata.info/pipermail/pd-dev/2018-11/021802.html

hey, you can make a PR to my branch or make a seperate PR.

nicolasdanet:
Ok, i'll make a new PR ASAP (and that probably means never).

Spacechild1:
you've already done the large work of reading through the sigmund code, commiting the changes should be a matter of minutes... I can do it as well but 1) you deserve the credit and 2) I don't understand what's going on in sigmund~ at all so I would be lost in subsequent discussions :-)

nicolasdanet:
I understand your reserve. I'm reticent about making a PR as I can not test the fixes as i never succeeded to compile Pd. The code and math involved are far away in my memory thus i would be lost in subsequent discussions also! I don't care about the credit. Probably that only Miller Puckette could review the patch ; thus to avoid to stall your PR i'll make a new one in the futur as soon as i'll have time to retry the compilation process. Sorry for the noise, and sorry again to hijacked your thread. 

Note that in my fork i make a version of sigmund~ without alloca. All the memory required is cached in the structure.

Spacechild1:
> soon as i'll have time to retry the compilation process.

compiling Pd should be straightforward on Linux, macOS and Windows (Msys2). you can drop me a PM anytime if you encounter any problems.

Issue #554, opened 2019-02-11
Issue receiving large quantity of MIDI messages


mariobuoninfante:
When Pd receives a large quantity of SysEx messages, the MIDI IN port 'gets stuck' and then you need to go to 'Media->MIDI settings...' and re-select the port, otherwise no MIDI messages will be received anymore.
Here's a video that shows the issue (pass: sysexissue):
https://vimeo.com/316624285
attached the Pd and Max patches used to repro.
This has been tested on a MacBookPro 2.7 GHz Intel Core i5 16 GB of ram using Pd 0.49-1
Similar results found on Windows 10 and Ubuntu Studio 16.04.

[Pd_SysEx_issue_repro.zip](https://github.com/pure-data/pure-data/files/2852774/Pd_SysEx_issue_repro.zip)

mariobuoninfante:
this happens also with different MIDI messages, not only with SysEx

mariobuoninfante:
Pd 0.47 can doesn't show the problem

umlaeute:
could you also try Pd-0.48?

umlaeute:
hmm, i cannot really reproduce this with a plain Pd-setup (see attached patches), using Pd-0.49-0 as shipped with Debian/buster (amd64), using the alsamidi backend.

at 2ms, i haven encountered *any* problems.

going down to 1ms, packages start getting dropped, but they keep getting in.


[sysex.zip](https://github.com/pure-data/pure-data/files/2855494/sysex.zip)

Lucarda:
@umlaeute 

Using a "Virtual loopback MIDI cable" (https://www.tobias-erichsen.de/software/loopmidi.html) i can send using your "sender.pd" to http://www.midiox.com/ @ **1ms** with no problem.


![sysex](https://user-images.githubusercontent.com/25239522/52634055-ef8cb100-2ea4-11e9-978b-580a56a01dc0.png)




When using your "receiver.pd" I cant go faster than **8ms**. Then it hangs and I have to reset the midi-in driver as in @mariobuoninfante  video.

Win8.1
Pd 0.49.0 64bit.

Spacechild1:
[...]

PR #598, opened 2019-04-29
fix MIDI overflow handling for portmidi

Spacechild1:
currently, [midiin] and related objects stop responding if too many MIDI messages are sent in short time intervals and you have to reset the MIDI device in the MIDI dialog. I've tested this with a MIDI loopback and it already happens after sending 3 MIDI bytes every 1 ms.

there were two problems in the Pd code: 
* sys_poll_midi used an if clause instead of a while loop so the midi queue doesn't get flushed properly (and the `throttle` variable was useless)
* the call to Pa_Poll is redundant and also doesn't report overflow errors. this means that sys_poll_midi would stop reading midi messages once an overflow occured.

fixes https://github.com/pure-data/pure-data/issues/554

### DIFF
--- a/src/s_midi_pm.c
+++ b/src/s_midi_pm.c
@@ -273,11 +273,10 @@ void sys_poll_midi(void)
     PmEvent buffer;
     for (i = 0; i < mac_nmidiindev; i++)
     {
-        while(Pm_Poll(mac_midiindevlist[i]))
+        while((nmess = Pm_Read(mac_midiindevlist[i], &buffer, 1)))
         {
             if (!throttle--)
                 goto overload;
-            nmess = Pm_Read(mac_midiindevlist[i], &buffer, 1);
             if (nmess > 0)
             {
                 int status = Pm_MessageStatus(buffer.message);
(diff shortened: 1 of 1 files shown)


Spacechild1:
although one can argue that the error was solely on the Pd side, I think it would make sense for Pm_Poll to return overflow errors and not simply say that the queue is empty...

danomatika:
Looks like my fault. I don't remember why I changed the original `while(1)` to an if statement.

danomatika:
If we test this on multiple platforms, I'd say let's merge into the 0.50 update branch.

Spacechild1:
This works fine for me on Windows: buffer overflows are vastly diminished (thanks to the while loop), remaining ones - when sending MIDI at very high rates - correctly reported and MIDI input continues to work. It's basically the original code before your changes but keeping the special handling of `pmBufferOverflow`. If this also works for you on macOS, I think you can merge it. Here's my test patch (using a MIDI loopback device): [midi-test.zip](https://github.com/pure-data/pure-data/files/3145520/midi-test.zip)

Spacechild1:
https://github.com/pure-data/pure-data/issues/554#issuecomment-494121449 confirms it's working fine on macOS.

millerpuckette:
merged

PR #604, opened 2019-04-30
add an inlet to [receive]

### LINKED ISSUE #603: request: inlet for [receive] to set receive symbol
this would free me from the dependency of using [r.mmb]

porres:
this makes [receive] behave just like [send] - when you load it without arguments, it creates an inlet that you can use to set a receive name by giving it a symbol message.

closes #603

requires https://github.com/pure-data/pure-data/pull/1760 to be safe.


porres:
this is particularly useful to create GUI abstractions that behave like other GUIs, where you can give it a 'receive' name via an input message from the inlet

src/x_connective.c
```diff
@@ -288,13 +288,45 @@ static void send_setup(void)
 /* -------------------- receive ------------------------------ */
 
 static t_class *receive_class;
+static t_class *receive_proxy_class;
+
+typedef struct _receive_proxy
+{
+    t_pd            l_pd;
```
Spacechild1: better: p_pd

src/x_connective.c
```diff
@@ -288,13 +288,45 @@ static void send_setup(void)
...
+    t_inlet        *p_inlet;
+    struct _receive  *p_owner;
+}t_receive_proxy;
 
 typedef struct _receive
 {
     t_object x_obj;
+    t_receive_proxy  pxy; // proxy inlet
```
Spacechild1: better: x_pxy

src/x_connective.c
```diff
@@ -288,13 +288,45 @@ static void send_setup(void)
...
     t_symbol *x_sym;
 } t_receive;
 
+static void receive_proxy_init(t_receive_proxy * p, t_receive *x){
+    p->l_pd = receive_proxy_class;
+    p->p_owner = (void *) x;
```
Spacechild1: why the cast to `void *`? `p_owner` actually is of type `t_receive *`

src/x_connective.c
```diff
@@ -288,13 +288,45 @@ static void send_setup(void)
 /* -------------------- receive ------------------------------ */
 
 static t_class *receive_class;
+static t_class *receive_proxy_class;
+
+typedef struct _receive_proxy
+{
+    t_pd            l_pd;
+    t_inlet        *p_inlet;
```
Spacechild1: not used

src/x_connective.c
```diff
@@ -288,13 +288,45 @@ static void send_setup(void)
...
+static void receive_proxy_init(t_receive_proxy * p, t_receive *x){
+    p->l_pd = receive_proxy_class;
+    p->p_owner = (void *) x;
+}
+
+static void *receive_proxy_new(void)
+{
+    t_receive_proxy *p = (t_receive_proxy *)pd_new(receive_proxy_class);
+    return p;
+}
```
Spacechild1: new method not used

src/x_connective.c
```diff
@@ -288,13 +288,45 @@ static void send_setup(void)
...
+
+static void receive_proxy_symbol(t_receive_proxy *p, t_symbol* s){
+    t_receive *x = p->p_owner;
+    pd_unbind(&x->x_obj.ob_pd, x->x_sym);
+    pd_bind(&x->x_obj.ob_pd, x->x_sym = s);
+}
+
+static void receive_proxy_setup(void)
+{
+    receive_proxy_class = (t_class *)class_new(gensym("receive_proxy"), (t_newmethod)receive_proxy_new, 0, sizeof(t_receive_proxy), 0, 0);
```
Spacechild1: better: 
`receive_proxy_class = (t_class *)class_new(gensym("receive proxy"), 0, 0, sizeof(t_receive_proxy), 0, 0);`

we don't want users to be able to create instances of the proxy class, so we name it "receive proxy" - which is more or less impossible to type - and get rid of the new method - which we don't need anyway.

doc/5.reference/receive-help.pd
```diff
@@ -17,12 +16,22 @@
 receives are named to tell them whom to connect to. They work across
 windows too. Also \, you can use message boxes as shown:;
 #X msg 84 233 \; help-rcv1 34 \; help-rcv2 67;
-#X obj 172 320 r;
+#X obj 177 324 r;
 #X obj 21 10 receive;
 #X text 79 10 -- receive messages without patch cords;
+#X text 222 235 If invoked without an argument \, "send" creates an
```
Spacechild1: do you mean "receive" instead?

Spacechild1 (changes requested):
this has been on my list for ages, so I would really like to see this happening! I've suggested a couple of minor changes to make the code cleaner.

doc/5.reference/receive-help.pd
porres: sure, I had copied from [send]'s help file but didn't change it, haha

porres:
hi, can you just send me the PR? I don't know how this works, if you have proposed the changes already, or if I have to do them myself...

I assumed my code wouldn't be great, as I was actually copying from some other object in cyclone. I don't really know how to do a proxy object, this is my first one ever, so I just copied. I can't answer why I used one thing or another, I'm just glad I tested it and it works. I'm also happy that I know I had to do a proxy for this to work :)

porres:
> this has been on my list for ages, so I would really like to see this happening!

Awesome, I assume many will be happy with this one :) - glad I can help

Spacechild1:
> I don't know how this works, if you have proposed the changes already, or if I have to do them myself...

I would appreciate if you could do the changes :-)

umlaeute:
iirc there were arguments by @millerpuckette why this is not such a good idea.
mostly because you could change the receiver-label while receiving, and get a segmentation fault.

this should be in the list-archives somewhere

Spacechild1:
thanks for the hint! I had a look and discovered that this is also true for the right inlet of [value]. the PR above attempts to solve this issue.

porres:
Hi, I cherry picked and made a new commit to cleanup the code and fix the documentation

porres:
> @porres porres#1

Hi, I cherry picked and made a new commit to cleanup the code and fix the documentation

millerpuckette:
This necessitates a complicated change to symbol binding/unbinding that I fear will never be completely debugged.

An alternative idea might be to make a multiple-symbol receive with an inlet to select among a (pre-determined) list of symbols by name or perhaps index.

porres:
> An alternative idea might be to make a multiple-symbol receive with an inlet to select among a (pre-determined) list of symbols by name or perhaps index.

Not sure I get, and my actual use case here is the need to set a [receive] object with a name that is not known or pre-determined.

porres:
> my actual use case here is the need to set a [receive] object with a name that is not known or pre-determined.

Let me give more details of this particular use case. Say you have a GUI abstraction and you want the user to be able to give it "send" and "receive" symbol names, like you can do with sliders and stuff. For [send] this is trivial, but the issue is if you want to set a receive name. And this name is not pre-determined.

For illustration, I have here a two dimensional slider from my ELSE library that's doing this kind of thing. It's an abstraction based on Data Structures.
Screen Shot 2019-12-07 at 17 30 53

millerpuckette:
> > my actual use case here is the need to set a [receive] object with a name that is not known or pre-determined.
> 
> Let me give more details of this particular use case. Say you have a GUI abstraction and you want the user to be able to give it "send" and "receive" symbol names, like you can do with sliders and stuff. For [send] this is trivial, but the issue is if you want to set a receive name. And this name is not pre-determined.
> 
[...]

That's a reasonble use case.  THe best way to do this given Pd's current state is to make the patch self-modify to make a subpatch with the desired receive and a local send in it.  Even though that's ugly it's less ugly than the code needed to allow dynamic symbol unbinding.

Spacechild1:
but the problem is still that you can't *change* the receive name because you can't (safely) unbind...

btw, my personal use case is to dynamically select event sources, similar to how you use `[receive~]` to dynamically select audio sources.

porres:
Something like dynamic patching to delete and create a new [receive] object with dynamic patching into a subpatch, right?

that's actually a solution and an external abstraction, purely vanilla, from 'mmb', I believe it's in https://puredata.info/downloads/mmb 

I made an external called [else/receiver], which seems to work fine for my abstractions. It's a compiled object and basically the same as this PR and if there were any issues, I thought https://github.com/pure-data/pure-data/pull/614 would take care of it.

But yeah, if a dynamic patching abstraction is what we need, I can think of working on something like that for my library instead, and it would also be a vanilla abstraction and something any one could use for dynamically setting a receive name.

UNLESS......... @Spacechild1 still tries to convice he's got the perfect simple and best solution for this in https://github.com/pure-data/pure-data/pull/614 ...

porres:
Hey, once again I did reply before I could see @Spacechild1 had replied before me.

Well, I'm out of this debate as I clearly have no idea what the issues involved are... so I'll let the grown ups do the talking

cheers

Spacechild1:
> still tries to convice he's got the perfect simple and best solution for this

it certainly isn't perfect, and I'm not even sure it's the best, but it's a proof of concept that it's possible to solve the issue with dynamic unbinding in the core.

porres:
> it certainly isn't perfect, and I'm not even sure it's the best, but it's a proof of concept that it's possible to solve the issue with dynamic unbinding in the core. it is something we should keep investigating.

seems https://github.com/pure-data/pure-data/pull/849 is better?

porres:
> this is particularly useful to create GUI abstractions that behave like other GUIs, where you can give it a 'receive' name via an input message from the inlet

Well, it occurs to me now that since IEMGUIs can set a "receive", whatever issues we have with this may already present itself in those cases, right?

Spacechild1:
yes

porres:
well, as I understand it then, this makes https://github.com/pure-data/pure-data/pull/849 necessary even if no inlet is added to [receive]

porres:
Hi, this has issues to be solved but I can't see them or do anything about it. Are you supposed to do something instead @Spacechild1 ?

Spacechild1 (approved):

Spacechild1:
Seems like I forgot to accept the changes.

You could mention in the PR description that it requires https://github.com/pure-data/pure-data/pull/1760 to be safe.

porres:
>

porres:
> You could mention in the PR description that it requires https://github.com/pure-data/pure-data/pull/1760 to be safe.

done

umlaeute:
I think this PR is a *no go* if applied without #1760.
Therefore i suggest to merge #1760 into this PR (rather than just mentioning it as a prerequisite).

(The only reason for not merging is, that there might be an alternative implementation of the issues fixed by #1760... But I don't see any such alternative implementation)

porres:
> Therefore i suggest to merge https://github.com/pure-data/pure-data/pull/1760 into this PR (rather than just mentioning it as a prerequisite).

Or the other way around? Get the work here on top of the other? What do you say @Spacechild1 ?

umlaeute:
Either way. 

(Though conceptually it makes more sense to have the PR that depends on another PR to include its dependencies. If you want to keep history logical (you *first* need the commits of #1760, then the commits of this PR), consider rebasing your PR on top of #1760 and then force push. (I really think you should learn how to do this))

porres:
well, as my oldest open PR, since https://github.com/pure-data/pure-data/pull/1760 got merged, can we look into this? :) 

ping @millerpuckette 

cheers

porres:
well, there are merging conflicts here but I can't act on it as the original branch is gone, so I'm rebooting with just the code change (which could setill be hand applied I guess) - see https://github.com/pure-data/pure-data/pull/2970

I left the help file change out as that might be the thing causing merge conflicts

PR #606, opened 2019-04-30
[feature]: add [send( method to [value] object

Spacechild1:
this method allows to directly send the stored value to a named object (e.g. [receive] or another [value]), analogous to the [send( method of [float], [int] and [pointer].

### DIFF
--- a/doc/5.reference/value-help.pd
+++ b/doc/5.reference/value-help.pd
@@ -1,58 +1,70 @@
-#N canvas 673 23 541 568 12;
-#X floatatom 21 59 5 0 0 0 - - -, f 5;
+#N canvas 673 23 541 623 12;
+#X floatatom 21 59 5 0 0 0 - - -;
 #X text 383 15 abbreviation:;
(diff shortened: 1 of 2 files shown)


Lucarda:
One other nice feature will be if **[value]** can receive a "bang" message :

`[;foo 12, bang(`

currently is not possible. we get on the console:

`value: no method for 'bang'`

Spacechild1:
unfortunately, this is not so easy to implement because with [; foo 12, bang( the messages are really sent to a common proxy object. the problem is that all [value foo] objects have a reference to this proxy object, but the proxy doesn't have backward references to the [value foo] objects, so even though the proxy could accept a [bang( message, it wouldn't be able to forward it to all the corresponding [value] objects. one *could* implement this but I don't think it's worth the complexity. 

for your use case, just do `[12, bang( -> [v foo]` instead. if "foo" can be any symbol, you can use the right inlet of [value] for that.

porres:
I think this is a nice addition that improves consistency between these objects :)

PR #607, opened 2019-04-30
pd_bind / pd_unbind bug fix

Spacechild1:
some externals (e.g. iemlib/iem_receive) bind/unbind symbols on the fly. this is dangerous in case the bound/unbound symbol is the same we're currently receiving messages for because the bindlist can get modified (or even deleted) while iterating over it.

[value] theoretically also has this issue but I couldn't get it to crash in practice. I've observerd situations where a bind element or the whole bind list gets deleted right before or after it receives a message but the freed memory most likely hasn't been overwritten or returned to the OS yet.

my solution is to add a flag in the bindlist class to tell if we're currently receiving messages for it. I check for this flag in `pd_bind`/`pd_unbind` and if true I schedule the bind/unbind request for the next tick and issue a warning.

NOTE: I can't simply ignore the bind/unbind request because this could lead to subtle errors (e.g. objects ending up bound to more than one symbol) and existing code assumes that pd_bind/pd_unbind never fails (expect for `pd_unbind` on an unbound symbol).

this PR is also needed for features like https://github.com/pure-data/pure-data/pull/604

### DIFF
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -52,6 +52,27 @@ void gobj_save(t_gobj *x, t_binbuf *b)
 actually bind a collection object to the symbol, which forwards messages sent
 to the symbol. */
 
+/* this is a helper class for scheduling a bind/unbind request to the next tick.
+we need this when someone wants to bind/unbind while we're receiving a message
+to very same symbol */
+typedef struct _schedbind
+{
+    t_clock *s_clock;
+    t_pd *s_who;
+    t_symbol *s_sym;
+    int s_action; /* 1: bind, 0: unbind */
+} t_schedbind;
+
+static void schedbind_tick(t_schedbind *x)
+{
+    if (x->s_action)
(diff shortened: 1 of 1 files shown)


umlaeute:
while the proposed patch should indeed fix the problem at hand (i haven't really checked the code, but trust your description for now), i'm not entirely sure about the implications of using "`[delay 0]`" for unbinding.
e.g. 

~~~
[t b b]
|     |
|     [delay 0]
|     |
|     [; foo foo; bar bar(
|
[s foo]

      [r rename]
      |
      [symbol bar(
      |
[r foo]
|
[route bang foo]
|              [print oops]
[s rename]

Spacechild1:
yeah, this only fixes the memory errors but doesn't really allow proper recursive binding/unbinding of symbols. forget this PR, here's something better: https://github.com/pure-data/pure-data/pull/614

PR #614, opened 2019-05-02
make pd_unbind safe and improve stack overflow protection

Spacechild1:
#### Problem:

the Pd GUI objects and some externals (e.g. [iem_receive] from iemlib) allow to bind/unbind symbols on the fly. this is dangerous in case the bound/unbound symbol is the same we're currently receiving messages from because the bind list can get modified (or even deleted) while iterating over it.

[value] theoretically also has this issue ~~although I couldn't get it to crash in practice - but I've observed situations where a bind element or the whole bind list gets deleted right before or after it receives a message (it only doesn't crash because the freed memory hasn't been overwritten or returned to the OS yet).~~

EDIT: here's a test patch which demonstrates the issue with value: [value-crash.zip](https://github.com/pure-data/pure-data/files/3139523/value-crash.zip) (for me this only crashes when I run Pd in a debugger.)

Also, the current overflow protection only works for a *single* recursion path; for more than one path (e. g. several outlets feeding back to the same inlet), Pd would end up in an infinite loop. The same problem applies to bind lists if more than one target sends back to the bound symbol.

#### Solution:

I added a (thread-local) stack of `t_bindelem` pointers. When sending a message to a symbol, I do the following for each bind list element:
1) put the *next* element on the stack
2) message the object
3) get the next element from the stack

The trick is that when `pd_unbind()` is asked to unbind an object from a symbol, it tries to locate the bind element in the stack and - if necessary - sets it to the next valid element (or 0). This means we can iterate over the bind list even recursively while removing elements (not that it's a good idea to do so). 

Also, t_bindlist has to be reference counted to avoid segfaults.

To improve the stack overflow protection, I set a flag when a stack overflow is detected so that no outlets or bind lists are messaged until the stack is completely unwound.

There is a tiny performance overhead but judging from the benchmarks I did it's pretty much neglectible.

---

Here's a crude testpatch using [iem_receive] from iemlib: [bindlist-test.zip](https://github.com/pure-data/pure-data/files/3145507/bindlist-test.zip)
Here you can test the improved stack overflow protection: [stackoverflow.zip](https://github.com/pure-data/pure-data/files/3145508/stackoverflow.zip)

Try both patches with and without this PR to see the difference!

This PR is needed by https://github.com/pure-data/pure-data/pull/604.

### DIFF
--- a/src/m_obj.c
+++ b/src/m_obj.c
@@ -299,6 +299,7 @@ void obj_init(void)
 /* --------------------------- outlets ------------------------------ */
 
 static PERTHREAD int stackcount = 0; /* iteration counter */
+static PERTHREAD int overflow = 0;
 #define STACKITER 1000 /* maximum iterations allowed */
 
 static PERTHREAD int outlet_eventno;
@@ -306,6 +307,26 @@ static PERTHREAD int outlet_eventno;
     /* set a stack limit (on each incoming event that can set off messages)
     for the outlet functions to check to prevent stack overflow from message
     recursion */
+
+int stackcount_increase(void)
+{
+    if (++stackcount >= STACKITER)
+    {
+            /* set overflow flag to prevent any further messaging */
+        overflow = 1;
+        return 0;
+    }
+    else
+        return !overflow;
+}
+
+void stackcount_release(void)
+{
+    stackcount--;
+        /* once the stack got completely unwound, we can clear the overflow flag */
+    if (stackcount == 0)
+        overflow = 0;
+}
 
 void outlet_setstacklim(void)
 {
@@ -356,19 +377,19 @@ static void outlet_stackerror(t_outlet *x)
 void outlet_bang(t_outlet *x)
 {
     t_outconnect *oc;
-    if(++stackcount >= STACKITER)
+    if(!stackcount_increase())
         outlet_stackerror(x);
     else
     for (oc = x->o_connections; oc; oc = oc->oc_next)
         pd_bang(oc->oc_to);
-    --stackcount;
+    stackcount_release();
 }
 
 void outlet_pointer(t_outlet *x, t_gpointer *gp)
 {
     t_outconnect *oc;
     t_gpointer gpointer;
-    if(++stackcount >= STACKITER)
+    if(!stackcount_increase())
         outlet_stackerror(x);
     else
     {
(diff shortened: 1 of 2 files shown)


umlaeute:
`value` doesn't crash because its receiver is *cold*, that is: you cannot trigger a re-bind while receiving.
(it would be different if https://github.com/pure-data/pure-data/pull/606#issuecomment-487926138  was implemented)

Spacechild1:
you can use a receive with the same name which binds value to another symbol. that's what I did. I printed the memory addresses and could see that there are possible cases of use after free. also, you have to create the receiver after the value object so it comes first in the bind list. everything together it's highly unlikely but possible :-)

Spacechild1:
the latest commits also allows stupid edge cases like in: 
[bindlist-test2.zip](https://github.com/pure-data/pure-data/files/3139379/bindlist-test2.zip)

Spacechild1:
hmmm, I think this is still not 100% safe because one could unbind/destroy an element in a deeper recursion level which is the "next" element on a higher level. maybe instead of storing/restoring "b_current" on the function stack I need to maintain an explicit stack of "next" values which is traversed down and adjusted whenever an element is unbound.

millerpuckette:
This is a ot of code to 'fix' what is actually a misuse of pd_unbind in the IEMGUIs.  (It's also dangerous to delete objects that are bound to symbols, as some self-editing patches do, but even this PR can't fix that as far as I can tell).

If it's even worth fixing the IEMGUI problem, I think a simpler workaround would be for the GUI itself to set a flag whenever it sends a message out, and then checking the flag when it gets an unbind message, and simply refusing to do it reentrantly with an error message.

porres:
> This is a ot of code to 'fix' what is actually a misuse of pd_unbind in the IEMGUIs.

iemguis are pd vanilla's gui objects, right? But Christof was mentioning about [iem_receive], which is basically a [receive] that you can set a receive name.

Now, the thing is that we have a PR that adds an inlet to [receive], see https://github.com/pure-data/pure-data/pull/604

 so it would work much like [iem_receive]. Hence, the idea here is to not fix an external, but fix things in Pd so we can add this functionality to [receive].

porres:
hmm, seems it'd also be necessary for this PR https://github.com/pure-data/pure-data/pull/606

Spacechild1:
> hmm, seems it'd also be necessary for this PR #606

no, it's not

Spacechild1:
@millerpuckette 

>  (It's also dangerous to delete objects that are bound to symbols, as some self-editing patches do, but even this PR can't fix that as far as I can tell).

hmmm, most objects bind itself (or a proxy) to a symbol and then unbind in the destructor. Do you have an example of an object binding *another* object to a symbol? You're right that such cases can't be fixed whatsover, but personally I haven't seen them.

Anyway, I think I've tried various solutions and this is the most complete fix I could come up with. It's not pretty, I admit, but it works :-) My main motivation was to allow "safe" settable receive objects.

millerpuckette:
The most horrible example I know of is that someone deletes a subpatch as a result of a message sent to a receive inside it.

Spacechild1:
Ah, right! The missing `[del 0]` problem :-) But I think this is a different issue, because it isn't primarily about bind lists. What you're describing is the general problem of deleting objects during message passing.

porres:
just linking to the new alternative PR https://github.com/pure-data/pure-data/pull/849

Spacechild1:
closing in favor of https://github.com/pure-data/pure-data/pull/849

Issue #273, opened 2017-12-18
suggestion to add an object like [canvasinfo] from purr data in vanilla


porres:
It could be a simple adoption of that object or a newly designed one with a different name and features.

For a previous discussion about it, see https://github.com/pure-data/pure-data/pull/150

porres:
I don't really care abbout that many features, mostly "args" and "dir", in which I designed two externals myself that take care of that

derekxkwan:
I would also find the "args" and "dir" functionality helpful as well as "name" (but I'd see no reason to exclude any of the apparent functionality of [canvasinfo], not so familiar with the object though). Another potential source of inspiration for canvas-oriented functionality would be the iemguts library which has [canvasargs], [canvasconnections], [canvasdollarzero], [canvaserror], [canvasindex], [canvasname]. [canvasobjectposition], [canvasposition], [canvasselect]. 

It looks like [canvasargs], [canvasdollarzero], [canvasindex], [canvasname], and [canvasposition] have siimilar functionality to [canvasinfo] (but the latter looks to be simply a "getter" and not a "setter" at least for name which is perhaps safer).

porres:
> but I'd see no reason to exclude any of the apparent functionality of [canvasinfo]

it'd be silly to have an object with the same name and less functionalities, indeed, but my point was that it could be a different object with different functionalities and a different name - like, it could be just [canvas]

HenriAugusto:
I back that especially because the |args( method is a must.

Currently the only vanilla solutions to get the argument list is either using a [tricky exploit + dynamic patch  routine](https://forum.pdpatchrepo.info/topic/10892/collect-all-arguments-as-a-list/10) 

or

actually passing everything in a single symbol, like arg1/arg2/arg3, and then splitting it into a list using [list fromsymbol/tosymbol]. But this is not compliant with PD's standard syntax!

One thing a miss from [namecanvas] is a selector. If you send |args( to it i would like it to prepend "args" to the list of arguments. So one would be able to use something like [route args dir toplevel] to easily send the information to the right place.
_________________________________

_On a side note:_ I see [namecanvas] on the "obsolete" (i'm reading deprecated) list of pd's object list so |name( might be a substitute, but i'm not sure. I like the ability to pass a canvas' name as an argument to an abstraction, like

|namecanvas $0-this]

[someAbstraction $0-this ]

[...]

PR #150, opened 2017-07-29
[getdir]

danomatika:
So just throwing this out there...

This PR integrates the [getdir] object by Guenter Geiger into Pd vanilla.

Being able to get the directory of the patch or it's parent(s) is useful for opening resources relative to an abstraction but also for supplying the directory to pd's open/close messages. This is one of the few externals I still need for use with libpd on mobile, mainly for playlist management for cycling through songs. Having functionality like this included with vanilla or a "core external" is one of the last things I would need to make my setup vanilla-only (one of these others being the sound file info).

This object is very small, works well, and Guenter uses a permissive [license](http://git.puredata.info/cgit/svn2git/libraries/ggee.git/tree/LICENSE.txt) for his [ggee externals](http://git.puredata.info/cgit/svn2git/libraries/ggee.git/tree/).

This may not be the best or most canonical solution, but it has worked well for me for years. It could also be included as an "extra."

Issue #503, opened 2018-09-26
$@ syntax addition


porres:
according to https://github.com/pure-data/pure-data/issues/273#issuecomment-415195064 the $@ syntax from Purr Data originated in the development of Vanilla, but never got into Extended and was never included in Vanilla. I wonder why it didn't make it, and if it has been already decided not be a good idea and rejected before, we can just close this issue. 

cheers

porres:
seems the relevant code in extended is in m_atom and m_binbuf only...

porres:
my main motivation for this was getting the arguments of an abstraction, which now can be done with [pdcontrol].

so only motivation now would be to be the same as pd extended 0.43, which purra data inherited, and that doesn't seem to be much of a cause to me, personally

HenriAugusto:
I still think it's a neat syntax.

Imagine you have an abstraction [listFoo] that deals with lists. You want the user to be able to initialize the internal list using arguments [listFoo a b c d].

Inside of [listFoo] you could just write 

`[listFoo $@]`

instead of

```
     [loadbang]
     |
     |args 0(
     |
     [pdcontrol]
     |
[list]
```

Also you can use it to "pass objects" to abstractions.

Example: i have a [elem-op] object that performs element-wise operations on lists.

Insite [elem-op] i have have a [$1] object in order to be able to pass the operator i want to use as an argument: [elem-op +], [elem-op -], [elem-op *], [elem-op &&], etc

But let's say i also have an unrelated object [doSomething] that takes two elements and process them in some specific way i want to use. [doSomething] takes up to 3 argument flags (-x -f -c) to control it's behaviour plus a number.

We want to be able to write [elem-op doSomething -x -f 32] so our [elem-op] will be able to process two lits element-wise using the [doSomething -x -f 32] abstraction. 

But we can't use [$1] for that because it would expand only to [doSomething]. 

We can't also use [$1 $2 $3 $4 $5] because we would be passing zeroes as arguments to [doSomething] in case the user don't supply all three flags and a number. 

But we would be able to do it if we had [$@].

porres:
it might feel nicer to do it that way instead with [pdcontrol], but once we have this new object, there's a way to do it and no need for another.

porres:
the other case feels like an edge case that I don't know exactly if I get it, but feels like there's an alternative as well...

[...]

PR #617, opened 2019-05-02
A canvas object to query information on canvases

porres:
A couple of years ago @danomatika thought about including [getdir] => see https://github.com/pure-data/pure-data/pull/150#issuecomment-318883169

But then the idea was that we'd create a new 'canvas info' object instead.

I also suggested we could port canvasinfo from Purr Data, but I now think it's best we do one from scratch! https://github.com/pure-data/pure-data/issues/273 - here's canvasinfo for reference, anyway, we could still get inspired to get some of the functions in it https://github.com/pd-l2ork/pd/blob/master/pd/src/x_interface.c#L174

So I started working on such object but, so far, I have only included the 'canvas dir' function. This is because I think this is the most crucial feature that we need in vanilla right now.

Issues that depend on a vanilla solution for this is found here: https://github.com/porres/pd-cyclone/issues/412

and also here: https://github.com/pure-data/pure-data/issues/557 

I'm now thinking of other functions to add.

cheers


porres:
I got my eyes on  '[canvas mouse]: get/send mouse interaction' that @danomatika suggests, seems like [canvasreceive] from iemguts lays all the grounds for that :) 

But I have great hopes for a [canvas arg] object, pretty much based on [canvasargs] as well. This would be crucially important to design vanilla abstractions in a more powerful way, being able to manage given args more flexibly

HenriAugusto:
Indeed those are welcome additions to vanilla. Some functionalities are very useful for creating more powerful GUIs, while _getdir_ and _canvas args_ are great for the reasons mentioned on #273 and #557

umlaeute:
could you please fix the issues reported by the appveyor build? (grep the code for `snprintf` to see how it is handled otherwhere)

porres:
> thanks for your efforts.

welcome :) and thanks back ;)

> could you please fix the issues reported by the appveyor build? (grep the code for snprintf to see how it is handled otherwhere)

sorry, I wanna figure this out, but I just can't seem to do it yet and need help. 

I can see in the details that the problem is the sprintf function, but I googled the error code around and couldn't find something enlightening.

I checked elsewhere how a similar sprintf function is used in Pd and copied from this https://github.com/pure-data/pure-data/blob/a9fd832e9b2ad355292b5eb1ecdb94139e08f2a7/src/s_print.c#L45

but I don't know why this code fails.

I also don't know how to 'grep the code' - at this point you might realise that me trying to do stuff here is actually a big effort :)

Spacechild1:
you have to add
```
#define snprintf _snprintf
#endif
```
somewhere at the top, have a look at s_loader.c for example.

porres:
Yay! Thanks.

Spacechild1:
> I also don't know how to 'grep the code'

`grep "snprintf" /path/to/pure-data/src/*.c`

this will list all occurances of snprintf in the C files in Pd's source folder, so you can check what they're doing differently.

Spacechild1:
(sorry, somehow Github only allowed me to do single comments instead of a review)

BTW, the diff of x_misc.c shows several changes in formatting which are not related to your code. maybe your editor does some auto formatting? if you use a GUI tool (e.g. git gui) you can see your changes in advance and you can only stage what you really *want* to commit :-)

danomatika:
>if you use a GUI tool (e.g. git gui) you can see your changes in advance and you can only stage what you really want to commit :-)

Or a decent text editor plus prodigious reviews with `git diff` before committing anything. Sublime Text is good about showing tabs and space character marks which is useful in fixing style & formatting issues.

Spacechild1:
@umlaeute do you know what's going with our code comments? I can only make single comments and they won't show the referenced section of the source code...

umlaeute:
@Spacechild1 full review works for me if I navigate to the [PR's commit](https://github.com/pure-data/pure-data/pull/617/commits/c874927fa42e7553439ee6fef34b36e422688b26) rather than [just the commit](https://github.com/pure-data/pure-data/commit/c874927fa42e7553439ee6fef34b36e422688b26)

porres:
Yeah, I'm just using XCode, and also saw now how the formatting changed, bummer... will look into that

porres:
> the beauty of something like $@ in contrast to [canvas args] is that you can pass the complete argument list down to several abstractions (e.g. [canvas-abstraction $@] will receive all arguments of the parent patch)

I remembered I opened an issue about it https://github.com/pure-data/pure-data/issues/503

umlaeute:
triggered by the discussion #660, it came to my attention that the `[canvas]` object mixes concepts of "abstractions" (with sub-commands like `dir` and `args`) and "canvases" (aka "windows"; with sub-command like `mouse`).

in order to avoid future confustion, i'd suggest to keep these concepts separate (that is: have different objects for interacting with different concepts).

porres:
please, can you elaborate more on how the concepts are defined and differ?

umlaeute:
- a *canvas* is contained within a single window (that is: either a subpatch (`[pd ...]`) or an abstraction)
- an *abstraction* is an object instance based on a `.pd` file.

Assuming you have an object `[foo 1 2 3]`, which is implemented by a file *foo.pd* living in the folder *~/Documents/Pd/* (so it's really an abstraction). This object contains another object `[pd bar zupi]` (that is: a subpatch).
Let the subpatch contain three objects `[canvas args]`, `[canvas dir]` and `[canvas mouse]`.
Furthermore, let the abstraction `[foo]` also contain instances of `[canvas args]`, `[canvas dir]` and `[canvas mouse]` directly.

- What's the output of `[canvas args]` that directly lives in `[foo 1 2 3]`? What's the output of `[canvas args]` that lives in the `[pd bar zupi]` subpatch of `[foo 1 2 3]`? What is "thing" that `[canvas]` refers to when calculating *args*?
- What's the output of `[canvas dir]` that directly lives in `[foo 1 2 3]`? What's the output of `[canvas dir]` that lives in the `[pd bar zupi]` subpatch of `[foo 1 2 3]`? What is "thing" that `[canvas]` refers to when calculating the *dir*?

however:

- What's the output of `[canvas mouse]` that directly lives in `[foo 1 2 3]`? What's the output of `[canvas mouse]` that lives in the `[pd bar zupi]` subpatch of `[foo 1 2 3]`? What is "thing" that `[canvas]` refers to when calculating the *mouse*?

porres:
I see!

[canvas mouse] refers to 'canvases', while [canvas arg] and [canvas dir] refer to 'abstraction' or, to use a technical term under the hood in the Pd code, to an "environment".

> i'd suggest to keep these concepts separate (that is: have different objects for interacting with different concepts).

What would you suggest to call the new object? I see a problem with calling it [abstraction] because you can have the 'dir' function not in an abstraction. On the other hand, the 'args' function only makes sense in an abstraction.

Calling it something like [environment] would make more sense, then, maybe.. the thing is that no one seems to know what that term refers to (I can't say I know it either and I might just saying something silly now).

Maybe [patch]?

porres:
Ok, now that I browse through the code, what I see is something called 'canvas environment'.

such as "/* --------- functions to handle the canvas environment ----------- */"

in https://github.com/pure-data/pure-data/blob/05178840f844e1e0411c1ce9a165613b3542f4fe/src/g_canvas.c

So there seems to be a convergence between these concepts in the code of Pd itself. Maybe we could just reorganize and revise the concept of a canvas and document it in the object or in the Pd documentation as well. By the way, is there a definition of the 'canvas' concept in Pd anywhere?

umlaeute:
the problem at hand is, that the `[canvas]` object relates to different things when using different verbs. which i think is counter-intuitive.

just split it into two objects (on for each "thing")

porres:
hey, I see now Miller has a new [pdcontrol] object that queries the directory of a patch! So [canvas dir] just became redundanct

porres:
maymiller says he can think of many other functionalities this [pdinfo] object could have. Maybe getting arguments can be another...

porres:
> BTW, the beauty of something like $@ in contrast to [canvas args] is that you can pass the complete argument list down to several abstractions (e.g. [canvas-abstraction $@] will receive all arguments of the parent patch)

hey, you can also achieve this with [canvas args], just put it in [canvas-abstraction] and have it ask for the parent's aerguments. What's the difference and the use case?

Anyway, this functionality may also just be part of the new [pd control] object - at least getting the arguments list, for setting the arguments, miller reasoned [savestate] can already handle something like that!

Spacechild1:
> hey, you can also achieve this with [canvas args], just put it in [canvas-abstraction] and have it ask for the parent's aerguments. What's the difference and the use case?

that's true!

Spacechild1:
actually, for this to work, the abstractions needs to know that it should get arguments from the parent. on the other hand, [myabs $@] is transparent. the use case is that some projects use a large number of nested abstractions where each abstraction *possibly* inherits parts of its arguments from its parent. but that's probably an edge case.

porres:
> but that's probably an edge case.

that could probably have another workaround, huh?

umlaeute:
i don't think it's an edge case at all.

oh, and btw: what's the difference between `[canvas args]` and `[canvasargs]`?

porres:
> what's the difference between [canvas args] and [canvasargs]?

the way depth is handled is different, I'm using "environment" and that seemed like an issue on why this was a confusing aspect of the canvas concept. If I recall correctly, canvas args is able to change the name of the abstraction, this one only changes the actual arguments, and also some code changes, but I'd have to look closely side by side to give you more details. I just simplified the code of else/args, which can also break arguments if you give it a breaking character, such as "-" so you can route "flags".

porres:
if this gets in https://github.com/pure-data/pure-data/pull/689 it makes this object almost obsolete...

all left would be the [canvas mouse], so we'd need to rethink what functionalities we're still missing for vanilla that we want, and if they could be in the new [pdcontrol] object, or into a new [canvas] object, or a [path] or whatever

porres:
Closing this for now, as both 'dir' and 'args' are now handled by the new [pdcontrol] object, see https://github.com/pure-data/pure-data/pull/689#event-2531097238

porres:
[canvas mouse], the thing that was still left here, can pretty much just reside in an external library like 'else', or I don't know, maybe we pick up on this discussion of what a new [canvas] object should be and present it with more options/functions.

PR #618, opened 2019-05-05
fix error message in s_audio_pa.c

Spacechild1:
* missing braces would put the second call to error() outside the else clause.
* also improve the error message itself.
* I don't think there's a reason to post the last error message only on Windows (and not on Linux for example)


umlaeute:
part of #650

PR #619, opened 2019-05-05
consistent key-events

umlaeute:
on linux, the `<KeyRelease>` event will not fill the `%A` value, leading to all kinds of inconsistent behaviour.

~~~tcl
sendkey .pdwindow.text 1 h h 0     # keypress 'h'
sendkey .pdwindow.text 0 h {} 0    # keyrelease 'h'
sendkey .pdwindow.text 1 plus + 0  # keypress '+' (%K="plus", %A="+")
sendkey .pdwindow.text 0 plus {} 0 # keyrelease '+'
~~~

This PR tries to fix the problem, by using a cached value of `%A` value (e.g. received through `<KeyPress>`) if necessary.

The solution uses `dict`, which requires Tcl>8.5; older tcl-versions will fall back to the original behaviour (although i haven't tested this yet).


umlaeute:
i've tested this on linux and OSX; probably some Windows user (@Spacechild1 , @Lucarda ?) could test as well.

my testing scheme is like this:

~~~Pd
[keyname ]      [key]        [keyup]
|        |      |            |
[pack f s]      [print key]  [print keyup]
|
[print keyname]
~~~

- open the test-patch
- clear the console (via the menu)
- press all buttons on the keyboard (from the upper-left corner to the lower-right corner)
- repeat with <kbd>⇧Shift</kbd> resp. <kbd>Alt Gr</kbd> (usually only for a subset, e.g. the numeric row)
- save the Pd-console (via the menu)

doing this for both the `updates/0.50` and the `bugfix/keys` and compare the results.

Lucarda:
@umlaeute 

I did the tests only replacing the tcl folder of both PRs (`updates/0.50` & `bugfix/keys`) on my 0.49. Is that OK or i missed something?

both tests gave the same results:

[pdwindow-key-all.txt](https://github.com/pure-data/pure-data/files/3154067/pdwindow-key-all.txt)

umlaeute:
that's perfect, thanks. (it shows that the patch didn't introduce a regression; since the Windows-version was fine in the first place, that's good enough)

Lucarda:
```
key: 59
keyname: 1 \\;
keyup: 59
keyname: 0 \\;
```
Is for my Spanish keyboard for the letter `ñ` but the OS is set to use an English keyboard.  `ñ` is an extra fisical key.

umlaeute:
i'm not sure i understand.
- afaict the <kbd>ñ</kbd> key (on **Spanish** keyboard) is right of <kbd>L</kbd>
  - this is the location of <kbd>;</kbd> on a **US** layout
- you have switched your keyboard layout (on software) to **US**
- if you press the <kbd>ñ</kbd> key in Pd, you get `;` (which I think is expected)
- which character do you get when you press the <kbd>ñ</kbd> in some other application (Word, notepad++,...)?

Lucarda:
Oh yes I get `;`.
`keyname: 1 \\;` 
the `\\ ` had fool me.
:)

danomatika:
Looks good to me. Yet another one of those cross-platform inconsistencies, *sigh*.

Issue #567, opened 2019-03-05
[array define -k foo] does not keep 'size' when size is not an argument


fdch:
Say you create an array with `[array define -k array2]`.

Then pass, for example, a `sinesum 64 1` message, and everything is working fine: array is resized to 64+3, as expected. Saving the patch does write 64+3 values (`#A 0 -0.0980171 0 0.0980171 ...`).

But, if you close and re open the patch, then the array is resized back to default, ignoring the new size, which looks like this:

![array_size_not_kept](https://user-images.githubusercontent.com/13335792/53813679-77268680-3f5e-11e9-92ee-f788ce4a7e1f.jpg)

This would be a problem when doing things dynamically, or when the new array size is greater than 100..., otherwise if you have your tabreads fit for the first 64 values, you won't notice it I guess....

fdch:
Can this be closed as completed?

porres:
I dunno, has it been completed? :)

fdch:
See the [draft PR](https://github.com/pure-data/pure-data/pull/655) and a comment by [Miller](https://github.com/pure-data/pure-data/pull/621#issuecomment-881788405) so I think it has a fix, but I'm not sure. That's why I asked :)

umlaeute:
i haven't looked at the draft PR nor at miller's comments, but in general we close issues once they are fixed in `master`.
This means that the mere existence of a PR is not enough to close the issue (esp. if it is a "draft PR"), but once it's merged into `master`, it can be closed. usually PRs should have a `Closes: https://github.com/pure-data/pure-data/issues/567` line (either in the top-level posting of the PR, or somewhere in the commit message of the relevant commit), to automatically close the issues once merged...

umlaeute:
apart from that: the issue appears to be fixed for good in current Pd(and documented in the array help-patch).

closing for good :-)

PR #621, opened 2019-05-08
Save size in [array define]

Spacechild1:
[array define -k foo] doesn't save the *actual* size of the array but only the initial size. This is especially confusing when no size is given as a creation argument (in which case the initial size is 100). 

For cases like `[array define -k foo]` (no initial size given), my solution is to simply add a `resize` message to the save binbuf.

For cases like `[array define -k foo 64]`, I replace the initial size argument in the saved obj binbuf and issue a notice (e.g. "array define: saving foo with size 1000") - but only if the size has actually changed, of course.

this fixes https://github.com/pure-data/pure-data/issues/567

### DIFF
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -700,6 +700,16 @@ static int garray_click(t_gobj *z, t_glist *glist,
 
 #define ARRAYWRITECHUNKSIZE 1000
 
+void garray_savesizeto(t_garray *x, t_binbuf *b)
+{
+    if (x->x_saveit)
+    {
+        t_array *array = garray_getarray(x);
+        binbuf_addv(b, "ssi", gensym("#A"), gensym("resize"), array->a_n);
+        binbuf_addsemi(b);
+    }
+}
+
 void garray_savecontentsto(t_garray *x, t_binbuf *b)
 {
     if (x->x_saveit)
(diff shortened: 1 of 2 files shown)


fdch:
Thanks for looking into this!

> * keep the original obj binbuf and just resize to another value (this is what I'm currently doing)

if you place [array define -k foo 64], then resize to 128, put some data, save, and close the patch, What would happen to the newly placed data when you re-open the patch? Unless data is kept in this first option, I think your second option might be better:

> * replace the initial size (if given) in the obj binbuf with the actual size

 Perhaps it might be useful, as well, to have a *notice* (as a *reminder* in this special case) saying that *foo has been resized* or something.

Spacechild1:
> What would happen to the newly placed data when you re-open the patch?

The data is kept, it's just that the object binbuf will show the wrong size. But I don't really like this, I'll try to update the binbuf.

> Perhaps it might be useful, as well, to have a notice (as a reminder in this special case) saying that foo has been resized or something.

Something like "saved array 'foo' with new size 1000". Of course only if an initial size had been given as a creation argument.

millerpuckette:
I did this a different way that I thought would be simpler (but now I think it was more complicated).  Now size changes are saved unless the size is given as a creation argument in which case that is respected.

Spacechild1:
Cool!

Issue #623, opened 2019-05-09
using `donecanvasdialog` switches to edit-mode


porres:
Hi, there's some bug now in Pd-0.49 that weren't there before.

The test patch explains itself. The thing is that 'donecanvasdialog' seems to force Pd to get into edit mode. 

This is not just an annoyance, it's hard for me to isolate the issue yet, but this can also make Pd crash and "blow up" (meaning, Pd closes and disappears immediately). I know I also had this when working on this test patch, but I couldn't replicate it. On the other hand I can easily see Pd crash with [else/mtx.ctl].

I'm abusing a lot of these features to design GUIs as externals  in ELSE. The [mtx.ctl] object is one of them and every time I open a patch it's contained in, the patch opens in Edit mode. And if I try to cut and paste the object => Booom! 

But all works fine in Pd-0.48-2 for instance

Anyway, here's the test patch.

[bug-edit.pd.zip](https://github.com/pure-data/pure-data/files/3160305/bug-edit.pd.zip)

and here is how it looks 
<img width="619" alt="Screen Shot 2019-05-09 at 01 04 05" src="https://user-images.githubusercontent.com/5360142/57426637-5dbbe800-71f6-11e9-9a2c-95c03726d8fb.png">

umlaeute:
- how do you trigger the "Boom" behaviour? assuming with "Boom" you mean a crash, i cannot reproduce this (or anything else really).
- what is the "donecanvasdialog" supposed to do in your "other bug"? can it be removed?

umlaeute:
the problem is a side-effect of the *infinite undo*.

the relevant code is near https://github.com/pure-data/pure-data/blob/0.49-0/src/g_editor.c#L2138

umlaeute:
on closer inspection:

- with the `donecanvasdialog` message you are actually modifying the patch.
- *typically*, any action that triggers a modification of the patch changes the patch's mode to "edit-mode".
- therefore i think it is (should be) actually *intended* behaviour to switch to edit-mode. (although previous versions of Pd did not exhibit this behaviour; but that was more a bug in the previous versions)

i acknowledge that there seem to be a couple of exceptions to the "'modifications of the patch' equals 'edit mode'", e.g. `obj 100 100 float` will create an object and not change the mode of the patch.
however,
[...]

Issue #624, opened 2019-05-09
changing an abstraction with `donecanvasdialog` switches the parent-patch to edit-mode.


umlaeute:
Sending a `donecanvasdialog` message to an abstraction to make it `GOP`, switches the containing patch to <kbd>Edit</kbd>-mode, even though the containing patch is not modified *at all* (it looks slightly differently, because the abstraction is now GOP, but that's a property of the abstraction rather than the containing patch).

this is quite similar to #623 (where it was reported for sub-patches, for which i think changing to <kbd>Edit</kbd>-mode is the expected behaviour, as the containing patch is indeed modified.)

umlaeute:
actually, the `donecanvasdialog` to the abstraction does *not* change the <kbd>Edit</kbd>-mode of the containing patch.

It's the `donecanvasdialog` message to a *subpatch* inside a GOP-abstraction (the abstraction seems to be skipped when it comes to switching to <kbd>Edit</kbd>-mode)

umlaeute:
closing this as really a duplicate of #623

PR #625, opened 2019-05-09
have "donecanvasdialog" add 'Undo' to the affected canvas

umlaeute:
the canvas-properties can be called
- from the canvas itself (by right-clicking on the empty canvas, and selecting "Properties")
- OR from the parent canvas (by right-clicking the object,...)
unfortunately, there's no way to find out how the canvas-properties have been changed.

because of this, there's some confusion about where the <kbd>Undo</kbd> should be added (the canvas itself or the parent canvas).
the original <kbd>Undo</kbd> implementation (as ported from Pd-l2ork), tried adding the <kbd>Undo</kbd> to the parent-canvas, but this lead to all kinds of unexpected situations 
- the parent canvas would go to <kbd>Edit</kbd>-mode if the GOP-state inside a GOP would 
be changed (#623 and it's dupe #624)
- the parent canvas would also get a new <kbd>Undo</kbd> entry, which - when triggered - would crash Pd :-( (the [relevant source code](https://github.com/pure-data/pure-data/blob/0.49-0/src/g_editor.c#L2138) makes it pretty clear that this was never tested)

The suggested fix for all these problems is to handle the <kbd>Undo</kbd> of the `donecanvasdialog` entirely inside the directly affected canvas.

This means that changing the canvas-properties via the parent-patch can only be undone inside the canvas.
So far, changing the size of a GOPified object (by dragging it's right corner in the parent canvas) is still attached to the parent canvas' <kbd>Undo</kbd> queue.

Closes: https://github.com/pure-data/pure-data/issues/623

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2131,18 +2131,9 @@ static void canvas_donecanvasdialog(t_glist *x,
            argument supplied. */
     if (fromgui && (!(graphme & 1)))
         graphme = 0;
-        /* parent windows are treated differently than applies to
-           individual objects */
-    if (glist_getcanvas(x) != x && !canvas_isabstraction(x))
-    {
-            /* JMZ: i don't know how to trigger this path */
-        t_canvas*x2 = glist_getcanvas(x);
-        canvas_undo_add(x2, UNDO_APPLY, "apply", canvas_undo_set_apply(x2, glist_getindex(x2, &x->gl_gobj)));
-    }
-    else
-    {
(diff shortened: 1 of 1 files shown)


porres:
Just tested this and can confirm this PR beautifully fixes the issues I was having :) now I can use [else/mtx.ctl] again without crashing Pd! Yay!

umlaeute:
included into the `update/0.50` branch.

umlaeute:
included as c85df5e62c09d0419899af050dd4c06905500558

PR #627, opened 2019-05-11
add "goprect" method to graphs for dynamically changing the GOP rect

Spacechild1:
The purpose of this PR is mainly to stop people from abusing internal messages like `[coords(` - or even worse: `[donecanvasdialog(` - for dynamically adjusting the GOP rect. Instead I want to offer a dedicated message for this purpose: `[goprect <xmargin> <ymargin> <width> <height>(`.

Width and height are optional, e.g. `[goprect 100 0(` will just move the GOP rect to position (100, 0) without changing the dimensions. This can be used for dynamically showing different areas of a subpatch and implementing "tabbed widgets" for example.

`[goprect(` redraws the parent canvas to avoid graphical artifacts ("ghost objects"). With `[coords(` and `[donecanvasdialog(`, you need to do `[map 0, map 1(` to achieve the same thing.

Just like `[bounds(`, `[goprect(` doesn't affect the undo queue.


LGoodacre:
Adding new dynamic patching capabilities is great, but could you please elaborate on what would be gained from breaking backwards compatibility with `coords` and `donecanvasdialogue`? Why exactly are they considered "abuse"?

Spacechild1:
> from breaking backwards compatibility with coords and donecanvasdialogue?

Just to be clear, I'm not trying to deprecate those messages. After all, they are internally used by Pd. 

> Why exactly are they considered "abuse"?

In a way, `[coords(` is less evil because it's used in Pd patch files and its behavior is more stable. On the other hand, `[donecanvasdialog(` is a reply message sent from a GUI dialog and certainly not meant to be called by anyone else (I'm guilty as charged :-). It's behavior is not consistent across Pd versions and it's quite easy to crash Pd in mysterious ways.

What I'm offering here is a *stable* and *practical* alternative for manipulating the GOP rect. With `[bounds`( we already have a documented and supported message to manipulate the x/y range; `[goprect(` only sets the GOP size/position. It's not only safer but also much easier to use!

danomatika:
>could you please elaborate on what would be gained from breaking backwards compatibility with coords and donecanvasdialogue?

The ability to add/change the values sent to `donecanvasdialog` for future GUI updates without breaking *everyone's* (currently unsupported) dynamic patching behavior?

Spacechild1:
@millerpuckette just in case you feel like sneaking a small feature into Pd 0.52.3 ;-). We keep having trouble with people (ab)using `[donecanvasdialog(` and `[coords(` and consequently complaining about various "bugs". A proper public message like this could finally solve it.

porres:
what about setting it to graph on parent, and show/hide object name and arguments?

Spacechild1:
> what about setting it to graph on parent, and show/hide object name and arguments?

Is there ever a situation where you wanted to this dynamically? Generally, we can always add arguments (later).

porres:
I've been using, in else/sample I have a [pd $0-buffers] subpatch and I create arrays to display audio from each channel dynamically. The way I know how to do this is how I tried to document today in my branch:

create a subpatch with [pd 'x'] and tell it to graph on parent hiding arg an name so I can then include an array in it.

Now, we could add a graph, but we can't choose the x/y coordinates of the graph this way, unline 'obj' message. Maybe that's an alternative... add this possibility, then resize it with this new message.

jmmmp:
The PR's purpose stated at the start isn't really achieved - at least in this page I didn't see any mention of the parameters GOP appearance, and all the parameters related to structs: X/Y units, X/Y range. Or are these covered by another new message? If not, the only option for dynamic changes is to use the old messages.

Spacechild1:
> parameters GOP appearance

do you mean "hide object name and arguments"? I never needed to change that dynamically. Either way, I think it should be another message, as it has nothing to do with the GOP rectangle.

> all the parameters related to structs: X/Y units, X/Y range.

There is already the `[bounds(` message which sets the X/Y range in GOP patches resp. the X/Y units in non-GOP patches. I should mention this in the PR description.

jmmmp:
> > parameters GOP appearance
> 
> do you mean "hide object name and arguments"? I never needed to change that dynamically. Either way, think it should be another message, as it has nothing to do with the GOP rectangle.

I use it in [jmmmp/multiarray], and also in several own patches but exceptionally. It's useful when redoing complex guis and save tcl/tk bottlenecks.

porres:
> I use it in [jmmmp/multiarray], and also in several own patches but exceptionally. It's useful when redoing complex guis and save tcl/tk bottlenecks.

can you give us a specific example/reason why?

Spacechild1:
> I use it in [jmmmp/multiarray], and also in several own patches but exceptionally. It's useful when redoing complex guis and save tcl/tk bottlenecks.

That's a nice trick! So it seems what you actually want is a way to temporarily disable sending GUI commands. I think you should ask for such a feature then. I don't think it makes sense to replace one hack with another, let's fix the actual issue instead!

jmmmp:
> That's a nice trick! So it seems what you actually want is a way to temporarily disable sending GUI commands. I think you should ask for such a feature then. I don't think it makes sense to replace one hack with another, let's fix the actual issue instead!

well, the real issue would be for the gui rendering to be more efficient, specially with arrays. everything else is a hack around it, be it an "official" or "unofficial" one. but this is a topic as long as Pd exists, so I think many minds have thought about this.

Spacechild1:
> well, the real issue would be for the gui rendering to be more efficient, specially with arrays.

That's true. GUI performance in Pd is horrible. Many things are getting redrawn all the time even if they don't need to be. But let's move this discussion somewhere else, e. g. the mailing list.

danomatika:
>Many things are getting redrawn all the time even if they don't need to be.

Note: This could be partially rectified by using the tkcanvas move method which was introduced in Tk 8.5, then Pd could simply move existing elements instead of destroying and creating them. This is something I'd like to look into now that we are firmly past Tk 8.4 where the method was not available and could also replace a good amount of redundant core messaging.

Spacechild1:
If there is some interest in merging, I will fix the conflicts.

millerpuckette:
I need to figure out how to fix number/symbol/list boxes within GOPs before adding any new functionality... maybe that's the next thing I should be working on since it's pretty badly broken.

Spacechild1:
>  I need to figure out how to fix number/symbol/list boxes within GOPs before adding any new functionality...

Hmmm... I'm not sure how these two things are related. Also, this PR does not really add any "new" functionality. It has always been possible to change the GOP rect, but only with a dialog. Now we would be able to do it with a message. That's really it.

millerpuckette:
Only in that fixing the GOP trouble will require a massive rewrite and probably everything about GOPs will have to be revisited.

Spacechild1:
> Only in that fixing the GOP trouble will require a massive rewrite

Oh...

> probably everything about GOPs will have to be revisited.

AFAICT, the GOP mechanism itself can't really change without breaking virtually every existing patch. (Think of the `coords` message in patch files.) It's just the implementation that could be improved.

So as long as GOPs stay rectangular, nothing would really change for this PR. But if you are not even sure about that, than better wait :-)

porres:
Hi, so, I guess number/symbol/list boxes within GOPs are finally sorted now, right?

But I wanted to revisit this PR for another thing. What if I want to change the size of an array?

The way I can do this now is add a [namecanvas] object inside the array canvas, then send a "coords" message to it. Another way is if I use hardcore and undocumented dynamic patching to create an array whose canvas has a given name...

Where I'm getting at and what I hope we can do is use the "goprect" sent to an array name so we can resize it. Is is possible?

Spacechild1:
This PR has been around for 6 years now. Once this is merged we can finally stop worrying about "regressions" in "donecanvasdialog" and tell people to use the new official method. This would solve a real long-standing issue and I don't think we should wait further for some hypothetical future GOP refactoring. See also https://github.com/pure-data/pure-data/issues/2653#issuecomment-3042551250.

porres:
Yay. This one requires a doc update though, so I added to https://github.com/pure-data/pddp/issues/292

PR #630, opened 2019-05-12
add namespace to classes from a multi-object-per-binary library

Spacechild1:
with multi-object-per-binary libraries (e.g. zexy, iemlib) it's sometimes impossible to avoid name collisions if the library authors didn't add alternative creators (e.g. "foo/bar"). on the other hand, with single-object-per-binary externs you can prepend the folder name as a kind of namespace.

this PR suggests to automatically add such a namespace to classes added by a multi-object-binary library, so that, for example, [atoi] from zexy could also be called as [zexy/atoi] to avoid a clash with another object of the same name.

currently I only use the library name, so loading zexy with `[declare -lib some/folder/zexy]` will lead to `[zexy/atoi]`. it's would also be possible to use the full name (by using `objectname` instead of `classname`) so you get `[some/folder/zexy/atoi]` but I don't think that people would want this...

of course it's still possible to use the short name (i.e. `[atoi]`)!

please test and comment!


porres:
> please test and comment!

Before testing, let me comment and say I always thought about proposing such a feature :) Thanks!

Spacechild1:
Unfortunately, there's no real way to distinguish a multi-object-per-binary external from a single-object-per-binary external which registers itself under a different name. let's take for example iemlib, compiled as multiple binaries:
`[iemlib/ii]` actually registers itself as "init" and adds "ii" as a second creator (which correctly registers "iemlib/ii" as an alias). but because the `class_loadsym` "iemlib/ii" doesn't end with "init", my code thinks we've loaded a library with multiple objects and therefore creates two bogus aliases: "ii/init" and "ii/ii".

I see two solutions:
* just live with it
* only create library aliases if `class_loadsym` is 0. this is true if a library is loaded via startup flags or `[declare]`, but it would prevent alias creation when loading the library as an object, e.g. `[zexy]`, because this sets `class_loadsym`.

what do you prefer?

src/m_class.c
```diff
@@ -529,6 +551,46 @@ void class_addcreator(t_newmethod newmethod, t_symbol *s,
...
+        {
+            char buf[MAXPDSTRING];
+            buf[MAXPDSTRING-1] = 0;
+            snprintf(buf, MAXPDSTRING-1, "%s/%s", loadstring, s->s_name);
+            /* post("add alias: %s", buf); */
+            class_addmethod(pd_objectmaker, (t_method)newmethod,
+                gensym(buf),
+                vec[0], vec[1], vec[2], vec[3], vec[4], vec[5]);
+        }
+    }
```
umlaeute, line 593: shouldn't that duplicate code go into a separate helper function?

Spacechild1: thanks, done!

Issue #632, opened 2019-05-12
[vu] has no "send symbol" and its outlets can't be hidden


porres:
[vu] only has "receive symbol".

I use this mostly for hiding inlets/outles, but I can't hide the outlets of [vu]...

Spacechild1:
to be honest, I never realized that [vu] actually has outlets... now I'll always notice, thanks a lot! :-p 

adding a send symbol means we would have to append it to the object binbuf for backwards compatibility which would be very ugly... also, the outlets of [vu] just pass the input, which doesn't seem to be very useful to me... maybe we can just hide them by default?

porres:
why ugly? just add an extra argument... it'll be a bit weird that it's in the end, far from the send argument, but no one will ever notice

porres:
it's hard to notice them because the background color is pretty dark by default. The way I dealt with this in [else/meter~] was by having it set to full blown black as the inlet color. But I actually wanted to mess with the color differently...

Anyway, what just stroke me is that there are two inlets and only one receive symbol. Of course if you really want to use a send address, you can send it a list of 2 values and it'll work.

But what about the send symbol? It'll be just one, it'll hide the two outlets and it'll send the value of just one outlet?

I think not...

Maybe the send address can actually send a list of 2 values.

So yeah, it's a bit tricky to mess with this, it's just a design flaw that's kinda cumbersome to fix.

Spacechild1:
> Anyway, what just stroke me is that there are two inlets and only one receive symbol. 

the receive symbol takes a list of 2 floats: rms + peak

> Maybe the send address can actually send a list of 2 values.

yes, so it would mirror the behavior of the receive symbol. anyway, I don't understand the purpose of the outlets at all, so I would just hide them :-D

porres:
>  so I would just hide them :-D

I know, they shouldn't exist! Perhaps just hiding them is not that bad of an idea...

porres:
> I don't understand the purpose of the outlets at all,  so I would just hide them :-D

I know, they shouldn't exist! Perhaps just hiding them is not that bad of an idea...

If an ugly fix is what we're looking ahead one way or the other, this is not the ugliest and it's just the easiest. The object will still have phantom outlets for those who ever used it (I doubt anyone ever did, though). But in practice this would be like removing them and not documenting it or giving people a hint it exists.

porres:
[vu] actually has a placeholder in the arguments for the send symbol, but it's always "empty".

[...]

PR #634, opened 2019-05-12
don't draw [vu]'s outlets

porres:
This is one solution that came up in #632 and closes it.

Another solution to that issue would be to add a send symbol and a new extra argument. But the discussion led to this solution as probbaly easier and more sensible.

In practice, this hides the outlets without removing them for backwards compatibility concerns. The object keeps its outlets, but these are undocumented and maybe no one ever used them anyway as there doesn't seem to be much of a point in using them...

If this PR is not accepted I can work on the other solution (adding a send symbol).


umlaeute:
> but these are undocumented and maybe no one ever used them anyway as there doesn't seem to be much of a point in using them...

hmm, well. they used to be documented (before your PR). so why do you think that nobody is using them? and why do you think they were added (and even documented!) in the first place, if there's not point in using them?

anyhow: your patch could also be simplified *a lot* by just forcing `x->x_gui.x_fsf.x_snd_able = 1;` (either changing the 2 lines in `vu_dialog` and `vu_new`; or introducing a single new line somewhere in `vu_draw_io`).
personally i'm not a big fan of patches that just uncomment chunks of code (as this is the best way to accumulate cruft), but your mileage my vary.

porres:
> why do you think they were added (and even documented!) in the first place

I never really said that

> hmm, well. they used to be documented (before your PR). so why do you think that nobody is using them? 

I agree this is ugly but it was one idea that felt best and was suggested to me in the discussion in https://github.com/pure-data/pure-data/issues/632

But now I have a new idea, which is just hide the outlets when you give it a send symbol.

umlaeute:
> > why do you think they were added (and even documented!) in the first place
> 
> I never really said that

no of course not.

the course of action is more like:
- the outlets are currently (as in: Pd-0.49-0) drawn, so somebody must have added them
- the outlets are currently documented in the help-patch, so somebody must have written that documentation
- you said that there doesn't seem to a reason to use them ("there doesn't seem to be much of a point in using them...")
- I asked you, why you think that somebody made the effort to both write code to display them *and* to document that feature if they are useless.

just because you (or me, or somebody else) doesn't see an immediate benefit in a feature, doesn't mean that nobody reasoned about it (and came to a certain conclusion).

porres:
closing this in favor of https://github.com/pure-data/pure-data/pull/654

PR #639, opened 2019-05-22
disable portmidi rate limit

Spacechild1:
portmidi added a speed limit for outgoing MIDI messages on macOS to prevent possible data loss because of an inofficial hard-coded 15 kB/s limit in CoreMidi (for feedback protection on the IAC). Unfortunately, this makes it impossible to continously send large amout of MIDI data to USB MIDI devices. It is not entirely clear if CoreMidi only drops messages on the IAC, but it seems so.

See this massive thread for more information: https://github.com/pure-data/pure-data/issues/581

This PR does the following:

a) patch portmidi so we can fully disable the rate limit for macOS with a compile time switch (`-DLIMIT_RATE=0`). the default is `#define LIMIT_RATE 1`
b) compile Pd without the rate limit to get the original behavior before the portmidi update. We might want to add a warning somewhere about possible message loss.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ tcl/pd-gui
 
 # build
 *.o
+*.obj
 lib*.a
 
 # libtool
@@ -55,6 +56,7 @@ src/*.a
 # distribution tarballs
 pd-0.*
 msw/0.*
+msw/pdprototype/*
 
 # gettext
 po/*.msg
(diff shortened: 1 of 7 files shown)


Spacechild1:
@danomatika if you have time, you might double check before I create a patch and send it to Roger.

danomatika:
If you send the the file, I can create  a patch from it and check at the same time. I want to split up the other patch we have anyway.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Should we close this in favor of #641?

Spacechild1:
sure

Issue #55, opened 2017-04-20
Conflict: incoming OSC/FUDI data and DSP on


NiccoloGranieri:
This problem occurs when, while receiving OSC data from inside the machine (thus using a localhost), DSP is turned on. The OSC data stops visually flowing (the GUI freezes) until the DSP isn't turned off. 

Even though it might seem like a GUI glitch, this has a major delay effect also on the digital sound processing. The delay rises up to 5 seconds when the glitch occurs. The same effects appear if the patch is launched with the -nogui command from the terminal.

The only workaround I found to solve this issue is to run two separate instances of Pd: the first one that will act as an OSC parser, not having the need to turn on DSP. Once the data is parsed, the data is sent to a second instance of Pd via FUDI protocol (netsend) where it can coexist with DSP on without causing any issues.

This unfortunately brings up the amount of computing power needed.

The bug can be replicated in two different ways. 
- On MacOS the glitch appears even by simply sending OSC data from pd and receiving it on the same patch at a high rate.

- On Linux this doesn't happen, but it can easily be reproduced by sending OSC data to Pd from another application, Processing for example, at a high frame-rate.

Example patches are attached below.

[Pd_Bug.zip](https://github.com/pure-data/pure-data/files/944269/Pd_Bug.zip)

**_# UPDATE_**

The same glitch appears with large amount of data incoming NOT formatted in OSC, but in Pd's native formatting FUDI. To replicate this, run this small python script in the terminal typing first:

`python`

and then pasting the code below:

`import socket`
`import time`

`IP="127.0.0.1"`
`PORT=5005`
`addr=(IP, PORT)`
`EOL=';\n'`
`sock=socket.socket(socket.AF_INET, socket.SOCK_DGRAM)`

`for i in range(10000000):`
`  msg="list foo "+str(i)+" bar"`
`  sock.sendto(msg+EOL, addr)`
`  time.sleep(0.001)`

This example sends data to PD via FUDI using the local host on the port 5005 every millisecond. As said previously if DSP is turned on the GUI freezes. The issue is avoided if the data is throttled once every 10 milliseconds. (This is achieved by changing the last line of the python code to: time.sleep(0.01)

umlaeute:
so the Mac-patch locks the GUI immediately on my linux laptop, afaict, regardless of DSP.

this happens, even if I disconnect the output of `[netreceive]` from `[parseosc]`.

[...]

PR #640, opened 2019-05-24
GUI messaging bug fixes

Spacechild1:
**a)**

`return (sys_domicrosleep(0, 1) || sys_poll_togui());`

if Pd is "flooded" with incoming messages, it might never get a chance to update the GUI because of the short-circuiting OR statement, so the GUI appears to "freeze".

I discovered this while investigating a specific issue (https://github.com/pure-data/pure-data/issues/55) but in hindsight it might explain a couple of other mysterious GUI "freezes" users have experienced in the past...

I think there's no problem with always trying to poll *and* update the GUI. If the GUI queue is empty, `sys_poll_togui` will be a no-op anyway.

---
**b)**

`sys_microsleep` should probably not poll sockets, at least I don't see why. people would expect this function to just sleep. also, looking at the places where it is used in the Pd code (m_sched.c, s_audio_alsa.c, s_audio_oss.c), this is outright dangerous because the Pd instance is unlocked.

---
**c)**

If DSP is on, `sys_pollgui` is called only once per DSP tick:

https://github.com/pure-data/pure-data/blob/7c27aa0ad505bb4802eee3fc40886836c814353f/src/m_sched.c#L523

This is not so much of a problem for incoming TCP traffic (max. MAXPDSTRING resp. INBUFSIZE bytes per socket) but rather for UDP (max. 1 packet per socket). 

See this test patch: [udp-vs-tcp.zip](https://github.com/pure-data/pure-data/files/3218070/udp-vs-tcp.zip)

Receiving 1000 messages takes ~10 ms with TCP but ~1500 ms with UDP because we only can receive one UDP packet per DSP tick (1.5 ms).

I tested two possible solutions, both with TCP and UDP, by sending 1000 short FUDI messages at once and receiving them on the same patch while playing a test tone.

1) in `sys_pollgui` call `sys_domicrosleep` in a loop to receive all pending UDP messages.
PRO: automatically works for all poll functions (also those added by externals)
CON: more prone to audio dropouts than the TCP version

2) make UDP sockets non-blocking and loop inside `socketreceiver_getudp` / `netsend_readbin`,
but limit it to max. INBUFSIZE bytes (to throttle it and make it behave roughly the same as the TCP version)
PRO: less prone to audio dropouts.
CON: has to be done for every UDP poll function, i.e. iemnet and mrpeach should be updated as well.

personally, I tend to solution 2). @danomatika  maybe this is something for the netobject updates. shall I make a PR?

EDIT: solution 2) is now in the netobject update branch https://github.com/pure-data/pure-data/pull/666

### DIFF
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -228,15 +228,9 @@ static int sys_domicrosleep(int microsec, int pollem)
         for (i = 0; i < pd_this->pd_inter->i_nfdpoll; i++)
             if (FD_ISSET(pd_this->pd_inter->i_fdpoll[i].fdp_fd, &readset))
         {
-#ifdef THREAD_LOCKING
-            sys_lock();
-#endif
             (*pd_this->pd_inter->i_fdpoll[i].fdp_fn)
                 (pd_this->pd_inter->i_fdpoll[i].fdp_ptr,
                     pd_this->pd_inter->i_fdpoll[i].fdp_fd);
-#ifdef THREAD_LOCKING
-            sys_unlock();
-#endif
             didsomething = 1;
         }
         return (didsomething);
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -903,7 +903,8 @@ void sys_unqueuegui(void *client)
 
 int sys_pollgui(void)
 {
-    return (sys_domicrosleep(0, 1) || sys_poll_togui());
+    int from = sys_domicrosleep(0, 1), to = sys_poll_togui();
+    return (from || to);
 }
 
 void sys_init_fdpoll(void)
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -255,7 +255,7 @@ static int sys_domicrosleep(int microsec, int pollem)
 
 void sys_microsleep(int microsec)
 {
-    sys_domicrosleep(microsec, 1);
+    sys_domicrosleep(microsec, 0);
 }
 
 #if !defined(_WIN32) && !defined(__CYGWIN__)


Spacechild1:
one idea for c): just before dsp_tick() we might do:

`while (countdown-- && sys_pollgui()) ;`

this would make sure that we receive all pending incoming UDP packets, but no more than the throttle limit (5000). it's a good idea to read UDP packets as fast as possible because otherwise they can easily pile up, overflow the read buffer and get lost.

umlaeute:
> `return (sys_domicrosleep(0, 1) || sys_poll_togui());`
> 
> if Pd is "flooded" with incoming messages, it might never get a chance to update the GUI because of the short-circuiting OR statement, so the GUI appears to "freeze".

i guess the idea is, to prioritise Pd-core at the expanse of Pd-GUI: if the GUI freezes, this is less of a problem than audio stuttering.

Spacechild1:
but we *do* always call `sys_domicrosleep`. is there a reason to prioritize `sys_domicrosleep` (polling the GUI) over `sys_poll_togui` (notifying the GUI)? I don't see why we should and it looks more like an oversight to me.

millerpuckette:
My intent was only to call sys_poll_togui() when Pd is truly idle.  sys_domicrosleep() is a bit
of a misnomer, since it doesn't sleep unless nothing is incoming on any ports.  It _should_ be the case that, when there are a bunch of packets coming in (UDP or TCP), that sys_domicrosleep()
gets called repeatedly as long as the dacs aren't hungry.  So I don't know why this appears to be throttling incoming UDP packets.  Perhaps it's just because checking sys_send_dacs() return result is too CPU intensive and one should poll several times for UDP input each time around?  Or perhaps because sys_send_dacs() actually does the sleeping, which would have the effect o fthrottling incoming packets.

Spacechild1:
> Or perhaps because sys_send_dacs() actually does the sleeping, 

yes! 

`sys_pollgui() / sys_domicrosleep()` is only called *once* after every DSP tick, no matter if  `sys_send_dacs()` has slept or not. This is not a problem because for every TCP socket we read up to 4096 bytes. The problem with UDP was that we would only read a single *packet*. This is actually fixed in https://github.com/pure-data/pure-data/pull/666 where I read several UDP packets up to 4096 bytes *in total* per socket.

The actual problem is this line:

`return (sys_domicrosleep(0, 1) || sys_poll_togui());`

It means `sys_poll_togui()` is only called if `sys_domicrosleep` didn't do anything. So if you send OSC/FUDI messages to Pd in short intervals (e.g. once per ms) when DSP is on, `sys_domicrosleep` will always be busy and the GUI won't see any updates. My suggestion was to always try to poll *and* update the GUI:

> I think there's no problem with always trying to poll and update the GUI. If the GUI queue is empty, sys_poll_togui will be a no-op anyway.

millerpuckette:
I believe I did it that way because otherwise Pd could easily swamp the GUI so that it became unresponsive.  OTOH, the real problem seems to be that the way I set up polling in s_audio_pa.c  sleeps instead of returning instantly when not ready for transfer.  That was appropriate for OSS because (I think) there was no way to poll, but all the other APIs should be returning immediately when not ready.

millerpuckette:
... and now reading the comment at head of s_audio_pa.c I see that ut uses usleep because apparently that gives much lower-latency wakeups than s_inter.c gets from select() - so the fix might be to have s_aufio_pa.c call sys_microsleep but fix sys_microsleep to use usleep (Mac) or Sleep (Windows) and never give select() a nonzero timeout.  Fooling with scheduling stuff is always a bit scary though...

Spacechild1:
> Fooling with scheduling stuff is always a bit scary though...

:-) I can do some experiments. What you're suggesting could actually improve performance!

But the actual issue at hand is that  `sys_pollgui` only notifiies the GUI if there was no socket to read. I doesn't matter where `sys_pollgui` is called.

I think that if there's something in the GUI queue, it has to be sent anyway... To be more exact: I don't see why the ability to *send* should depend on the ability to *receive*. 

If you're worried that there might be too frequent GUI updates, we can add some throttling, e.g. `sys_poll_togui()` will be called if either 
a) `sys_domicrosleep(0, 1) `returned 0 or 
b) `sys_poll_togui()` hasn't been called for N times.

On the other hand, could it be that the shortcircuiting in
`return (sys_domicrosleep(0, 1) || sys_poll_togui());`
is only by accident?

Spacechild1:
The more I think about, the less I think that `sys_poll_togui()` should dependent on `sys_domicrosleep(0, 1)` at all. Why should sending messages to Pd implicitly control the GUI update rate? If I send a burst of messages, the GUI buffer will fill up and it will be flushed the next time it has a chance to do so. I don't think this is better for performance than calling `sys_poll_togui()` regularly.

millerpuckette:
But it's better to call sys_poll_togui() very frequently when idle, adn rarely (but at least occasionally) when Pd is fully loaded (either through its own computation or by being blinded by net packets or incoming MIDI).  So Maybe: call it as is now, but also call it twice a second or so even if Pd is struggling.

millerpuckette:
I tried a fix on new branch "portaudio-calls-microsleep".  Hopefully this doesn't increase latency on any platform and stops GUI freezes (and packet-dropping!) when receiving thousands of UDP packets per second.

Spacechild1:
> I tried a fix on new branch "portaudio-calls-microsleep".

I left some comments. Basically, the `sys_pollgui` part looks good and you can probably merged it, as it solves the issue at hand. I just think you can set a lower time limit to maintain a somewhat reasonable framerate - even when Pd is flooded with messages.

I think the scheduler part needs more work (and it's actually not related to the issue at hand). I have some ideas I want to try out.

millerpuckette:
Yep, saw them (and I think I've got things working now).  But I can't yet reproduce the original problem - I tried sending 10K packets from one Pd to another (on same machine) and it worked fine either before or after "fixing".  Perhaps I need two machines to get the problem to show up?

Spacechild1:
Can you post your test patch?

millerpuckette:
http://msp.ucsd.edu/tmp/net-test.tgz

Spacechild1:
on my machine, your patch freezes the GUI. here's a minimal patch which shows the same behavior: [netsend-gui-freeze.zip](https://github.com/pure-data/pure-data/files/3465406/netsend-gui-freeze.zip)

As you can see, it also "works" with TCP, you just have to make sure that you don't give `sys_pollgui` a chance to notify the GUI.

I guess you're testing on Linux + ALSA? Maybe it only concerns the portaudio backend (for whatever reason)

PR #641, opened 2019-05-27
Disable macOS Portmidi rate limit

danomatika:
This PR is an updated version of #639 and is included as a patch to the Portmidi sources:

> portmidi added a speed limit for outgoing MIDI messages on macOS to prevent possible data loss because of an inofficial hard-coded 15 kB/s limit in CoreMidi (for feedback protection on the IAC). Unfortunately, this makes it impossible to continously send large amout of MIDI data to USB MIDI devices. It is not entirely clear if CoreMidi only drops messages on the IAC, but it seems so.
> 
> See this massive thread for more information: #581
> 
> This PR does the following:
> 
> a) patch portmidi so we can fully disable the rate limit for macOS with a compile time switch (-DLIMIT_RATE=0). the default is #define LIMIT_RATE 1
> b) compile Pd without the rate limit to get the original behavior before the portmidi update. We might want to add a warning somewhere about possible message loss.

### DIFF
--- a/portmidi/Makefile.am
+++ b/portmidi/Makefile.am
@@ -26,8 +26,9 @@ libportmidi_a_SOURCES += \
     portmidi/pm_linux/pmlinuxalsa.c
 endif
 
+# disable portmidi rate limiting on macos
 if MACOSX
-AM_CPPFLAGS += -I$(top_srcdir)/portmidi/portmidi/pm_mac
+AM_CPPFLAGS += -I$(top_srcdir)/portmidi/portmidi/pm_mac -DLIMIT_RATE=0
 libportmidi_a_SOURCES += \
     portmidi/porttime/ptmacosx_mach.c \
     portmidi/pm_mac/pmmac.c \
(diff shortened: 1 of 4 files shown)


danomatika:
Ping @Spacechild1

danomatika:
@rbdannenberg Here is Christof's modifications to allow overriding LIMIT_RATE in patch form.

Spacechild1:
cool, thanks!

danomatika:
@ch-nry @Spacechild1 Here isa build from this branch if y'all can do a quick confirmation test: http://docs.danomatika.com/pdbuilds/0.50/Pd-0.49-2-no-midi-limit.zip

ch-nry:
the midi speed is ok in this build, but I get some problem with midi in that get stuck (and need to be reset), did you apply the patch that address the midi in bug?

danomatika:
>did you apply the patch that address the midi in bug?

No. This is only the changes in this PR which disable the Poertmidi rate limit. If your output is ok, then I'd take it that the patch works.

ch-nry:
yes, output is ok.

Spacechild1:
merged with 3a68a948f2301b2aea6abffb6b4b52fccedff5ec

Issue #622, opened 2019-05-08
literal '\' freeze iemguis


porres:
If I try to name a label with a literal backlash, Pd freezes

<img width="254" alt="Screen Shot 2019-05-08 at 17 33 11" src="https://user-images.githubusercontent.com/5360142/57406649-aef8b700-71b7-11e9-9533-8f875d9402df.png">

here's a test patch, click on the message and watch Pd become unresponsive

[crash-test.pd.zip](https://github.com/pure-data/pure-data/files/3159202/crash-test.pd.zip)

Spacechild1:
a single escaped backslash seems to generally cause weirdness. try `[send \\(` for example. I agree that it should be possible to display such a symbol. I think this could be part of the iemgui update.

porres:
It seems it can't handle literal characters in general inside a label...

For instance, this gets expanded

<img width="202" alt="Screen Shot 2019-05-09 at 14 23 33" src="https://user-images.githubusercontent.com/5360142/57473429-30a71e00-7266-11e9-82e3-c787dac6acc7.png">

porres:
> It seems it can't handle literal characters in general inside a label...

Not true, not "in general", this works, for instance

<img width="249" alt="Screen Shot 2020-04-16 at 14 22 37" src="https://user-images.githubusercontent.com/5360142/79486880-c0645900-7fed-11ea-9b8f-2cebe2113e99.png">

porres:
> this works, for instance

not fully... well, many issues in iemguis and I'm trying hard to make sense of it.

- literal \ freezes 
- literal $0 gets expanded, but this is a separate issue and not a "general" issue related to the above
- literal things like comma kinda work, but they trigger a tcl error when you try to open the properties dialog afterwords, and this seems to be yet another unrelated issue (see image below for the error)

<img width="513" alt="Screen Shot 2020-04-16 at 14 27 19" src="https://user-images.githubusercontent.com/5360142/79487296-6ca63f80-7fee-11ea-872b-f2004c2c73b5.png">




 even though it is unrelated to a literal $0 getting expanded in the label

porres:
> literal $0 gets expanded, but this is a separate issue and not a "general" issue related to the above

it's probably related to this https://github.com/pure-data/pure-data/issues/558

porres:
> I think this could be part of the iemgui update.

I'm trying to organize all issues related to updating iemguis, I edited this issue and included more issues under a newly named issue regarding escaping characters in general.

[...]

PR #647, opened 2019-06-03
escape curlies

### LINKED ISSUE #615: [print] adds a backlash
Hi, when using '\' to escape a character like ',' or something, the [print] object actually adds another backlash that's not in there. I think this is a bug, right? Or is there a reason why it has to be this way?

Screen Shot 2019-05-02 at 13 41 01

umlaeute:
this PR is to allow the printout of objects/msgboxes/comments/symbolboxes that contain curly braces without locking up Pd.

A simple test-patch to lock up Pd is like this:

~~~
[123(
|
[list tosymbol]
|
[symbol\
~~~

this (displaying an open curly "`{`" on a patch-canvas) locks up the receiving thread in the GUI, until a matching `}` is received.
to the user it just locks up Pd.
if the patch is currently "dirty", Pd won't react on <kbd>Ctrl</kbd>-<kbd>Q</kbd> anymore (as this requires the GUI to pop up a window), although it would react on <kbd>Ctrl</kbd>-<kbd>Shift</kbd>-<kbd>Q</kbd> (but not everybody knows about this).

what the patch does:
- make the already existing `pdgui_strnescape()` non-static, so it can be used for escaping object-labels.
  - this also adds a new argument to the `strnescape()` function for the size of the destination buffer (rather than just assuming that the destionation buffer is big enough; you can still pass `0` as the size for the old behaviour)
 - as a quick recap: the `strnescape` function replaces `('{', '}', '\')` with `('\{', '\}', '\\')` and is already used for escaping curlies and backslashes for Pd-console messages.
- use this `pdgui_strnescape()` when creating/renaming objects (via `rtext_senditup()`)
   - i'm sure this could be done more efficiently, as `rtext_senditup` already mangles the buffer.
- once the label has been received successfully on the GUI-side, it is unescaped
   - this now also happens for messages to the Pd-console (prior, printouts where escaped before sending them up, but never unescaped; Closes #615).

### DIFF
--- a/src/g_rtext.c
+++ b/src/g_rtext.c
@@ -186,6 +186,8 @@ static void rtext_senditup(t_rtext *x, int action, int *widthp, int *heightp,
     int inindex_c = 0;
     int selstart_b = 0, selend_b = 0;
     int x_bufsize_c = u8_charnum(x->x_buf, x->x_bufsize);
+    char smallescbuf[400], *escbuf = 0;
+    size_t escchars = 0;
         /* if we're a GOP (the new, "goprect" style) borrow the font size
         from the inside to preserve the spacing */
     if (pd_class(&x->x_text->te_pd) == canvas_class &&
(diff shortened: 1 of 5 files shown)


Spacechild1:
> use this pdgui_strnescape() when creating/renaming objects (via rtext_senditup())

does this mean we can type curly braces in object boxes and symbol atoms? (for some reason, you already can in [text define])

umlaeute:
not that this PR does not fix #622, even though it is closely related. (i gave up, not wanting to put the same code again and again in about 60 or so places; iemguis ought to be modularized first)

umlaeute:
@Spacechild1 this PR only fixes the display of curly braces. it's "unrelated" (well, it's a prerequisite) to the *input* of such characters.

so to answer your question: no.

umlaeute:
merged with c037533e2f3515598f17c0a339f54e1258c32c0a

PR #653, opened 2019-06-10
Use .gitignore trick to include empty 'm4/generated' directory

jeremiahrose:
This allows the project to be built using autotools without running autogen.sh.
This makes it easier to package Pure Data for Buildroot (for example) without having
to define post-extraction hooks to manually create the 'm4/generated' directory.

Tested on Ubuntu 18.04, normal build with autogen.sh works.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@ configure
 config.log
 config.status
 libtool
-m4/generated
 m4/config
 makefile.dependencies
 dist
--- a/autogen.sh
+++ b/autogen.sh
@@ -2,12 +2,6 @@
 
 PATH=/sw/bin:$PATH
 
-PWD=${0%/*}
-
-## git cannot really handle empty directories
-## so let's create the missing ones
-mkdir -p ${PWD}/m4/generated
-
 case `uname -s` in
     MINGW*)
         # autoreconf doesn't always work on MinGW
--- a/m4/generated/.gitignore
+++ b/m4/generated/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore


umlaeute:
you don't need to run `autogen.sh` if you want to create the directory `m4/generated` (just run `mkdir` instead).

or put otherwise: `autogen.sh` does more things than just creating this directory for you (most notably it also runs `autoreconf`)

jeremiahrose:
Yes you're right - currently the puredata package I am making for Buildroot runs a post-extraction script to generate `m4/generated` manually (it doesn't run `autogen.sh` because `autoreconf` is usually run directly by Buildroot). But it isn't the most elegant solution - having the empty `m4/generated` dir already there is simpler.

jeremiahrose:
Or to put it another way, when `autoreconf` is run on a freshly downloaded repo, it fails with an error.

umlaeute:
sure, but the `INSTALL.txt` tells you to run `./autogen.sh`, not `autoreconf`.

(there's a number of things that you cannot run from a fresh clone, including `make` and `configure`).

danomatika:
If the dir is *really* required, using `.gitkeep` would be simpler.

danomatika:
Also, this would be moot if the source release packages were created with `make list` which has everything generated and ready to go, but that's another topic...

umlaeute:
but with `.gitkeep` the directory would be marked as "dirty" as soon as autogen puts things into it.

umlaeute:
i do not know which autotools version was used back then, but here `autoreconf -i` will create the `m4/generated/` directory just fine.

~~~cmd
~$ git clone https://github.com/pure-data/pure-data
Cloning into 'pure-data'...
remote: Enumerating objects: 47375, done.
remote: Counting objects: 100% (1406/1406), done.
remote: Compressing objects: 100% (602/602), done.
remote: Total 47375 (delta 1107), reused 835 (delta 804), pack-reused 45969 (from 3)
Receiving objects: 100% (47375/47375), 73.20 MiB | 10.77 MiB/s, done.
Resolving deltas: 100% (35342/35342), done.
zmoelnig@xenakis:/dev/shm$ cd pure-data/

~/pure-data$ ls m4/generated
ls: cannot access 'm4/generated': No such file or directory

~/pure-data$ autoreconf -fiv
autoreconf: export WARNINGS=
autoreconf: Entering directory '.'
autoreconf: configure.ac: no obvious need to run autopoint
autoreconf: running: aclocal --force -I m4/generated -I m4
aclocal: warning: couldn't open directory 'm4/generated': No such file or directory
autoreconf: configure.ac: tracing
autoreconf: configure.ac: creating directory m4/config
autoreconf: running: libtoolize --copy --force
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'm4/config'.
libtoolize: copying file 'm4/config/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'm4/generated'.
libtoolize: copying file 'm4/generated/libtool.m4'
libtoolize: copying file 'm4/generated/ltoptions.m4'
libtoolize: copying file 'm4/generated/ltsugar.m4'
libtoolize: copying file 'm4/generated/ltversion.m4'
libtoolize: copying file 'm4/generated/lt~obsolete.m4'
autoreconf: configure.ac: not using Intltool
[...]
autoreconf: Leaving directory '.'

~/pure-data$ ls m4/generated
libtool.m4  ltoptions.m4  ltsugar.m4  ltversion.m4  lt~obsolete.m4

~~~

umlaeute:
which i take to mean that you just need to add something like this to your buildroot config:
~~~
PUREDATA_AUTORECONF_OPTS=-fi
~~~

PR #654, opened 2019-06-11
hide [vu]'s outlets

porres:
This is one solution that came up as a way to deal with what's discussed in https://github.com/pure-data/pure-data/issues/632 and closes it.

The issue is that we don't have a send symbol to [vu] and that adding it would be too clumsy. Therefore, this PR does not actually add any send symbol, but at least deals with an annoyance that we cannot hide [vu]'s outlets, just the inlets.

In practice, this hides the outlets whenever the inlets are also hidden. The ideal solution would be to have a send symbol, but this seems just more sensible.

This is also a better solution than just hiding the outlets by default and not document them as suggested in https://github.com/pure-data/pure-data/pull/634


porres:
no idea why AppVeyor build failed - help?

Spacechild1:
looks like a server error. you try to close and reopen the PR, this will trigger a rebuild.

Spacechild1:
there you go :-)

porres:
> there you go :-)

Yay! Thanks :)

Issue #578, opened 2019-04-02
Tcl/Tk Error with Pd-0.49


oliverstotz:
sorry for a not so precise bug-report, but it's all i can offer:

recently i discovered a strange behaviour in Windows PD 0.49 (32bit) when the closing of a patch window (or even the closing of a subpatch window) would result in a Tcl/Tk error that i have not excperienced up until PD 0.49.

so, i close a window (main or subpatch - mostly with STRG+w), the patch window turns white with some inlet relics still being displayed and the console prints out a Tcl/TK error message in red colour.

the console says for example this:

~~~
(Tcl) UNHANDLED ERROR: bad window path name ".x3b212c8"
    while executing
"wm title $mytoplevel "$name$dirtychar$arguments - $path""
    (procedure "pdtk_canvas_reflecttitle" line 15)
    invoked from within
"pdtk_canvas_reflecttitle .x3b212c8 {D:/pd_0.49/_MYPATCHES_/MFPOW} {metronome} { [edit]} 0"
    ("uplevel" body line 718)
    invoked from within
"uplevel #0 $docmds".x2596028: no such object
.x2596028: no such object
.x2596028: no such object
.x2596028: no such object
~~~

......................................  ({metronome} is the name of a subpatch with GOP enabled)

from this point on nothing can be done in PD anymore, the only way to get out of this is to kill the PD process from the outside. These "crashes" seem to occur completely random, so unfortunately i can not supply an example patch, but i noticed this:

the patch contains several GOP subpatches and also some of my own GUI abstractions, that contain GOPs. the "crash" only happens when a window is closed.

(the patch is very involved so i couldn't really check what would happen, if i replaced all of my abstractions with something else ...)

Spacechild1:
probably related: https://sourceforge.net/p/pure-data/bugs/1290/

bobczz:
I also encounter this error many times when I use and/or modify GOP, frankly I've never see it before 0.49. I hope someone finds a solution.


[TCL error.txt](https://github.com/pure-data/pure-data/files/3133104/TCL.error.txt)

umlaeute:
i acknowledge the issue (having seen it at least once), but we really need to be able to reproduce it, before we can fix it.

so if anybody can describe how they got the error, *please* share this info.

so far i only know that it is related to *edit*ing *GOP*s (i doubt that merely *using* them - as indicated by @bobczz - is enough to trigger the issue)

umlaeute:
[...]

PR #658, opened 2019-06-12
[WIP] only call canvas_redraw() if we do have a window

umlaeute:
don't call it, if there's no window but we are visible due to GOP

might close (or not) https://github.com/pure-data/pure-data/issues/578

want review from @danomatika and/or @Spacechild1

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -759,7 +759,8 @@ int canvas_undo_cut(t_canvas *x, void *z, int action)
                 }
             }
                 /* LATER disable redrawing here */
-            canvas_redraw(x);
+            if (x->gl_havewindow)
+                canvas_redraw(x);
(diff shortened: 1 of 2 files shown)


src/g_editor.c
```diff
@@ -1449,7 +1447,8 @@ int canvas_undo_canvas_apply(t_canvas *x, void *z, int action)
             glist_noselect(x);
             gobj_vis(&x->gl_gobj, x->gl_owner, 0);
             gobj_vis(&x->gl_gobj, x->gl_owner, 1);
-            canvas_redraw(x->gl_owner);
+            if (x->gl_havewindow)
```
Spacechild1: shouldn't this be `x->gl_owner->gl_havewindow`?

umlaeute: yes. 2c7b4c1937c0e57d24d48de3d83969004e080bab

PR #661, opened 2019-06-17
[poly] features

Spacechild1:
These are my personal patches to the [poly] object. I've found them so useful that I think they could make it into Pd vanilla.

* dynamically set number of voices
if the new number is smaller than the old one, excess voices are flushed, 
otherwise existing voices are left untouched.
* dynamically enable/disable voice stealing
* add `[sustain <f>(` message which mimicks a piano sustain pedal. 
`[sustain 1(` - note-offs are sustained
`[sustain 0(` - all sustained notes are flushed
* add `[stack <f>(` message which enables/disables stacking of sustained notes of the same pitch
`[stack 0(` - can't sustain more than one voice of the same pitch (like a piano)
`[stack 1(` - sustain as many notes of the same pitch as you like (like many polyphonic synthesizers)

### DIFF
--- a/src/x_midi.c
+++ b/src/x_midi.c
@@ -979,11 +979,13 @@ static void stripnote_setup(void)
 
 static t_class *poly_class;
 
+#define PLAY 1
+#define SUSTAIN 2
+
 typedef struct voice
 {
     t_float v_pitch;
-    char v_used;
-    char v_sustain;
+    char v_state;
     unsigned long v_serial;
 } t_voice;
 
@@ -997,7 +999,7 @@ typedef struct poly
     t_outlet *x_velout;
     unsigned long x_serial;
     char x_steal;
-    char x_retrigger;
+    char x_stack;
     char x_sustain;
 } t_poly;
 
@@ -1006,8 +1008,7 @@ static void poly_flushvoice(t_poly *x, t_voice *v, int index)
     outlet_float(x->x_velout, 0);
     outlet_float(x->x_pitchout, v->v_pitch);
     outlet_float(x->x_obj.ob_outlet, index);
-    v->v_used = 0;
-    v->v_sustain = 0;
+    v->v_state = 0;
     v->v_serial = x->x_serial++;
 }
 
@@ -1022,7 +1023,7 @@ static void poly_resize(t_poly *x, t_float fnvoice)
     {
         /* flush excess voices */
         for (i = n, v = x->x_vec + n; i < x->x_n; i++, v++)
-            if (v->v_used)
+            if (v->v_state)
                 poly_flushvoice(x, v, i+1);
     }
 
@@ -1031,7 +1032,7 @@ static void poly_resize(t_poly *x, t_float fnvoice)
     if (n > x->x_n)
     {
         for (i = x->x_n, v = x->x_vec + x->x_n; i < n; i++, v++)
-            v->v_pitch = v->v_sustain = v->v_used = v->v_serial = 0;
+            v->v_pitch = v->v_state = v->v_serial = 0;
     }
 
     x->x_n = n;
@@ -1047,7 +1048,7 @@ static void *poly_new(t_symbol *s, int argc, t_atom *argv)
     x->x_vel = 0;
     x->x_sustain = 0;
     x->x_steal = (atom_getfloatarg(1, argc, argv) != 0);
-    x->x_retrigger = argc > 2 ? atom_getfloat(argv+2) : 1; /* default: true */
+    x->x_stack = argc > 2 ? (atom_getfloat(argv+2) != 0) : 1; /* default: true */
     floatinlet_new(&x->x_obj, &x->x_vel);
     outlet_new(&x->x_obj, &s_float);
     x->x_pitchout = outlet_new(&x->x_obj, &s_float);
@@ -1056,7 +1057,7 @@ static void *poly_new(t_symbol *s, int argc, t_atom *argv)
 
     x->x_vec = (t_voice *)getbytes(x->x_n * sizeof(*x->x_vec));
     for (v = x->x_vec, i = x->x_n; i--; v++)
-        v->v_pitch = v->v_sustain = v->v_used = v->v_serial = 0;
+        v->v_pitch = v->v_state = v->v_serial = 0;
 
     return (x);
 }
@@ -1068,23 +1069,30 @@ static void poly_float(t_poly *x, t_float f)
     t_voice *firston, *firstoff;
     unsigned int serialon, serialoff, onindex = 0, offindex = 0;
 
-        /* turn off oldest matching voice for
-         * a) note-off message
-         * b) sustain + retriggering disabled */
-    if (x->x_vel <= 0 || (x->x_sustain && !x->x_retrigger))
+        /* note-off message */
+    if (x->x_vel <= 0)
     {
         for (v = x->x_vec, i = 0, firston = 0, serialon = 0xffffffff;
             i < x->x_n; v++, i++)
-                if (v->v_used && v->v_pitch == f && v->v_serial < serialon)
+                if (v->v_state == PLAY && v->v_pitch == f && v->v_serial < serialon)
                     firston = v, serialon = v->v_serial, onindex = i;
         if (firston)
         {
-                /* note-off only: mark voice as sustained */
-            if (x->x_vel <= 0 && x->x_sustain)
-                firston->v_sustain = 1;
+            if (x->x_sustain)
+                firston->v_state = SUSTAIN;
             else
                 poly_flushvoice(x, firston, onindex+1);
         }
+    }
+        /* prevent stacking of sustained notes */
+    else if (x->x_sustain && !x->x_stack)
+    {
+        for (v = x->x_vec, i = 0, firston = 0, serialon = 0xffffffff;
+            i < x->x_n; v++, i++)
+                if (v->v_state == SUSTAIN && v->v_pitch == f && v->v_serial < serialon)
+                    firston = v, serialon = v->v_serial, onindex = i;
+        if (firston)
+            poly_flushvoice(x, firston, onindex+1);
     }
 
     if (x->x_vel > 0)
@@ -1093,9 +1101,9 @@ static void poly_float(t_poly *x, t_float f)
         for (v = x->x_vec, i = 0, firston = firstoff = 0,
             serialon = serialoff = 0xffffffff; i < x->x_n; v++, i++)
         {
-            if (v->v_used && v->v_serial < serialon)
+            if (v->v_state && v->v_serial < serialon)
                 firston = v, serialon = v->v_serial, onindex = i;
-            else if (!v->v_used && v->v_serial < serialoff)
+            else if (!v->v_state && v->v_serial < serialoff)
                 firstoff = v, serialoff = v->v_serial, offindex = i;
         }
 
@@ -1104,8 +1112,7 @@ static void poly_float(t_poly *x, t_float f)
             outlet_float(x->x_velout, x->x_vel);
             outlet_float(x->x_pitchout, firstoff->v_pitch = f);
             outlet_float(x->x_obj.ob_outlet, offindex+1);
-            firstoff->v_used = 1;
-            firstoff->v_sustain = 0;
+            firstoff->v_state = PLAY;
             firstoff->v_serial = x->x_serial++;
         }
             /* if none, steal one */
@@ -1115,11 +1122,7 @@ static void poly_float(t_poly *x, t_float f)
             outlet_float(x->x_velout, x->x_vel);
             outlet_float(x->x_pitchout, firston->v_pitch = f);
             outlet_float(x->x_obj.ob_outlet, onindex+1);
-            firston->v_used = 1;
-        #if 0 /* see poly_flushvoice */
-            firston->v_sustain = 0;
-            firston->v_serial = x->x_serial++;
-        #endif
+            firston->v_state = PLAY;
         }
     }
 }
@@ -1129,9 +1132,9 @@ static void poly_steal(t_poly *x, t_float f)
     x->x_steal = (f != 0);
 }
 
-static void poly_retrigger(t_poly *x, t_float f)
+static void poly_stack(t_poly *x, t_float f)
 {
-    x->x_retrigger = (int)f;
+    x->x_stack = (f != 0);
 }
 
 static void poly_sustain(t_poly *x, t_float f)
@@ -1142,7 +1145,7 @@ static void poly_sustain(t_poly *x, t_float f)
     if (!f) /* sustain released */
     {
         for (v = x->x_vec, i = 0; i < x->x_n; v++, i++)
-            if (v->v_sustain)
+            if (v->v_state == SUSTAIN)
                 poly_flushvoice(x, v, i+1);
     }
     x->x_sustain = (f != 0);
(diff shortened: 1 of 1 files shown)


porres:
a sustain feature is terribly missing in vanilla :) - highly welcome!

porres:
hmm, it seems you'd need to flush the voices when resizing, huh?

Spacechild1:
> hmm, it seems you'd need to flush the voices when resizing, huh?

Resizing simply clears all voices. If you want to flush all voices, you just have to send [stop( before [resize(.

porres:
forgot baour "stop". But what would be the point to not use "stop"? Cause it's possible that notes will be left hanging

Spacechild1:
[poly] also has a [clear( message because sometimes you don't want to flush notes. My thinking was that it's easier to flush notes manually when needed than preventing undesired flushed notes.

pkstone:
This is an excellent, much-needed enhancement to [poly] (i.e., dynamic number of voices).

porres:
> sometimes you don't want to flush notes

like when? can't think of a use case

Spacechild1:
honestly, I don't know, I just assume there must be some use case :-) Another alternative would be to neither flush nor clear voices and instead only flush superfluous notes when resizing to a lower number of voices. Someone on the mailing list has brought this up and it's something you can't easily achieve otherwise, so I might give it a shot.

porres:
> Another alternative would be to neither flush nor clear voices and instead only flush superfluous notes when resizing to a lower number of voices. 

That's what I also thought. I do offer voice resizing in [else/voices], but I just flush the notes. I thought it was too hard to flush the extra notes :/ 

But if you achieve this, I'll check how you did it and try to do the same for me :)

jaffasplaffa:
"dynamically set number of voices" is a must! From one from here :) 

I miss that a lot, the ability to set number of voices with for example a message.

alfonso73:
@Spacechild1 yeah! I strongly agree with your suggestions. For poly instrument building they would be essential additions to vanilla's [poly].

jaffasplaffa:
Hey again :) 

Am I correct to understand that to change number of voices dynamically was implemented?

How does one actually do it?

Thanks in advance!

Spacechild1:
> Am I correct to understand that to change number of voices dynamically was implemented?

This PR hasn't been merged (yet).

BTW, I just realized I need to resolve merge conflicts.

jaffasplaffa:
Ah okay. I was looking for way to do it and didn't find it, so thought I'd ask.

Thanks for making it, I guess we have to wait a bit more for the merge :)

Spacechild1:
Will fix conflicts once there is interest in merging.

pkstone:
I’m not sure why I’m cc’d on this.


Best,

Phil

porres:
> Ok @Spacechild1 I'm not sure how to test 'stack', can you help me?

Ok, I got it, it sends a note off!

porres:
you know, for ELSE, I have a separate [sustain] object, and I also have [voices] that replaces [poly]. So I'm basically thinking of extra ideas they have but I know that maybe it's just too much and I won't pitch them all (and I'll keep them for feature freaks).

That is to say I don't need this myself and that I'm trying to think of what is actually important. So retriggering seems cool and one extra thing could be a "release time". What it is it for?

This makes [poly] hold its voice for a given time that would, in principle, match the release time of an envelope. 

Or maybe that's too much for [poly] and it's ok I have it in [voices]... I'm not sure. Just leaving this on the table...

millerpuckette:
I'm trying to keep objects single-purpose to the extent possible - note that 'bag' in vanilla can be used as a sustain pedal (and for other purposes as well I hope).  Having an extra decay time period is a good idea, but since different voices would often have different decay times we would want a way for the voice to communicate back to the poly object.  To make this easier one could add messages to ploy allowing to address specific voice numbers, for instance to undo a note-off and then to redo it later.  Or something like that.

porres:
> I'm trying to keep objects single-purpose to the extent possible - note that 'bag' in vanilla can be used as a sustain pedal (and for other purposes as well I hope).

To be honest, I would expect sustain to be treated externally, with a new object. But now this is in [poly] :) and I guess we can have sustain and how it manages stacked/repeated notes on a new object. I'd be fine with that, but we'd need now to remove this from poly and create a new object. 

As for my other suggestions, they weren't about the sustain mode, but more of how poly generally works. The retriggering wouldn't care if the note was sustained or not. If the current note repeats the last one, then it retriggers on the same voice number, and this basically prevents the same note from being set to a release phase on a voice allocation while it attacks on a new one. This can't be dealt in an external object.

> but since different voices would often have different decay times we would want a way for the voice to communicate back to the poly object

I say usually it is the same envelope for all voices, but hey, different release for each would be fine by me as well :) 

As for "communicating back", the way I see it is that you have to set the release time of an envelope and send the same setting value as a message to [poly].

> for instance to undo a note-off and then to redo it later. Or something like that.

I didn't get that though

Spacechild1:
> I can confirm and see that 'stack' only works if the sustain mode is on. Not only that, but for notes that have been sustained - that is, whose note off message has been received and 'held',

This is entirely on purpose. The 'stack' method solves a very specific issue:

What should happen if you hold the sustain pedal and repeatedly press and release the same note:
1. every new sustained note is stacked on top of previous sustained notes (of the same pitch)
2. a new sustained note cancels a previous sustained note (of the same pitch)

Either answer can be correct, depending on the use case. Since the sustaining is handled internally by `[poly]`, we need to tell it what to do. That's what the 'stack' method does.

> meaning that a repeated note on message without a note off does not get on the radar in "sustain 1, stack 0" mode.

This can't happen if you play a keyboard instrument, which is what I had in mind for the 'stack' mode. 

---

> So, "retrigger 1" ('mode 1') would just retrigger (resend) the note on message on the same voice allocation. And "retrigger 2" ('mode 2') would be 'stack', which would retrigger it on a new voice allocation, like it already does but sending a note off message first.

'mode 2' cannot be done in `[poly]` because the original note-off message of the first (overlapping) note would accidentally stop the second note. If you don't want overlapping notes of the same pitch, this has to be handled *before* `[poly]`.

'mode 1' is actually what I would expect 'retrigger' to do. It sounds like an interesting use case, but IMO it's not really related to 'stack'. Feel free to make a feature request!

porres:
I agree that sustain logic—whether integrated into [poly] or externally—should handle repetitions, either by stacking and holding note-offs or freeing them first. For me, this is a retriggering issue. 

In ELSE’s [suspedal], I have four modes:
- mode 0: ignore repeated notes (no retrigger),
- mode 1: repeat note-on and free one at the end (I don't see the point, really, but I copied this from cyclone cause it had it),
- mode 2: free sustained note of same pitch (also in [cyclone/sustain])
- mode 3: stack and release all later. (also in [cyclone/sustain])

> This can't happen if you play a keyboard...

But in Pd, it can—like with [makenote], which may send repeated notes without note-offs.

> Once you send overlapping notes of the same pitch, you are obviously interested in stacking behavior.

Not sure what you mean, really. Anyway, I don't know what is "obvious" in Pd. What I'd expect, though, is that [poly] would reuse the same voice number for repeated pitches. It only does so if that’s the only available voice. Otherwise, it allocates a new one. I wish this retrigger behavior could be addressed too, even if it's separate from sustain.

> 'mode 2' cannot be done in [poly]

Actually, that’s exactly what [poly] does—it always allocates a new voice (unless one is forced by availability). So [poly] does "mode 2" by default, as I see it.

> Not really related to 'stack'

True, I guess, but I’m analyzing retriggering separately, both with and without sustain. Your view seems to intertwine the two, but I see that as a retriggering issue, and it is. ELSE’s [voices], by the way, does no sustaining, but also has retriggering options, where repeated notes can go to the same voice or another. 

So, if sustain and voice handling were separate, we still could address repeated notes in [poly]. Combining both functions into one object could reduce complexity and options, as ELSE has 4 retriggering modes with [suspedal] + 3 modes with [voices] - that gives us many combinations! But looking at all this again, maybe I think it can all be much simpler with [poly] in another way.

So what does [poly] currently do? It repeats notes without freeing them, and into new voice numbers, of course. So a retrigger mode that forces freeing the voice makes sense. So what I think is: 

retrig 0 (default): don't free,  move to next.
retrig 1 ("same"): free and repeat in the same voice.
retrig 2 ("next"): free and move to a new voice.

These modes handle repeated note-ons, regardless of sustaining. Now, combining with sustain, we can forget about the "stack" message, and we'd have:

retrig 0 = like using with "stack 1".
retrig 1 = same as stack 0 (but repeating on same voice)
retrig 2 = like "stack 0" is now.

So, I' proposing to:
1) Handle computer-triggered repeated notes (e.g. from [makenote]).
2) Add a retrig message with 3 modes (0/1/2) and remove "stack".
3) Possibly move sustain logic to its own object (like my [bag]-based abstraction), which stacks note offs by default and wouldn't need to have any other setting because this is taken care of in [poly]... and when stacked note-offs get release, [poly] would just ignore extra note-offs.

I shold get my hands into the code and see if I can put this to test on a new PR, but I hope I am clear

millerpuckette:
OK, I suggest we get rid of all the "sustain" and "stack" stuff but keep the dynamic number of voices and steal/nosteal behavior.  Sometime later, we can do it a better way I think: messages to query which voice a given pitch corresponds to (allowing either the oldest or the newest one as a message flag in case of multiples) and to alter the status of a specific voice (for instance, if the corresponding synth voice wants to delay being turned off).

Spacechild1:
> OK, I suggest we get rid of all the "sustain" and "stack" stuff

😢 Can we at least add an example to `bag-help.pd` for how to implement a sustain pedal (and add a link to `poly-help.pd`)? I think this is not entirely obvious to many people.

Note that `[bag]` currently only allows to build a sustain pedal *with* stacking because it keeps copies of the same value. If you don't want note stacking, you're out of luck.

Here's an idea: what if we add an option to `[bag]` so that it does not accumulate? I guess that might be useful for other things as well.

The only thing I really care about is that we are *somehow* able to build a simple piano sustain pedal in Pd without any external objects.

millerpuckette:
Perhaps a more generally powerful ability would be to be able to query bag without changing its state (same for poly).
But adding an element only if it's not already there does sound pretty elemental too.

Another thing I would add if I were working on this just now would be a separate method to flush the output in the order it was input, not the reverse order that now holds.

porres:
I'd hate to be the one who made the sustain get out of [poly] and then we stay with nothing :) I wouldn't rest until it'd be back one way or another.

I wouldn't mind having "sustain" in [poly], at least it's there, but I have to say I think it makes sense to have it as a dedicated object. Though I believe the usage on its own is rare in practice, and I may have never used it like that. Only thing coming to mind now is sending MIDI from Pd where you can set notes to sustain or not.

> Note that [bag] currently only allows to build a sustain pedal with stacking because it keeps copies of the same value. If you don't want note stacking, you're out of luck.

That's a good catch, it'd ruin the idea of allowing retrigerring of repeated notes as things are.


> Here's an idea: what if we add an option to [bag] so that it does not accumulate? I guess that might be useful for other things as well. 

that'd be the least to do to offer an abstraction, but maybe this should grant its own coded object.

> Can we at least add an example to bag-help.pd for how to implement a sustain pedal (and add a link to poly-help.pd)? I think this is not entirely obvious to many people.

I agree it's not obvious and I have to say it's not trivial at all. It's tru [bag] can be used as one of the foundations to implement a sustain pedal, but you also need [stripnote] and more stuff, at least that's how I did it...

I'd rather have a ready made and proper sustain object, an "extra" abstraction or not, whatever. But having it as an abstraction based on [bag] right now will miss "stack 0".

> The only thing I really care about is that we are somehow able to build a simple piano sustain pedal in Pd without any external objects.

I'd really care about having it as a ready made object. If hacking Pd with [bag] or whatever to build a sustain pedal, I bet you can do something and find another solution without [bag] (btw, I wouldn't see any reason to use this besides implementing a sustain pedal).

If you think this grants its own coded object, I can suggest my work with [else/suspedal] as a reference. See https://github.com/porres/pd-else/blob/master/Source/Control/suspedal.c

Besides basic sustaining and maybe too many retriggering modes, it has a very cool thing that I don't know why is so overlook in music synthesis: I also implemented the "third" piano pedal, with is the tonal sustain pedal. When activated, it only holds the keys that are currently pressed, then you can make it hold a chord use both hands to play on top of it...

Maybe it's considered feature bloated, so if a basic/common sustain pedal is desired and good enough, well, so be it.

cheers

Spacechild1:
> Perhaps a more generally powerful ability would be to be able to query bag without changing its state (same for poly).

I just realized that we would actually need such a method for implementing a sustain pedal without note stacking!

> But adding an element only if it's not already there does sound pretty elemental too.

> Another thing I would add if I were working on this just now would be a separate method to flush the output in the order it was input, not the reverse order that now holds.

I will try to make a PR for all of these. But only after I finish the other stuff :)

umlaeute:
btw, there's already https://github.com/pure-data/pure-data/pull/2425 which implements a
> bang method to `[bag]` for flushing without emptying

Spacechild1:
> bang method to [bag] for flushing without emptying

+1

@millerpuckette Please also add a "unique" method so we can set the behavior at runtime!

millerpuckette:
> > bang method to [bag] for flushing without emptying
> 
> +1
> 
> @millerpuckette Please also add a "unique" method so we can set the behavior at runtime!

I thought about that.. but then you'd have to go through and clean out multiples when 'unique' got turned on (or maybe offer teh choice) and it turned into feature creep again.  And I can't think of a use case besides simplifying the help
window :)

Spacechild1:
> And I can't think of a use case besides simplifying the help
window :)

If I implement a sustain pedal, I want to change the behavior of `[bag]` depending on the stacking mode. Without a "unique" message I have to use two different `[bag]` objects and route the messages to the right instance...

And generally, we shouldn't limit behavior to creation arguments only, unless there is a very good reason.

> I thought about that.. but then you'd have to go through and clean out multiples when 'unique' got turned on (or maybe offer teh choice) and it turned into feature creep again. 

I see... I think we can go the lazy way and just set the `x_unique` member, as long as we document it. If we interpret "unique" as "don't add a number if it already exists", this can be justified :)

millerpuckette:
yeah, OK... I'll do it.

Spacechild1:
Just for reference, here's an abstraction that uses the new `[bag]` features to implement a sustain pedal, with or without note stacking: [csustain.zip](https://github.com/user-attachments/files/20445743/csustain.zip)

(Not purely Vanilla because I just cannot make patches without `[zexy/demux]` :)

Spacechild1:
> I just cannot make patches without [zexy/demux] :)

What I really meant is: I don't *want* to make patches without [zexy/demux] :)

Also, the `[list prepend]` trick breaks down as soon as you mix lists and "messages" because [route] unfortunately strips the selector of both... It would indeed work in this particular case, but it's not a general purpose solution.

Another workaround is to use a `[send]` and dynamically set the destination. I don't really like that either.

It would be nice if we could have something `[mux]?  and `[demux]` in Pd vanilla ;-)

porres:
Ok, so I built Pd and tried this. Here's my patch 

[susnote-bag.pd.zip](https://github.com/user-attachments/files/20649689/susnote-bag.pd.zip)

It's about the same as the one I did before based on [text]. And, well, I think it's a cleaner version. I can include this example in the help file of bag soon.

By the way, with the new "bang" method, I can see that it may be used to create an arpeggiator. Here's my patch

[arp-vanilla.pd.zip](https://github.com/user-attachments/files/20649764/arp-vanilla.pd.zip)

It also uses the "unique" flag, and what it does is that it outputs all notes in the collection into a [text] object and sorts them, so it's based on text as well. My idea is to add both examples to the help file.

Now, this might be too much to ask, but what if I want to also store the on velocity inside bag? A "clean" way to do it would be to add another outlet to output it. In this case it'd be a new mid inlet, since a right outlet was also created. This way I could use the velocity on the arpeggiator.

What do you say @Spacechild1 and @millerpuckette ?

PR #664, opened 2019-06-20
fix [text get] out of range behavior

Spacechild1:
Always output empty list + 2 if line number is out of range (not only for whole lines).

@millerpuckette I was wondering about the purpose of that explicit `else if` clause because I can't really see a reason to handle out-of-range errors for whole lines only.

### DIFF
--- a/src/x_text.c
+++ b/src/x_text.c
@@ -638,7 +638,7 @@ static void text_get_float(t_text_get *x, t_floatarg f)
             ATOMS_FREEA(outv, nfield);
         }
     }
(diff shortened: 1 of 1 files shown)


millerpuckette:
I'm not sure what the correct behavior should be - perhaps it should bang outlet 2 but _not_ output an empty list in that case (the empty list makes some sense if asking for a whole line but not for a field that's out of range).

Spacechild1:
Hmmm... I'm not sure if there's a need to distinguish between "whole line" and "some fields" in this case. Both can actually be the same.

PR #665, opened 2019-06-21
Feature/netobject updates

Spacechild1:
mostly minor bugfixes but also two "features":

* allow to receive more than 1 UDP packet per scheduler tick to match the behavior of TCP. With this fix, receiving 1000 short UDP packets (with DSP on) only takes ~1 ms instead of ~1.5 seconds (!)

* properly notify TCP clients when a socket has been closed, i.e. `[listen 0( -> [netreceive]` will cause all connected `[netsend]`s to close and output a 0. 
(currently, `[netsend]` is not notified when its connection is closed from the other side, so you get a cryptic socket error when you send it a message.)

successfully tested on Windows 7 and Debian 9. @danomatika @umlaeute

### DIFF
--- a/doc/5.reference/netsend-help.pd
+++ b/doc/5.reference/netsend-help.pd
@@ -70,12 +70,12 @@ request should specify the name or IP address of the other host and
 the port number. There should be a "netreceive" on the remote host
 with a matching port number., f 132;
 #X text 147 12 netsend -- send Pd messages over a network;
-#X msg 447 258 timeout 3;
-#X text 526 251 TCP connect timeout, f 11;
 #X text 35 554 As of 0.50+ \, Pd supports IPv6 addresses \, netsend
 -u (UDP) is fully "connectionless" and no longer closes if no one receives
 a UDP message \, and netsend (TCP) has a settable connect timeout which
 defaults to 10 seconds., f 115;
+#X msg 437 255 timeout 3000;
+#X text 532 250 TCP connect timeout (ms), f 13;
 #X connect 0 0 8 0;
 #X connect 0 1 38 0;
 #X connect 1 0 0 0;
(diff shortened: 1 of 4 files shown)


danomatika:
I thought Miller's requirement was building for Win XP?

Spacechild1:
> I thought Miller's requirement was building for Win XP?

Are you talking about -D_WIN32_WINNT=0x0600? This won't do any harm because the Windows headers do a >= check. Note that currently we already set a higher version than Windows XP (namely 0x0502 = Windows Server 2003).

Spacechild1:
I think we just have to this (compile on some higher Windows version and try to run on XP).

The problem was that your branch didn't compile on recent Msys2 because with `-D_WIN32_WINNT=0x0502` some IP6 constants (AI_ADDRCONFIG, AI_V4MAPPED) are not defined (which is actually a bug in Msys2). Alternatively, we could also define them manually like:
```
#ifndef AI_ADDRCONFIG
#define AI_ADDRCONFIG 0x00000400
#define AI_V4MAPPED 0x00000800
#endif
```

whatever is less evil.

danomatika:
The defines could work as I have refactored the networking includes and helper functions into a s_net.h & s_net.c files. I haven't finished the commit yet, but this would be a central place for handling this. We already do a define for missing error constants.

Spacechild1:
OK, after you've merged this you can just revert 9f6f8fb once you've got the `s_net.h/s_net.c` part finished. I just had to do *something* so I could compile and test the branch on my system :-)

Spacechild1:
Or I can revert it and add the defines. I'll probably have to resolve merge conflicts anyway when you push the `s_net.h/s_net.c` thing (in case you don't merge before).

danomatika:
Keep it as it is / clean up what you have. We'll do a squash in a finalized PR anyway as there are too many small commits by now anyway, largely thanks to cross-platform stuff.

danomatika:
At this point, it's easier for me to handle the merge as I moved the stuff to different files.

PR #666, opened 2019-06-22
[WIP] Feature/netobject updates

Spacechild1:
- [x] more bug fixes
- [x] update pdsend
- [x] update pdreceive
- [x] IPv4 vs IPv6 issues (see below)
- [x] update sys_do_startgui


Spacechild1:
@danomatika I've tested everything on Windows and I will test tomorrow on Mac and Linux. 

Naturally, `pdsend/pdreceive` now also have the localhost problem :-(

I have to investigate this further but the situation I'm currently observing is that on my Windows 7 machine, requests made to an IPv6 address are not handled by sockets bound to the IPv4 any address (`0.0.0.0`) and vice verca. On my system, the IPv6 addresses come first in the address list, so `[netreceive]` is bound to `::` and therefore I can't connect with `[connect 127.0.0.1 9999(`. The same happens when using the IP of my local network, I can connect with the IPv6 version, but not with the IPv4 version.

On Pd 0.49, `[netreceive]` is bound to the IPv4 any address, and localhost is resolved to `127.0.0.1`, but on Pd 0.50 `[netreceive]` is bound to the IPv6 any address, and localhost is resolved to `::1`, which basically means that Pd 0.49 can't make any TCP connection to Pd 0.50.

For UDP there is a hack: `[netreceive 9999 0.0.0.0]` actually allows to receive IPv4 messages.

Spacechild1:
OK, here's some more context: 

https://stackoverflow.com/questions/27480094/ipv6-is-equivalent-to-0-0-0-0-when-listening-for-connections

https://docs.microsoft.com/de-de/windows/desktop/WinSock/dual-stack-sockets

https://docs.microsoft.com/de-de/windows/desktop/WinSock/ipproto-ipv6-socket-options

I've tested with a couple of other apps on Windows (e.g. REAPER) and there it's the same: I can't send OSC messages to REAPER with `[connect localhost(` -> `[netsend -u -b]` (I have to explicitly use `127.0.0.1`) and I can't receive them with `[netreceive -u -b]` (unless I do `[netreceive -u -b 9999 0.0.0.0]`).

This leads me to believe that it's not a good idea for Pd to use IPv6 addresses by default and that IPv6 should be rather explicitly requested by the user. The current implementation would break most existing setups, at least on Windows. 

@umlaeute's thoughts are highly appreciated, as I am not a networking specialist.

danomatika:
Than what we could do is set the gerad sri go hunt for IP version based on the address type, however this partially negates the point of using it. Maybe there is some way to request that IPv4 addresses are first in the list?

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
> gerad sri go hunt

Autocomplete? Google led to some interesting results, but none of them computer related :-)

danomatika:
iPhone spaghetti: “getaddrinfo hint flag”

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
@danomatika: On my Windows 7 system, `getaddrinfo` (used in `addrinfo_get_list`) can't resolve "localhost" if I'm not connected to the internet (but it *does* work if I'm offline). Removing `AI_ADDRCONFIG` solves this but I'm not sure about the consequences, as I don't fully understand what it does. But the following lines seems to explain the issue I'm experiencing:

> If the AI_ADDRCONFIG bit is set, getaddrinfo will resolve only if a global address is configured [...] The IPv4 or IPv6 loopback address is not considered a valid global address. 

https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-getaddrinfo

Spacechild1:
@danomatika I tested the branch now on my old MacBook (OSX 10.11) and the situation is the same as on my Debian and Windows machines: sockets can only listen to either IPv4 or IPv6 adapters.

What was your test setup? Did `[netreceive]` really listen to both IPv4 or IPv6 adapters?

Anyway, I now also have the explanation for the "weird" behavior on Debian where I can *only* connect with `[connect 127.0.01( `: For the "any" address, the IPv4 address comes first in the address list, but for "localhost" the IPv6 comes first! This means that `[netreceive]` would only listen to IPv4 adapters and I have to explicitly connect via the IPv4 address because hostnames are resolved to IPv6 addresses.

Simply taking the first working entry in the address list obviously leads to various issues. Some thoughts:
* `addrinfo_get_list` needs a parameter for allowing/disallowing IPv6 addresses
* for hostname resolution, IPv6 has to be explicitly requested by the user (another flag?), maybe falling back to IPv4.
* the "any" address should be explicitly set to IPv4 (default) or IPv6 (if requested)

We can still search for a *reliable* cross-platform way to create dual-stack sockets for `[netreceive]`, though.

Spacechild1:
@danomatika Anyway, the commits up to here have been successfully tested on Windows/macOS/Linux and can be merged into your branch.

Spacechild1:
@danomatika I had a sessions today with @umlaeute where we discussed a couple of things.

* AI_ADDRCONFIG seems to be redundant for our use case and it causes troubles on Windows ("localhost" can't be resolved when offline)
* enable/disable IPv6 networking with a message (maybe this should be rather a creation flag?). For now, the default is *off* because it causes too much troubles with existing setups/programs which only listen to IPv4 adapters. Maybe in 10 years we can change the default to *true* :-)
* make sure to disable `IPV6_V6ONLY` on IPv6 sockets, so that they also listen to IPv4 adapters (maybe this was already disabled on your system, but on all the systems I've tested it was enabled by default)

First I've thought `[netreceive]` (without hostname) could always bind to the IPv6 "any" address and disable `IPV6_V6ONLY`, but this is not supported on Windows XP, so for now we better only do this if explicitly requested.

Need to update pdsend/pdreceive accordingly (for now I've just set them to IPv4), probably adding the new arguments as flags.

danomatika:
I think the best course of action is to keep current behavior as it is and use IPv6 for explicit IPv6 addresses:

* "localhost" is IPv4 by default
* IPv4 address patterns are automatically IPv4
* IPv6 address patterns are automatically IPv6

I suppose an explicit flag would help, but I think it makes the most sense that this is transparent and revolves around the addresses themselves.

Spacechild1:
> IPv4 address patterns are automatically IPv4
> IPv6 address patterns are automatically IPv6

this is already the case. 

> "localhost" is IPv4 by default

why single out "localhost" here? the question is really how to deal with hostname resolution in general. my proposal above is to use IPv4 by default and only allow IPv6 if requested. what do you think?

Spacechild1:
or are you suggesting to only use IPv6 for explicit IPv6 addresses for now? (and later expand this to hostname resolution?)

danomatika:
>"localhost" is IPv4 by default

Is it? "localhost" on my system sets up an IPv6 address "::1".

>why single out "localhost" here?

It's used in the examples and by probably 90% of the existing patches out there.

>the question is really how to deal with hostname resolution in general. my proposal above is to use >IPv4 by default and only allow IPv6 if requested. what do you think?

I think I basically said the same thing already, just a different approach. To make this work, we need to set the getaddrinfo() ai_family hint explicitly. What I'm is:

* do a string compare for "localhost", if found then AF_INET
* check if the address is IPv4, if so then AF_INET
* check if the address is IPv6, if so then AF_INET6

I had some code which returned the IP version number for a given address. I just need to dig it back out again.

As for a flag, I would say we could add is as an override for this automatic recognition which explicitly specifies AF_INET*. I would not want to force the user to put *both* a flag *and* an IPv6 address to get IPv6. If we plan for the LOOONG term, then if the addresses already specify the IP version then only thing we'd need to change for an IPv6 default work would be what to use for the "localhost" string.

(Sure, a default IPv6 world will probably never come, but it's still probably good to give yourself some room.)

Spacechild1:
> Is it? "localhost" on my system sets up an IPv6 address "::1".

sure, on mine too. I didn't say that it doesn't. I've understood that you've proposed to resolve "localhost" to an IPv4 address.

> check if the address is IPv4, if so then AF_INET
> check if the address is IPv6, if so then AF_INET6

yes, I've already agreed with this! 

But let's talk about *hostname resolution*. We need a consistent way how to handle this in a backwards compatible way. (And I don't think we need to treat "localhost" specially at all). My current proposition is to only resolve to IPv4 addresses by default - but give the user the possibility to *allow* IPv6 addresses. In the latter case, addrinfo_get_list() could return both IPv4 and IPv6 addresses and we try them until one of them works. What do you think about this?

Spacechild1:
Note that the problem with your current implementation is not only with "localhost" but any hostname in a local network, that's why I wouldn't single out "localhost" but rather think about hostname resolution in general.

Spacechild1:
or simply keep hostnames IPv4 for now and later figure out how to deal with it?

danomatika:
The how about:

* IPv4 by default
* IPv6 used when address is definitively IPv6, ie "::1"
* user can override with flag to set family

I mainly don't want to set AF_INET for what is obviously an IPv6 address as I think the intent of requesting "::1" is clear.

Spacechild1:
> IPv4 by default
> IPv6 used when address is definitively IPv6, ie "::1"

agreed.

> user can override with flag to set family

in my meeting with @umlaeute, he pointed out that this kind of defeats the point of using hostnames. especially when connecting to a remote network, you often don't know what kind of family is supported. that's why I've suggested a flag/message to *allow* IPv6, so addrinfo_get_list() could return both. With TCP, we can simply iterate and try to connect until we succeed, but with UDP the situation is not so clear because we don't get any feedback if the address actually works.

Maybe we can actually offer three modes for hostname resolution: 
1) "ipv4" (IPv4 only, the default), 
2) "ipv6" (IPv6 only)
3) "any" (IPv4 and IPv6)

The last one could work fine for e.g. making web requests (we iterate and try to connect till success). Maybe not so useful for UDP because of the reasons mentioned above, but it would be fine if we *know* that the other end listens to both IPv4 and IPv6 adapters.

Spacechild1:
I think I've found a backwards compatible solution which doesn't require any flags! I'll implement it when I'm home and push it here.

danomatika:
Ok. I'm curious to see. If you need a function to check an address string for family:

~~~
int addr_get_ipver(const char *hostname) {
    char buf[16];
    if(inet_pton(AF_INET, hostname, buf)) {
        return AF_INET;
    } else if(inet_pton(AF_INET6, hostname, buf)) {
        return AF_INET6;
    }
    return AF_INET;
}
~~~

Spacechild1:
@danomatika check this out:

* I added a `addrinfo_sort_list` function to manually resort the address list, by either putting IPv6 or IPv4 addresses first. The order within a family is preserved. For explicit IPv4/IPv6 addresses this is a no-op, because the list usually (always?) consists of a single entry.
* `[netreceive]`: create a dual stack socket by default (if possible). The address list is sorted with IPv6 first. If an IPv6 address is encountered then `IPV6_V6ONLY` is disabled, so the socket also listens to IPv4 adapters. If a system doesn't support this socket option (e.g. Windows XP) , we fall back to IPv4.
* `[netsend]`: the address list is sorted with IPv4 first. This means that hostnames are preferably resolved to IPv4, to avoid issues with single stack IPv4 listeners. To force IPv6 networking, you simply have to provide an explicit IPv6 address. (We might later add an option for *not* sorting the address list, to keep the original order returned by the system).

AFAICT, this is fully backwards compatible.

umlaeute:
> in my meeting with @umlaeute, he pointed out that this kind of defeats the point of using hostnames. especially when connecting to a remote network, you often don't know what kind of family is supported.

if I ever use `[connect ip6-localhost 7777(` I will be *highly* disappointed if this doesn't work (on my Debian system, this faux hostname resolves to `::1`, it's the IPv6-only variant of `localhost`).

however, if a host resolver returns multiple addresses for a name, i *think* that the client is under no obligation to pick the first one, and instead can choose any (iirc, in reality most clients will nevertheless pick the first address, so a DNS-server that uses a round-robin for resolving a single name to multiple IP addresses is a cheap load balancer; but i don't think this behaviour is mandated).
if this is true, we can pick the first IPv4 address of the result contains both IPv4 and IPv6 addresses to connect to a remote, and otherwise just pick the first available address.

Spacechild1:
> i think that the client is under no obligation to pick the first one

yes, I think it's only a recommendation. To quote the documentation for `getaddrinfo`:

> Normally, the application should try using the addresses in the order in which they are returned.

http://man7.org/linux/man-pages/man3/getaddrinfo.3.html

This doesn't read like a must do me. The Windows documentation is even more open worded:

> The list can be processed by following the pointer provided in the ai_next member of each returned addrinfo structure until a NULL pointer is encountered.

https://docs.microsoft.com/en-us/windows/desktop/api/ws2tcpip/nf-ws2tcpip-getaddrinfo

Spacechild1:
> if this is true, we can pick the first IPv4 address of the result contains both IPv4 and IPv6 addresses to connect to a remote, and otherwise just pick the first available address.

I think this is more or less what I've implemented in 9b537df. For `[netsend]` I simply reorder the list to put IPv4 addresses first. This is obviously a no-op if the list only contains a single entry.

Note that for `[netreceive]` I prefer IPv6 addresses to create dual stack sockets (if possible).

Spacechild1:
I made a much better version of `addrinfo_sort_list` which takes a compare function. This seems to work fine with 1-2 entries, couldn't test yet with more entries.

umlaeute:
> Note that for `[netreceive]` I prefer IPv6 addresses to create dual stack sockets (if possible).

ah yes; i only commented on client-functionality (where you pass a *to*-address, as in `[netsend]`). i think that the receivers should always use dual-stack if indicated (as we discussed) and possible.

danomatika:
This is looking good now.

I started to update the GUI socket setup in s_inter.c but it quickly got  too complicated and I need to redo it. Do you have time to try?

Spacechild1:
> Do you have time to try?

sure! 

BTW, in the original code of `sys_do_start_gui` I don't understand the point of trying 20 hardcoded port numbers https://github.com/pure-data/pure-data/blob/c0773945fcd9af3645799f2de94e35970503b25f/src/s_inter.c#L1094
Why don't we just let the OS pick a free port number in the first place - which is done here only as a fallback: https://github.com/pure-data/pure-data/blob/c0773945fcd9af3645799f2de94e35970503b25f/src/s_inter.c#L1108

danomatika:
Probably 2 reasons:

1. this was written before you could easily ask the system for a free port
2. Miller wanted an established port range for Pd?

The free ports chosen by my system all tend towards the extreme upper range (40000). I wouldn't if the original pattern stayed for now...

Spacechild1:
> The free ports chosen by my system all tend towards the extreme upper range (40000)

but that's still in the "User Port" range, just like 5400+. Personally, I don't see any advantages of trying a bunch of fixed port numbers. @umlaeute what do you think?

umlaeute:
> Probably 2 reasons:
> 
>     1. This was written before you could easily ask the system for a free port

i seriously doubt that. (i think the ability to request a free port is as old as the `bind()` interface [citation needed]). although it's possible that the original code was written before someone *knew* that it is possible to ask for a free port.

>     2. Miller wanted an established port range for Pd?

more likely (although no attempt has ever been made to register that port(range) at the [IANA](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?&page=113)).

i guess using a known port just made it easier to do manual debugging (as in `netcat localhost 8900`).


the "extreme upper range (>40000)" is the typical [private port range](https://en.wikipedia.org/wiki/List_of_TCP_and_UDP_port_numbers#Dynamic,_private_or_ephemeral_ports).

as for the code, i added the fallback to port:0 as I sometimes ran out of free ports in the range 8900++. (debugging crashing Pds :-)), so that's why "we already do this as a fallback".
i don't see much advantage in getting rid of the 8900 ports (so as @danomatika put it: "I wouldn't mind if the original pattern stayed for now")

Spacechild1:
> i don't see much advantage in getting rid of the 8900 ports

do you mean 5400?

> as for the code, i added the fallback to port:0

I see! Note that we already do this on the Tcl side (when we start Pd from the GUI): https://github.com/pure-data/pure-data/blob/c0773945fcd9af3645799f2de94e35970503b25f/tcl/pd_connect.tcl#L39

So I don't see why we shouldn't do the same thing on the core side... 

> i don't see much advantage in getting rid of the 8900 ports

Well, it would simplify the code. But if everyone else thinks we should keep the existing behavior, then I'll accept this. I don't have a problem with the code, it just felt redundant to me. Maybe @millerpuckette wants to share this opinion?

umlaeute:
> do you mean 5400?

yes of course.

umlaeute:
> Note that we already do this on the Tcl side (when we start Pd from the GUI):
> 
> https://github.com/pure-data/pure-data/blob/c0773945fcd9af3645799f2de94e35970503b25f/tcl/pd_connect.tcl#L39

speaking of which: this binds to `localhost`, which could resolve to *either* `127.0.0.1` or `::1`.
see the clumsy hack:
https://github.com/pure-data/pure-data/blob/c0773945fcd9af3645799f2de94e35970503b25f/src/s_inter.c#L78-L82

Spacechild1:
Ah, this is of course obsolete with the IPv6 additions.

Spacechild1:
@millerpuckette is there still a reason for trying predefined port numbers before letting the OS give us a free one? See: https://github.com/pure-data/pure-data/pull/666#issuecomment-507203732

Just asking. If it's not necessary, I could remove it when I update of the GUI startup code, otherwise I just keep it.

millerpuckette:
Cool to get rid of the old port-scanning code - I never looked at the fallback code but it should work in any case I can think of.

millerpuckette:
I'm not sure what to do now - should I just go ahead and try to merge what you've got so far?  (There will be a conflict and a compile ouch or two to fix which will knock master slightly out of sync with this branch.)

danomatika:
We can ha sleep the conflicts by merging Christoph’s branch i to mine first. Once everything’s finalized, we can make it clean to merge.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
> We can ha sleep the conflicts by merging Christoph’s branch i to mine first. Once everything’s finalized, we can make it clean to merge.

@danomatika sounds good. I'll ping you when this PR is ready to merge.

Spacechild1:
@danomatika Basically, I'm finished. I've successfully tested `sys_do_startgui` in both directions (wish.exe starts pd.exe / pd.exe starts wish.exe) on Windows and Linux. However, on macOS Pd hangs on startup when I open the .app created with `make app`... Can you confirm this? Do you have an idea what's going on? I'll investigate tomorrow.

Spacechild1:
@danomatika strange, after a recompile the .app works just fine... I think you can merge now.

danomatika:
>However, on macOS Pd hangs on startup when I open the .app created with make app... Can you confirm this?

I am seeing this now.

danomatika:
Seems to be working if I start the core inside the app bundle first, so the issue is most likely the core receiver code.

Also, I notice that listening on "localhost" doesn't  work for [netreceive -u] on my system.

danomatika:
The "localhost" issue is the problem. Changing s_inter.c's LOCALHOST define from "localhost" to "127.0.0.1" allows Pd to start correctly from the GUI.

danomatika:
As a follow-up: I think we need to look at the GUI side in pd_connect.tcl as well. I bet on my system "localhost" is being interpreted as IPv6 and the GUI is sending to IPv4 localhost. I thought it should allow for handling receiving IPv4 messages on a properly configure IPv6 socket tho?

danomatika:
(I suppose we could move this conversation to the other PR now...)

Ok, some more observations:

* Sorting the ailst for IPv4 first on s_inter.c line 1065 allows the GUI start ok. I assume this is because it will respond with the correct IPv4 localhost message as opposed to responding with IPv6 which the GUI ignores.
* Tcl 8.5 sockets are still IPv4 only. It seems that IPv6 support was added to Tcl 8.6 in 2010.
* If I change x_net.c to always sort IPv4 addresses first, "localhost" and IPv4 sending.recieving works fine but receiving IPv6 to "localhost" is broken.

danomatika:
For debugging, you can check the configuration of an open port with:

    netstat -lan | grep 5000

which outputs:

    tcp4       0      0  *.5000                 *.*                    LISTEN

It would be "tcp6" for IPv6 or "tcp46" for bridged.

Spacechild1:
> Seems to be working if I start the core inside the app bundle first, so the issue is most likely the core receiver code.

Yes, I only get this when I start the GUI first. It's because when we start the GUI, we explicitly pass the IP address.

> Tcl 8.5 sockets are still IPv4 only. It seems that IPv6 support was added to Tcl 8.6 in 2010.

Yes, see below.

> Maybe we need to call bind twice when listening to ensure both IPv4 and IPv6 messages are received? 

you can't bind a single socket to two addresses. My idea was to simply iterate over addrinfo results of "localhost" and pick the first we can connect to. On macOS, sometimes this works, sometimes it doesn't... Very strange.

Basically, we can't know which protocol the GUI supports. I had another solution in mind, where the GUI would also pass the host name, e.g. with "-guihost <IP>". This should be foolproof. Do you mind giving it a try?

PR #667, opened 2019-06-25
use gpointer to cache float array in [tab*] objects

Ant1r:
This PR proposes, for` [tabread] [tabwrite] [tabosc~]` etc..., to cache float arrays through a gpointer, for two distinct goals:

 - speeding up table access in control domain by avoiding reacquiring the array each time; for example `[tabread]` is roughly 3 times faster once the array is acquired.

 - preventing the signal objects from setting the array's `usedindsp` flag; this is strongly related to #655: if `usedindsp` is not set, then rebuilding the whole DSP graph is not needed anymore after swapping the array vector. This flag is not needed anymore for signal objects, because the pointer is checked before processing each block, and refreshed if necessary.

The CPU overhead for "setting" control objects is about 16% from my approximative measurements, and seems negligible for signal objects.

This idea was @Spacechild1 's one actually ;-)

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -7,7 +7,42 @@
 /* LATER make tabread4 and tabread~ */
 
 #include "m_pd.h"
+#include "g_canvas.h"
 
+int gpointer_acquire_floatarray(t_gpointer *gp, t_symbol *arrayname, 
+    int *npoints, t_word **vec, t_garray **garray, t_object *owner, int recover)
+{
+    t_garray *a;
+
+    if(gpointer_check(gp, 0))
+    {
+        *vec = (t_word *)gp->gp_stub->gs_un.gs_array->a_vec;
+        *npoints = gp->gp_stub->gs_un.gs_array->a_n;
+        return 1;
+    }
+    else if(recover || gp->gp_stub)
+        /* when the pointer is invalid: if "recover" is true or if the array has
+        already been successfully acquired, then try re-acquiring the array */
+    {
+        if (!(a = (t_garray *)pd_findbyclass(arrayname, garray_class))) {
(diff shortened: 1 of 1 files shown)


Spacechild1:
Awesome! You've been reading my thoughts :-)

Have you tested this in DSP? I can imagine it will lead to some cache misses because of the pointer chasing involved but we have to see if it is significant. Note that it's important to test with objects reading from several different tables (not only a single one!) to simulate situations where the memory is not cached. Obviously, the overhead would become more significant with smaller block sizes. I'll test myself and I'm curious about the results :-)

Ant1r:
I tried to do some profiling myself, comparing current **master** with the present **tabobjs-gpointer** branch.

- reading and setting multiple tables in the control domain:
tabpointer-ctl2

- CPU usage in signal domain:
tabpointer-sig

Here are the patches I used:
[tabpointer-tests.tar.gz](https://github.com/pure-data/pure-data/files/3936185/tabpointer-tests.tar.gz)

Tell me if you see any more relevant tests to do.

Spacechild1:
nice, the benchmark doesn't look too bad! I'll also test once I have more time.

Spacechild1:
BTW, I think `gpointer_acquire_floatarray` should be exported in `m_pd.h` so that externals can use it too!

Ant1r:
It appears that @millerpuckette partly implemented this PR with 35beb41eb72fa24998852c19ea352ef2e6b9ab05, but only the changes to DSP objects.

However, this PR was also using gpointer-caching with control objects, which had the nice effect of dividing the CPU cost of e.g [tabread] by 2.4.

So I guess this PR will have to be rewritten, concerning control objects only.

PR #673, opened 2019-07-03
Fix for issue with t_int on 64-bit Windows for d_math objects

NiallMoody:
This is a fix for an issue I encountered when using libpd in my own [Unity wrapper](https://github.com/LibPdIntegration/LibPdIntegration) for the library (you can read the original issue [here](https://github.com/LibPdIntegration/LibPdIntegration/issues/6)).

In d_math.c there are a number of objects which, in their perform function, define the block size as a t_int variable. This causes problems on 64-bit Windows binaries, because the block size is originally defined as an int (32-bits on Windows). Because t_int is defined as a long long (64-bits) this results in the perform function of these objects reading a junk value in the upper 32 bits of the t_int and attempting to process the wrong number of samples.

The affected objects are:
rsqrt~
sqrt~
wrap~
mtof~
ftom~
dbtorms~
rmstodb~
dbtopow~
powtodb~

Without the fix, any patches running via my wrapper in Unity (that use one of these objects) crash. For reasons that aren't clear to me, the currently-available 64-bit Windows executable of Pure Data seems to work fine without the fix.

### DIFF
--- a/src/d_math.c
+++ b/src/d_math.c
@@ -138,7 +138,7 @@ static void *sigrsqrt_new(void)
 static t_int *sigrsqrt_perform(t_int *w)
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    int n = *(int *)(w+3);
+    int n = (int)w[3];
     while (n--)
     {
         t_sample f = *in++;
@@ -197,7 +197,7 @@ static void *sigsqrt_new(void)
 t_int *sigsqrt_perform(t_int *w)    /* not static; also used in d_fft.c */
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    int n = *(int *)(w+3);
+    int n = (int)w[3];
     while (n--)
     {
         t_sample f = *in++;
--- a/src/d_math.c
+++ b/src/d_math.c
@@ -138,7 +138,7 @@ static void *sigrsqrt_new(void)
 static t_int *sigrsqrt_perform(t_int *w)
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    t_int n = *(t_int *)(w+3);
+    int n = *(int *)(w+3);
     while (n--)
     {
         t_sample f = *in++;
@@ -197,7 +197,7 @@ static void *sigsqrt_new(void)
 t_int *sigsqrt_perform(t_int *w)    /* not static; also used in d_fft.c */
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    t_int n = *(t_int *)(w+3);
+    int n = *(int *)(w+3);
     while (n--)
     {
         t_sample f = *in++;
@@ -253,7 +253,7 @@ static void *sigwrap_new(void)
 static t_int *sigwrap_perform(t_int *w)
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    t_int n = (t_int)w[3];
+    int n = (int)w[3];
     while (n--)
     {
         t_sample f = *in++;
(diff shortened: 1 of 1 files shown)


src/d_math.c
```diff
@@ -138,7 +138,7 @@ static void *sigrsqrt_new(void)
 static t_int *sigrsqrt_perform(t_int *w)
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    t_int n = *(t_int *)(w+3);
+    int n = *(int *)(w+3);
```
Spacechild1: should also be `int n = (int)w[3];`

Spacechild1: should also be `int n = (int)w[3];`

src/d_math.c
```diff
@@ -197,7 +197,7 @@ static void *sigsqrt_new(void)
 t_int *sigsqrt_perform(t_int *w)    /* not static; also used in d_fft.c */
 {
     t_sample *in = (t_sample *)w[1], *out = (t_sample *)w[2];
-    t_int n = *(t_int *)(w+3);
+    int n = *(int *)(w+3);
```
Spacechild1: should also be `int n = (int)w[3];`

Spacechild1:
> Because t_int is defined as a long long (64-bits) this results in the perform function of these objects reading a junk value in the upper 32 bits of the t_int and attempting to process the wrong number of samples.

The question I've immediately asked myself is: why doesn't this happen on Linux and macOS (where 64-bit Pd has been the default for several years now)?

After some headscratching here is my attempt of answering this question:

The problem isn't `t_int n = (t_int)w[3];` because `w` really *is* an array of `t_int`s. The real problem is that we push `s_n` to `dsp_add` as an `int` (32-bit) but inside `dsp_add` we retrieve all arguments as `t_int` (64-bit). This is where the actual size mismatch happens. **The only correct way is to cast `s_n` to `t_int` before pushing it to `dsp_add`!** 

Ok, nobody does that and it works nevertheless. In the case of Windows, luckily everybody casts `n` to `int` in the perform routine. Because Windows is little endian, this gets rid of the subsequent garbage bytes and all is good. (Note that this only works if `s_n` is the *last* argument, otherwise all subsequent arguments would be been turned into garbage inside `dsp_add`!).

But why don't we have to do this on macOS and Linux? Everything I've just said is only true if we assume that arguments are really passed on the stack. This is true for varargs on x86, but on x64 the situation is different (and complicated): arguments are passed in registers (if possible) - even in varargs functions! See this article: https://blog.nelhage.com/2010/10/amd64-and-va_arg/

Registers don't have a byte ordering and it seems that for 32bit operations, the remaining upper bits are zeroed out: https://stackoverflow.com/questions/11177137/why-do-x86-64-instructions-on-32-bit-registers-zero-the-upper-part-of-the-full-6 That's why you can push an `int` and read an `t_int` without getting garbage (at least for positive numbers I suppose).

However, it seems that the calling convention is slightly different on Windows: https://docs.microsoft.com/en-us/cpp/build/x64-calling-convention?view=vs-2019#varargs

> Arguments five and higher are passed on the stack.

Aha, there we have it! In most DSP objects (with at least 1 inlet and 1 outlet), `s_n` will come after the 4th argument: `dsp_add(foo_perform, x, sp[0]->s_vec, sp[1]-s_vec, sp[0]->s_n);`

All I've said is pure speculation and *not tested*, but it might explain this strange situation. Please correct me if I'm wrong!

NiallMoody:
> All I've said is pure speculation and _not tested_, but it might explain this strange situation. Please correct me if I'm wrong!

Ah, I think that does make sense. I'd got it in my head that it was down to Windows using an LLP64 data model and everyone else using LP64, but actually that would only make sense if there were long variables involved.

(also apologies for implementing your suggested changes as a separate commit. I'm still kind of new to github and didn't realise I could just click a button on your changes to implement them instead of doing my own commit until after the fact)

Spacechild1:
> also apologies for implementing your suggested changes as a separate commit.

That's fine!

> didn't realise I could just click a button on your changes to implement them

Didn't know that either :-) personally, I never use the github web interface. I always make changes locally where I can test them.

millerpuckette:
merged

Issue #583, opened 2019-04-15
Request: allow [trigger] to convert "anythings" to other message types.


porres:
I don't get why trigger can't convert anything to something other than a bang.

I think it should be converted to a list just by adding the list selector, and treated as a list so it could be converted to the other message types (symbol, float)

porres:
this closes this issue https://github.com/pure-data/pure-data/pull/582

Issue #681, opened 2019-07-17
[trigger anything] and bang/pointer input


porres:
A bang input will not quite come out as a bang in [trigger anything]

trigger_bang calls trigger_list in https://github.com/pure-data/pure-data/blob/master/src/x_connective.c#L1051

so the  output is coded here: https://github.com/pure-data/pure-data/blob/master/src/x_connective.c#L1033

what we have as an output is then a "list" message without any elements, what becomes a bang in Pd, but that doesn't work 100%

I have an external that has both a bang and a list method, and this output triggers is treated as a "list" input instead of a "bang" input.

porres:
In practice, there's no difference of a bang input into [trigger list] and [trigger anything].

I guess it is "ok"  to have an empty list come out from [trigger list] with a bang input though...

porres:
I realize now that another a request I had already fixes this issue. Much of the code comes actually from purr data, so I guess they were aware of this issue and fixed it over there.

The Pull Requests also offers constant floats and allows anythings to be converted to other types as well. Regarding the issue reported here, it seems this is taken care in this line https://github.com/pure-data/pure-data/blob/ba93a6447038216b8092a705d0145e28c66237f2/src/x_connective.c#L1039

porres:
Note that the PR https://github.com/pure-data/pure-data/pull/682

still outputs an empty list come out from [trigger list] with a bang input... which is "ok" as I said

Spacechild1:
I think I ran into this very issue with [mrpeach/midifile] which defines both a bang and a list method. 

While I agree that `[t a]` should pass messages unchanged, I would also suggest that you check for empty lists in your external (+ call your bang method) because there are other objects around which output empty lists as bangs. Unfortunately, Pd's semantics are a bit ambiguous here, so it's probably good to take care of it on both sides.

porres:
trying to make things more obejctive, so I rebooted this issue into https://github.com/pure-data/pure-data/issues/683

PR #682, opened 2019-07-17
Add "constant float" to [trigger] and allow "anythings" to be converted

porres:
This is a reboot of https://github.com/pure-data/pure-data/pull/600

This Pull Request closes https://github.com/pure-data/pure-data/issues/583 and https://github.com/pure-data/pure-data/issues/681

And it also partially deals with the request https://github.com/pure-data/pure-data/issues/354 and takes care only of constant floats (not symbols). This is is also a simplification of https://github.com/pure-data/pure-data/pull/582 that implements constant values for floats and symbols. In this alternative, the the behaviour for symbols is exactly the same as before and this option seemed to encounter less resistance, so I thought I'd offer this option as well (see discussion in https://github.com/pure-data/pure-data/pull/582 for more details).


src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
     }
 }
 
 static void trigger_anything(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
 {
+    t_atom *av2 = NULL;
     t_triggerout *u;
-    int i;
-    for (i = (int)x->x_n, u = x->x_vec + i; u--, i--;)
+    int i, j = 0;
```
umlaeute: you could much reduce the scope of the `j` variable by declaring it later (near L1092)

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+                    freebytes(av2, (argc + 1) * sizeof(t_atom));
+                }
+        }
     }
 }
 
 static void trigger_bang(t_trigger *x)
 {
-    trigger_list(x, 0, 0, 0);
+    trigger_list(x, &s_bang, 0, 0);
```
umlaeute: +1

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+    for(i = x->x_n, u = x->x_vec + i; u--, i--;)
     {
         if (u->u_type == TR_FLOAT)
             outlet_float(u->u_outlet, (argc ? atom_getfloat(argv) : 0));
         else if (u->u_type == TR_BANG)
             outlet_bang(u->u_outlet);
         else if (u->u_type == TR_SYMBOL)
             outlet_symbol(u->u_outlet,
-                (argc ? atom_getsymbol(argv) : &s_symbol));
+                          (argc ? atom_getsymbol(argv) : (s != NULL ? s : &s_symbol)));
```
Spacechild1: are you sure about this? I think the original code makes sense (empty list -> "empty" symbol). I don't know if we want to output the "list" selector instead...

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+        else if (u->u_type == TR_ANYTHING)
+            outlet_anything(u->u_outlet, s, argc, argv);
         else if (u->u_type == TR_POINTER)
         {
             if (!argc || argv->a_type != TR_POINTER)
                 pd_error(x, "unpack: bad pointer");
```
Spacechild1: "unpack" -> "trigger"

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+                else outlet_pointer(u->u_outlet, argv->a_w.w_gpointer);
+                else if (u->u_type == TR_STATIC_FLOAT)
+                    outlet_float(u->u_outlet, u->u_float);
+                else
+                { // Ico: don't have to worry about zero element case (AFAICT)
+                    av2 = (t_atom *)getbytes((argc + 1) * sizeof(t_atom));
+                    SETSYMBOL(av2, s);
+                    if (argc == 0)
+                        outlet_list(u->u_outlet, &s_symbol, argc+1, av2);
+                    else
```
Spacechild1: the code path in the else clause is all we need (as Ico has pointed out in his comment), so you can remove the if check for `arg == 0`.

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+        {
+            // copying trigger_list behavior except that here we keep
+            // the outlet number and therefore avoid redundant printouts
+            if (u->u_type == TR_FLOAT)
+                outlet_float(u->u_outlet, (argc ? atom_getfloat(argv) : 0));
+            else if (u->u_type == TR_BANG)
+                outlet_bang(u->u_outlet);
+            else if (u->u_type == TR_SYMBOL)
+                outlet_symbol(u->u_outlet,
+                              (s != NULL ? s : (argc ? atom_getsymbol(argv) : &s_symbol)));
```
Spacechild1: I think you can simply do:

`outlet_symbol(u->u_outlet, (s ? s : &s_symbol));`

I don't see why we should consider the arguments here. Actually, if the selector here is NULL, I would consider this a bug.

Spacechild1:
I'm not so sure about the anything conversions to float and pointer. IMO, we should only consider the selector - which can never be a float or pointer - so these two cases should remain an error. 

I don't see why `[foo 1( -> [t f]` should print 1...

The symbol and list conversion is cool, btw!

src/x_connective.c
```diff
@@ -952,9 +952,13 @@ static t_class *trigger_class;
 #define TR_LIST 4
 #define TR_ANYTHING 5
 
+#define TR_STATIC_FLOAT 6
+
 typedef struct triggerout
 {
     int u_type;         /* outlet type from above */
+    t_symbol *u_sym;     /* static value */
```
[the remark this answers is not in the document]
porres: yeah, that was for the static symbol, which I had removed but I missed this one...

src/x_connective.c
porres: ok, reverting back to original

porres: that came from purr data world, not my idea to be sure about it :) thanks for revising this

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+            else if (u->u_type == TR_BANG)
+                outlet_bang(u->u_outlet);
+            else if (u->u_type == TR_SYMBOL)
+                outlet_symbol(u->u_outlet,
+                              (s != NULL ? s : (argc ? atom_getsymbol(argv) : &s_symbol)));
+            else if (u->u_type == TR_ANYTHING)
+                outlet_anything(u->u_outlet, s, argc, argv);
+            else if (u->u_type == TR_POINTER)
+                if (!argc || argv->a_type != TR_POINTER)
+                    pd_error(x, "unpack: bad pointer");
```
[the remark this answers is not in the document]
porres: oops

src/x_connective.c
porres: done, thanks again for catching these errors I imported from purr data

src/x_connective.c
porres: done

src/x_connective.c
```diff
@@ -1015,63 +1026,104 @@ static void trigger_list(t_trigger *x, t_symbol *s, int argc, t_atom *argv)
...
+                else outlet_pointer(u->u_outlet, argv->a_w.w_gpointer);
+                else if (u->u_type == TR_STATIC_FLOAT)
+                    outlet_float(u->u_outlet, u->u_float);
+                else
+                { // Ico: don't have to worry about zero element case (AFAICT)
+                    av2 = (t_atom *)getbytes((argc + 1) * sizeof(t_atom));
```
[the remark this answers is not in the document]
porres: could you do that for me in a PR?I dont think I can do that by myself

src/x_connective.c
porres: ok

porres:
so, I tried the suggested changes, but ran into errors :)

porres:
> I'm not so sure about the anything conversions to float and pointer. IMO, we should only consider the selector - which can never be a float or pointer - so these two cases should remain an error.
> 
> I don't see why [foo 1( -> [t f] should output 1...

I see

doc/5.reference/trigger-help.pd
```diff
@@ -1,37 +1,60 @@
...
-#X obj 112 242 print x2;
-#X obj 196 240 print x3;
-#X obj 43 26 trigger;
-#X obj 286 241 print x4;
-#X text 114 27 - sequence messages in right-to-left order;
-#X text 417 331 updated for Pd version 0.33;
-#X text 81 290 the above can be abbreviated as:;
-#X msg 172 152 symbol dog;
-#X text 39 59 The trigger object outputs its input from right to left
+#N canvas 554 23 536 576 12;
```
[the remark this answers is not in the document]
porres: > this is meant as a general advise.

ok, so I don't need to change anything, cool, got it

porres:
> I'm not so sure about the anything conversions to float and pointer. IMO, we should only consider the selector - which can never be a float or pointer - so these two cases should remain an error.

I think a pointer will always output an error, right? it's the same thing as in the list inoput. Or we could just don't even bother checking for what's the message like and just go straight to pd_error(x, "trigger: bad pointer")

> I don't see why [foo 1( -> [t f] should output 1...

I agree, but in other cases, like when we have a symbol input into a [t f], we have a "0" as the output. So I say we just force a "0" output as well.

porres:
that's what I did for https://github.com/pure-data/pure-data/pull/682/commits/67f59bbb70aa878532af817f2e4c6b30658e5eaf

pointer goes straight to an error, and float always becomes 0

doc/5.reference/trigger-help.pd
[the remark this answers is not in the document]
umlaeute: but you should remember it *next* time :-)

Spacechild1:
> symbol input into a [t f], we have a "0" as the output.

actually, since Pd 0.48 `[float]` can convert symbols, e.g. `[4.5( -> [makefilename %f] -> [float]` works, so we could do the same for `[t f]`.

BUT, now that I think of it, I'm not sure if the PurrData approach is the right one. Basically, it gets rid of any type errors, so `[t]` will happily take any kind ouf input and convert it to the given type (the only exception is `[t p]` where the input has to be/have a gpointer).

Basically, this means we lose any kind of type checking. There is a reason why someone writes `[t b f]` instead of just` [t b a ]`. The former is a kind of contract that the input type should be trivially convertable to a float. Now the question is what "trivially convertible" means and this is where programming languages differ a lot. Pd doesn't have a formal type system, so we can only guess. AFAICT, for the current behavior of `[trigger]` it looks like this:
* everything converts to a bang
* list is convertable to float, symbol or pointer if the first argument is of the correct type
* float, symbol and pointer can be converted to a list

Now, for every implicit conversion we add, we might gain some convenience (e.g. the ability to sequence and convert in one object) but we lose a bit of type safety. IMO, we should think carefullly about *every* implicit conversion we add to `[trigger].`

Just to give an example: When I write `[t b f]`, I would be disappointed if I accidentally send` [foo(` or `[symbol foo(` and don't get an error message which helps me to catch my programming mistake.

Generally, I think we should be careful with introducing automatic float <-> symbol conversion. In a largely weakly typed language like Pd, it's good to keep at least *some* amount of type safety.

Just look at the madness in Javascript:
`"5.7" + 1` -> `"5.71"` (a string)
`"5.7" * 2` -> `11.4` (a number)
`"foo" + 1` -> `"foo1"` (a string)
`"foo" * 2` -> `NaN` (a number - no error!)
(well, what should I expect from a language designed in 10 days...)

Python and Ruby get it partially right:
`"5.7" + 2` -> `TypeError: can only concatenate str (not "int") to str`
`"5.7" / 2` -> `TypeError: unsupported operand type(s) for /: 'str' and 'int'`
BUT:
`"5.7" * 2` -> `"5.75.7"` (because the `*` operator is overloaded for strings...)

Lua tries to coerce strings to numbers for arithmetic operations and throws an error if it can't:
`"5.7" + 1`-> `6.7` 
`"foo" + 1` ->  `attempt to perform arithmetic on a string value`
BTW, this is the same behavior as in `[float]`.

The examples above just show what can go wrong and that this stuff needs a lot of thinking :-) I'm not critizing your PR (as I think some of the conversion operations might make sense), I just want to stimulate some discussion.

Spacechild1:
> pointer goes straight to an error, and float always becomes 0

Generally, for `[t s]` and `[t f]` I would prefer error messages over silent conversion to some default value to help catching programming mistakes. From all the conversions added by PurrData, I only tend to agree with anything -> symbol and anything -> list. But again, all of this is highly debatable and I'm only throwing my (current) opinion into the ring.

Spacechild1:
BTW, `[trigger]`'s type error messages could be improved. Currently, it's 
`trigger: can only convert 's' to 'b' or 'a'` 
for all types :-)

porres:
> actually, since Pd 0.48 [float] can convert symbols, e.g. [4.5( -> [makefilename %f] -> [float] works, so we could do the same for [t f].

but when I say a symbol input into [t f] gives us "0" as the output, I also mean an input like "symbol foo". This becomes "0" with no error or warning, so silently dealing with such things in [trigger] is already a current practice. So I don't see why "foo" cannot silently become "0" as I've managed to do so.

Spacechild1:
> This becomes "0" with no error or warning, so silently dealing with such things in [trigger] is already a current practice.

ouch, you're right. Actually, I consider this a bug in `[trigger]`... at least I don't see the purpose of the silent conversion in those two cases... Of course we can keep the current behavior, but I'm not sure if it's a good idea to add even more silent conversions. I mean, what's the use case? It just makes debugging Pd code harder than it already is :-)

porres:
> BUT, now that I think of it, I'm not sure if the PurrData approach is the right one. Basically, it gets rid of any type errors

It does, all but pointer! But then, as discussed, I think we already had some inconsistency where some errors that made sense (i.e. symbol to float), so if we're not to get rid of, we might as well change the behaviour and include an error to that conversion for consistency, which I think is bad, it's better the other way around.

> list is convertable to float, symbol or pointer if the first argument is of the correct type

yeah, but then again, you can have a list of symbols into [t f], then the output is "0", no error type/check.

> we should think carefully about every implicit conversion we add to [trigger].
> 
> Just to give an example: When I write [t b f], I would be disappointed if I can accidentally send [foo( or [symbol foo( and don't get an error message

I'm just repeating myself but yeah, this already happens for "symbol foo", we're not adding this implicit conversion, it already exist, without error - so you seem to be proposing we change this current behaviour to add an error instead. 

> Generally, I think we should be careful with introducing automatic float <-> symbol conversion.

Oops, that's another one. This already happens, we're not introducing it. There's an implicit conversion, without errors or type check, so currently a float input becomes "symbol float"

So my whole point is that, first, there was no point to not convert something like an anything message to a list message, I think we all agree. Now, the only type error checking regarded conversion from anything, but this was inconsistent, because other conversions did not have any kind of check. Therefore, I just think it makes sense to allow "foo" become "0" instead of adding all these types of error checking and basically changing the whole thing and even potentially breaking patches, as the behaviour would change.

For instance, a patch that used to output "0" from a symbol input would not anymore. Where I'm getting at is that this is a rather serious change that I don't know if it has chance to be accepted. In the best case scenario, it would require one of those famous compatibility flags - and @umlaeute would not be happy at all about that (go ahead, ask him how he feels about it).

So, we add conversions that made sense and I don't know why they didn't happen in the first place, and we now allow a silent conversion ("foo" to "0") in a consistent way as it already happens ("symbol foo" to "0").

Spacechild1:
> so you seem to be proposing we change this current behaviour to add an error instead.

after some more thought, yes!

> For instance, a patch that used to output "0" from a symbol input would not anymore. Where I'm getting at is that this is a rather serious change that I don't know if it has chance to be accepted.

I don't know, why would someone do that? I mean, seriously. Anyway, it wouldn't cause silent errors, it would scream to people in the console: fix your patch! So I don't really see this as a backwards compatibility problem.

Personally, I would be happy to get *more* type checking in `[trigger]` instead of - in my opinion - useless conversions to default values. I'm not dismissing your viewpoint, it's just as valid, it only goes into the opposite direction.

Spacechild1:
Also, `[symbol foo( -> [t f] -> 0` is even more absurd when you consider that `[symbol foo( -> [float]` gives you an error.

porres:
> [symbol foo( -> [float] gives you an error.

you sure? I get "0" from that, like I said before.

Spacechild1:
> you sure?

yes, it will print: `Couldn't convert foo to float.`

porres:
oh, you meant into [float], not [trigger float], I see!

Spacechild1:
also note that `[unpack]` does proper type checking...

porres:
> I don't know, why would someone do that?
> what's the use case?

I know there are conversions that don't make sense. So if any of them are being done, I agree is for no good reason! But the fact is that they happen, so there's a chance any of such 'nonsensical' behaviour is being abused. 

So, why add this conversion from "foo" to "0"? It's just a matter of keeping things more consistent. I don't see why, of all things, we just worry that "foo" cannot become "0". It feels silly to hold on to this type of checking. for me it's like, either we obsess and worry about every type check, or just let it go...

i don't actually personally mind a deep overhaul of [trigger] where now "symbol foo" would say "Couldn't convert foo to float" just like in [float], but it seems like a big change, that requires at least a compatibility flag, and from what I know about Pd's development, I just think things have more of a chance to stay the way they are. And in this context/scenario, this conversion I'm now allowing without errors is not intrusive at all, and is also consistent to the current behaviour of the object.

Spacechild1:
> but it seems like a big change, that requires at least a compatibility flag,

I don't know... IMO, patches which rely on such behavior are just broken. Again, at least you would get a clear error. 

I mean, what we could do as a compromise is keep the default conversion **and** post an error message, so we don't "break" existing patches but instead just nag users to change them (if they want to get rid of the red lines in their console).

> And in this context/scenario, this conversion I'm now allowing without errors is not intrusive at all, and is also consistent to the current behaviour of the object.

I agree that your proposal would add consistency, it's just the opposite direction from where I want to go ;-)

porres:
> what we could do as a compromise is keep the default conversion and post an error message

hey, that actually seems like a perfect compromise!

we can have the cake and eat it too in the end.

So, we allow things to happen and bug people to fix their silly patches :) I'm onboard

porres:
but, regarding consistency, when Pd gives an error, it usually doesn't have any output!

maybe it's the case to add a "warning"? Like when we have arrays with the same name? ("warning: array1: multiply defined")

This would point to the fact that the conversion did not go very well or that it doesn't really make sense, but the best thing would still make this quite clear in the documentation/help file!

porres:
Now that I think of it, it just seems like a pain to deal with all the issues when converting message types. I was thinking to add warnings that explicitly deal with all conversions that are not pertinent (such as "symbol message was converted to '0'"), but it's kind of a nightmare. The code needs to be changed for that to a more complex check. This hit me when I just realised how a bang cannot be converted to float or symbol. So each input type would need to be taken care in site, and not like the code currently works, which just creates a hub and sends almost all input to a single code section that deals with everything. 

Maybe we just accept this object has always been confusing when dealing with conversions. I can try and make a more extensive documentation that deals with it in depth, showing which are the pertinent conversions and what happens when something not clever is attempted (convert a bang to a float). Extended/purr data tried to detail this as well, but I'd do it differently.

porres:
so, there's also no use case for a pointer being converted to list or anything, right?

porres:
I think it'd make sense that if [trigger anything] receives a pointer, it'd just let it through, but what happens is that it gets corrupted in the same way a bang input into [trigger anything] got corrupted.

Spacechild1:
> but what happens is that it gets corrupted in the same way a bang input into [trigger anything] got corrupted.

yes, that's a bug as well. ideally, `[t a]` should pass any message unchanged.

> so, there's also no use case for a pointer being converted to list or anything, right? 

`float, symbol, pointer -> list` conversion and vice verca is what I would call "trivially convertable": the atom type doesn't change, only the selector does. Conceptually, a float/symbol/pointer is the same as a list with a single element of the same type. For the vast majority of objects, the behavior is also equivalent. Actually, I think single item lists are one of the most ambiguous parts of Pd...

I guess I'm mostly critical of silent convertions between different atom types, e.g. `[symbol foo( -> [t f]`. 
99% of the time, sending a symbol to `[t f]` is a programming error, so we should at least post a warning. And maybe not add more silent conversions...

porres:
I could see a bang to symbol conversion, where it'd output "symbol bang", and not "symbol symbol"...

porres:
here's what I have prepared for the documentation, and what I think makes sense

- bang: any input message can be converted to a bang.
- float: a list that starts with a float lets it through, other message types convert to "0".
- symbol: a list that starts with a symbol lets it through. An anything message like "dog my cats" will convert the selector to a symbol. A float message or a list that starts with a float become "symbol float", and bang becomes "symbol bang".
- list: any message that is not a list will have a list selector instead, in the case "dog my cats", the list selector is added. Note that a bang comes as a zero element list message, which gets automatically converted to a bang. A one element float list also becomes a float and a one element symbol list becomes a symbol message as well.
- pointer: no other message can be converted to a pointer.
- anything: whatever comes in, comes out, there's no conversion.

porres:
I see I didn't cover pointer to other message types conversion...

porres:
maybe this does it?

- pointer: no other messages can be converted to a pointer, and a pointer should only be converted to a bang.

porres:
> I could see a bang to symbol conversion, where it'd output "symbol bang", and not "symbol symbol"...

seems purr data does that!

Spacechild1:
Again, I don't think `[trigger]` is the place for message conversion. We already have objects for this. It's not worth the loss of type safety... And we should rather add more warnings than more conversions. But I notice I only keep reinforcing my position, so maybe I should give it a break for a while :-) 

But please think about my ramblings about type safety and what it means for programming and debugging in general. There is a reason why more and more people choose TypeScript over plain JavaScript, or why Python introduced type annotations, why people have to use linters, etc.

Spacechild1:
To sum it up: I think `[t s s]` means "I want to sequence a message which is trivially convertable to a symbol" whereas the PurrData folks think it means "I want to sequence a message and convert it to a symbol no matter what". The question is now: who's right? :-)

porres:
> And we should rather add more warnings than more conversions. 

for the conversions that don't make sense, but what about converting a bang to "symbol bang", do you mean that would deserve a warning as well and not a more sensible and reasonable conversion as done in purr data?

Well, the object does serve for the purpose of converting message types, so one of the points was to add the conversion from anything to a list. 

I can see the deal of how to manage conversions that don't make sense, though these are already happening a lot. For this matter, I don't really have a dog on this fight, it wouldn't be up to me in the end, as well. All I can say is that adding proper warnings require quite a good amount of surgery in the code - and I'm kinda shy to attempt that :) 

And anyway, I guess we opened a big can of worms when trying to revise [trigger]. I'm already kinda lost with everything in discussion here. Maybe I'll reboot this PR once again and try to better organize different commits that relate to different issues so it all gets easier for Miller to check out.

So one idea would be to open an issue regarding how [trigger] should post warnings, and a dedicated commit or PR could deal with that as well.

porres:
> Well, the object does serve for the purpose of converting message types

At least that's how it presents itself, but we know how it does not do quite a good job, and how tricky it all is

Spacechild1:
> for the conversions that don't make sense, but what about converting a bang to "symbol bang", do you mean that would deserve a warning as well

Probably... The thing is: when would you actually *need* the output of `[bang( -> [t s] `to be `[symbol bang(`? When typing `[t s]`, you basically say you expect a symbol. A bang is not a symbol, neither is a float or a pointer. `[list foo(` can be easily interpreted as a symbol, since it's first atom is a symbol. Honestly, I'm not sure what to do with `[foo(`... I tend to think it should stay an error.

> Maybe I'll reboot this PR once again

That's probably a good idea. Let's not copy all code from PurrData but make changes incrementally.

> All I can say is that adding proper warnings require quite a good amount of surgery in the code - and I'm kinda shy to attempt that :)

I think it's not so bad. It's a bit more code when we deal with every combination of input/output types explicitly, but it also makes the code easier to reason about. I think there's no other way if we want to introduce meaningful warnings/error messages.

porres:
on it

PR #688, opened 2019-07-21
Emscripten support for libpd

claudeha:
The main fixes are because Emscripten (a cross-compiler for C and C++ to Javascript and Web Assembly for web) is a lot stricter about types of function pointers.

Worked fine early this year, but not tested very recently, still work in progress - but perhaps less controversial parts can be cherry-picked and included in main Pd?


src/m_class.c
```diff
@@ -919,22 +919,179 @@ t_pd *pd_newest(void)
     calls in typedmess(), to keep the compiler quiet. */
 typedef t_pd *(*t_newgimme)(t_symbol *s, int argc, t_atom *argv);
 typedef void(*t_messgimme)(t_pd *x, t_symbol *s, int argc, t_atom *argv);
+typedef void*(*t_messgimmer)(t_pd *x, t_symbol *s, int argc, t_atom *argv);
```
danomatika: Maybe a comment is good for these huge masses of typedefs. Maybe there is a way to have these in a separate file?

claudeha: added some commentary in 1c9385abd1790ebcc13cbc1e24fb55b8b4a9a1db

src/m_class.c
```diff
@@ -933,24 +933,194 @@ t_pd *pd_newest(void)
...
+    t_floatarg d1);
+typedef t_pd *(*t_fun31)(t_int i1, t_int i2, t_int i3,
+    t_floatarg d1);
+typedef t_pd *(*t_fun41)(t_int i1, t_int i2, t_int i3, t_int i4,
+    t_floatarg d1);
+typedef t_pd *(*t_fun51)(t_int i1, t_int i2, t_int i3, t_int i4, t_int i5,
+    t_floatarg d1);
+typedef t_pd *(*t_fun61)(t_int i1, t_int i2, t_int i3, t_int i4, t_int i5, t_int i6,
+    t_floatarg d1);
+
```
Spacechild1, line 975: Some of those function prototypes (and the corresponding case labels in `pd_typedmess`) are redundant. Pd only allows up to 5 typed arguments, so we don't need `t_fun53`, for example.

Also, why do we need `t_fun6*` or `t_fun*6` at all if there are only max. 5 arguments? The original implementation also did this, but I don't really understand why. Can you figure it out and add an explanation? :-)

Spacechild1: And just in case anyone worries: having 21 case labels instead of 6 doesn't make a difference in performance because the compiler turns the switch statement into a jump table. Here's my version of the code with the assembly output:
https://godbolt.org/z/49vbq3x13

GCC produces a jump table, as expected.

Clang, however, does some crazy optimizations which I don't 100% understand. I *think* it recognizes that all those function signatures only differ in the number of integer resp. float arguments. With the System V AMD64 ABI, the first 6 integer/pointer arguments and the first 8 floating point arguments are passed in registers (https://en.wikipedia.org/wiki/X86_calling_conventions#System_V_AMD64_ABI). This means we can simply iterate over the `t_atom` argument vector, push the values to the appropriate registers and then call the function pointer *unconditionally*!

BTW, I encode the number of symbol/gpointer and float arguments in hex instead of decimal, because it is a bit more efficient and also more elegant :-)

Jonathhhan:
@claudeha 
If I replace this files in ofxPd:
```
d_array.c 
d_osc.c
d_ugen.c
m_class.c
m_class_dispatcher_1.h
m_class_dispatcher_2.h
m_imp.h
m_obj.c
m_pd.c
m_pd.h
s_inter.c
x_connective.c
x_list.c
x_misc.c
x_text.c
```
with the file from your branch, I can run ofxPd with Emscripten, which is really great.
Are there any plans to merge it into the master branch?

The only thing, that does not seem to work is closePatch(): https://github.com/danomatika/ofxPd/issues/88

claudeha:
Rebased on pure-data/master again.

> Are there any plans to merge it into the master branch?

I think merging is a good idea, saves me having to do so much maintenance work - it's probably not perfect, but good enough, and wider testing will help bugs get fixed faster.

> To be precise in this line: `if (c->c_freemethod) (*(t_gotfn)(c->c_freemethod))(x);`

Does this change fix it?  If so I'll commit and push.

```
diff --git a/src/m_pd.c b/src/m_pd.c
index d2c4ea08..da33ef52 100644
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -27,10 +27,12 @@ t_pd *pd_new(t_class *c)
     return (x);
 }
 
+typedef void (*t_freemethod)(t_pd *);
+
 void pd_free(t_pd *x)
 {
     t_class *c = *x;
-    if (c->c_freemethod) (*(t_gotfn)(c->c_freemethod))(x);
+    if (c->c_freemethod) (*(t_freemethod)(c->c_freemethod))(x);
     if (c->c_patchable)
     {
         while (((t_object *)x)->ob_outlet)

```

danomatika:
Please note: there will be a major update to libpd which adds internal multi-instance support. Included is a `libpd_freehook` as well.

Jonathhhan:
@claudha thank you very much, 
```
diff --git a/src/m_pd.c b/src/m_pd.c
index d2c4ea08..da33ef52 100644
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -27,10 +27,12 @@ t_pd *pd_new(t_class *c)
     return (x);
 }
 
+typedef void (*t_freemethod)(t_pd *);
+
 void pd_free(t_pd *x)
 {
     t_class *c = *x;
-    if (c->c_freemethod) (*(t_gotfn)(c->c_freemethod))(x);
+    if (c->c_freemethod) (*(t_freemethod)(c->c_freemethod))(x);
     if (c->c_patchable)
     {
         while (((t_object *)x)->ob_outlet)

```
this fixes the issue and now its possible to open and close patches with Emscripten.
One thing to mention: I still use this libpd branch: https://github.com/cuinjune/Ofelia/tree/master/libs/ofxPd/libs/libpd
for using it with Open Frameworks and ofxPd, because its the only branch that works. But I think that the only difference is in `m_class.c`.

Jonathhhan:
@claudeha  I updated the current libpd for Emscripten. Its basically your changes, but I had to adapt m_class.c from ofxOfelia.
Not sure about other platforms, but OF and Emscripten work very well with this libpd branch.
Here are my changes compared to the official libpd branch: https://github.com/Jonathhhan/ofxOfelia/compare/update_compare...Jonathhhan:ofxOfelia:update

Jonathhhan:
One thing I recognized is that it is not possible to build with SAFE_HEAP=1 because of `class_addcreator` and `class_addmethod` in `m_class.c`. Maybe it is caused by the vargs and maybe this issue is related: https://github.com/emscripten-core/emscripten/pull/14026/files

Jonathhhan:
And here is a branch updated from the current Pure Data Master branch (0.52-2) which seems to work well: https://github.com/Jonathhhan/ofxOfelia/compare/update_compare_0.52-2...Jonathhhan:ofxOfelia:update_0.52-2

umlaeute:
@millerpuckette pinh

charlesneimog:
Hi everyone.

I'd like to endorse this Pull Request. The possibility of running PureData on WebBrowser seems to be particularly important in Brazil, where not all universities have electroacoustic composers, and simplifying configurations is crucial mainly for the point of view of performers (Plug and Play using websites it is amazing). 

Currently, I am working on [pd4web](https://github.com/charlesneimog/pd4web) where we have a Python script that makes all the config for the compilation of PureData embedded externals, as well as other externals to run in web browsers. It already supports else, cyclone, pmpd, TimbreIDLib, earplug~ and others that I use in my work.

Obs.: I did the update for pd0.54 with no problems at all. 

Here are some examples of the works currently in development:

* It uses plaits~ from `else` am `pmpd`: https://charlesneimog.github.io/Algorithm-Music/Piece-I/
* https://www.charlesneimog.github.io/Compiled-I/webpatch/index.html <= Under development, this is more complex and yet seems to work very well.

If you see the console of the browser, you can see all the info about Puredata.

It would be great if this could be merged for `pd0.55` or next release...

Thank you very much!!!!

Jonathhhan:
@claudeha it would be really nice, if this will work with ofxPd together. Right now I have this branch: https://github.com/Jonathhhan/ofxPD_Emscripten

millerpuckette:
... and as a note to self .... in future I should test Pd releases by running huge ugly patches under emscripten hoping to catch run-time errors caused by typecast hoohas.

PR #689, opened 2019-07-24
pdcontrol args method

porres:
here's a suggestion to get the patch's arguments as a list with a new 'args' method with the new [pdcontrol] object.

PR #690, opened 2019-07-24
[pdcontrol] fixes

porres:
fixed two issues I found

see also: https://github.com/pure-data/pure-data/pull/689 for an addition of the 'args' method


porres:
these have been merged, see https://github.com/pure-data/pure-data/pull/698#issuecomment-517883632

Issue #696, opened 2019-07-25
delay size limit


porres:
There's an old bug I've reported once on the limit of delay lines. I plan to give it a try to fix it. 

For reference: https://sourceforge.net/p/pure-data/bugs/1217/

porres:
in short, the issue is that the delay size is hardcoded to 64, so we have issues when using larger block sizes, where the given delay size is always: buffer_size - block_size + 64, and it should be buffer_size + block_size

porres:
by the way, this is kinda way overdue, as I reported this bug while in the 0.46 cycle :)

Issue #697, opened 2019-07-25
different blocks delay (documentation and error)


porres:
Delays with different block sizes (or also different overlaps/oversampling). More specifically, they should be under the same block tree to avoid any issue. 

We should document this in the help file and maybe give an error in the code, like in https://github.com/pure-data/pure-data/blob/master/src/d_delay.c#L230

Now, I see there's a suppressed error when checking for different block sizes, maybe we could do something like that it https://github.com/pure-data/pure-data/blob/master/src/d_delay.c#L74

Spacechild1:
> More specifically, they should be under the same block tree

what do you mean by that?

> maybe give an error in the code

certainly not! delread~ and delwrite~ *do* work with different block sizes, overlap and upsampling. the result is well-defined, although it's not always immediately obvious. some algorithms require this behavior. making it an error is not necessary and will break existing patches. I would say it's a feature for expert users :-)

> We should document this in the help file

I agree this should be documented, maybe something in the line of: "usually, delread~ should have the same block size, overlap and upsampling factors as the corresponding delwrite~. otherwise you might get unexpected results."

porres:
>  "usually, delread~ should have the same block size, overlap and upsampling factors as the corresponding delwrite~. otherwise you might get unexpected results."

sounds good.

> some algorithms require this behavior. making it an error is not necessary and will break existing patches. I would say it's a feature for expert users :-)

hey, I like to think I'm an expert, but I don't understand how this works :( 

how about an advanced tutorial for experts to understand and make sense of this? :)

Spacechild1:
> hey, I like to think I'm an expert, but I don't understand how this works :(

as a consolation, I keep forgetting the details myself all the time... 

overlap itself doesn't do anything special, but it goes hand in hand with reblocking. reblocking only leads to troubles if the [delwrite~] has a larger blocksize than the [delread~], e.g.:

[delwrite~] with blocksize 256
[delread~] with blocksize 64

[delwrite~] will update the phase of the ringbuffer every 256 samples, but [delread~] reads from the buffer every 64 samples, so it will always read the same 64 samples 4 times in a row (assuming that the delay time doesn't change).

[...]

PR #694, opened 2019-07-24
Trigger updates

porres:
- commit: https://github.com/pure-data/pure-data/pull/694/commits/e3fbd92026c6e6bccb81614391e1723d3c1e243b closes https://github.com/pure-data/pure-data/issues/693;

- https://github.com/pure-data/pure-data/pull/694/commits/58e664d20b91823648e37a7be4fcb6f839e96bb2 & https://github.com/pure-data/pure-data/pull/694/commits/51b128384a25e4bb184eb1dbfeea45142ff87b1d close https://github.com/pure-data/pure-data/issues/683;

- https://github.com/pure-data/pure-data/pull/694/commits/1fc9a5baefbf8975144c21b6ef54c4bd7cb602f9 closes https://github.com/pure-data/pure-data/issues/684;

- https://github.com/pure-data/pure-data/pull/694/commits/c46cc08dd93480a4f71160c6274dab0cfccd11da & https://github.com/pure-data/pure-data/pull/694/commits/e8da4c45a2ff8ebb5291eeb87294100edc75796a close https://github.com/pure-data/pure-data/issues/695;

- https://github.com/pure-data/pure-data/pull/694/commits/cf0b2843537f921e29f0006e93be6d66e2c8e1c6 closes https://github.com/pure-data/pure-data/issues/685;

- https://github.com/pure-data/pure-data/pull/694/commits/f2617fb00273b03025a77d6b674f24669919a392 & https://github.com/pure-data/pure-data/pull/694/commits/cf0ff631f89afea4e7807179d4e6af696b3ef9b6 revise the documentation and includes changes from above;

- https://github.com/pure-data/pure-data/pull/694/commits/5fa2b5fc7ecc693dff069762d370458158b3fe5a closes https://github.com/pure-data/pure-data/issues/691;

- https://github.com/pure-data/pure-data/pull/694/commits/89a159ea678c42c7a981dcd8aafb168886a08e47 adds "constant float" feature to the documentation;

PR #698, opened 2019-07-25
Delay fixes

porres:
the most important commit closes https://github.com/pure-data/pure-data/issues/696

I'm yet to give it a try to https://github.com/pure-data/pure-data/issues/697

### DIFF
--- a/src/d_delay.c
+++ b/src/d_delay.c
@@ -9,7 +9,7 @@
 extern int ugen_getsortno(void);
 
 #define DEFDELVS 64             /* LATER get this from canvas at DSP time */
-static const int delread_zero = 0;    /* four bytes of zero for delread~, vd~*/
+static const int delread_zero = 0;    /* four bytes of zero for delread~, vd~/delread4~*/
 
 /* ----------------------------- delwrite~ ----------------------------- */
 static t_class *sigdelwrite_class;
@@ -244,7 +244,7 @@ static void sigdelread_setup(void)
 }
 
 
-/* ----------------------------- vd~ ----------------------------- */
+/* ----------------------------- vd~ / delread4~ ----------------------------- */
 static t_class *sigvd_class;
 
 typedef struct _sigvd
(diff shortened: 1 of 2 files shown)


millerpuckette:
I applied first 3 of these - the last one needs more work (I think) to check that delread objects stay consistent with delwrite.

porres:
thanks, the first two commits were actually from another PR (and I screwed up and had it also in here), guess I'll close it https://github.com/pure-data/pure-data/pull/690

porres:
> the last one needs more work (I think) to check that delread objects stay consistent with delwrite.

not sure what you mean, maybe that blocksizes should be of the same size? Anyway, I think this is also to be considered https://github.com/pure-data/pure-data/issues/697

porres:
> thanks, the first two commits were actually from another PR (and I screwed up and had it also in here)

I also have them as a leftovers in this other PR => https://github.com/pure-data/pure-data/pull/694  not sure how it'll work now because they've already been merged... I'll be extra careful not to mess this up again.

millerpuckette:
no worries.  git is deeply confusing, but at least it seems to deal with duplicate commits well.

src/d_delay.c
```diff
@@ -30,6 +30,7 @@ typedef struct _sigdelwrite
     int x_sortno;   /* DSP sort number at which this was last put on chain */
     int x_rsortno;  /* DSP sort # for first delread or write in chain */
     int x_vecsize;  /* vector size for delread~ to use */
+    int x_blocksize;
```
Spacechild1: no need to add `x_blocksize`, just use `x_vecsize`

src/d_delay.c
```diff
@@ -123,6 +125,7 @@ static t_int *sigdelwrite_perform(t_int *w)
 
 static void sigdelwrite_dsp(t_sigdelwrite *x, t_signal **sp)
 {
+    x->x_blocksize = sp[0]->s_n;
     dsp_add(sigdelwrite_perform, 3, sp[0]->s_vec, &x->x_cspace, sp[0]->s_n);
     x->x_sortno = ugen_getsortno();
     sigdelwrite_checkvecsize(x, sp[0]->s_n);
```
Spacechild1: `sigdelwrite_checkvecsize` will already set `x_vecsize` to the block size.

you would need to make sure that `sigdelwrite_checkvecsize` comes *before* `sigdelwrite_checkvecsize` in `sigdelread_dsp/sigvd_dsp`

src/d_delay.c
```diff
@@ -89,6 +90,7 @@ static void *sigdelwrite_new(t_symbol *s, t_floatarg msec)
     x->x_sortno = 0;
     x->x_vecsize = 0;
```
Spacechild1: we should also do `x_rsortno = 0;`

src/d_delay.c
Spacechild1: > basically, the first object in the DSP chain determines the blocksize

additionally, subsequent calls to `sigdelwrite_checkvecsize` should update `x_vecsize` if `vecsize` is *larger*. This makes sure that we resize the delay line according to the *largest* blocksize.

porres:
well @Spacechild1  , I tried your suggestions, I could compile and test (and it didn't help), but now there are building errors here :/

Spacechild1:
you've missed some details. try this one: https://github.com/pure-data/pure-data/pull/708

porres:
I will, but first let me thank you for all the great work you've been doing :) 

you rock!

porres:
Obviously https://github.com/pure-data/pure-data/pull/708 works like a charm, so I'm just closing this :)

porres:
it seems https://github.com/pure-data/pure-data/pull/708 fixes this, right?

PR #708, opened 2019-08-04
delwrite~: allocate extra space based on actual block size

Spacechild1:
in the case of conflicting block sizes, use the largest (some algorithms require different block sizes for delwrite~ and delread~).


porres:
closes https://github.com/pure-data/pure-data/issues/696

porres:
this closes https://github.com/pure-data/pure-data/issues/697

PR #710, opened 2019-08-05
improve portaudio scheduler

Spacechild1:
* `sys_microsleep` only sleeps (because it is called in contexts where Pd is not locked)
* `sys_domicrosleep` -> `sys_pollsockets` - only polls sockets
* `pa_send_dacs`: first try `sys_pollgui()` before going to sleep. This makes sure we actually use the idle time! 

This PR implicitly fixes a problem where excessive incoming network traffic would freeze the GUI (when DSP is on): https://github.com/pure-data/pure-data/issues/55

### DIFF
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -436,6 +436,7 @@ the audio I/O system is still busy with previous transfers.
 
 void sys_pollmidiqueue(void);
 void sys_initmidiqueue(void);
+int sys_pollsockets(void);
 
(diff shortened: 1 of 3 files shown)


Spacechild1:
ping @millerpuckette

Spacechild1:
ah, just saw you new commits. they should the same thing. personally, I think that getting rid of `sys_domicrosleep` makes things easier to reason about, but that's probably a matter of taste.

feel free to close this.

Spacechild1:
will make a new PR later

Issue #401, opened 2018-07-13
Limit drawing into table/array/canvas borders


alfonso73:
Hi all, as it is now any canvas related object (array, table) let the mouse draw outside its borders.
I think is quite unconfortable and quite ugly to see on a GUI.
Would be great to just trim any drawing that happens outside the canvas borders.
best!

alfonso73:
Hi, is there any advancement on this topic? Is really hard to use tables and arrays at runtime when doing sound design.

porres:
> Hi, is there any advancement on this topic?

Please don't assume all issues will be reviewed and taken care of, unfortunately that's not how it works. I pointed this issue in a PR that adds this functionality to arrays. Let's see what the developer thinks. I think this can be useful, not sure if I can make an attempt on implementing this. I think there should be a setting to clip values to the range, btw.

if the table isn't too large, I think you can have some workarounds in the meantime, by getting all values of the table at a fixed rate and then cliping the values manually and resseting them.

PR #711, opened 2019-08-06
[WIP] Array features: dynamically set color, width, style and visibility

Spacechild1:
This PR adds the following messages for garrays:

* `[color <f>(` - set line/point color (in data structure color format)
* `[width <f>(` - set line/point width
* `[style <f>(`- switch between point, line and bezier style
* `[vis <f>(` - show/hide array. Very useful if there's more than 1 array in a graph!

I've documented these features in `2.control.examples/16.more.arrays.pd`.

Small example:
graph

### DIFF
--- a/doc/2.control.examples/16.more.arrays.pd
+++ b/doc/2.control.examples/16.more.arrays.pd
@@ -1,4 +1,4 @@
-#N canvas 288 82 910 558 12;
+#N canvas 288 82 910 685 12;
 #N canvas 0 50 450 250 (subpatch) 0;
 #X array array99 5 float 1;
 #A 0 0.32 0.499999 -0.406667 -0.753333 0.00666714;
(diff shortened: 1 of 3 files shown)


porres:
Wow... you know what would be really awesome? a "bar graph" representation as well :)

Spacechild1:
oh, that would be nice. I can give it a try.

porres:
Check Purr Data, they have actually implemented that... ;)

porres:
though I'm not sure if this was in Pd-l2ork or only in purr data with the new GUI, hopefully in pd-l2ork...

porres:
see https://puredata.info/Members/jancsika/bargraph-array/view

Spacechild1:
Ha! I'm currently working on adding the color option to the properties dialog. I think I'll have a look at how the GUI callback message looks like in PurrData to avoid incompatibilites (if possible).

porres:
It really seems this feature was available before "purr data", so here it is https://github.com/pd-l2ork/pd/search?q=bargraph&unscoped_q=bargraph

jmmmp:
Sidestep: I made a new abstraction [jmmmp/multiarray] with these features. But surely more efficient having them in the core.

esaruoho:
what about being able to set the array background colour? that'd be really useful for me ( 
image ) - so that i'd be better able to differentiate between which one i am sampling to or modifying

Spacechild1:
@esaruoho Just use a `[cnv]` behind the array and change the color with the `[color(` message?

jmmmp:
You can see jmmmp/multiarray, that and many other details are possible.

Christof Ressi <notifications@github.com> schrieb am So., 2. Aug. 2020,
15:02:

esaruoho:
> You can see jmmmp/multiarray, that and many other details are possible. 

@jmmmp is one `jmmmp/multiarray` less heavy on the cpu than 16 arrays?

jmmmp:
[multiarray] will always be heavier on the cpu as any built-in solution,
because it's an abstraction with GOP. you can save some computation by
switching off the gop if there is some intensive array editing happening.
but because it's heavier doesn't mean that it's "thaaat" heavy. you'll have
to test with your patch what you consider to be heavy.

porres:
Do we have to set the color in the Data Structures style? I think its limited resolution is pretty bad. Not only I think we should have a better resolution in arrays, but also in Data Structures in general (there might be a PR for that somewhere, right?)

Spacechild1:
> I think its limited resolution is pretty bad

Well, garrays *are* data structures. I've had the idea to allow using hex color strings for all data structures (as symbols), but that would be another PR.

I can see how the RGB style could be handy: `[color <r> <g> <b>(`, but this doesn't do much, unless data structures can actually handle high resolution colors. For example, we could take negative numbers meaning high res colors, like the iemguis. But this should be really a seperate PR because it applies to all data structure drawing instructions.

porres:
what do you think about a "clip" message or setting in the properties, to deal with this request? https://github.com/pure-data/pure-data/issues/401

umlaeute:
> I've had the idea to allow using hex color strings for all data structures. 

yes please.

> Also, we could take negative numbers meaning high res colors, like the iemguis.

please don't.

> But this should be really a seperate PR

yes please.

porres:
> I've had the idea to allow using hex color strings for all data structures.

+1   🎉

PR #715, opened 2019-08-08
Feature/netobject updates

Spacechild1:
@danomatika the thing is that you're not supposed to bind to the multicast address. bind only determines which interfaces you listen to. binding to the multicast address causes an error (at least on Windows).

here's a reference: https://www.ibm.com/support/knowledgecenter/en/ssw_ibm_i_71/rzab6/xmulticast.htm

also, it's possible to join more than 1 multicast group. you can also leave them.
so my suggestion is:
1) use `[listen <port> <host>(` only to select the interface (TCP and UDP), e.g. to only allow traffic from "localhost"
2) add [multicast_join( and [multicast_leave( methods to join/leave multicast groups

With this PR I can successfully join/leave several multicast groups. I've tested locally and with another machine in the local network.

The only caveat is that the IP family of `[listen(` and `[multicast_*(` must match. This means you might have to explicitly do `[listen 0.0.0.0(` if you want to join IPv4 multicast groups. I don't like this but I don't see an easy way around this. FWIW, I post a warning if the IP family doesn't match and suggest to bind to the correct "any" address.

Need to update the docs, but only if you're happy with the implementation.

### DIFF
--- a/src/s_net.c
+++ b/src/s_net.c
@@ -369,6 +369,29 @@ int socket_join_multicast_group(int socket, const struct sockaddr *sa)
     return -1;
 }
 
+int socket_leave_multicast_group(int socket, const struct sockaddr *sa)
+{
+    if (sa->sa_family == AF_INET6)
+    {
+        struct sockaddr_in6 *sa6 = (struct sockaddr_in6 *)sa;
+        struct ipv6_mreq mreq6 = {0};
+        memcpy(&mreq6.ipv6mr_multiaddr, &sa6->sin6_addr,
+            sizeof(struct in6_addr));
+        return setsockopt(socket, IPPROTO_IPV6, IPV6_LEAVE_GROUP,
+            (char *)&mreq6, sizeof(mreq6));
+    }
+    else if (sa->sa_family == AF_INET)
+    {
+        struct sockaddr_in *sa4 = (struct sockaddr_in *)sa;
+        struct ip_mreq mreq = {0};
+        mreq.imr_multiaddr.s_addr = sa4->sin_addr.s_addr;
+        mreq.imr_interface.s_addr = INADDR_ANY;
+        return setsockopt(socket, IPPROTO_IP, IP_DROP_MEMBERSHIP,
+            (char *)&mreq, sizeof(mreq));
+    }
+    return -1;
+}
+
 int socket_errno(void)
 {
 #ifdef _WIN32
--- a/src/s_net.h
+++ b/src/s_net.h
@@ -125,6 +125,9 @@ int socket_set_nonblocking(int socket, int nonblocking);
 /// join a multicast group address, returns < 0 on error
 int socket_join_multicast_group(int socket, const struct sockaddr *sa);
 
+/// leave a multicast group address, returns < 0 on error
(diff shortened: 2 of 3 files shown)


Spacechild1:
Give it a try. We can talk again tomorrow or so. I'm offline now.

danomatika:
ping @umlaeute 

I remember reading somewhere that closing a socket bound to multicast handled leaving the group for you, but you want to make this more hot-swappable and allow for receiving multiple multicast messages on the same socket?

I know it's all *possible* but is that something Pd needs?

Spacechild1:
>  a socket bound to multicast

again, you should not *bind* to the multicast address. On Windows, I even get an error when trying to do so. You bind to an interface and then join one (or several) multicast groups. 

The listening interface and a multicast address are two different things and shouldn't be mixed in the same argument. That's why I ripped multicast out of `[listen(` and added dedicated messages. Joining several groups comes for free.

danomatika:
>again, you should not bind to the multicast address.

Semantics. The multicast stuff happens after bind, I get it.

I guess this is a windows difference as the current approach is working on macOS (?). I did check the mrpeach network objects but decided the multicast interface stuff was way too much OS-specific stuff, ie. listing interfaces and making them selectable.

Also, how about a modification of the messaging as I'm not aware of other core objects using underscore naming:

* multicast join ...
* multicast leave ...

Spacechild1:
> Semantics. The multicast stuff happens after bind, I get it.

In the articles I've read, people *don't* bind to the multicast address. It makes sense, because you listen to *one* interface, but you can join *several* multicast groups.

> Also, how about a modification of the messaging as I'm not aware of other core objects using underscore naming:

Ah, sure. I don't care about the naming :-)

danomatika:
>In the articles I've read, people don't bind to the multicast address. It makes sense, because you listen to one interface, but you can join several multicast groups.

Ok. I focused on trying to add the *minimum* required to make it work but if there is more needed than so be it. What's in the main PR is working for me, but maybe I don't really know how *well*.

Spacechild1:
I'll test on Linux.

Spacechild1:
OK, Linux doesn't complain about binding to the multicast address either, so this might be a Windows thing. 

Basically, I see three alternatives:

1) set the multicast address as the second argument to `[listen(` and bind to the correct "any" address (at least on Windows).
*Pro:* very simple
*Contra:* can't set the interface, less flexible
2) use the [multicast*( methods to join/leave multicast groups
*Pro:* flexible, can additionally set the interface
*Contra:* have to explicitly listen to the correct "any" address
3) a combination of 1 + 2
*Pro:* simple default + additional features
*Contra:* ?

Honestly, 2) might be a bit hairy but 3) could be quite powerful. 1) might also do it.
I would wait for @umlaeute's opinion as I'm not a networking expert by any means.
But when we decide for something, I can implement it and test it on all 3 platforms.

Spacechild1:
@danomatika here's the simple 1) version: https://github.com/pure-data/pure-data/pull/716

actually, for me this looks good enough. joining/leaving multicast groups could be added if needed.

Spacechild1:
closing for now

PR #716, opened 2019-08-09
Feature/netobject updates2

Spacechild1:
This fixes multicast for Windows. Basically, we can't bind to the multicast address (https://stackoverflow.com/questions/6140734/cannot-bind-to-multicast-address-windows) so we bind to the "any" address instead.

This is a lightweight alternative to https://github.com/pure-data/pure-data/pull/715

### DIFF
--- a/src/s_net.c
+++ b/src/s_net.c
@@ -369,6 +369,29 @@ int socket_join_multicast_group(int socket, const struct sockaddr *sa)
     return -1;
 }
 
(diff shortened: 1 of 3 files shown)


danomatika:
Should we call `sys_sockerror("bind");` in `netreceive_listen()`? Since the bind can fail before trying the next address which may ultimately succeed, I feel that it's not needed as the :listen failed" error is printed later on anyway.

PR #723, opened 2019-08-11
Feature/netobject updates3

Spacechild1:
1) add a warning in netsend-help.pd that "timeout" value shouldn't be too small
2) `[netsend]/[netreceive]`: add list method as an alias for "send". This is especially handy when sending binary data, e.g. `[oscformat] -> [netsend -u -b]`

For 2) I've added a notice in the help patches that this is Pd 0.50+.


danomatika:
Will a list starting with "connect ..." be sent as a list or interpreted as the control message. I can see a situation where you build commands via lists and forget the [list trim]. This is more likely to happen when *not* using binary.

Spacechild1:
A list is always sent as a message. I don't think it should be restricted to binary.

> I can see a situation where you build commands via lists and forget the [list trim].

Doesn't this happen with all objects which accepts lists and messages, e.g.` [list store]`?

danomatika:
I'm not sure, so I'm asking. I know the mrpeach net objects work this way, but I'd like to double-check. If there is a risk, we probably need to communicate it some way and suggest using the explicit "send" prefix for the edge case.

Spacechild1:
The question was rethorical :-)

Spacechild1:
`[list append foo bar( -> [list store]` -> Oops

Spacechild1:
come to think of it: [netsend]/[netreceive] should probably print an error when trying to send while not connected/listening, like the mrpeach objects.

Spacechild1:
Anyway, it's easy to revert if Miller doesn't like it.

doc/5.reference/oscformat-help.pd
```diff
@@ -56,6 +56,7 @@ made for the rest.;
 #X text 306 666 Note: there's no way using oscparse to distinguish
 between floats and integers \, nor to see blobs unambiguously., f
 40;
+#X text 463 304 ("send" not needed in Pd 0.50+);
```
danomatika: I would rather say:

> "send" is optional for lists in Pd 0.50+

PR #724, opened 2019-08-11
Feature/netobject updates4

Spacechild1:
1) multicast: always bind to "any" address. Binding to an IPv6 multicast address doesn't work on Linux, either, so I think it's another indicator that this isn't good practice. Feel free to change `#if 1` to `#ifndef __APPLE__` if this reliably works on macOS, though!
2) improve pdsend/pdreceive print out
3) pdreceive: allow TCP hostname and move "host" argument *after* tcp/udp (it's less confusing)

### DIFF
--- a/src/u_pdreceive.c
+++ b/src/u_pdreceive.c
@@ -106,20 +106,20 @@ int main(int argc, char **argv)
         {
             /* stream (TCP) sockets are set NODELAY */
             if (socket_set_boolopt(sockfd, IPPROTO_TCP, TCP_NODELAY, 1) < 0)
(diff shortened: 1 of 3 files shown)


danomatika:
You could remove the "netreceive: " and (possibly) "netsend: " fprintf prefixes.

Spacechild1:
ouch, of course. will do.

danomatika:
I already got the quotes. Do a pull.

Spacechild1:
I don't see any merge conflicts

Spacechild1:
@danomatika now I see what you mean. Also, something was missing in my last commit. I've rebased on your branch and force pushed.

PR #725, opened 2019-08-12
[text define]: add field count argument to [sort( method

Spacechild1:
This adds an argument to the `[sort(` message so you can restrict the number of fields to compare.

Practical example: Let's say I have lists of "timestamp + event" and I only want to sort based on the timestamp but *don't* change the order of events with the same timestamp (e.g. a note-off followed immediately by a note-on)

I've decided to turn the "-k" flag into an argument, so the "sort" method takes two optional arguments: `[sort <onset> <count>(`

Of course, it would've been possible to implement the field count as a flag. Personally, I prefer positional arguments in such cases, as it matches the interface of `[text get], [text set]` etc.

@millerpuckette Sorry for the nagging :-) "sort" is great, BTW!

### DIFF
--- a/doc/5.reference/text-object-help.pd
+++ b/doc/5.reference/text-object-help.pd
@@ -12,7 +12,7 @@
 #X obj 280 217 text define;
 #X text 81 173 The text object's first argument sets its function:
 , f 30;
-#N canvas 373 69 684 718 define 0;
+#N canvas 788 67 684 718 define 0;
 #X msg 38 80 clear;
 #X msg 38 106 read text-object-help.txt;
 #X msg 38 132 write text-object-help.txt;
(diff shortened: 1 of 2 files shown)


millerpuckette:
difficulty: qsort and qsort_r don't define what they will do to lines for which the compare
function returns "equal".  So other than writing my own sort function I don't know how to do
this :)

Spacechild1:
But we never return 0 from the compare function! If the fields compare equal, you do a pointer comparison to maintain the original order.

Spacechild1:
BTW, if you prefer a flag, e.g. "-n", instead of positional arguments, I can rework the PR (if you want).

millerpuckette:
right (I now remember thinking quite hard about that earlier and totally forgot what I did.)
Unless there's no way to add the range argument compatibly later, I'd rather hold onto this for 0.51.

Spacechild1:
my actual question is, do you prefer:
a) `-k` (= onset) and `-n` (= number of fields) or
b) `[sort <onset> <fields>(`

if you prefer a) then of course the "-n" flag could be added later, but if you prefer b) then this has to be changed *before* the release.

I know that you're very busy with this release, but the range argument is rather trivial (especially when implemented as a flag) and I'm offering to do that for you. But if you want to wait, that's fine as well!

Spacechild1:
c) `-k (= onset) + [sort <fields>(`
this also a possibility which can be added later

millerpuckette:
I think the best way is "sort -k 1 3" (etc).  This leaves room for adding more features later (and I can think of lots - this is a slippery slope :)

Spacechild1:
looks good! shall I do a PR? and would you mind changing "-k" to "-s" (like "start")? "-k" is mostly used for "keep".

Spacechild1:
sorry for the bikeshedding!

Spacechild1:
just to be sure: "sort -k 1 3" means that 3 is a positional argument and 1 belongs to "-k", right?

millerpuckette:
I think "-k" is better since it matches posix sort command line usage (which seems important enough to justify possible confusion about re-using "-k").

Spacechild1:
ah, I see! I've wondered where it comes from.

millerpuckette:
I'm thinking "-k <first field> [number of fields]" .  So for instance "sort -k 3 2 -r" would reverse-sort
and  key on fields 3 and 4, ignoring the rest.

Spacechild1:
Ah, I see. Does the -k flag take an *optional* second argument? Because would rule out any future positional float arguments (probably not happening anway). Or does the -k flag always take 2 arguments? IMHO, a seperate -n flag for the number of fields might be more intuitive...

Spacechild1:
>  "-k [number of fields]"

where's the onset argument? ;-)

Spacechild1:
> IMHO, a seperate -n flag for the number of fields might be more intuitive...

ah, forget it. I think `-k <onset> [number of fields]` is just fine.

Spacechild1:
want to do a new PR

PR #733, opened 2019-08-16
zoom bugfixes

jyg:
Some fixes (in a try-and-error approach) for enabling patch editing with zoom x 2.
Due to changes in xtopixels, ytopixels, pixelstox, pixelstoy fonctions, we loosed the ability to editing patch with zoom x2, resulting in huge differences  between mouse cursor position and created object position. (the same for bng, hsl, hdial, etc.)
In this fix proposal, I modified g_editor.c/glist_getnextxy(), g_editor.c/canvas_startmotion() and g_text.c/canvas_howputnew() functions to take care of those changes.
Further, I modified g_canvas.c/*canvas_new() and g_canvas.c/canvas_coords() in order to enable creating and displaying  a GOP abstraction( or a graph) with correct scale in a patch that is already with zoom mode = 2.
At least, I made a change in drawnumber_vis() to display correctly font size in datastructures with zoomx2 and integrated a bugfix mentionned here (https://github.com/pure-data/pure-data/pull/563#issuecomment-519444185) and here (https://github.com/pure-data/pure-data/commit/f1516d89c133791f6bfd5f444bf41496058a35d0#commitcomment-34674313)

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -407,6 +407,11 @@ t_canvas *canvas_new(void *dummy, t_symbol *sel, int argc, t_atom *argv)
     x->gl_edit = !strncmp(x->gl_name->s_name, "Untitled", 8);
     x->gl_font = sys_nearestfontsize(font);
     x->gl_zoom = 1;
+
+/////zoomfix
+    if(owner)
+	x->gl_zoom=owner->gl_zoom;
+
     pd_pushsym(&x->gl_pd);
     return(x);
 }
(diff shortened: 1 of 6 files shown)


millerpuckette:
Applied them by hand (with some changes)... would like to hear if this fixed all the problems you had.

jyg:
For me it seems  ok, exept for the case you use autoconnect feature :
create a new patch, set zooming == 2,
 then create an object (for example [f ]), 
select it and press ctrl+1 to create another autoconnected object.
 It is not shifted the right way on the patch.
The fix I proposed seems still to work with your changes : 
In g_text.c, replace lines 162 / 163  (was 
```
*xpixp = x1;
*ypixp = y2 + 5;
```  )
with those : 
 ```
*xpixp = x1 / x->gl_zoom ;
  *ypixp = (int)(0.5 + (float)y2 / x->gl_zoom + 5);
```

PR #735, opened 2019-08-25
fix PD_LONGINTTYPE for 64-bit windows

Spacechild1:
The correct check for 64-bit Windows is `_WIN64`, not  `__x86_64__` which just happens to be defined by MinGW. 

Yesterday I've been scratching my head why my 64-bit MSVC build of a external would crash Pd (with a wiped out stack) while the MinGW build would not...

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -83,7 +83,7 @@ typedef unsigned __int64  uint64_t;
 
 /* signed and unsigned integer types the size of a pointer:  */
 #if !defined(PD_LONGINTTYPE)
-#if defined(_WIN32) && defined(__x86_64__)
+#if defined(_WIN32) && defined(_WIN64)
 #define PD_LONGINTTYPE long long
 #else
 #define PD_LONGINTTYPE long


umlaeute:
do you know whether this will yield the correct value for the `arm64` platform as well (as opposed to `amd64` aka `x86_64` resp `x64`)?

or put simpler: is `long long` on *Win/arm64* 64bit or something else?

Spacechild1:
I don't think that Window's LLP64 model depends on the actual CPU architecture (Intel vs. ARM)...

OTOH, MSVC supports "stdint.h" since 2010, so we could get rid of the #ifdef altogether and just do `#define PD_LONGINTTYPE intptr_t` - maybe even for all platforms.

I think I've read that `intptr_t` isn't officially required to be part of "stdint.h", but it certainly is on MSVC and MinGW: https://docs.microsoft.com/en-us/cpp/c-runtime-library/standard-types?view=vs-2019

Spacechild1:
Actually, `long long` is guaranteed to be at least 64 bits by the C99 standard and I don't think it's larger than that on Windows/Arm64 ;-) Also, the reason for Window's use of LLP64 is backwards compatibility with legacy code, so it's unlikely they would want to break things on ARM...

Issue #732, opened 2019-08-15
Can't input Tilde (~) character with pt-br keyboard


HenriAugusto:
I'm running pd-0.50-0test1 and i can't type tildes anymore while using the PT-BR keyboard.

I can only type them if i change my system's keyboard setting to "EN".

Here are images of both layouts, just in case somebody wants to see them
[US keyboard](https://en.wikipedia.org/wiki/British_and_American_keyboards#/media/File:KB_United_States-NoAltGr.svg) | [PT-BR keyboard](https://en.wikipedia.org/wiki/Portuguese_keyboard_layout#/media/File:KB_Portuguese_Brazil.svg)

In english you just press **shift + '** and the tilde is output directly.
In portuguese we have chars like "ã" and "õ"so here is how it works: first you type the "~" key (and nothing happens) and then _a_ or _o_ so it ouputs _ã_ or _õ_. Or if you want to type the "~" alone (as we do in PD) you just press the "~" key twice. 

When pressing it twice, as reported by the [key] obj:

In pd 0.49 the 1st time press the "~" key i get keycode 0 and then the 2nd time i get 126. An "~" is output

In pd 0.50 the 1st time press the "~" key i get keycode 0 and then the 2nd time i get 0 again. Nothing is output.

danomatika:
Does it work if you type into a GUI text field, ie. the find window? If it doesn't;t, then the problem is likely with Tk.

sevdh:
What happens if you press ~ followed by a space?

HenriAugusto:
> Does it work if you type into a GUI text field, ie. the find window? If it doesn't, then the problem is likely with Tk.

Yes. It outputs two tildes `~~` which is the same behavior throghout windows. I'm not sure how in PD it would only output one tilde `~`, which was way better.

> What happens if you press ~ followed by a space?

In an object (pd-0.50): it outputs a space
In an object (pd-0.49): outputs only a single tilde.
Throghout windows: outputs only a single tilde.

umlaeute:
i can reproduce the issue on Windows 10.

Pressing the <kbd>~</kbd> key (just left of <kbd>Return</kbd> ) twice, the GUI sends:

~~~
.x24a680 key 1 Multi_key 0;
.x24a680 key 0 Multi_key 0;

.x24a680 key 1 Multi_key 0;
.x24a680 key 0 Multi_key 0;
~~~

(the empty indicates the pause between the two keypresses)

[...]

PR #736, opened 2019-08-29
refactored sendkeys to properly handle dead keys

umlaeute:
attempt on fixing the issue with not being able to type `~` character on portuguese/brazil keyboard layouts.

this still requires testing!

Closes: https://github.com/pure-data/pure-data/issues/732

### DIFF
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -115,7 +115,7 @@ proc ::pd_bindings::global_bindings {} {
     bind all <$::modifier-Shift-Key-v> {menu_send %W vslider}
     bind all <$::modifier-Shift-Key-w> {::pd_bindings::window_close %W 1}
     bind all <$::modifier-Shift-Key-z> {menu_redo}
-    bind all <KeyPress-Escape>         {menu_send %W deselectall; ::pd_bindings::sendkey %W 1 %K %A 1}
+    bind all <KeyPress-Escape>         {menu_send %W deselectall; ::pd_bindings::sendkey %W 1 %K %A 1 %k}
 
     # OS-specific bindings
     if {$::windowingsystem eq "aqua"} {
@@ -136,10 +136,10 @@ proc ::pd_bindings::global_bindings {} {
         bind all <$::modifier-less>        {menu_raisepreviouswindow}
     }
 
-    bind all <KeyPress>         {::pd_bindings::sendkey %W 1 %K %A 0}
-    bind all <KeyRelease>       {::pd_bindings::sendkey %W 0 %K %A 0}
-    bind all <Shift-KeyPress>   {::pd_bindings::sendkey %W 1 %K %A 1}
-    bind all <Shift-KeyRelease> {::pd_bindings::sendkey %W 0 %K %A 1}
+    bind all <KeyPress>         {::pd_bindings::sendkey %W 1 %K %A 0 %k}
+    bind all <KeyRelease>       {::pd_bindings::sendkey %W 0 %K %A 0 %k}
+    bind all <Shift-KeyPress>   {::pd_bindings::sendkey %W 1 %K %A 1 %k}
+    bind all <Shift-KeyRelease> {::pd_bindings::sendkey %W 0 %K %A 1 %k}
 }
 
 # bindings for .pdwindow are found in ::pdwindow::pdwindow_bindings in pdwindow.tcl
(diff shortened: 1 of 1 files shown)


HenriAugusto:
Dead keys are working now. When you press them twice it appears two of them (which is what actually happen throghout windows so it seems correct)

It is a bit annoying though having to type `~~` and then BackSpace everytime you want to input a single `~`. For that reason i would add an _ad_hoc_ solution to the new code to only output a single `~`. It's way more convenient since `~` plays a special role in PD.

In the case of not coding this _ad_hoc_ solution for `~~` i would at least write in the release notes of the new version that you can output a single `~` by pressing <kbd>~</kbd>+<kbd>Space</kbd>. Everyone i knew just used `~` twice. Almost nobody here know about the <kbd>DeadKey</kbd>+<kbd>Space</kbd> combo  heh :)

umlaeute:
> In the case of not coding this ad_hoc solution for `~~` I would at least write in the release notes of the new version that you can output a single `~` by pressing <kbd>~</kbd>+<kbd>Space</kbd>.

but that's simply not true for the majority of users.
it's *only* true for users with keyboard layouts  where the `~`-key is used as diacritic and therefore set as a dead key.

we could add a release note telling people that dead-keys are finally fixed, and if they need to type them, they should type them as in any other application :-)

otoh, who ever reads release-notes?

millerpuckette:
I had already put on a band-aid here: 65d1dcbea654092d3fc6c129def98d0357170c4f
(just now pushed to github) - it seems to manage the immediate problem at least.  Maybe that will suffice for 0.50-1 and this nice complicated rewrite should wait for 0.51?

HenriAugusto:
> it's _only_ true for users with keyboard layouts where the `~`-key is used as diacritic and therefore set as a dead key.

I may be missing something but can't we check if we received a `~~` and change it to `~` only in this special case? I can't imagine any case where someone would actually type two tildes but typing a single one is bread and butter. And if someone, with a US kbd for example, sent a single `~` it would pass untouched.

But it is just a Quality-of-life suggestion anyway.

> otoh, who ever reads release-notes?

I do 🤓. Some people don't but it was an idea for a "courtesy" of letting at least the ones who read them know.

Btw do we have testers with spanish/french/etc keyboards? I figure their keyboard works in a similar fashion but i've never actually used them.

Lucarda:
> Btw do we have testers with spanish/french/etc keyboards? 

Tilde "`~`" is a _no-go_ for the Spanish layout (even the Latin American variant). I can’t get it in any app unless I do ALT+126 (on Windows).
On my translation of Kreidler’s tutorial I suggest people to switch to the US-layout when using Pd.

Spacechild1:
>  I can’t get it in any app unless I do ALT+126 (on Windows).

ouch... maybe we can add a possibility for people to set their own shortcut for the tilde symbol?

millerpuckette:
Can you try msp.ucsd.edu/tmp/pd-tmp.msw.zip ? It should work on Portuguese (and I think Spanish)
keyboards that you can hit the "~" modifier twice to type a tilde character into Pd.

Lucarda:
> Can you try msp.ucsd.edu/tmp/pd-tmp.msw.zip ? It should work on Portuguese (and I think Spanish)...

Didn't work for me with the "**Spanish**" layout (<kbd>Alt Gr</kbd> + <kbd>4</kbd> , <kbd>space</kbd>). With the "**Spanish Latin American**" variant is a _no-go_ across all apps.

It only works for me with this PR by @umlaeute  and the "**Spanish**" layout (Alt Gr + 4 + space).

millerpuckette:
OK, I just loaded up the Spanish (espana) layout and can confirm I can't type a tilde.  #$&^%ing hell... will look at what tcl/tk is seeing and see if I can work around it

millerpuckette:
(I don't really think alt-gr-4 + space is intuitive :)

Lucarda:
> (I don't really think alt-gr-4 + space is intuitive :)

Yes but is had always been like that (there is no regression for people with the Spanish layout).
It is really easy to add a keyboard layout in windows and switch to the US layout when using Pd.
This is what I had always suggested for readers of Kreidler’s tutorial. One of the very few out there in Spanish and it has considerable traffic. I know because the translation is hosted on my server. So people had always been aware of the "~" thing.

umlaeute:
I did some Windows10 tests with Spanish keyboard layouts, running in a VM (which runs on a Debian machine, that has a German keyboard (and accoriding keyboard layout) attached to it):

This is what i get in *notepad++* (so i assume  all "proper" applications behave like this):
- Windows10 *Spanish (United States)*: <kbd>AltGr</kbd>+<kbd>+</kbd> (the <kbd>+</kbd> is located left of the <kbd>↵</kbd>, first row)
- Windows10 *Spanish (Latin America)*: <kbd>AltGr</kbd>+<kbd>+</kbd> (as above)
- Windows10 *Spanish (Spain)*: <kbd>AltGr</kbd>+<kbd>4</kbd>-<kbd>Space</kbd>
     I don't think this is "totally unintuitive", given that on a [spanish keyboard](https://user-images.githubusercontent.com/214034/64098196-8ac47580-cd66-11e9-89f0-fc1ee99374de.jpg), the <kbd>4</kbd> is marked with a `~` as the third (AltGr-) option.

Testing with Pd I get:
- <kbd>AltGr</kbd>+<kbd>+</kbd> (for *Spanish (US)* and *Spanish (Latin America)* works on Pd-0.49-0, Pd-0.50-0test1 with this PR.
- <kbd>AltGr</kbd>+<kbd>4</kbd>-<kbd>Space</kbd> (for *Spanish (Spain)*) works with Pd-0.49-0 and this PR. However, it does not work on ~~Pd-0.50-0test1~~

I did not test 65d1dcb

Lucarda:
@umlaeute  I can confirm all your test on a Win8.1 real machine. (my previous tests were wrong because I was using a Spanish variation instead of Latin American (this appear as an option once adding the spanish language.)

I can confirm 65d1dcb is not working (alt gr + 4 , space) in the Spanish layout (in this layout is a dead-key)

umlaeute:
@Lucarda so does this mean that you  have now found a way to type `~` without having to resort to ALT+126?

umlaeute:
@millerpuckette since the problem affects portuguese, spanish and french (these i have confirmed) and probably others (which i have not confirmed), i don't think that the trivial hack in 65d1dcb suffices to fix the problem.

i'd suggest to either use "this nice complicated rewrite" or switch back to the `::pd_bindings::sendkey` implementation as found in Pd-0.49-1.

switching back to the original implementation will of course revert to the old inconsistent behaviour between KeyPress and KeyRelease messages (which is what the new `::pd_bindings::sendkey` implementation was trying to fix in the first place)

Lucarda:
> @Lucarda so does this mean that you have now found a way to type ~ without having to resort to ALT+126?

- With "**Spanish (Latin American)**" is simply <kbd>Alt Gr</kbd> + <kbd>+</kbd>. This works on all Pd versions and was _not affected_ because it is _not_ a dead key.

- With "**Spanish (Spain)**" the dead key <kbd>Alt Gr</kbd> + <kbd>4</kbd> , <kbd>space</kbd> was affected in Pd 0.50 but works with this PR.

PR #739, opened 2019-09-05
Feature (GUI): Snap to Grid while dragging objects (menu toggle)

eeropic:
Snaps objects to a coarse x,y-grid, the grid size is defined by a shared constant, 
now set to 10px, like the paste offset & shift+nudge values.

The checkbox is now under Edit menu, but could be moved next to "Zoom new windows" toggle  in preferences if Edit gets too cluttered.

I'd like to get feedback and discuss the idea, as it's been discussed on the mailing lists before and various implementations exist in forked repos.

snaptogrid_pd

TODO:
- Different Tidy Up behaviour when snapping is enabled (?)
- Refine behaviour when zoomed - currently snaps to 0.5 grid unit steps when zoomed.
- ?

NOTE:
Tested now only on Mac OSX


esaruoho:
this is awesome! i needs it! :)

esaruoho:
@porres @agraef any thoughts guys?

Ant1r:
I tested it on Linux, works as expected at first sight, I like it!

Will need allowing to change the grid size of course.

The addition of `gl_snaptogrid` in the `struct _glist` may be refused though, because it introduces a potential incompatibility at binary level with existing compiled externals... What do you think @umlaeute @millerpuckette @danomatika ?

danomatika:
Yes, please.

>and various implementations exist in forked repos

I was thinking of porting one of those but this approach looks pretty clean to me.

Spacechild1:
> The addition of gl_snaptogrid in the struct _glist may be refused though, because it introduces a potential incompatibility at binary level with existing compiled externals...

`gl_privatedata` might be a good place

eeropic:
> > The addition of gl_snaptogrid in the struct _glist may be refused though, because it introduces a potential incompatibility at binary level with existing compiled externals...
> 
> `gl_privatedata` might be a good place

Hello, long time no update - but would now have time to finish this. How about adding this to the _editor struct?

Spacechild1:
Actually, now that I have a closer look, I realize that `gl_snaptogrid` would be part of a bit field. I *think* you can safely add this right after `unsigned int gl_isclone:1;`. Right now we only use 12 from 32 available bits, so adding a new bit wouldn't change the struct layout.

However, the C standard says: `The order of allocation of bit-fields within a unit (high-order to low-order or low-order to high-order) is implementation-defined.`

I have to investigate whether this implies that adding a bitfield (after the last one) could change the layout of existing bitfields. It would be a shame if we couldn't add flags without breaking the ABI...

On the other hand, we have already added flags in Pd in the past (e.g. `gl_isclone`) and didn't run into problems...

porres:
@esaruoho - i like this

virusys:
Nice I would definitely find this useful!

eeropic:
How would one proceed testing the bit fields? 
I.e. the possible problem could occur in running older external binaries on platform X that might then think that a particular bit is for some other property?

To end on a high note / a terrible joke, how bout always using only 16 of the available bits, and make it a palindrome so every bit is set from both LSB and MSB direction? :D Or can they be in arbitrary order, not just reversed in implementation?

Spacechild1:
> I.e. the possible problem could occur in running older external binaries on platform X that might then think that a particular bit is for some other property?

Yes, exactly. I know that some externals like `iemguts` rely on `g_canvas.h`, but I don't know whether any of them directly access the bitfields.

Anyway, the bitfield in `t_glist` is obviously meant to be expanded (and *has* been expanded), so don't worry too much!

src/g_canvas.h
```diff
@@ -189,6 +191,7 @@ struct _glist
     unsigned int gl_loading:1;      /* am now loading from file */
     unsigned int gl_willvis:1;      /* make me visible after loading */
     unsigned int gl_edit:1;         /* edit mode */
+    unsigned int gl_snaptogrid:1;   /* snap to grid mode */
```
Spacechild1, line 194: move this after `unsigned int gl_isclone:1`

esaruoho:
@eeropic any progress with the suggestions sent to this PR?

eeropic:
> @eeropic any progress with the suggestions sent to this PR?

Hey - it is almost done, but! Looking still into whether the whole thing could be implemented as Tcl GUI plugin, instead of touching the core ... but some of the things are quite deep inside the source, have to see.

porres:
I'd love to see this in the core, I don't know why there should be a concern in including it in the core, what do you say?

danomatika:
>Looking still into whether the whole thing could be implemented as Tcl GUI plugin, instead of touching the core ... but some of the things are quite deep inside the source, have to see.

I'd say it's actually not bad for this to stay in the core. In a future libpd with GUI communication, I can imagine being able to much more easily implement a Pd UI in different frameworks. Having an internal snap top grid would allow for it to be available in non-TCL platforms.

sebshader:
I think all gui stuff should hypothetically be in a separate thread/process from the scheduler/audio stuff. The only geometry pd really needs to know about is the order of inlets and outlets.

src/g_canvas.h
eeropic: Done thanks

eeropic:
Okay this is ready from my side I think. But, I will still explore the tcl-plugin possibilities.

bmitc:
Is there any update on this? I would love to use this in Pure Data and would prefer not to use a fork.

porres:
I'm afraid this one got not much attention, but I'd like to have it in and will keep it in my radar for lobbying it.

right now it has conflicts that need to be solved, what do you say @eeropic?

esaruoho:
let's try and lobby it in, right?

eeropic:
Rebased + resolved the conflicts

I tried recreating all of this purely with Tcl/Tk, but couldn't get the behaviour to match 1:1 this one (yet)

Even if this would be replaced later with the work from the incremental GUI-separation process, I think it would have a big positive impact on usability.

I know lot of people who don't mind "messy" patches, but also know people who arrow-nudge objects to achieve "a perfect layout" 😄

danomatika:
I think most of us are for this. It's just a matter of finding the implementation that is most likely to be acceptable & understandable. Don't be discouraged that it hasn't been merged yet.

bmitc:
Sounds great! Thanks for the quick turnaround on the ping! I was just pinging because this seemed pretty far along and will be well-received. Coming from another visual language, I've been sorely missing some alignment help.

porres:
no, I got nothing to do with this :)

esaruoho-hbo:
@porres  i meant that last time u said it should be rebased, now it had been, is it ok now?

esaruoho:
> no, I got nothing to do with this :)

Yep. I know. @porres  but last time u said it needs a rebasing to resolve conflicts, they have been done, what do you reckon is needed now?

porres:
@millerpuckette needs to check it :)

pingarelho:
Firstly, thank you for this PR, I have been using it extensively for the past month.

I noticed that when creating objects/messages with other objects/messages selected - to connect them automatically - the positioning of the new objects will not obey to the grid, and instead will be the stock Pd behavior. Video below for reference.

https://github.com/pure-data/pure-data/assets/42239760/0e5b5b78-b59c-49cb-a933-fab8afe71319

umlaeute:
i wonder what the proper behaviour actually should be.
e.g.  in your video i find it obvious (but othres might have different opinions), that the snap-to-grid alignment of chained object is too narrow (that is: I would like to have snap-to-grid but the grid should allow to auto-place the objects where they are currently put)

pingarelho:
> i wonder what the proper behaviour actually should be. e.g. in your video i find it obvious (but othres might have different opinions), that the snap-to-grid alignment of chained object is too narrow (that is: I would like to have snap-to-grid but the grid should allow to auto-place the objects where they are currently put)

I, too, agree that it is too narrow. Would be nice to have a way of controlling it.

eeropic:
Perhaps instead of 10px increments, a 8px grid would work better.
Here's an example with 8px grid placement on the left, and autopatch placement on the right.

image

The autopatch uses ~5px offset from bottom of the object
https://github.com/pure-data/pure-data/blob/54f42a918bc4c01d33aabfcd078a6105418143f7/src/g_text.c#L174

Duplicating objects uses PASTE_OFFSET defined at https://github.com/pure-data/pure-data/blob/54f42a918bc4c01d33aabfcd078a6105418143f7/src/g_editor.c#L39

So there's several places where these offset values (for different purposes though) are stored,
and my original idea was at least to use the same GRID_UNIT value for keyboard nudge offset and snapping.

One could easily do a preference value for the grid units, like in design software, but personally I don't see point as most of the components have fixed size, thus some kind of optimal grid value for spacing them could exist

umlaeute:
i *strongly* suggest that the grid be made configurable by the user.

esaruoho:
> i _strongly_ suggest that the grid be made configurable by the user (even if there _is_ an optimum spacing value)

for sure, i need the snap to be "one row of pixel space between objects", otherwise i gotta do that tweak by hand

danomatika:
Some possible options:
* use the object sizing font metrics to calculate a default grid spacing
* provide a simple mechanism to set grid size via a command line arg

A question comes to mind: should the grid size be in pixels/points or relative to the default object size (for instance if using font metrics)?

umlaeute:
Use gui preferences, rather than a cmdline option.

On the pd-core side, it's enough to provide a method that allows setting the grid (dx/dy), the rest can be handled on the gui side (including the decision of whether to use pixels, points or furlongs)

Issue #579, opened 2019-04-07
quotes in settings string causing issue with TCL


danomatika:
Singles quotes, as used by loading a font string in the start up settings cause issues with Tcl. There probably needs to be some explicit string escaping going on when processing these commands, either in the GUI or core if the startup settings are sent as a string to the GUI. (I haven't looked yet.)

Relevant copy/paste from Pd-list:

>> I am no expert, but after your first email, I played around a bit. I
>> am on Ubuntu 18.10 running Pd 0.48.2, and I cannot reproduce your
>> problem. I tried:
>> 
>>     pd -font-face "Celeste Small"
>> 
>> Pd started OK, and Celeste Small was loaded as the default font (see
>> screenshot — just a labelled canvas).
> 
> strange ... this is what happens on my system when i do this:
> 
> WARNING: Font family '"Celeste' not found, using default (DejaVu Sans Mono)
> 
> Small": can't open
> -font-size: can't open
> 12: can't open
> -font-weight: can't open
> normal: can't open
> (Tcl) UNHANDLED ERROR: extra characters after close-quote
>     while executing
> "pdtk_startup_dialog .gfxstub5648e27e9080 0 "-font-face "C"
>     ("uplevel" body line 7)
>     invoked from within
> "uplevel #0 $docmds"
> 
> from that point on PD doesn't respond to anything anymore.
> 
> Tested on both PD 0.48 and PD 0.49
> 
> BTW: my system is:
> Ubuntu 19.04 (Disco Dingo) with an XFCE desktop
> 
> (but the same happened on a Ubuntu 18.04.2 LTS)

umlaeute:
moved milestone to corresponding PR #741

PR #741, opened 2019-09-09
allow quotes (and spaces) in PdSettings-flags

umlaeute:
the parser for the flags in the PdSettings is currently very naive: it breaks the `flags` strings on each whitespace, and passes these tokens to `sys_argparse`.

Unfortunately, this gives ugly results, when people try to pass strings-with-spaces using their knowledge from the shell (https://github.com/pure-data/pure-data/issues/579):

E.g. while starting Pd with `pd -font-face "Latin Modern Mono Caps"` works nicely, adding the string `-font-face "Latin Modern Mono Caps"` to the *Startup Flags* via the <kbd>File</kbd>-><kbd>Preferences</kbd>-><kbd>Startup...</kbd> will not work at all:
Starting Pd will result in the following errors, indicating that Pd really splits the `"Latin Modern Mono Caps"` into 4 tokens, the 1st of which will be interpreted as the requested font-face, and the rest as patches to open:
```
WARNING: Font family '"Latin' not found, using default (DejaVu Sans Mono)
Using font: DejaVu Sans Mono bold
Modern: can't open
Mono: can't open
Caps": can't open
```

worse: when trying to fix the problem via the <kbd>File</kbd>-><kbd>Preferences</kbd>-><kbd>Startup...</kbd> menu, it will refuse to open with

~~~
(Tcl) UNHANDLED ERROR: extra characters after close-quote
    while executing
"pdtk_startup_dialog .gfxstub55ba2e3998f0 1 "-font-face "L"
    ("uplevel" body line 3)
    invoked from within
"uplevel #0 $docmds"
~~~

The proposed PR fixes all these issues by:
- using a better parser for the `flags` field of the PdSettings, that handles quotes (single, or double) and escapes similar to bash. so the string `-font-face "Latin Modern Mono Caps"` is really parsed into two tokens.
- by escaping the any quotes and curly braces in both `flags` and `font-face` whenever they are sent to the Pd-GUI (mostly `pdtk_startup_dialog` and `pdtk_pd_startup`)

Closes: https://github.com/pure-data/pure-data/issues/579


umlaeute:
@millerpuckette i think this fixes a rather serious problem with the pdsttings parsers.
it would be cool to have this (or the alternative implementation in #1066) included in the next Pd-release

Spacechild1:
@umlaeute I think your PR is better but it can be simplified with https://github.com/pure-data/pure-data/pull/1070

Spacechild1:
@millerpuckette I think this PR is badly needed.

chainjazz:
Just to add my experience. Flag parsing will accept backslash escaped spaces,. and accept equal signs as fillers between flagname and value.

chainjazz:
After some consideration, why have command line flags in the GUI anyway?

Issue #746, opened 2019-09-18
Change "save as" starting point from "untitled" to last saved name.


jaffasplaffa:
As discussed on the pd developers mailing list:

When you do a "save as", instead of starting from the name, "Untitled", Pd should start from last saved name, for example "CombFilter1" and then you can go on and change the name from there.

See this clip:
https://www.youtube.com/watch?v=lv3a83yzmJc

danomatika:
What OS and PD version are you using? I can confirm on macOS 10.14 and Pd 0.50.

umlaeute:
while `pdtk_canvas_saveas` received the filename, but did'nt use it.

it seems that this is intentional: b6b19c7afbdba37701d38cd4cac98dc31fb116ea

danomatika:
I did a quick test on my system (macOS 10.14) by adding a post to `pdtk_canvas_saves`:

    ::pdwindow::post "$initialfile"

The expected filename is printed and I don't see anything weird happening with saveas for both new and existing files.

umlaeute:
to clarify:
Pd-core sends the filename to Pd-GUI's `pdtk_canvas_saveas` just fine.
It's only that the `pdtk_canvas_saveas` doesn't use it when calling `tk_getSaveFile` (or anything else).

This is because "Tk 8.5 on Macintosh" has hickups when the `-initialfile` flag is used (as per b6b19c7).

I don't know which Tk-version is used for which Macintosh packages.
(I'm pretty sure that the PowerPC package still has such an old version.)

danomatika:
>Pd-core sends the filename to Pd-GUI's pdtk_canvas_saveas just fine.

Yes.

>This is because "Tk 8.5 on Macintosh" has hickups when the -initialfile flag is used

Apparently. The commit in question basically applies the same bug to all other Ones.

>I don't know which Tk-version is used for which Macintosh packages.

8.5.19 as embedded frameworks in the .app bundle

>(I'm pretty sure that the PowerPC package still has such an old version.)

8.4 system frameworks which probably work ok

danomatika:
Here is a 32 bit Pd build with the savers fix and an embedded Tk 8.6.7: http://docs.danomatika.com/pdbuilds/0.51/Pd-0.50-0-saveas-fix-tk867.zip

It seems 8.6 has fixed the underlying issue as the save panel filename is set as expected. yet another reason to move forward on macOS...

umlaeute:
removed milestone (it's in #747)

[...]

PR #747, opened 2019-09-19
"Save as" defaults to the original filename

umlaeute:
as discussed at lengths on the mailinglist...

however, while answering #746, i found the commit that introduced the current behaviour: b6b19c7afbdba37701d38cd4cac98dc31fb116ea

it seems that after all this was intentional. quoting from the commit message:
> "save as" no longer sends an initial file name to the `tk_getSaveFile` function
> 
> this seems to have stopped working in tcl 8.5 on Macintosh (the
> `initialfile` gets incorporated into the `initialdir` and the dialog gets badly
> confused.)  Workaround is simply not to supply it.

this PR has been tested on linux with Tcl/Tk-8.6, but not on any other system.
Given the comment, i would highly recommend testing this PR on various ("all") deployments, before accepting it.


danomatika:
Ok on macOS 10.14 and Pd 0.50-0 w/ embedded Tk 8.5.19.

danomatika:
Actually, I'm wrong. The filename received by `pdtk_canvas_saveas` is ok, but the save panel still show's "Untitled.pd".

umlaeute:
I can confirm the non-functioning of my patch on OSX (10.12 and 10.14), with the default Tcl/Tk that comes with Pd-0.50 (Tcl/Tk-8.5.19).

With this Tcl/Tk version, `tk_getSaveFile` ignores the `initialfile` completely, and i'm not sure i fully understand what `initialdir` dir.

Using the Tcl/Tk that comes with Mojava (`$ echo 'puts $tk_version' | wish` returns *8.6*) it works as expected.
Starting Pd via `wish pd-gui.tcl` makes the `Save As...` dialog work (although other things seem to be broken, like rendering patches...)

So I guess this is just another bug in Tcl/Tk on OSX.

umlaeute:
now that Pd-0.51 ships with Tcl/Tk-8.6, this PR seems to work nicely.

@millerpuckette i think this is ready for inclusion. if you do want to support the broken Tk8.5 on macOS, we *could* special handle this case explicitely. just tell me ;-)

umlaeute:
hand-applied with 27cf05e

PR #753, opened 2019-09-24
fix GOP-Canvas Properties dialog issues when zoom = 2 (and fix another zoom bug)

jyg:
With pd 0.50, Canvas Properties dialog values are inconsistent whether you open it  with zooming = 1 or with zooming = 2 (the margins and offset are halfened with zoom = 2).
The proposed fix restores consistent displaying of the margins and offset values in canvas properties dialog box, when gop enabled, independently of zoom displaying factor.
This fix may also restore consistent behaviour of donecanvasdialog message between pd 0.49 and pd 0.50, with zoom = 2.

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2098,15 +2098,15 @@ void canvas_properties(t_gobj*z, t_glist*unused)
                 0., 0.,
                 glist_isgraph(x) ,
                 x->gl_x1, x->gl_y1, x->gl_x2, x->gl_y2,
-                (int)x->gl_pixwidth/x->gl_zoom, (int)x->gl_pixheight/x->gl_zoom,
-                (int)x->gl_xmargin/x->gl_zoom, (int)x->gl_ymargin/x->gl_zoom);
+                (int)x->gl_pixwidth, (int)x->gl_pixheight,
+                (int)x->gl_xmargin, (int)x->gl_ymargin);
     else sprintf(graphbuf,
             "pdtk_canvas_dialog %%s %g %g %d %g %g %g %g %d %d %d %d\n",
                 glist_dpixtodx(x, 1), -glist_dpixtody(x, 1),
                 0,
                 0., -1., 1., 1.,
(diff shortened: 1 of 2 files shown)


jyg:
commit https://github.com/pure-data/pure-data/pull/753/commits/dc293e35f424de0f0c0e268437d1c3848d8ba460 fixes wrong mouse-displacing of labels with zooming=2 for floatatom & symbolatom objects.

danomatika:
Does this handle opening the dialog with one room level, chaining zoom level, then closing the dialog? That was what I was trying to handle by multiplying then dividing by the zoom level between dialog communication.

umlaeute:
and would you Mond also fixing the zoom-issue with inserting an object into a connection via triggerize, you mentioned on the mailinglist?

danomatika:
@umlaeute This may be the better way to do this as the dialog has no way of knowing wha the zoom level is. This solution returns to only sending the dialog the "virtual size/pos" and applies zoning to the actual draw command.

danomatika:
@umlaeute I will look into this tonight as I was the original author of the zoom changes.

jyg:
I cannot check this right now and look further in triggerize + zoom. maybe in one or two days.

jyg:
@danomatika > Does this handle opening the dialog with one zoom level, changing zoom level, then closing the dialog? 
If I understand your question, yes, my fix handles when you change zooming while canvas properties dialog remains opened

danomatika:
Ok I check this and it looks good.

The canvas dialog issue was clearly a bug as I was applying the zoom (incorrectly as well) to the value when it clearly didn't need it. I think this came form the fact that I needed to do this for all the IEM gui dialogs and I probably just added it quickly here, by mistake.

I missed the mouse issue as I did not do much testing other than mouse over checks.

Note: There is a test patch which displays all of the guis for alignment and size testing: Browser->Pure Data/7.stuff/tools/sizingtest.pd

danomatika:
I have a fix for the triggerize issue coming as well.

danomatika:
The triggerize issue should be fixed now. Please test the `update/0.50` branch.

PR #756, opened 2019-09-24
Enabling IEMGuis autoconnect

jyg:
The autoconnect feature enables new objects and messages  to be  automatically connected to any previously created (and selected) object.
This patch enables other gui's (vsl, hsl, bang, toogle, hradio, etc...) to get this feature.
I also checked the behaviour with zoom = 2.

PS: Sorry for my misuse of git :  g_text.c only is concerned by this pull request, I don't know how to mask po/template.pot that is not related to this change

EDIT(@umlaeute):
- Closes:https://github.com/pure-data/pure-data/issues/826

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -234,6 +234,15 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
     // enable autoconnect
     int connectme, indx, nobj;
     canvas_howputnew(gl, &connectme, &xpix, &ypix, &indx, &nobj);
+    // align vsl and hsl like others objects
+    if (guiobjname== gensym("hsl"))
+    {
+        xpix+=3;
+    }
+    else if (guiobjname== gensym("vsl"))
+    {
+        ypix+=3;
+    }
 
     pd_vmess(&gl->gl_pd, gensym("editmode"), "i", 1);
 
@@ -247,7 +256,7 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
     }
     else 
     {
-	canvas_startmotion(glist_getcanvas(gl));
+        canvas_startmotion(glist_getcanvas(gl));
     }
     if (!canvas_undo_get(glist_getcanvas(gl))->u_doing)
 	canvas_undo_add(glist_getcanvas(gl), UNDO_CREATE, "create",
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -231,15 +231,32 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
     t_binbuf *b = binbuf_new();
     int xpix, ypix;
 
+    // enable autoconnect
+    int connectme, indx, nobj;
+    canvas_howputnew(gl, &connectme, &xpix, &ypix, &indx, &nobj);
+
+
     pd_vmess(&gl->gl_pd, gensym("editmode"), "i", 1);
-    glist_noselect(gl);
+ //   glist_noselect(gl);
+
     SETSYMBOL(&at, guiobjname);
     binbuf_restore(b, 1, &at);
-    glist_getnextxy(gl, &xpix, &ypix);
-    canvas_objtext(gl, xpix/gl->gl_zoom, ypix/gl->gl_zoom, 0, 1, b);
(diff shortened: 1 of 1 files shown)


umlaeute:
this is not a misuse of github, but of git.

just don't commit the changes to unrelated files: rather than `git commit .` do `git commit src/g_text.c` (or whatever you did to stage your commits).

You probably should fix your commit (see https://stackoverflow.com/a/42450337/1169096) and force-push to get the changes into the PR.

jyg:
Ok. Done. Thanks for the advice.

src/g_text.c
```diff
@@ -231,15 +231,32 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
...
     int xpix, ypix;
 
+    // enable autoconnect
+    int connectme, indx, nobj;
+    canvas_howputnew(gl, &connectme, &xpix, &ypix, &indx, &nobj);
+
+
     pd_vmess(&gl->gl_pd, gensym("editmode"), "i", 1);
-    glist_noselect(gl);
+ //   glist_noselect(gl);
```
umlaeute: why?

src/g_text.c
```diff
@@ -231,15 +231,32 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
...
     SETSYMBOL(&at, guiobjname);
     binbuf_restore(b, 1, &at);
-    glist_getnextxy(gl, &xpix, &ypix);
-    canvas_objtext(gl, xpix/gl->gl_zoom, ypix/gl->gl_zoom, 0, 1, b);
-    canvas_startmotion(glist_getcanvas(gl));
-    canvas_undo_add(glist_getcanvas(gl), UNDO_CREATE, "create",
-        (void *)canvas_undo_set_create(glist_getcanvas(gl)));
+    if (connectme)
+    {
+	canvas_objtext(gl,  xpix, ypix, 0, 1, b);
```
umlaeute: minor nitpick, but you might want to kep your indentation consistent with the one that Pd uses)

src/g_text.c
```diff
@@ -231,15 +231,32 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
...
+        canvas_connect(gl, indx, 0, nobj, 0);
+    }
+    else 
+    {
+	glist_getnextxy(gl, &xpix, &ypix);
```
umlaeute: that's superfluous. `canvas_howputnew()` already calls `glist_getnextxy()` (and dezooms the coordinates) if it detects that autoconnect is not desired.

so the code could be much simplified, see e.g. the implementation in `canvas_msg()`: https://github.com/pure-data/pure-data/blob/master/src/g_text.c#L490-L504

src/g_text.c
```diff
@@ -231,15 +231,32 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
...
+        // glist_getnextxy gives mouse pixel coords, that need to be "dezoomed"
+	canvas_objtext(gl,  xpix/gl->gl_zoom, ypix/gl->gl_zoom, 0, 1, b);
+	canvas_startmotion(glist_getcanvas(gl));
+    }
+    if (!canvas_undo_get(glist_getcanvas(gl))->u_doing)
```
umlaeute: :+1:

src/g_text.c
```diff
@@ -237,7 +237,6 @@ void canvas_iemguis(t_glist *gl, t_symbol *guiobjname)
 
 
     pd_vmess(&gl->gl_pd, gensym("editmode"), "i", 1);
- //   glist_noselect(gl);
```
umlaeute: oh, i asked "why?" because i didn't understand why you disabled the call to `glist_noselect()`, rather than why you didn't just remove the entire line. I agree that it should be removed rather than just commented out; but the question remains: why it you don't want to call `glist_noselect()`

jyg: Because I couldn't figure the effect of this call,  because it is not called in `canvas_obj()`, and overall because `glist_noselect()` is already called inside `canvas_howputnew()` ;-)

umlaeute:

Issue #751, opened 2019-09-22
multiple "save as" windows allowed


jmmmp:
Try the following (tested on windows)
- open/create a patch window
- press save as
- use alt-tab to focus on any other program
- go back to the patch window
- press save as 
- now you have 2 open save as windows
- keep doing it, until you have as many "save as" windows as you might find necessary...

umlaeute:
do you want to provide a PR to document this?

jmmmp:
what is a PR?

Am So., 22. Sept. 2019 um 18:09 Uhr schrieb umläute <
notifications@github.com>:

Spacechild1:
PR = pull request

> do you want to provide a PR to document this?

@umlaeute why document and not fix?

IMO, the open/save dialogs should simply be modal, like in any other application...

BTW, a similar problem occurs with `[openpanel] `and `[savepanel]`, but they probably should stay non-modal. Instead we can fix it in the core (don't create the dialog if the object already has one).

umlaeute:
> why document and not fix?

because @jmmmp described a behaviour, which he (most likely) finds odd.

there's nothing inherently and/or obviously *wrong* with how the "Save As..." currently behaves. I don't know whether the OP meant this as a feature request, a bug report or as a documentation of the current behaviour.

as for @Spacechild1's proposed solution: i find modal dialogs rather cumbersome in multi-window applications like Pd: it's just too easy to hide the modal behind one of the many open windows, which leaves you with a non-responsive Pd (that you can't even quit). this is think is much more problematic than having 23 open "Save As..." windows for a single patch.

Spacechild1:
> there's nothing inherently and/or obviously wrong with how the "Save As..." currently behaves.

@umlaeute  I'm pretty sure that being able to create more than 1 open/save dialog for the *same* document is a bug.

>  i find modal dialogs rather cumbersome in multi-window applications like Pd: it's just too easy to hide the modal behind one of the many open windows, which leaves you with a non-responsive Pd

First, a modal dialog should *always* be in front. The problem you're describing (dialogs hidden by other windows) really applies to the current implementation and modal dialogs would actually solve that! Also, the dialogs should probably be window-modal (and not application-modal), so it wouldn't block the user from interacting with the rest of app.

umlaeute:
> > there's nothing inherently and/or obviously wrong with how the "Save As..." currently behaves.
> 
[...]

PR #764, opened 2019-10-03
make openpanel/savepanel + "save as" dialogs window modal

Spacechild1:
On Windows, the Tk dialogs are non-modal which means that they can get hidden by the parent window and the user can open the same dialog more than once. This PR makes those dialogs window-modal, so they always stay on top of the parent. They only block interaction with the parent window, other windows are not affected. As a long-time Windows user, this is the behavior I would expect of such dialogs. Fixes https://github.com/pure-data/pure-data/issues/751

I've tested on OSX 10.11 and Debian 9 + GNOME and there's practically no difference in behavior because dialogs already seem to be *application-modal* by default. There's one subtle difference on OSX: dialogs show up in front of the parent window, which I find rather nice, but I'm not a regular OSX user (@danomatika probably has something to say). 

I'm not sure about the implications for Linux + other desktop environments... (@umlaeute: what is the current behavior on your desktop?) If this change is too risky, I could set the parent window only on Windows (and maybe also OSX).

### DIFF
--- a/tcl/wheredoesthisgo.tcl
+++ b/tcl/wheredoesthisgo.tcl
@@ -36,7 +36,7 @@ proc open_file {filename} {
 # ------------------------------------------------------------------------------
 # procs for panels (openpanel, savepanel)
 
-proc pdtk_openpanel {target localdir {mode 0}} {
+proc pdtk_openpanel {target localdir {mode 0} {parent .pdwindow}} {
     if { $::pd::private::lastopendir == "" } {
         if { ! [file isdirectory $::fileopendir]} {
             set ::fileopendir $::env(HOME)
@@ -49,9 +49,12 @@ proc pdtk_openpanel {target localdir {mode 0}} {
 
     # 0: file, 1: directory, 2: multiple files
     switch $mode {
-        0 { set result [tk_getOpenFile -initialdir $localdir] }
-        1 { set result [tk_chooseDirectory -initialdir $localdir] }
-        2 { set result [tk_getOpenFile -multiple 1 -initialdir $localdir] }
+        0 { set result [tk_getOpenFile -initialdir $localdir \
+            -parent $parent] }
+        1 { set result [tk_chooseDirectory -initialdir $localdir \
+            -parent $parent] }
+        2 { set result [tk_getOpenFile -multiple 1 -initialdir $localdir \
+            -parent $parent] }
         default { ::pdwindow::error "bad value for 'mode' argument" }
     }
     if {$result ne ""} {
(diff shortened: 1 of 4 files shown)


danomatika:
Sounds good to me. The open and save panels are already modal on macOS and *should* open on top anyway.

Spacechild1:
> and should open on top anyway.

Ah sorry, I was rather speaking about the position, not the stacking order. With this patch, the dialog appears in front of the parent window - "growing" out of the window's top border -, at least that's the case on OSX 10.11. Before this patch, the dialog would just appear somewhere on the screen. I didn't intend this behavior, it's just a byproduct of setting the `-parent` option on macOS, but if it's acceptable (or even desirable), I would just keep it.

danomatika:
That's just opening the window in a sheet attached to the parent's titlebar which is standard practice for Apple HIG: https://developer.apple.com/design/human-interface-guidelines/macos/windows-and-views/sheets/

Spacechild1:
Cool, thanks for the link!

danomatika:
As you mentioned, since the open and save panels are modal already, this is really an improvement on macOS as it associates the panel with the document window it affects. If the parent doesn't exist, opening a sheet defaults to its own window anyway. The only issue I could think of is when triggering multiple [openpanel] objects at the same time, but they should already be processed in order, one at a time anyway.

danomatika:
Also, the close confirmation dialogs already work this way.

tcl/wheredoesthisgo.tcl
```diff
@@ -28,28 +28,28 @@ proc open_file {filename} {
 # ------------------------------------------------------------------------------
 # procs for panels (openpanel, savepanel)
 
-proc pdtk_openpanel {target localdir} {
+proc pdtk_openpanel {target localdir parent} {
```
umlaeute: probably the `parent` argument should be made optional:

~~~tcl
proc pdtk_openpanel {target localdir {parent .pdwindow}} {
~~~

tcl/wheredoesthisgo.tcl
```diff
@@ -28,28 +28,28 @@ proc open_file {filename} {
...
-    set filename [tk_getOpenFile -initialdir $localdir]
+    set filename [tk_getOpenFile -initialdir $localdir -parent $parent]
     if {$filename ne ""} {
         set ::fileopendir [file dirname $filename]
         pdsend "$target callback [enquote_path $filename]"
     }
 }
 
-proc pdtk_savepanel {target localdir} {
+proc pdtk_savepanel {target localdir parent} {
```
umlaeute: same here

PR #765, opened 2019-10-05
Feature/netobject-updates: documentation fixes

Spacechild1:
* clarify IP version detection (hostname resolution favors IPv4 results)
* `[netreceive]` hostname argument can't be a remote host
* on some systems (e.g. Windows), dual-stack sockets only work for the `any` address, i.e. `[netreceive 9999 ::1]` won't receive IPv4 messages.

ping @danomatika


umlaeute:
merged into `feature/netobject-updates`

Issue #675, opened 2019-07-05
`qsort_r()` is not portable


claudeha:
The new text sorting in c399093d adaaf9a3 is not portable to Emscripten, which currently has neither `qsort_r()` nor `qsort_s()`.  See <https://github.com/emscripten-core/emscripten/pull/5424#issuecomment-327145679>.  It is more likely that Emscripten would in the future have a C11 compatible `qsort_s()` than anything else, afaict, but it has been open a couple of years.

claudeha:
https://github.com/claudeha/pure-data/commit/e8023866ef8be3475cc101cfb1656f5249030875 is an horrible hack that got it to compile here, I haven't tested it for correctness yet (needs a nogui test patch).

PR #771, opened 2019-10-08
text sort: apply MSVC hack to Emscripten and Android

Spacechild1:
use thread-local storage to make `stupid_sortcompare` thread-safe, so we can use it in multi-instance libpd builds. also fixes https://github.com/pure-data/pure-data/issues/675

### DIFF
--- a/src/x_text.c
+++ b/src/x_text.c
@@ -470,17 +470,18 @@ equal:
     else return (1);
 }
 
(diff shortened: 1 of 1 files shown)


Spacechild1:
@umlaeute is this innocuous enough to push on develop?

umlaeute:
i guess so ;-)

Spacechild1:
@millerpuckette do you want to merge this? I just resolved the conflicts.

PR #776, opened 2019-10-12
Search for hex-munged filenames when hexmunge is true

myQwil:
While the setup methods for special-character objects are hex-munged, the raw object name still gets passed to the method sys_trytoopenone, which means that it searches for a filename with the special characters still intact. Here, the hex-munged name is passed instead, with the assumption that the filename is also going to be hex-munged.

A couple of extra conditions for special cases have been added. One allows for a single tilde by itself to be seen as an object name rather than part of a path name, and another allows for a forward slash to be seen as part of an object name if it's the last character in the name.

### DIFF
--- a/src/s_loader.c
+++ b/src/s_loader.c
@@ -122,7 +122,7 @@ static int sys_do_load_lib(t_canvas *canvas, const char *objectname,
     const char *classname, *cnameptr;
     void *dlobj;
     t_xxx makeout = NULL;
-    int i, hexmunge = 0, fd;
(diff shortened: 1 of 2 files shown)


umlaeute:
@myQwil any specific reason to close this PR?
if so, please mention it on close.

myQwil:
Yes, sorry, I wasn't happy with the structure of my git fork and wanted to start again from scratch. I might repost this pull request at some point after getting everything sorted out.

PR #777, opened 2019-10-27
fix 'multiply defined' error in help of [array]

reduzent:
by renaming arrays in the [pd quantile+random] and [pd min+max] subpatches.


umlaeute:
i've done a slightly less intrusive fix in 87c058e

Issue #790, opened 2019-11-14
Number box with a zero won't update when given a zero of opposite sign.


danfuzz:
If you have a number box which shows a positive or negative zero, and you then send a zero of the opposite sign to it, the value stored by and transmitted through the box is properly updated, but the display is not. Here's a quick demo:

![number-box-zero-display](https://user-images.githubusercontent.com/1090235/68825954-df327e00-0650-11ea-8a92-fc18b215cc0f.png)

Note that when you hit the bang for the positive and negative zeros, the value gets propagated correctly, but neither of the intermediate number boxes ever changes its display. And when you hit the bang to resend the value of the top number box, the correct stored value is sent.

This was observed on Pd 0.50.0.

Spacechild1:
as optimization, graphics are only updated if the new value is different from the old one, and as you've mentioned in another issue, positive and negative zero compare equal. we could add additional logic, but I'm not sure it's worth it...

BTW, since you're an experienced software engineer, many of your feature requests could be pull requests ;-)

danfuzz:
>many of your feature requests could be pull requests ;-)

I'm certainly up for submitting PRs, but it hasn't been clear to me whether the project would actually be receptive. (Same goes with the other issue you commented on similarly, though I think that one was perhaps using "PRs gladly accepted" more in the euphemistic sense :-p .)

Spacechild1:
>  whether the project would actually be receptive.

it certainly is! if you make a good PR, the chances for the feature to get included is much higher than just making a feature request and wait for someone to implement it ;-) I'm just telling you this because I know you're an experienced software developer

Spacechild1:
Pd needs more developers, so you're highly welcome to dive into the source code! if you have any questions, you can always ask me.

danfuzz:
BTW I just tested the new-style number box, and in addition to not changing its display when presented with a zero of opposite sign, it also fails to store or propagate the new value (that is, not just a display bug). E.g., if it stores a positive zero and you send a negative zero to it, it will re-output a positive zero.

porres:
This has been fixed, we can close this!!!

umlaeute:
i think so

PR #791, opened 2019-11-20
Fix how number boxes perform sameness comparison (fix bug due to surprising IEEE754 outcome).

danfuzz:
This PR fixes issue #790 ("Number box with a zero won't update when given a zero of opposite sign."). I made the equivalent fix for both old- and new-style number boxes, which were both buggy though in somewhat different ways.

What I specifically did was make the "equality" comparison (of old and new values) be based on value-sameness instead of the IEEE754 definition of numeric equality. The most notable reason for doing this is because IEEE754 (a) defines both positive and negative zeros, and (b) considers them to be "equal." Since number boxes (as currently written) _don't_ completely ignore this distinction (e.g. you can type `-0` into a box and it behaves as expected, etc.), it was bad news to ignore the distinction in the value sameness comparison. The other consistent option would of course be to more completely ignore the `+0/-0` distinction in number boxes, but this struck me as inconsistent with the rest of Pd, which seems to (implicitly) fully embrace IEEE754 math (or at least reifies whatever math rules are implemented by the C compiler that compiled it, which by and large will be IEEE754 compliant these days).

In addition to fixing the aforementioned bug, this also improves sameness checking of NaNs (which IEEE754 _also_ treats surprisingly, with respect to the `==` operator). Interestingly enough, the new-style number box already had an apparent attempt to do something special with NaNs, but as it turns out what was written was effectively a no-op, so I removed the code in question.

As part of this PR I wrote a new utility function, which is called `float_strict_equal`. I have questions: (1) Is this a good enough name? (2) I couldn't figure out a good file to put it in, so I made a new file for it. Do (the collective) you have a suggestion for a better place? (3) If it stays in its own file, I presume you'd prefer me to assign copyright to Puckette, as his name seems to appear on the rest of the files. Do you have a form (CLA or what-have-you) for administration of project contribution / copyright?

Final questions:

* I tested this to my satisfaction using the attached patch, but I couldn't spot any unit tests (or similar) to update. Did I miss them? If so, please let me know and I'll edit / augment as necessary.

* I updated all the makefiles, and the auto-builds seem to be passing, but I have no practical way of testing anything but the macOS build locally. Is there something I should do to gain assurance of the build's soundness on other platforms?

Here's the patch I used for testing:
[numbox-test.pd.txt](https://github.com/pure-data/pure-data/files/3867335/numbox-test.pd.txt)

numbox-test

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -584,9 +584,7 @@ static void gatom_set(t_gatom *x, t_symbol *s, int argc, t_atom *argv)
     if (x->a_atom.a_type == A_FLOAT)
     {
         x->a_atom.a_w.w_float = atom_getfloat(argv);
-        changed = ((x->a_atom.a_w.w_float != oldatom.a_w.w_float));
-        if (isnan(x->a_atom.a_w.w_float) != isnan(oldatom.a_w.w_float))
-            changed = 1;
+        changed = !float_strict_equal(x->a_atom.a_w.w_float, oldatom.a_w.w_float);
     }
     else if (x->a_atom.a_type == A_SYMBOL)
         x->a_atom.a_w.w_symbol = atom_getsymbol(argv),
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -9,6 +9,7 @@
 #include <stdio.h>
 #include <ctype.h>
 #include "m_pd.h"
+#include "m_imp.h"
 #include "g_canvas.h"
 
 #include "g_all_guis.h"
@@ -588,7 +589,7 @@ static int my_numbox_newclick(t_gobj *z, struct _glist *glist,
 
 static void my_numbox_set(t_my_numbox *x, t_floatarg f)
 {
-    if(x->x_val != f)
+    if(!float_strict_equal(x->x_val, f))
     {
         x->x_val = f;
         my_numbox_clip(x);
--- a/src/m_imp.h
+++ b/src/m_imp.h
@@ -90,6 +90,9 @@ EXTERN int obj_sigoutletindex(const t_object *x, int m);
 void pd_globallock(void);
 void pd_globalunlock(void);
 
+/* x_misc.c */
+EXTERN int float_strict_equal(t_float f1, t_float f2);
+
 /* misc */
 #define SYMTABHASHSIZE 1024
 
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -852,7 +852,18 @@ static void fudiformat_setup(void) {
   class_addanything(fudiformat_class, fudiformat_any);
 }
 
-
+/* This compares two floating point values for "strict equality," in the usual
+sense, that is sameness-of-value rather than mathematical-equality. That is,
+unlike `==` which obeys (by and large) IEEE754 rules for equality and hence
+sometimes leads to surprising results (such as the nonreflexivity of NaN, and
+negative and positive zeros being considered equal to each other), this function
+operates on the bitwise representation of its arguments. */
+int float_strict_equal(t_float f1, t_float f2) {
+    /* Unfortunately, the most reliable way to perform the requisite comparison
+    in C is to use `memcmp()`. Fortunately, many compilers understand this
+    particular idiom, so it's typically reasonably performant (just ugly). */
+    return memcmp(&f1, &f2, sizeof (t_float)) == 0;
+}
 
 void x_misc_setup(void)
 {


src/g_text.c
```diff
@@ -584,9 +584,7 @@ static void gatom_set(t_gatom *x, t_symbol *s, int argc, t_atom *argv)
     if (x->a_atom.a_type == A_FLOAT)
     {
         x->a_atom.a_w.w_float = atom_getfloat(argv);
-        changed = ((x->a_atom.a_w.w_float != oldatom.a_w.w_float));
-        if (isnan(x->a_atom.a_w.w_float) != isnan(oldatom.a_w.w_float))
```
danfuzz: This clause is a no-op, because if either value were a NaN, then the `!=` check on the previous line would already result in `true`.

src/m_misc.c
```diff
@@ -0,0 +1,20 @@
...
+unlike `==` which obeys (by and large) IEEE754 rules for equality and hence
+sometimes leads to surprising results (such as the nonreflexivity of NaN, and
+negative and positive zeros being considered equal to each other), this function
+operates on the bitwise representation of its arguments. */
+int float_strict_equal(t_float f1, t_float f2) {
+    /* Unfortunately, the most reliable way to perform the requisite comparison
+    in C is to use `memcmp()`. Fortunately, many compilers understand this
+    particular idiom, so it's typically reasonably performant (just ugly). */
+    return memcmp(&f1, &f2, sizeof (t_float)) == 0;
+}
```
Spacechild1: This function will yield true if both floats happen to be NaNs with the same bit representation - which is against the IEEE rules. I think you would have to add an explicit NaN check here.

Spacechild1:
> Do you have a form (CLA or what-have-you) for administration of project contribution / copyright?

No. Pd is BSD-licensed anyway.

> I tested this to my satisfaction using the attached patch, but I couldn't spot any unit tests

there are no unit tests :-)

> Is there something I should do to gain assurance of the build's soundness on other platforms?

We have CI builds for Windows, macOS and Linux. The green tick next to your last commit means that the builds succeeded.

Spacechild1:
> but this struck me as inconsistent with the rest of Pd, which seems to (implicitly) fully embrace IEEE754 math (or at least reifies whatever math rules are implemented by the C compiler that compiled it, which by and large will be IEEE754 compliant these days).

Note that Pd is compiled with `-ffast-math` which turns on "unsafe" math optimizations. This means that the compiler is allowed to violate certain IEEE rules to generate faster code. So Pd generally favors speed over correctness. Just something to keep in mind.

src/m_misc.c
danfuzz: Well, that's the thing: In this case I would presume you want two NaNs to be considered the same, so you don't superfluously update the UI.

danfuzz:
> your commits contain many formatting changes which are not relevant to this PR. ammend your commits and do a force push.

Very sorry! I failed to notice.

src/m_misc.c
danfuzz: That is, explicitly wanting to _not_ obey IEEE754 math is why (something like) this function is appropriate to use in this context.

danfuzz:
Sorry again. I just force-pushed a cleaner diff on the makefiles. Hopefully the auto-builds will succeed in a minute or two.

And, I did my best to figure out the best `p_`refix to use but am not surprised I guessed wrong. Let me know if you do want me to rehome the function and I'll do another turn of the crank.

Spacechild1:
I just want to restate my opinion that we shouldn't add any special treatment of negative zeroes. After all, the IEEE 754 standard says that negative and positive zeroes should compare equal. I still don't see why "preserving" negative zeros is necessary and I think that any code which relies on the signedness of zeroes is flawed. But this is just my personal opinion. Eventually, it's up to Miller to decide.

danfuzz:
Thanks for the approval.

FWIW I think the camel's nose is already under the tent here: Pd already exposes negative zeros, including accepting them as syntax (and as things different than positive zeros) as well as showing them in the UI. What I'd say this PR is doing is making Pd _more consistent with itself_ with regards to treatment of signed zeros, as opposed to "add(ing) any special treatment."

Going the other direction would be reasonable too (though against the grain of most modern systems that I'm aware of), e.g. making the text `-0` parse to a (positive) zero, not ever displaying `-0` in a number box, not ever transmitting a `-0` through a connection, etc. That is, ultimately, I think the high-order bug is _inconsistency of treatment_.

Spacechild1:
> Pd already exposes negative zeros, including accepting them as syntax (and as things different than positive zeros)

You can *create* negative zeroes, but it doesn't mean that Pd treats them as different things when it comes to computations! So I don't see how your PR would lead to more consistency.

> The other consistent option would of course be to more completely ignore the +0/-0 distinction in number boxes, but this struck me as inconsistent with the rest of Pd, which seems to (implicitly) fully embrace IEEE754 math

Well, the standard says +0 and -0 should compare equal. You want to be more IEEE 754 compliant by violating the IEEE 754 standard? I think the issue you have is rather cosmetic (the GUI doesn't update), but technically the current behaviour is 100% fine.

The *real* problem is that your patch produces negative zeroes :-)

danfuzz:
Ok, I think we're talking past each other, but I'll try one last time:

* The two number boxes behave differently than each other with respect receipt / retransmission of zeros. This in and of itself is surprising.

* IEEE754 specifies that when comparing for equality `-0 == +0` which is different than saying that the two values are indistinguishable. In fact, IEEE754 defines `copysign()` and `signbit()`, which both have different results given positive vs. negative zero. And you get a different-signed infinity when dividing `1/+0` vs. `1/-0`.

* Pd already _mostly_ treats the two zeros as the different values that they are. What's weird with the number boxes is that they implicitly distinguish the two values in _some_ but not _all_ cases. This inconsistency is what I'm fixing in this PR, and an essential part of that fix is using something other than IEEE754 numeric equality to compare the values.

* Whether or not you think I _should have_ had a negative zero in my code, the fact is that I _did_. And I didn't expect it! And it was when I was trying to debug the problem that I got sidetracked because the number boxes I was using gave me confusing signals. I fixed the problem in my code, but it was _despite_ wacky surprising behavior from number boxes. I am here with this PR in hand to help save the next poor sap who gets an unexpected negative zero and can't figure out what's going wrong.

src/g_text.c
Spacechild1: BTW, this is not a no-op when compiling with -ffast-math because the compiler might generate assembly which can give wrong results for comparisons involving NaNs (I've been hit by this personally: https://stackoverflow.com/questions/57894032/floating-point-comparison-with-quiet-nan-on-gcc-with-ffast-math-always-yields)

But with the `memcmp` it becomes obsolete, of course.

Spacechild1:
I'm not totally against your PR, as it might make sense *visually* (like the special NaN handling). I just don't think that "preserving" negative zeroes is a desirable thing in itself. If we keep it to those two GUI objects, I guess that's fine.

danfuzz:
Thanks, and gotcha. I'll move the function and repush.

Spacechild1:
> I am here with this PR in hand to help save the next poor sap who gets an unexpected negative zero and can't figure out what's going wrong.

That's fair!

src/g_text.c
danfuzz: TIL! Thanks for the info.

danfuzz:
Ok, all done, tested locally, and repushed. Hopefully the auto-builds will go green in a couple of minutes. Thanks again for working with me on this.

Spacechild1:
> Pd already mostly treats the two zeros as the different values that they are.

Except for every a `==` or `!=` in the Pd code and numerous externals ;-)

umlaeute:
hand-applied with 137d6d73a2f6ad3aad69e6078a2d76057063f7e9

danfuzz:
Thanks for the merge!

PR #797, opened 2019-11-30
read and write to mac plist on directly

danomatika:
This PR introduces direct plist preference reading and writing on macOS. This greatly increases performance as compared to the previous method of calling the `defaults` command through `system()`.

In (informal) testing, read and write times go from around 3-4 seconds to barely imperceptible.

This uses the CoreFoundation C framework and should build back to OS X 10.6 without issue. I did some reorganization by splitting out the various platform-specific implementations into separate sections.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -194,7 +194,7 @@ endif
 
 # need Carbon framework for PA on Mac
 if MACOSX
-LIBS += -framework Carbon
+LIBS += -framework Carbon -framework CoreFoundation
 endif
 
 endif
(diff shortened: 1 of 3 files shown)


millerpuckette:
no review needed on this one

danomatika:
Mmm I just noticed something. I will probably push a couple more commits to the branch.

millerpuckette:
dojh... maybe put them in the 0.51 running updates instead?

danomatika:
Doing it now. The values need to be escaped, otherwise garbage is written to the file.

millerpuckette:
OK let me know if you put it here or in the running updates

danomatika:
Worst case: you can disable the plist by returning early in `sys_initloadpreferences()` and `sys_initsavepreferences()`.

millerpuckette:
all good.  It's going to be at least a month before the big freeze.

danomatika:
Got it. The keys were getting corrupted as I wasn't making a copy of them for the dictionary, only using a reference, so the memory could get corrupted. That didn't show up in testing since the strings were static at compile-time.

Issue #806, opened 2019-12-03
double-precision: PortAudio callbacks make only noise


umlaeute:
when compiling Pd for double-precision (`PD_FLOATSIZE=64`), we cannot use *PortAudio* in **callback** mode, as it only makes noise.

the non-callback-mode works fine though.

Issue #805, opened 2019-12-03
double-precision: [vcf~] broken


umlaeute:
[vcf~] is severly broken when compiling Pd for double-precision

Issue #804, opened 2019-12-03
double-precision: writes soundfiles broken


umlaeute:
after compiling Pd in double-precision mode (`CPPFLAGS="-DPD_FLOATSIZE=64"`), attempting to write a soundfiles with `[writesf~]` in 16bit mode, produces a silent soundfile (all samples are set to 32767 (aka 0dBFS).

reading soundfiles with `[readsf~]` works without problems

umlaeute:
also, writing soundfiles with `[soundfiler]` produces unwanted results:
writing one second of 440Hz in *4-byte mode* (`float`) produces white noise, for all formats (`.wav`, `.aiff`, `.snd` (*big* and *little* endian))

PR #807, opened 2019-12-03
more double-precision fixes

umlaeute:
this is yet another round of bringing @katjav's double-precision exoperiment to mainline Pd.

this was triggered by Purr-Data spending a full "Google Summer of Code" on bringing katja's patches to pd-l2ork...

# included stuff

### fixing the broken stuff:
   - reading/writing soundfiles in float-mode was broken (#804)
   - `[vcf~]` was broken for `t_sample!=float` (#805)
   - PortAudio's callback mode was broken for `t_sample!=float` (#806)

### safety measures
   - protect `[wrap~]` against integer overflow
   - check that the **q** for `[vcf~]` is in a sane range

### having fun with doubles
   - use double-precision variants of math-functions when compiled with `PD_FLOATSIZE=64` for added precision
   - more digits for constants
   - use`t_float` for intermediate variables, in order to not lose precision during calculations

### other goodies
   - print the floatsize on startup (so the user knows it)
   - use the `[cos~]` creation argument (as input value if there's no signal connected)

### cleanups
   - another round of fixing the types: `t_sample`, `t_float`,...
   - declare `sigvcf_setup` as *static* like the rest of the setup_functions in *d_osc.c*

# things left out
this patch-set does not include the following changes from the original proposal:

- replacing höldrich's phase-wrapping algorithm with a more generic one
   - my tests show a slight performance degradation (in terms of CPU) and many more audio glitches (my benchmark test would eat ~45% with the original implementation and ~55% with the generic implementation; probably it was just using more than 50% of available CPU that triggered the audio glitches - i didn't investigate further); i wouldn't mind the performance degredation so much, but the increase of audio glitches was disturbing.

- changing the printout-format for numbers to contain more digits.
  - this is really just eye-candy, and it slightly annoyed me to see `439.99999243520523` instead of `440` (even though the former was certainly more precise)


umlaeute:
i've tested this on linux only.

it would be nice if some windows (@Spacechild1, @Lucarda ) and macOS (@danomatika ) could test the changes as well (esp. the PortAudio stuff)

to do so, simply recompile Pd with `-DPD_FLOATSIZE=64`, e.g.:

~~~sh
./configure --enable-jack CPPFLAGS=-DPD_FLOATSIZE=64
~~~

danomatika:
i’ll check it out. Should this also be a configure option ala —enable-double?

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
i thought about adding such a configure option, but didn't do it because:
- adding `CPPFLAGS=-DPD_FLOATSIZE=64` to the configure-flags is not much more complicated than adding `--enable-double`
- i would expect the latter to have "a more lasting effect", so all externals compiled against a Pd that was configured with this flag to automatically build for double-precision.
  however, this means that we have to change `m_pd.h` based on this flag, which i didn't want to do just now.
  otoh, using `CPPFLAGS` gives the configure-option an air of volatility...

Spacechild1:
I'll give it a try, but note that the callback scheduler is broken anyways on Windows (didn't check on macos). I want to fix it, but I guess I better make a seperate PR?

umlaeute:
of course i 'm happy if you can test other backends as well.. 

portaudio with callbacks works for me when usind Debians portaudio...

Lucarda:
I'm only doing the "testtone.pd"(**Pd-64-64** on Windows) and get _nice_ sound using ASIO but with MMIO i get a loud and distorted tone.

Yeah!

umlaeute:
thanks.
even though you already found an issue, i would really welcome tests that go beyond `testtone.pd`

umlaeute:
@Lucarda hopefully the MMIO problem has been fixed. could you try again?

Lucarda:
> hopefully the MMIO problem has been fixed. could you try again?

MMIO working nice.

katjav:
On this page (at the bottom) is an old set of pd-double 'test' patches which may still be useful:

http://www.katjaas.nl/doubleprecision/doubleprecision.html

umlaeute:
thanks @katjav, i checked your test patches, and afaict everything works.

only `[phasor~]` is still drifting like it used to - which is to be expected as we are still using the old fast-but-inaccurate algorithm....

umlaeute:
however, i think these test-patch are synthetic; i'd like people to do some real tests, preferably using some larger project and see whether everything works.

the project should be Pd-vanilla though (unless you have your externals ready for double-precision)

it would also be nice if people could test the single-precision variant as well (to make sure we don't introduce regressions).

danomatika:
I've not bought about what needs to change for libpd either...

umlaeute:
@danomatika  well, you can always just continue with single-precision.

danomatika:
I think my question was partially philosophical. Concretely, I think it really only needs changes to the sample process functions and possibly the ring buffer messaging.

If you want to integrate libpd into this repo, I would think we would want to have it support double precision as well. I'm working on it now for a new stable release, then we can make deeper changes like this.

umlaeute:
with *libpd* i really mean the "low-level C API".
the various wrapper do not concern me at the moment.

## what works

afaict, libpd (as imported into PR #519)  will not have any problems with double-precision as is.
the `libpd_process_double` will already allow you to have full benefits of a double-precision build.
so no changes required for sample-processing!

however, the message-subsystem as exposed via libpd only offers a single-precision interface.
so you might want to add double-precision variants of the following functions, so that libpd can make *full* use of a double-precision Pd:

- `libpd_read_array`, `libpd_write_array`
- `libpd_float`
- `libpd_set_float`, `libpd_get_float`

you also might want to implement a `doublehook` (that probably should precedence over the `floathook`).

danomatika:
I'm more of the opinion to change the given type to t_float as opposed to add companion functions. Anyway, I'll step out of this discussion as this is a related, but tangential topic.

Lucarda:
This patch from @katjav test patches works for me with the _three_ "framesize" options. It states: ` "No matter what the floating point precision of Pd is, fft~&Co. will be limited to a maximum framesize of 2^17 = 131072"`. If I run this patch on single precision it crashes Pd (w32 or w64) if I select "framesize" [131072( or [262144( .

Untitled-3

Spacechild1:
>  If I run this patch on single precision it crashes Pd (w32 or w64) if I select "framesize" [131072( or [262144( .

should be fixed with https://github.com/pure-data/pure-data/pull/81

Lucarda:
Also I have a patch that was limited by single precision but with Pd-double it works perfect. It's just a +1 counter doing stuff on big arrays. Now this arrays can be bigger than 2^24.

Lucarda:
@umlaeute :

Is it OK if I upload this **Pd-0.52-2-double** (windows64) to my puredata.info folder and we ask the Pd-list to do some testing with vanilla patches?

umlaeute:
@Lucarda i think its good to upload an experimental build somewhere. but you might revisit the version number...

although i would just attach such an experiment to this ticket.
or better yet, instruct the AppVeyor CI to also create double-builds...

umlaeute:
Am 4. Dezember 2019 23:31:47 MEZ schrieb katjav <notifications@github.com>:
>katjav commented on this pull request.
>
>This is not good. 


indeed.

as a sidenote: the new perform routine doesn't wrap very large numbers (eg 1e10 or somesuch, cannot test right now), but passes them through....
so we need fixes anyhow...

umlaeute:
@katjav :woman_facepalming: my sidenote is actually the thing that your patch was attempting to fix in the first place.

anyhow, a8c4099 is a simple replacement for your original algorithm (and seems to perform better on my system; i haven't investigated though)

katjav:
> things left out

> this patch-set does not include the following changes from the original proposal:
> [...]
> changing the printout-format for numbers to contain more digits.
>  this is really just eye-candy, and it slightly annoyed me to see 439.99999243520523 instead of 440 (even though the former was certainly more precise)

Sorry to disagree here (if 'printout-format' means all GUI representations). Optimal precision in the GUI is important not only for visual verification but, in Pd GUI widgets, also because the representation determines how a number will be saved with the patch. In pd-double if you create an object like this:

`[* 3.1415926535]` 

the number will be saved without losing precision. In the current double precision branch of pure-data the same number will be truncated to 3.14159 in the visual representation. While this may not affect the binary number under the hood before stopping Pd, it is the truncated number that will be saved and later reloaded. 

Pd-double displays a maximum of 14 significant digits for double precision, based on the same rationale as 6 decimal digits in single precision Pd. Those are the maxima that can be displayed without causing apparent inconsistencies, where computer calculations would differ from those done 'by hand' due to binary <> decimal conversions. By the way this means we lose up to 2 decimal digits in the text format anyhow, for single and double precision alike.

Number `439.99999243520523` is thus rounded to `440` in single precision and `439.99999243521` in double precision. That is a consequence of the float/double precision difference for the specific example. If you type `[* 439.999}` or `[* 440]`, both represent the number as typed. It is not like extra precision makes everything look annoying.

danomatika:
I agree with both points of view:

* I do not want to see extra-long precision in boxes everywhere, aka truncated numbers
* I do want to have precision kept when saving and loading patches

Does this call for some sort of *explicit* way to "round" precision by the user? Maybe an object?

###

Recently, I had a similar issue with a project at work where math is done using doubles internally but values are saved and loaded via XML files using float precision. In this case, a computed angle rounded from "-29.9999999998" to "-30" makes for a more readable output and, for our use case, no real loss of discernible precision.

My solution so far is to explicitly round to 6 digits while saving using:
```C
/// round double to a float with 6 digit precision: -0.00000000001 -> 0.0
float roundtof(double value) {
    double v = value * 1000000.0;
    v -= remainder(v, 1.0);
    return v / 1000000.0;
}
```

Thus intermediate values are kept, ie. "-29.9998" while much smaller precision is dropped.

umlaeute:
what i find more important, is that the string-to-float (e.g. `[symbol [1.2345678901( -> [float]`) uses single-precision `float`...

katjav:
Let's distinguish between GUI widget types.

In object boxes and messages boxes you can type numbers as you want them. In those cases there's no need for Pd to truncate more than float precision requires, i.e. maximally 6 resp. 14 significant decimal digits depending on float type.

In numboxes you may  see results of calculations that are more precise than you need. In this case you can set the box width, which almost determines the number of significant digits apart from the fact that a decimal point is also a character. If set to 4 digits, number `439.999` will show as `439.`,  no matter how much we'd prefer to see `440.`, and regardless how many significant digits Pd allows to represent.

The only real 'issue'  that I see with extra precision (in GUI widgets and saved patches): Pd will behave different depending on float type. One could distribute a 'double precision patch' which will be rewritten to single precision patch when re-saved in single precision Pd even without editing.

danomatika:
macOS test1 build is here: http://docs.danomatika.com/pdbuilds/0.51/

I noticed a buffer size issue when Pd is running at a different sample rate than the system, in my case Pd 48k and system 44.1k. This works fine with single precision Pd as Portaudio handles the sample rate conversion, so it might be something to do with #806.

Lucarda:
> I noticed a buffer size issue when Pd is running at a different sample rate than the system

Couldn't replicate on Windows. I set Pd to 48k and works fine using MMIO or ASIO.

danomatika:
It might something with my system as the float build is also acting strangely. When in doubt, reboot!

katjav:
> anyhow, a8c4099 is a simple replacement for your original algorithm (and seems to perform better on my system; i haven't investigated though)

@umlaeute ,  I tried your simple replacement in pd-double's 'naive' [phasor~] method to check performance penalty as compared with Hoeldrich method. Performance of 'naive method' is frequency-dependent since branch prediction misses more often for high frequencies. Up till 1KHz it is now (with check against INT_MIN / INTMAX) similar to Hoeldrich. At 10KHz it is 1.5 times more expensive. Tested on Linux x86_64.

Lucarda:
> , i would really welcome tests that go beyond testtone.pd

I have gone through all **doc/3.audio.examples** and it's all working nice.

millerpuckette:
is this PR ready to merge? (looks good to me so far but I haven't looked closely).

umlaeute:
i think so.

i just pushed a small fix for the maximum input to `[exp]`.

PR #825, opened 2019-12-09
basic PDDEPRECATED macro

danomatika:
This PR adds a basic PDDEPRECATED macro and applies it to *obvious* legacy functions in m_pd.h, ex. :

~~~
PDDEPRECATED EXTERN t_float *value_get(t_symbol *s); /* use value_getfloat() */
~~~

This is just a suggestion / proof of concept as I know there are varying approaches to this, including the existing version compatibility checks. Currently, warnings are printed for *any* usage of the affected functions, so legal internal usage such as `clock_getsystime()` will also print a warning.


danomatika:
Ping @Spacechild1.

danomatika:
Also, the macro define could be wrapped to allow for disabling it ala:

~~~
#ifndef PDDEPRECATED
...
#endif
~~~

This would allow for disabling via: `make CFLAGS="-DPDDEPRECATED=\"\"" or more simply (but counter-inutituvely): `make CFLAGS="-DPDDEPRECATED"`

src/m_pd.h
```diff
@@ -681,15 +691,15 @@ EXTERN int garray_getfloatwords(t_garray *x, int *size, t_word **vec);
...
+PDDEPRECATED EXTERN void garray_resize(t_garray *x, t_floatarg f); /* use garray_resize_long() */
 EXTERN void garray_resize_long(t_garray *x, long n);   /* better version */
 EXTERN void garray_usedindsp(t_garray *x);
 EXTERN void garray_setsaveit(t_garray *x, int saveit);
 EXTERN t_glist *garray_getglist(t_garray *x);
 EXTERN t_array *garray_getarray(t_garray *x);
 EXTERN t_class *scalar_class;
 
-EXTERN t_float *value_get(t_symbol *s);
+PDDEPRECATED EXTERN t_float *value_get(t_symbol *s); /* use value_getfloat() */
```
Spacechild1: As noted in https://github.com/pure-data/pure-data/pull/817#issuecomment-563009409, `value_getfloat` is not a simple replacement of `value_get`. While the former just returns the current float value and returns false if the value object doesn't exist, the latter creates a strong reference to the value object (which has to be released with `value_release`) and also creates the value object itself if needed. Until we have a better replacement, like my proposed `value_retain` method, we shouldn't deprecate it.

Spacechild1:
> 
> 
> Also, the macro define could be wrapped to allow for disabling it ala:
> 
> ```
> #ifndef PDDEPRECATED
> ...
> #endif
> ```
> 
> This would allow for disabling via: `make CFLAGS="-DPDDEPRECATED=\"\""` or more simply (but counter-intuitively): `make CFLAGS="-DPDDEPRECATED"`

Or use a seperate define to enable/disable deprecation warnings :-)

src/m_pd.h
umlaeute: i would recommend setting the deprecated flag in multiple steps:
- first mark functions as deprecated where we *know* that they are indeed deprecated because they are already marked as deprecated in the headers (e.g. `clock_getsystime`, `garray_resize`)
- in a separate commit mark functions that are deprecated that say so when being called (e.g. `garray_getfloatarray`)
- finally mark functions that are supposedly deprecated (~`value_get`~, if you really think it is)

umlaeute:
btw, `clock_getsystime()` and `clock_getlogicaltime()` really have identical implementations, so I guess we should replace the "legit" internal uses of the former with the non-deprecated version...

danomatika:
>As noted in #817 (comment), value_getfloat is not a simple replacement of value_get.

I know. It's more of the *spirit* of the usage: as in use get/set instead of *f.

Spacechild1:
The point is that value_get

Spacechild1:
(Sorry, accidentally closed PR when typing on my phone)

umlaeute (changes requested):
i like this.

but i agree with @Spacechild1 that there's no apparent reason to mark `value_get` as being deprecated, so i would remove this.

umlaeute:
I've cleaned up this PR, incorporated the changes discussed here and renamed the macro to `PD_DEPRECATED` (so it matches the other Pd-specific macros, which are mostly prefixed with `PD_` rather than just `PD`).

the changes are included in
(so once #2223 is merged, this one can be closed as well)

danomatika:
This looks good to me.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Mar 18, 2024, at 9:06 AM, umläute ***@***.***> wrote:﻿
I've cleaned up this PR, incorporated the changes discussed here and renamed the macro to PD_DEPRECATED (so it matches the other Pd-specific macros, which are mostly prefixed with PD_ rather than just PD).
the changes are included in

#2223
(so once #2223 is merged, this one can be closed as well)

umlaeute:
closing this, as the followup PR #2223 was merged

PR #827, opened 2019-12-09
[WIP] implement list support for [value]

Spacechild1:
This PR adds the possibility to store any number of atoms (float, symbol, gpointer) in a value.

It is a generalization of https://github.com/pure-data/pure-data/pull/817.

Here's a simple test patch: [value-list-test.zip](https://github.com/pure-data/pure-data/files/3942253/value-list-test.zip) Basically it works but it could need more testing in case I've overlooked something.

One general problem is that `value_get` would become unsafe (see https://github.com/pure-data/pure-data/pull/817#issuecomment-563009409) and would have to be deprecated. I've introduced `value_retain` as a replacement for creating a strong reference to a value object. The rest can be done with the getter/setter methods.


danomatika:
As the `value-send` branch was merged, do you think this branch is up to date or needs work, ie. for after 0.51?

Spacechild1:
The branch already contains the `[send(` method, I just need to do more extensive testing and update the documentation.

Spacechild1:
OK, did some more stress tests (including gpointers) and everything looks fine.

porres:
tried this and it looks good, I also worked a bit on the documentation and added a bit more info, I think you can merge it to this branch, check it out https://github.com/Spacechild1/pure-data/pull/3

porres:
here's a better one :) https://github.com/Spacechild1/pure-data/pull/4

Spacechild1:
TBH, I'm not 100% convinced of this PR myself because you can already achieve the same thing with `[text define]` + `[text get]` + `[text set]`. I just realized this after finishing the PR...

porres:
I don't think I ever needed this, but I've seen people want it to at least hold symbols and a list expansion seems a natural. Well, I guess you can find more than one workaround but the point is that this is quite convenient. I don't really get how to do the same thing with text but it seems you need many many objects to achieve the same thing.

Spacechild1:
> I don't really get how to do the same thing with text but it seems you need many many objects to achieve the same thing.

well, you define a "variable" with `[text define]` and then set/get the "variable" with `[text set]` / `[text get]`. A `[value]` holding a symbol or list is not much different than a `[text define]` holding a single line of text.

IMO, the only place where symbol support for `[value]` would be really justified is for `[expr]`.

But if Miller and other people like this feature, I won't oppose :-)

porres:
> But if Miller and other people like this feature, I won't oppose :-)

I see :) well, just accept my help update PR then just in case ;) => https://github.com/Spacechild1/pure-data/pull/4

danomatika:
Hah, well then neither value-symbol or value-list branch are needed. Just add a  item about using [text] for symbols and lists to the value-help.pd. Done.

In the end, the simplest is best.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
OTOH value-list would bring parity to value with the Max version. We’ll see what Miller thinks.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
Another problem I just realized: since "bang" outputs the stored value, and "bang" is equivalent to an empty list, there would be no way to set the value to an empty list - which feels very inconsistent to me. I am tempted to close this PR...

Ant1r:
I usually use [list( to initialize an empty list, wouldn't it work in this case?

Spacechild1:
> I usually use [list( to initialize an empty list, wouldn't it work in this case?

Nope, because `[list(` is the same as `[bang(`.

porres:
Ok, just wanted to say that after checking things out, this seems like a better choice than https://github.com/pure-data/pure-data/pull/817 just wanted to also say that we should make it work with the new string functions in [expr].

Also note that I suggested a list type in [expr] in https://github.com/pure-data/pure-data/issues/2108 and maybe we could one day integrate these ideas...

porres:
Hmmm, i suggested this but I see now we got some conflicts with [expr], I could try and see if I can resolve them

PR #828, opened 2019-12-10
add depth argument to 'isvisible' message

### LINKED ISSUE #803: request: include depth argument for 'isvisible' message in [pdcontrol]
a depth message can query if parent window is visible or not, in a similar fashion to 'dir/args' messages.

porres:
closes #803

### DIFF
--- a/src/x_gui.c
+++ b/src/x_gui.c
@@ -474,9 +474,16 @@ static void pdcontrol_browse(t_pdcontrol *x, t_symbol *s)
     sys_gui(buf);
 }
 
-static void pdcontrol_isvisible(t_pdcontrol *x)
+static void pdcontrol_isvisible(t_pdcontrol *x, t_floatarg f)
 {
-    outlet_float(x->x_outlet, glist_isvisible(x->x_canvas));
+    t_canvas *c = x->x_canvas;
+    int i;
+    for (i = 0; i < (int)f; i++)
+    {
+        if (c->gl_owner)    /* back up one more into an owner if any */
+            c = c->gl_owner;
+    }
+    outlet_float(x->x_outlet, glist_isvisible(c));
 }
 
 static void pdcontrol_setup(void)
@@ -490,7 +497,7 @@ static void pdcontrol_setup(void)
     class_addmethod(pdcontrol_class, (t_method)pdcontrol_browse,
         gensym("browse"), A_SYMBOL, 0);
     class_addmethod(pdcontrol_class, (t_method)pdcontrol_isvisible,
-        gensym("isvisible"), 0);
+        gensym("isvisible"), A_DEFFLOAT , 0);
 }
 
 /* -------------------------- setup routine ------------------------------ */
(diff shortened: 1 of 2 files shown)


porres:
I see many other applications of the commit https://github.com/pure-data/pure-data/pull/828/commits/f4816e9a31261c0efcfbf6ec62feb4be132e34f8 . I think it's a more powerful/configurable alternative solution to the key issue in https://github.com/pure-data/pure-data/pull/570 but it can also be used for other things, like filtering out mouse events from [cyclone/mousestate] for instance

porres:
commit https://github.com/pure-data/pure-data/pull/828/commits/31d293142ed186766a95434e0d36164b193f3354 closes https://github.com/pure-data/pure-data/issues/1027

src/x_gui.c
```diff
@@ -497,6 +497,17 @@ static void pdcontrol_version(t_pdcontrol *x)
...
+static void pdcontrol_dollarzero(t_pdcontrol *x, t_symbol *s, t_floatarg f)
+{
+    t_canvas *c = x->x_canvas;
+    for (int i = 0; i < (int)f; i++)
+    {
+        if (c->gl_owner)  /* back up one more into an owner if any */
+            c = c->gl_owner;
+    }
+    outlet_symbol(x->x_outlet, canvas_realizedollar(c, s));
+}
```
umlaeute, line 509: this will realize *all* dollargs, not just dollarzero.

so the method name `dollarzero` is plainly wrong.

porres: I guess the best way out is to rename it then, so I did change it to "realizedollar"

Lucarda:
@porres could you append an atom to the output of `[version(`

so we get:

`print: list version 0 54 4`

And we can use a [route version]

What do you think?

porres:
yeah, in order to use [route version] we need the output to start with 'version' instead of list, and [pdcontrol] doesn't really send tagged data to tell you what you asked for, so it'd be a special case as an odd one out. I guess the idea is that you know what you're asking to [pdcontrol] so you know what you'r getting. And if you want different things from [pdcontrol] you have to use more than one.

umlaeute:
i agree with @porres.
`[pdcontrol]` is a design nightmare, but the intended use for querying a version is to use an other instance of `[pdcontrol]` than for querying e.g. the path

porres:
as long as this doesn't have conflicts, I can update the help file later, I'm always strugglng on solving conflicts :(

porres:
> as long as this doesn't have conflicts, I can update the help file later, I'm always strugglng on solving conflicts :(

yeah, with all the doc updates I did I'm having to solve many conflicts in lots of old PRs of mine. I don't know why this now fails in one of the checks, same thing happened to another PR but not all... I just put the original help file from master and didn't change the code

Ant1r:
The CI check_file_permissions output is pretty clear:

src/x_gui.c
Data files should never have the EXECUTABLE BIT set!

porres:
sorry I didn't read it :/ I was just surprised that I hadn't touched the code, but then, maybe this check was added later and my old mistake just wasn't caught before ;) there's also another error on another PR that I made the same conflict resolving but read the error and it wasn't this at all... so that remains a mystery.

thanks

PR #836, opened 2019-12-12
[WIP] multi-instance hook update

danomatika:
This PR aims to introduce per-instance hooks for printing and MIDI out for libpd. This involves changes to the core, hopefully none too involved.

* [x] per-instance print hooks; working via `sys_printhook` -> `STUFF->st_printhook`
* [ ] per-instance float, symbol, & message hooks
* [ ] per-instance MIDI out hooks

The message and MIDI hooks could be added via a void* context pointer added to the instance STUFF.

This helps rectify one of the issues with libpd's current implementation: you can send to specific instances but multi-instance receiving is to the same, shared hooks.

See companion [libpd PR](https://github.com/libpd/libpd/pull/282).


danomatika:
Originally, I wanted to remove `sys_printhook` but it's required as libpd already set the convention that it's possible to set any of the hooks *before* pd/libpd has been inited. I brought it back as the default per-instance hook when an instance is inited.

A second solution, which would allow for removing sys_printhook, would be to move the default hook storage into libpd and set it manually in libpd init and whenever a pd instance is created.

Right now, I feel the first option is the safest as this ensures sys_printhook is always set.

danomatika:
I was thinking about the MIDI functions and one solution would be to add MIDI callback function pointers to the midi instance struct. By default, these would point to the `outmidi` functions but could be overridden by libpd. I'd prefer not to add NULL checks, so care would be needed to ensure to reset the function pointers back to the default if/when disabling in libpd.

danomatika:
I'm looking at this again now that libpd is integrated into the main repo.

danomatika:
@millerpuckette Can you consider this for 0.52? I think I have the MIDI hooks sorted in the libpd layer and this is the only change needed in the Pd core.

danomatika:
I've now added an implementation-specific opaque data pointer `st_impdata` to the pdinstance which libpd can use to store the per-instance hooks. This allows simplifying the libpd side which simply grabs its implementation data for the current instance.

Ant1r:
Could you also change in x_interface.c:
```C
#define print_startlogpost(object, fmt, ...) do{ \
    if (sys_printhook || sys_printtostderr) \
```
to
```C
#define print_startlogpost(object, fmt, ...) do{ \
    if (STUFF->st_printhook || sys_printtostderr) \
```
?
Now the [print] output is still prefixed with `verbose(2):` when using mutli-instance printhooks.

danomatika:
Ok, I will look at this tonight.

umlaeute:
probably the `[WIP]` could be dropped from the description.

umlaeute (approved):
looks good.

danomatika:
Thanks for merging this. I have integrated these changes with the libpd `features/multi-hooks` branch. We'll do some testing, then probably release libpd 0.13 with *full multi-instance receiving support*.

Issue #835, opened 2019-12-12
send messages to a (parent) canvas


Spacechild1:
Sometimes we need to a way to send messages to the current canvas or one of its parents. Currently, this is only possible if you know the symbol the canvas is bound to. `[namecanvas]` does allow to bind an abstraction to a (unique) symbol, but still gives no way to talk to the parent.

I see four basic ways how to implement this:

1) have a dedicated object, similar to `[iemguts/sendcanvas]` (with the parent level as a creation argument)
2) use a `[pointer]` and expand the `[traverse(` message to (alternatively) accept the parent level as a float argument, so we can send a message to *this* canvas with `[traverse 0, send-window foo bar( -> [pointer]` and to the *parent* canvas with `[traverse 1, send-window foo bar( ->[pointer]`
3) add a message to some existing class (e.g. `[pdcontrol]` or maybe even `t_canvas`) like `[send-window <parent-level> <message...>( `
4) add a second argument to `[namecanvas]` to specify the parent level, as suggested by @porres 

Are there more ways?

As @umlaeute points out in https://github.com/pure-data/pure-data/issues/794#issuecomment-558531123, using a receive symbol (as in the `[namecanvas]` approach) might not be the cleanest approach.

Personally, I don't care much because I use `iemguts` anyway :-) Conceptually, I like 2) but other people might find it too arcane.

porres:
> As @umlaeute points out in #794 (comment), using a receive symbol (as in 4) might not be the cleanest approach.

Not, not "as i 4". As I see it, this comment talks about getting a file name so you can send messages to it, not about setting a name with [namecanvas] (as in 4). 

[sendcanvas] is merely a conjunction of [namecanvas] and [send], but it bypasses the need of setting a name. [sendcanvas] can also refer to a parent patch, and this is why I'm proposing [namecanvas] could do the same.

My main opposition here is with creating a whole new object (option 1). I think we can find a way to use an existing object. 

> Conceptually, I like 2) but other people might find it too arcane.

yeah, I can't understand this one at all.

As for the others, I feel [namecanvas] seems the best candidate as it's closely related to [sendcanvas] and its purpose is exactly this, allowing communication to a patch. Therefore, if we include the functionality to talk to a patch with another object like [pdcontrol], it becomes completely and actually obsolete. 

[...]

Issue #833, opened 2019-12-10
request: make pd ask if I want to save changes in the patch when I altered an abstraction


porres:
So, we have the savestate object, but we can accidentally close the patch without saving something cool we wanted.

Not sure how's the best way to do this, if [savestate] could receive a message for that, or if we could have another functionality in another object (a new object just for this would be silly). As for the second option, maybe a way to abuse dynamic patching and make the abstraction talk to the parent patch (and send it a "dirty 1" message) could do the trick, but it's a rather low level solution. Something like that could be achieved with the same request as https://github.com/pure-data/pure-data/issues/794 - where an object like [pdcontrol] could give us the filename of a parent patch.

The advantage of having a functionality built in somehow in savestate is that it'd be easir. The advantage of the second option is that people could use this to abuse it in other ways.

porres:
For instance, changing parameters in inemguis (color, size, etc) makes pd asks if the user wants to save changes. It'd be interesting if you could have a way to do that for abstractions as well

Spacechild1:
> So, we have the savestate object, but we can accidentally close the patch without saving something cool we wanted.

? Pd already asks if you want to save changes made to an abstraction... what are you missing?

porres:
> ? Pd already asks if you want to save changes made to an abstraction... what are you missing?

I meant like changing parameters, like when it receives data that can be stored. Currently if you mess with your abstraction's parameters, [savestate] won't bother to ask you if you want to save it before closing.

porres:
> For instance, changing parameters in inemguis (color, size, etc) makes pd asks if the user wants to save changes. 

I just saw that this is not true when you change these parameters via messages (say, a "size" message for instance). It feels weird that it asks you if you want to save changes if you altered via the properties but not from messages - but that'd be a separate issue

Spacechild1:
Because the dialog sets the dirty flag, while sending a message to an object doesn't. The rationale is that whenever you change something from the GUI, the canvas gets dirty. On the other hand, changing the internal state of an object with a message obviously doesn't do that. Imagine that changing the value of `[float]` would always set the dirty flag...

If you want to make the canvas dirty when the user changes a certain value, just send `[dirty 1( `to the abstraction itself and you're done.

porres:
I can see it makes no sense for [float] but not for GUIs, but for that I opened a new issue - https://github.com/pure-data/pure-data/issues/834

[...]

PR #837, opened 2019-12-12
Add a 'depth' argument to [namecanvas]

porres:
This PR adds a 2nd float argument that sets a name to a parent patch (like with 'dir'/'args' methods in [pdcontrol]). The use case is that sometimes abstractions need to talk to the parent patch. 

This PR closes https://github.com/pure-data/pure-data/issues/835 as it's one of its possible solutions. This feature has recently been requested a couple of times and https://github.com/pure-data/pure-data/issues/835 is where the current discussion is.

I don't mean to impose this solution, and maybe I can close this PR in favor of another one. And, as I mentioned in https://github.com/pure-data/pure-data/issues/835#issuecomment-565108439 I'm not sure how @millerpuckette feels about [namecanvas], so this is my main concern.

The code says "Delete later" in the comments, but obviously it won't be deleted anymore. Therefore, my first commit here only removes that in the comment. The 2nd commit implements the feature.

### DIFF
--- a/doc/5.reference/namecanvas-help.pd
+++ b/doc/5.reference/namecanvas-help.pd
@@ -1,13 +1,19 @@
-#N canvas 402 55 524 371 12;
-#X obj 39 20 namecanvas;
-#X obj 39 108 namecanvas bonzo;
-#X text 291 309 updated for Pd version 0.44;
-#X text 131 20 - attach this canvas to a name;
-#X text 34 150 This is sometimes the only way to send a message to
-a canvas when making graph-on-parent abstractions \, but its use in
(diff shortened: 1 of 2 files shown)


porres:
One use case I have is sending a "dirty 1" message to the parent patch, so Pd asks you if you want to save changes in parent patch before closing it. This way I can have the parent patch keep changes I made to the abstraction (with [savestate] for instance) - this had been discussed in https://github.com/pure-data/pure-data/issues/833

Other use cases are possible, like designing abstractions that can create objects on the parent patch or anything else people feel like doing with dynamic patching, which was requested in https://github.com/pure-data/pure-data/issues/833

PR #841, opened 2019-12-13
several gpointer bugfixes

Spacechild1:
* `[print]`: print a single gpointer as `(pointer)` for consistency with all other objects
* `[print]`: correctly print a list containing a single gpointer (the current code accidentally prints a symbol selector)
* `[route]`: correctly forward a single gpointer (the current code accidentally sends the gpointer as a symbol!)
* `[trigger]`: fix gpointer error message
* `pd_typedmess`: check for gpointer, so we can't type bogus messages like `[pointer foo(`, just like we can't type `[float foo(`. This also ensures that `gpointer -> [pointer $1(` stays a gpointer message.
* `binbuf_eval`: check for gpointer selector, so we can pass a gpointer to `[$1 ...(` (the current code just silently ignores the message).

### DIFF
--- a/src/m_binbuf.c
+++ b/src/m_binbuf.c
@@ -791,7 +791,12 @@ void binbuf_eval(const t_binbuf *x, t_pd *target, int argc, const t_atom *argv)
                 if (nargs == 1) pd_float(target, mstack->a_w.w_float);
                 else pd_list(target, 0, nargs, mstack);
                 break;
+            case A_POINTER:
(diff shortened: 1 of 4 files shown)


Spacechild1:
@millerpuckette maybe this also something for Pd 0.51-1? Since those bugs have been around for years it is not really urgent, but they have always bothered me when working with data structures. The `[route]` bug is especially nasty.

PR #842, opened 2019-12-14
make sys_lock/sys_unlock recursive

Spacechild1:
Currently, `sys_lock()/sys_unlock()` uses a simple non-recursive mutex which is prone to dead locking. 

This causes problems in libpd (https://github.com/libpd/libpd/issues/274). Although libpd could use its own recursive (per-instance) mutex instead of `sys_lock() / sys_unlock()`, I guess the easiest solution is to use a recursive mutex in Pd itself.

It would also help to simplify the Pd scheduler, where we have to be very careful not to deadlock.

I've tested my changes with both single-instance Pd and multi-instance Pd (`-DPDINSTANCE`), using a custom external which calls `sys_lock() / sys_unlock()` in one of its methods (while the message system is being locked) and I could verify that it doesn't deadlock.

I also had a quick look at the winpthreads implementation and the overhead of a recursive `pthread_mutex_t` over a non-recursive one is very small.

### DIFF
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -125,7 +125,7 @@ struct _instanceinter
     LARGE_INTEGER i_inittime;
     double i_freq;
 #endif
-#if PDTHREADS && PDINSTANCE
+#if PDTHREADS
     pthread_mutex_t i_mutex;
 #endif
 };
@@ -1551,14 +1551,12 @@ void s_inter_newpdinstance(void)
 {
     pd_this->pd_inter = getbytes(sizeof(*pd_this->pd_inter));
 #if PDTHREADS
-#if PDINSTANCE
     {
         pthread_mutexattr_t attr;
         pthread_mutexattr_init(&attr);
         pthread_mutexattr_settype(&attr, PTHREAD_MUTEX_RECURSIVE);
         pthread_mutex_init(&pd_this->pd_inter->i_mutex, &attr);
     }
-#endif
     pd_this->pd_islocked = 0;
 #endif
 #ifdef _WIN32
@@ -1585,13 +1583,9 @@ void s_inter_freepdinstance(void)
     s_inter_free(pd_this->pd_inter);
 }
 
-#if PDTHREADS
-#ifdef PDINSTANCE
+#if PDTHREADS && defined(PDINSTANCE)
 static pthread_rwlock_t sys_rwlock = PTHREAD_RWLOCK_INITIALIZER;
-#else /* PDINSTANCE */
-static pthread_mutex_t sys_mutex = PTHREAD_RECURSIVE_MUTEX_INITIALIZER;
-#endif /* PDINSTANCE */
-#endif /* PDTHREADS */
+#endif
 
 #if PDTHREADS
 
@@ -1626,12 +1620,10 @@ first.  The "pd_this"  variable can be written and read thread-safely as it
 is defined as per-thread storage. */
 void sys_lock(void)
 {
-#ifdef PDINSTANCE
     pthread_mutex_lock(&pd_this->pd_inter->i_mutex);
+#ifdef PDINSTANCE
     pthread_rwlock_rdlock(&sys_rwlock);
     pd_this->pd_islocked = 1;
-#else
-    pthread_mutex_lock(&sys_mutex);
 #endif
 }
 
@@ -1640,10 +1632,8 @@ void sys_unlock(void)
 #ifdef PDINSTANCE
     pd_this->pd_islocked = 0;
     pthread_rwlock_unlock(&sys_rwlock);
-    pthread_mutex_unlock(&pd_this->pd_inter->i_mutex);
-#else
-    pthread_mutex_unlock(&sys_mutex);
 #endif
+    pthread_mutex_unlock(&pd_this->pd_inter->i_mutex);
 }
 
 int sys_trylock(void)
@@ -1662,7 +1652,7 @@ int sys_trylock(void)
     }
     else return (ret);
 #else
-    return pthread_mutex_trylock(&sys_mutex);
+    return pthread_mutex_trylock(&pd_this->pd_inter->i_mutex);
 #endif
 }
 
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -125,7 +125,7 @@ struct _instanceinter
     LARGE_INTEGER i_inittime;
     double i_freq;
 #endif
-#if PDTHREADS
+#if PDTHREADS && PDINSTANCE
     pthread_mutex_t i_mutex;
 #endif
 };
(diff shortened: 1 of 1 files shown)


danomatika:
This looks good to me.

danomatika:
To reiterate: this is pretty important for libpd.

millerpuckette:
I think the correct solution would be for "hooks" in libpd to be called with the lock released.  (One
knows at compile time that the lock is in place when someone (for instance) sends a message to a
print object, so instead of calling the printhook one could release the lock, then call it, then relock.

I found an entertaining screed about recursive mutexes here:
.http://www.zaval.org/resources/library/butenhof1.html
the upshot is that it's a good idea, if it's possible, to see to it that at every line of code we know at
compile time whether the lock is held or not.  I find that persuasive.

Spacechild1:
Thanks Miller, that was a good read :-) I agree that this PR is mainly a crutch that the need of recursive locks is often (always?) a sign for a design flaw. On the other hand, it's basically free. I don't have a strong opinion...

> I think the correct solution would be for "hooks" in libpd to be called with the lock released

I guess this could work. Float atomes are fine, symbol atoms also (the don't move in memory and the `s_name` part is immutable) and pointer atoms are not supposed to be used by libpd clients AFAICT.

I think the even better approach would be not to lock a mutex inside libpd *at all*. In virtual all libpd methods, they currently do the following:

1) lock a mutex
2) do some stuff
3) unlock the mutex

This is really an anti-pattern because 1) and 3) can be easily handled by the user. In fact, they might prefer to use their own mutex type (e.g. libwinpthread sucks) or send messages via lockfree FIFOs (the good way). BTW, libpd already has an (optional) lockfree queue for *receiving* messages, so I guess it could also offer one for *sending* messages.

Having the synchronization *outside* the library makes thread synchronization easier to reason about. Sometimes it's better not to hide things just for the sake of user convenience.

So I think libpd could actually wrap all those calls to `sys_lock` and `sys_unlock` in a #define, which is off by default, but can be enabled for old codebases.

Spacechild1:
Even if Pd's lock is recursive, calling a callback function while holding a lock can still deadlock, even if there's no reentrancy. Here's a real life example, which I've encountered more than once:

Let's say we have Thread A, Thread B, Ressource C and Ressource D.

1) A locks C and calls a callback function while holding the lock.
2) B locks D
3) the callback function attempts to lock D but has to wait
4) B attempts to lock C and has to wait.
5) deadlock

This is really nasty because it is mostly a timing issue. Now if one of the locks is *inside* a library, this can be extremely difficult to debug, especially if you can't look at the source code.

I once had this issue with a (buggy) VST plugin which called the host callback while locking the audio thread. In my client code, both the audio thread and the host callback needed a shared ressource and it would occasionally deadlock. Took me a while to figure out what was going on... I didn't have the source code for the plugin, I only realized later when I made a similar mistake in my own code.

So the takeaway is: never, never, never lock a (hidden) mutex while calling a callback function, unless you can prove 100% that it's safe.

Spacechild1:
Some more thoughts. When you design a library where parts of it could/should be called from seperate threads, there's always the question whether to handle thread synchronization inside or outside the library itself. My personal conclusion so far has been, that it's best to leave thread synchronization *outside* and only do it in the library if:

1.  there's no need for locks (atomic variables, lock free data structures, message queues, etc.)

and/or

2.  you can use fine grained locking with short critical sections

Regarding 2:

If all of the library methods look like this:
```
void somemethod()
{
    lock()
    // do stuff
    unlock()
}
```
it is certainly better to do
```
lock()
somemethod()
unlock()
```

Both effectively do the same thing and I think explicit is better than implicit. As I've already mentioned, it gives the user the possibily to user their own synchronization primitives. Also, the user can control the atomicity of method calls!
```
foo() // internal locking
call bar // internal locking
```
is different from
```
lock()
foo()
bar()
unlock()
```

Using internal locks might be OK if the methods instead look like this:
```
void methodA()
{
    // do lots of stuff
    lock()
    // short critical section
    unlock()
    // do more stuff
}
```

danomatika:
If want the locking outside of the libpd C api, then we are back to how it was before where locking is done in the language wrappers. I would still add an optional layer which does the locking as  ist projects don’t need to do it manually.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
> If want the locking outside of the libpd C api, then we are back to how it was before where locking is done in the language wrappers.

Not really, because the locking shouldn't simply be moved to the language wrappers.

> I would still add an optional layer which does the locking as  ist projects don’t need to do it manually.

This is basically what I was suggesting. Personally, I would only keep the locking for backwards compatibility. I mean, if someone is writing a multithreaded programm, it's not too much to ask to create a mutex and lock it before calling a method from a different thread...

On the other hand, simply releasing the lock while calling the message hooks as suggested by Miller could be enough.

And of course, one should encourage users to use the queued hooks, because that's generally the better solution. And it would be cool to have queued sending as well :-) Lockfree FIFOs for the win!

danomatika:
I don’t think removing the locking by default is a good idea. Most everyone relies on it now, mostly through the language wrappers. If people want to ha dlenit directly, then issuing a default layer which can be overridden makes sense.

If you want to check the queues implementation and improve it so it can be the default, please feel free. :)

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Spacechild1:
> Most everyone relies on it now,

Yeah, I see that the transition would be difficult to manage.

> If people want to ha dlenit directly, then issuing a default layer which can be overridden makes sense.

Agreed.

> If you want to check the queues implementation and improve it so it can be the default, please feel free. :)

I will :-)

In the meantime, what do think about releasing the lock in the message hooks as Miller suggested? AFAICT, this would solve the deadlock issue at hand.

danomatika:
>In the meantime, what do think about releasing the lock in the message hooks as Miller suggested? AFAICT, this would solve the deadlock issue at hand.

Hrmm that could work. If the variables are cached, it should be fine, although incoming atoms are probably provided as pointers.

Spacechild1:
I think it's definitly fine for the float, symbol and bang hooks.

Regarding list/message hooks:

The atom lists received by `libpdreclist` and `libpdrecanything` are always valid till the end of the function call. Like any other Pd method call, the caller always has to put them on the stack or allocate on the heap and free after the call, otherwise reentrant function calls could invalidate the memory. 

Float atoms are always safe. Symbol atoms should also be ok, because symbols never change their memory address and the `s_name` part is immutable.

Actually, when you look at `internal_listhook`/`internal_messagehook` in `z_queued.c` you'll see that it simply does a *shallow* copy of the atom list, which is then later passed to the user-provided hook in `libpd_queued_receive_pd_messages` (without a lock!). So if that is safe, then calling the (non-queued) hook with the lock released should be fine as well.

Spacechild1:
@danomatika I am tempting to close this and rather fix this in libpd (release the lock in the message hooks). what do you think?

danomatika:
>@danomatika I am tempting to close this and rather fix this in libpd (release the lock in the message hooks). >what do you think?

If you think everything would be safe without needing further core changes, sure. I'm just glad to consider both sides and find a good solution that also happens to not involve a ton of core changes. :)

Spacechild1:
I'll take a close look, but I think it's safe (see also https://github.com/pure-data/pure-data/pull/842#issuecomment-615933959). I'll make a PR to libpd.

Spacechild1:
This is now fixed in libpd itself.

PR #849, opened 2019-12-26
make pd_unbind safe (+ improve stack overflow protection)

Spacechild1:
### a) make `pd_unbind` safe

Currently, it is unsafe to dynamically unbind symbols because some other object could be sending to them.

The solution is to simply make a copy of the bind list before iterating over it. A counter keeps track of the total number of items. We use stack allocation up to a certain limit and then switch to heap allocation to avoid a stack overflow.

To make copying as fast as possible, I changed the bind list implementation to a dynamically sized array instead of a linked list. (This is also better for cache coherency).

In a way, this not much different than how `[list]` has to output a *copy* of the stored list to prevent other objects to modify the list while iterating over it.

This PR would finally make it possible to safely implement dyamically settable `[receive]` objects (https://github.com/pure-data/pure-data/pull/604), and make existing similar objects safe, for example iemgui's `[receive(` method.

---

### b) improve stack overflow protection

The current overflow protection only works for a *single* recursion path. For more than 1 path (e.g. several outlets feeding back to the same inlet), Pd would end up in an infinite loop. Once the stacklimit is reached, I set an overflow flag to prevent any further messaging until the stack got completely unwound.

### DIFF
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -6,6 +6,20 @@
 #include "m_imp.h"
 #include "g_canvas.h"   /* just for LB_LOAD */
 
+#include <string.h>
+
+#ifdef _WIN32
+# include <malloc.h> /* MSVC or mingw on windows */
+#elif defined(__linux__) || defined(__APPLE__)
+# include <alloca.h> /* linux, mac, mingw, cygwin */
+#else
+# include <stdlib.h> /* BSDs for example */
+#endif
+
+#ifndef HAVE_ALLOCA     /* can work without alloca() but we never need it */
+#define HAVE_ALLOCA 1
+#endif
+
     /* FIXME no out-of-memory testing yet! */
 
 t_pd *pd_new(t_class *c)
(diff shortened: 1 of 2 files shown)


src/m_pd.c
```diff
@@ -129,21 +176,19 @@ void pd_bind(t_pd *x, t_symbol *s)
...
         {
             t_bindlist *b = (t_bindlist *)s->s_thing;
-            t_bindelem *e = (t_bindelem *)getbytes(sizeof(t_bindelem));
-            e->e_next = b->b_list;
-            e->e_who = x;
-            b->b_list = e;
+            int oldsize = b->b_n++;
+            b->b_vec = (t_pd **)resizebytes(b->b_vec, oldsize * sizeof(t_pd *),
+                                   b->b_n * sizeof(t_pd *));
+            memmove(b->b_vec + 1, b->b_vec, oldsize * sizeof(t_pd *));
```
umlaeute, line 182: one of the CI-builds rightfully fails, as `memmove` and `memcpy` are not declared tue to a missing `#include <string.h>`

umlaeute (changes requested):
please fix the build-failures and warnings :-)

src/m_pd.c
Spacechild1: d'oh. thanks!

Spacechild1:
Actually, I don't know why I grouped these 2 commits together in a single PR... Closing in favor of https://github.com/pure-data/pure-data/pull/1760 and https://github.com/pure-data/pure-data/pull/1759

PR #855, opened 2019-12-31
soundfile updates: refactor, bugfixes, AIFC, & CAF

danomatika:
This PR is focused on d_soundfile.c and thus the [soundfiler], [readsf~] and [writesf~] objects.

My initial focus was on adding support for CAF and AIFC files but I also realized it made sense to refactor the file type handling into separate .c implementations which will make it easier to maintain/extend the code base in the future.

CAF files offer a couple of advantages:

* 64-bit offsets for nigh-unrestricted durations (so file sizes > 4 GB)
* recording could be faster as it allows for appending data
* open format that is supported outside of macOS/iOS

## TODOs

* [x] refactor d_soundfiler.c file type specific code:
  - separate aiff, wave, and next .c files
  - shared d_soundfiler.h header
  - promote t_soundfile_info struct to t_soundfile as basic format info data type used throughout
  - consolidated buffer & byte size defines
  - max file type header size now 128 instead of MAXPDSTRING which should be large enough (I mainly did this as each file type implementation now has a buffer as well for reading the header and 1k seems excessive)
  - moved & renamed `garray_ambigendian()` to `sys_isbigendian()` as it was only used in d_soundfile.c
  - print error if file type format is not supported
  - set NEXT data size when saving (was easy to do)
  - UPDATE: larger byte and frame sizes using `size_t` and `ssize_t`
  - UPDATE: handle max size differences between file types, ie. AIFF chunk sizes are signed
* [x] general cleanup and consistency (naming, format/style, removing unused vars/functions, etc)
* [x] fix 32 bit float file crashes discovered during testing (seems to affect Pd 0.50-2 *and* this branch)
* [x] add WAVE extended fmt chunk support
* [x] add AIFF 32 bit float support
* [x] add AIFF extensible format header support (including "sowt" little-endian sample format)
* [x] add AIFC support (uncompressed formats only)
* [x] add CAF support (uncompressed formats only) -> see PR #879 (now merged)
* [x] ~add 32-bit int support?~ _probably not needed, this would also require an additional flag to distinguish between 4 byte int and 4 byte float_
* [x] further refactor soundfile type handling: t_soundfiler_type struct with function pointers, t_soundfile_info now t_soundfile with fd and type
* [x] expanded soundfiler's ascii .txt read into full read/write .txt type
* [x] extensive, cross-platform testing

## References

* [NEXT](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AU/AU.html)
* [WAVE](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/WAVE/WAVE.html)
* [AIFF/AIFC](http://www-mmsp.ece.mcgill.ca/Documents/AudioFormats/AIFF/AIFF.html)
* [CAF](https://developer.apple.com/library/archive/documentation/MusicAudio/Reference/CAFSpec/CAF_spec/CAF_spec.html)

Additionally, I will do some more testing with the test files linked to the WAVE/AIFF/NEXT references above.

## Notes/Ideas

The sample conversion functions check bytes per sample and endianness continually during processing. Since these only change when the format or file is changed, I think there could be a small performance improvement by having specific function pointers for byte conversion that are set when the format info is set/detected. This is also something that could be overridden/provided by the file type implementations. Another option might be macros [ala libpd](https://github.com/libpd/libpd/blob/master/libpd_wrapper/z_libpd.c#L172).

I think it would be *relatively* easy to add A-law and u-law companding support, but I suspect the need for this is pretty low. See [example C implementation](https://dystopiancode.blogspot.com/2012/02/pcm-law-and-u-law-companding-algorithms.html).


Lucarda:
some feedback with this test patch:

[update-soundfile-tester.zip](https://github.com/pure-data/pure-data/files/4132918/update-soundfile-tester.zip)


it works Ok (win32 and linux64) on this branch and on 0.50-2. It does not cover testing raw formats or ascii.

<hr>

**May be out of scope for this thread:**

This patch also triggers some problem with [writesf~] when trying to open 2 of the recorded files on Soundforge (but they open fine on Audacity).

File 1 (24bit-mono.wav) gives this warning but it opens:

Untitled



File 2 (16bit-mono.aiff) gives this warning and opens a file full of zeros

Untitled2


The rest of the written files with [writesf~] and [soundfiler] opens correctly on Soundforge.

umlaeute:
running @Lucarda 's patch on a vanilla Pd-0.50-2, i get the following problems:

~~~sh
$ sndfile-info 16bit-mono.aiff | egrep '^\*'
*** Frame count read from 'COMM' chunk (220480) not equal to frame count
*** calculated from length of 'SSND' chunk (220476).

$ sndfile-info 16bit-stereo.aiff | egrep '^\*'
*** Frame count read from 'COMM' chunk (220480) not equal to frame count
*** calculated from length of 'SSND' chunk (220478).

$ sndfile-info 24bit-mono.aiff | egrep '^\*'
*** Chunk size 4293599231 > file length 661494. Exiting parser.
*** Frame count read from 'COMM' chunk (220477) not equal to frame count
*** calculated from length of 'SSND' chunk (220474).

$ sndfile-info 24bit-mono.wav | egrep '^\*'
*** 'data' chunk should be an even number of bytes in length.
*** Unknown chunk marker (EB60FFEB) at position 661476. Exiting parser.
*** Chunk size 4293635071 > file length 661484. Exiting parser.

$ sndfile-info 24bit-stereo.aiff | egrep '^\*'
*** Unknown chunk marker D0FF0D2C at position 1322856. Exiting parser.
*** Chunk size 788540461 > file length 1322934. Exiting parser.
*** Frame count read from 'COMM' chunk (220467) not equal to frame count
*** calculated from length of 'SSND' chunk (220465).

$ sndfile-info 24bit-stereo.wav | egrep '^\*'
*** Chunk size 3531800620 > file length 1322924. Exiting parser.

$ sndfile-info 32bitfloat-mono.wav | egrep '^\*'
**** All non-PCM format files should have a 'fact' chunk.

$ sndfile-info 32bitfloat-stereo.wav | egrep '^\*'
**** All non-PCM format files should have a 'fact' chunk.
~~~

I get similar problems when running the tests with 4e136f63f27203a0cd7eeb3588db4b304913ed03:

~~~sh
$ sndfile-info 16bit-mono.aiff | egrep '^\*'
*** Frame count read from 'COMM' chunk (220480) not equal to frame count
*** calculated from length of 'SSND' chunk (220476).

$ sndfile-info 16bit-stereo.aiff | egrep '^\*'
*** Frame count read from 'COMM' chunk (220480) not equal to frame count
*** calculated from length of 'SSND' chunk (220478).

$ sndfile-info 24bit-mono.aiff | egrep '^\*'
*** Chunk size 2017536 > file length 661494. Exiting parser.
*** Frame count read from 'COMM' chunk (220477) not equal to frame count
*** calculated from length of 'SSND' chunk (220474).

$ sndfile-info 24bit-mono.wav | egrep '^\*'
*** 'data' chunk should be an even number of bytes in length.
*** Unknown chunk marker (1B2C001C) at position 661476. Exiting parser.
*** Chunk size 1665024 > file length 661484. Exiting parser.

$ sndfile-info 24bit-stereo.aiff | egrep '^\*'
*** Unknown chunk marker 1400A7E1 at position 1322856. Exiting parser.
*** Chunk size 2298470437 > file length 1322934. Exiting parser.
*** Frame count read from 'COMM' chunk (220467) not equal to frame count
*** calculated from length of 'SSND' chunk (220465).

$ sndfile-info 24bit-stereo.wav | egrep '^\*'
*** Chunk size 277479390 > file length 1322924. Exiting parser.

$ sndfile-info 32bitfloat-mono.wav | egrep '^\*'
**** All non-PCM format files should have a 'fact' chunk.

$ sndfile-info 32bitfloat-stereo.wav | egrep '^\*'
**** All non-PCM format files should have a 'fact' chunk.
~~~

umlaeute:
it turns out that the differences between the two `sndfile-info` runs are due to misaligned phases (the test-patch doesn't reset the phase before recording a new soundfile).

so the "unknown chunk marker" error we get, is because the size of the `DATA` chunk is too small, making libsndfile attempt to read another chunk after the `DATA` chunk, but since it still finds sample data there, and misinterpretes them as unknown chunk IDs.

anyhow, this is not a regression.

however, some files produced by Pd-0.50-2 differ from ones produced with 4e136f63f27203a0cd7eeb3588db4b304913ed03:

### 32bitfloat-stereo.wav
in my test-run i get a `0x4B` vs `0x4C` in the LSB of a single sample. i guess this is some rounding error, and could be ignored (i haven't done any repetitive tests, to see whether the behaviour is consistent)

### 24bit-mono.au, 24bit-stereo.au
the size of the data-chunk is not written correctly in `0.50-0`; this has been fixed in 4e136f63f27203a0cd7eeb3588db4b304913ed03

### 32bitfloat-mono.au, 32bitfloat-stereo.au
again, the size of the data-chunk is not written correctly in `0.50-0`; this has been fixed in 4e136f63f27203a0cd7eeb3588db4b304913ed03
however, here i also got a few imprecision errors (with 4e136f63f27203a0cd7eeb3588db4b304913ed03 adding one least-significant **bit** in a total of 5(!) samples.

### 16bit-mono.wav
it seems that there was some glitch in my phase-reset :-), so this case should probably be ignored...

danomatika:
I was think I need to double check the cached format info in the read and write thread functions. I tried to simplify it by using `soundfile_info_copy()` but I'm now not 100% certain all of the values copied to and from the object instance *shoukld* be copied. I will do that now.

danomatika:
Also, many of the missing chunk warnings come from revisions to the format specs which simply aren't handled by Pd yet. I'll start on that next.

danomatika:
>I think I need to double check the cached format info in the read and write thread functions.

Just checked through and it should be ok.

danomatika:
AIFF support has now been overhauled to add support for AIFF-C which brings little endian and the 32 bit float samples. I also took the opportunity to change some variables to explicitly types based on their function, ie. buffer * byte sizes are `size_t` and file offsets are `off_t`. Making this change through out the file read/write sections should make it far easier to allow for native large file support in a pd double build.

Lucarda:
> Any feedback on implementing this a use way? Perhaps as some sort of global flag ->` ; pd debug soundfile 1` enables verbose soundfile header info printing on the fly.


The global flag seems Ok for me. The other options I can think of are no better.

danomatika:
WAVE and NEXT have been overhauled. WAVE now supports the extended format header and fact chunk, as required by the Rev. 3 spec. Both now post header info when verbose is on. AIFF chunk sizes are actually signed, and this is now handled.

I've converted byte and file offset variables to `size_t`, `ssize_t`, and `off_t`. Commenting style has been updated to keep "light doxygen" but to better conform to Miller's styling, ie. `/** blah blah */` with extra indentation.

As Pd versions before this soundfile update cannot read AIFF-C or WAVE extended format headers, writing these formats is disabled if the pd compatibility level is < 0.51. Reading works fine. This is mainly meant to make sure newer versions of Pd cannot create files that older Pd's cannot read if explicit compatibility is required.

Also, I think I fixed the 32 bit read crashes with 8c2e568.

danomatika:
Note: As Pd compatibility is now checked for writing AIFC and WAVE extended format headers, you'll have to run pd with `-compatibility 0.51` to test these features. I recommend also running with `-verbose` to see the header info posts when reading/writing the files.

Lucarda:
This is going well on my tests on Windows.

I have cleaned up my test patch and also include a "Sample-length" test. 

[soundfile-test3-for-PR-855.zip](https://github.com/pure-data/pure-data/files/4183513/soundfile-test3-for-PR-855.zip)


Now the files generated opens fine on other software (no error warnings as before) but the 24bit files (generated with [readsf~] ) have less samples than the 16 or 32 bit files. : 


..


patch1

Lucarda:
Here is an updated test-patch that includes 32bit AIFC in the tests.

[PR-855-soundfile-test-v4.zip](https://github.com/pure-data/pure-data/files/4183705/PR-855-soundfile-test-v4.zip)


Now I notice that [soundfiler]'s 2nd outlet is no longer reporting "sample-rate" for all AIFF files.

danomatika:
Thanks, I'll check it out. As Pd doesn't have any formal unit testing, patches like this might be good to add to the distribution so people with different systems can also test and troubleshoot via some standard method.

I'm working on updating and merging CAFF support currently.

umlaeute:
> As Pd versions before this soundfile update cannot read 32 bit float AIFF or AIFF-C and WAVE extended format headers, writing these formats is disabled if the pd compatibility level is < 0.51. 

i think this is misguided.
yo are using the `compatibility` flag to indicate compatibility with an *external application* (another instance of Pd, running somewhere else).
however, imo the compatibility flag is about being able to run the patch at hand within the *current* Pd-runtime, not about compatibility with external applications.

If i want to share a soundfile with an old version of Pd (or some other software), then i should refrain from writing a CAF file (on the patch level) rather than setting some obscure flag (on the patch level).

danomatika:
>If i want to share a soundfile with an old version of Pd (or some other software), then i should refrain from writing a CAF file (on the patch level) rather than setting some obscure flag (on the patch level).

CAF yes, but writing 32 bit float WAVE files *requires* the extended format chunk as per the Rev 3 spec. This happens automatically unless the compatibility mode is < 51. Before these updates, the header was being written with basic format which *works* but sendfile-info, for instance complains about them missing extended format.

danomatika:
>Now I notice that [soundfiler]'s 2nd outlet is no longer reporting "sample-rate" for all AIFF files.

I fixed this in the `feature/caff` branch.

danomatika:
CAF file support is now merged.

After feedback from @umlaeute, I removed the compatibility checks as either A. old Pd's would not write new type files (and expect to read them) or B. old Pd's were actually able to read extended format WAVE files as they ignore the format flag\*.  I documented the specific version differences in the file format implementation files.

@Lucarda AIFF [soundfiler] sample rate output is fixed.

_\*In the B situation, this would only be a problem if we supported 32 bit int samples, but Pd currently does not.\*

danomatika:
@Lucarda Regarding the number of samples for 24 it files, I'm seeing the same numbers with the current Pd release as well, so nothing has really changed regarding this so far.

danomatika:
There has been some noise on the pd-list about hopes/wishes for [soundfiler] to dump loop marker info, etc. At this point, it would be relatively trivial for me to add info chunk output. The only question I have is regarding the patch-level implementation: should it A. come out of the existing 2nd outlet via an extra read flag, ie. `[meta somefile.wav<` dumps meta chunk info, or B. come out of a new 3rd outlet which could possibly only appear from a creation flag like `[soundfiler -i]`?

Lucarda:
>  AIFF [soundfiler] sample rate output is fixed.

I get it on Linux but not on Windows.

I have updated the test-patch to use CAF files and to choose a dir for the soundfiles:

[soundfile-test-v7.zip](https://github.com/pure-data/pure-data/files/4191701/soundfile-test-v7.zip)


CAF files are working nice in Pd (I don't have other software that supports CAF to double-check the files).

> Regarding the number of samples for 24 it files, I'm seeing the same numbers with the current Pd release as well, so nothing has really changed regarding this so far.

Sure, nothing bad changed, on the contrary. I have slightly worse figures with current Pd and AIFF on Windows. :)
Anyway I wonder why 24bit files don't have the same samples as the 16/32bit.

danomatika:
>I get it on Linux but not on Windows.

Did you rebuild after pulling the latest commits? I'm not sure how this would work on 1 OS and not another as there should be nothing OS-specific going on.

>(I don't have other software that supports CAF to double-check the files).

Try VLC.

Lucarda:
> Did you rebuild after pulling the latest commits?

I build it again today until a600c50 and the issue[*] was still there **but** it worked OK if I start Pd _without_ the `-verbose` flag.



[*] [soundfiler]'s second outlet not reporting "sample rate" for AIFF/C files.

Lucarda:
I have installed libsndfile with the win32 installer from their web page and using it inside MSYS (same results as using it with CMD).

I get this troubles with all the CAF files (note the nice results with an AIFF file):

```

Lucarda@T410 MSYS /c/Program Files (x86)/Mega-Nerd/libsndfile/bin
$ ./sndfile-info.exe D:/00-volatil/sbsf/v7/16bit-mono.caf
Error : Not able to open input file D:/00-volatil/sbsf/v7/16bit-mono.caf.
File : D:/00-volatil/sbsf/v7/16bit-mono.caf
Length : 441028
caff
  Version : 1
  Flags   : 0
desc : 44
  Sample rate  : 44100.000
  Format id    : lpcm
  Format flags : 0
  Bytes / packet   : 2
  Frames / packet  : 1
  Channels / frame : 1
  Bits / channel   : 16
Have 0 marker at position 76 (0x4C).
**** Error, could not find 'data' chunk.

Supported file format but file is malformed.

Lucarda@T410 MSYS /c/Program Files (x86)/Mega-Nerd/libsndfile/bin
$ ./sndfile-info.exe D:/00-volatil/sbsf/v7/16bit-mono.aiff
========================================
File : D:/00-volatil/sbsf/v7/16bit-mono.aiff
Length : 441014
FORM : 441006
 AIFF
 COMM : 18
  Sample Rate : 44100
  Frames      : 220480
  Channels    : 1
  Sample Size : 16
  Encoding    : NONE
 SSND : 440968
  Offset     : 0
  Block Size : 0

----------------------------------------
Sample Rate : 44100
Frames      : 220480
Channels    : 1
Format      : 0x00020002
Sections    : 1
Seekable    : TRUE
Duration    : 00:00:05.000
Signal Max  : 32767 (-0.00 dB)


```
I also installed the latest VLC but it refuses to open the CAF files. I'm not sure if it supports CAF files.

Lucarda:
I'm also having the same problems with **sndfile-info** on Linux:
```

lucarda@debian10qt:~/Downloads/v7$ sndfile-info /home/lucarda/Downloads/v7/16bit-mono.aiff
========================================
File : /home/lucarda/Downloads/v7/16bit-mono.aiff
Length : 441014
FORM : 441006
 AIFF
 COMM : 18
  Sample Rate : 44100
  Frames      : 220480
  Channels    : 1
  Sample Size : 16
  Encoding    : NONE
 SSND : 440968
  Offset     : 0
  Block Size : 0

----------------------------------------
Sample Rate : 44100
Frames      : 220480
Channels    : 1
Format      : 0x00020002
Sections    : 1
Seekable    : TRUE
Duration    : 00:00:05.000
Signal Max  : 32767 (-0.00 dB)

lucarda@debian10qt:~/Downloads/v7$ sndfile-info /home/lucarda/Downloads/v7/16bit-mono.caf
Error : Not able to open input file /home/lucarda/Downloads/v7/16bit-mono.caf.
File : /home/lucarda/Downloads/v7/16bit-mono.caf
Length : 441028
caff
  Version : 1
  Flags   : 0
desc : 44
  Sample rate  : 44100.000
  Format id    : lpcm
  Format flags : 0
  Bytes / packet   : 2
  Frames / packet  : 1
  Channels / frame : 1
  Bits / channel   : 16
Have 0 marker at position 76 (0x4C).
**** Error, could not find 'data' chunk.

Supported file format but file is malformed.


lucarda@debian10qt:~/Downloads/v7$ sndfile-info /home/lucarda/Downloads/v7/soundfiler-32bitfloat-stereo.caf
Error : Not able to open input file /home/lucarda/Downloads/v7/soundfiler-32bitfloat-stereo.caf.
File : /home/lucarda/Downloads/v7/soundfiler-32bitfloat-stereo.caf
Length : 8068
caff
  Version : 1
  Flags   : 0
desc : 44
  Sample rate  : 44100.000
  Format id    : lpcm
  Format flags : 1
  Bytes / packet   : 8
  Frames / packet  : 1
  Channels / frame : 2
  Bits / channel   : 32
Have 0 marker at position 76 (0x4C).
**** Error, could not find 'data' chunk.

Supported file format but file is malformed.
```

danomatika:
>I build it again today until a600c50 and the issue[*] was still there but it worked OK if I start Pd without the -verbose flag.
>
>[*] [soundfiler]'s second outlet not reporting "sample rate" for AIFF/C files.

Ok, I think I've fixed it. The sample rate bytes were getting changed when it was being read.

Lucarda:
> Ok, I think I've fixed it. The sample rate bytes were getting changed when it was being read.

Yep, it works now with or without the `-verbose` flag.

danomatika:
There was indeed a problem with the CAF desc chunk size. This is now fixed and VLC and sendfile-info should be bale to read the files created by Pd.

Lucarda:
Now CAF files are played nice with VLC.

Output from sndfile-info is perfect for CAF files created with [soundfiler] but there is a message for all the [readsf~] files : "Request for header allocation of  "some number" denined.".

full output: 
[sndfile-info-caf-1.txt](https://github.com/pure-data/pure-data/files/4198836/sndfile-info-caf-1.txt)

danomatika:
>"Request for header allocation of "some number" denined."

I have seen that too, but I think it's a warning and/or a bug in their output. I double checked the file output in a hex editor the first time I saw the print, but the files are fine.

danomatika:
Wit the latest commits, the basic backend architecture is now back ported from the `feature/soundfile-backend` branch. This is the minimal `t_soundfile_type` and `t_soundfile` implementation which makes the type handling much more elegant. I've also brought back various improvements such as soundfiler ascii write, error strings, verbose debug posts now enabled by compile time defile, and various other smaller bug fixes and improvements.

This has the *minimal* type framework required by the existing code without flexibility for extended/external "plugins." After working on the other branch for a while, I feel that needs some more testing and work while what is here is mainly refactoring of the existing code base.

I have a few details left over from the soundfiler ascii work, but in the meantime I think this is ready for a look.

danomatika:
The soundfiler ascii read/write has now been promoted to a "full" file type with as associated .txt extension and supports the skip and sample length flags. It basically works like diff, wave, caf, and next files, either implicitly through the file extension (file.txt) or via the explicit "-ascii" flag. The "-raw" flag will override it as with the other types.

@Lucarda You could update you test patch to read/write ascii txt files with soundfiler, although I would recommend making the table length short, ie. not 5 seconds.

One useful idea for the ascii format: an optional header with the number of channels and number of frames. This would make it easier to work with these files to & from other software.

What we currently have is:

~~~
f f f f f f f f...
~~~

My thought was something like:

~~~
#A headersize channels frames f f f f f f f...
~~~

Once the file is read via the binbuf, we can simply check if the "#A" symbol is there. The headersize is how many atoms follow before the data starts and allows for optionally extending the "format" in the future.

danomatika:
Ok, I think this is ready for a round of testing. Other than the ascii header idea, I think this branch has pretty much all that it needs now. I will leave the larger questions to other the WIP branch(es).

Lucarda:
Pd-w64 built as of 09b10f7

2 things about -ascii:

- If read a 12 values txt into a 100 points array some junk is added afterwards point 12 in the array. (this does not happen in current master)

- If i do `[write -ascii foo.txt fooarray( `the file gets written in Pd's directory instead of the patch dir.

here's a test-patch:

[ascii-2-things.zip](https://github.com/pure-data/pure-data/files/4296450/ascii-2-things.zip)

Lucarda:
Also this patch:

[ascii-mem-leak.zip](https://github.com/pure-data/pure-data/files/4297517/ascii-mem-leak.zip)


triggers garbage.

danomatika:
Latest commit should fix ascii reading bad values after num atoms in file and writing to canvas dir.

danomatika:
Also, this branch handles .txt natively as ascii so the "-ascii" flag is not needed with a filename + extension, ie. "a.txt" or "aw.txt"

Lucarda:
Two things:

- I'm not sure if the current behavior of combining `-skip` & `-resize` is the expected one. See test patch:

[ascii-9.zip](https://github.com/pure-data/pure-data/files/4315007/ascii-9.zip)


- This is just a question. Does this overhaul will truncate double-precision arrays when writing a 32bit float file?

Lucarda:
@danomatika 

Based in all of what you were doing in this overhaul, I think you can set an 64bit-floating-point audio file format spec in less than 30 minutes.

:)

danomatika:
@Lucarda That was part of the idea, but I held off on additional sample conversion. However, if you compile Pd with double precision, then it would be possible to use 64 bit sample indices with CAF files, which allows for working with HUGE files.

Lucarda:
> You could update you test patch to read/write ascii txt files with soundfiler, although I would recommend making the table length short, ie. not 5 seconds.

@danomatika  I have updated the test patch to :

[tester9.zip](https://github.com/pure-data/pure-data/files/4329479/tester9.zip)

danomatika:
@Lucarda I don't quite get what you mean with:

>I'm not sure if the current behavior of combining -skip & -resize is the expected one. See test patch

Can you explain?

Lucarda:
> Can you explain?

1) I naively expect that if I have a 100 points txt and I do` [read -skip 25 -resize foo.txt array(` I'll get an 75 points array. This subtraction can be done at the patch level but may be it can be resolved in the C side?



2) I get some strange behaviors randomly (I think its triggered sometimes when I write an ascii). If I do `[read -skip (say 50) & -resize(` a 100 points txt I could get any of this 3 states in the array:

- The 50 points and 50 more zeros. (may be this is the expected behavior)

- The 50 points and less than 50 zeros plus some garbage.

- the 50 points and again the 50 points

Testing on Windows64

danomatika:
Ok, I see what you mean. I didn't understand that from the patch itself.

danomatika:
Latest commit should fix ascii read -skip behavior. The onset frame (ie. skip) wasn't beefing used to calculate the resize length, plus there was an error in calculating the first frame with more than 1 table.

Lucarda:
As of c325b22 all very nice. I have updated the test patch that now includes -flags testing on all file formats:

[PR-855-tester-11.zip](https://github.com/pure-data/pure-data/files/4336786/PR-855-tester-11.zip)


So far all good except 2 things when writing ascii:

- If I do `write -skip 10 foo.txt array` on a 50 points array: I get a 30 values file starting with the correct initial skip but also the last 10 values are missing. (if I `-skip 15` I also miss the last 15 values).

- If I do `write -skip 10 -nframe 20 foo.txt array` I get an empty file. Seems the 2 flags can't be used combined.

This two issues are shown in the patch.

Lucarda:
I have corrected some bad information i wrote in the comments of the ascii section and included  tests for [readsf~] with onset flags.

[PR-855-tester-12.zip](https://github.com/pure-data/pure-data/files/4368105/PR-855-tester-12.zip)

danomatika:
This branch is up to date with 0.51-0.

danomatika:
I will check over this again and then I would submit it for Pd 0.52. This would help at work who basically paid for my time to make this PR, among other things, as we need CAF file support and improved WAV, AIFC, etc.

danomatika:
There is an issue with CAF that I am looking into now.

danomatika:
CAF issue fixed with replacing a `memcpy` with `memmove`. This was not a problem with debug builds but crept back up after optimization.

danomatika:
@Lucarda The ascii write issues should be fixed now.

Lucarda:
> @Lucarda The ascii write issues should be fixed now.

Yes it is working Ok. Everything is working Ok with this test patch:

[audio-file-objects-tester.zip](https://github.com/pure-data/pure-data/files/5061422/audio-file-objects-tester.zip)


It mostly test everything. Should it go to `doc/7.stuff/tools` ?

danomatika:
@umlaeute Is the Travis build failing due to a misconfiguration?

umlaeute:
afaict this is because brew is updated much faster than the travis-ci images.

danomatika:
> afaict this is because brew is updated much faster than the travis-ci images.

Ok, so it's not an actual build error.

umlaeute:
actually, on closer inspection i think the build-failure was due to https://github.com/pure-data/pure-data/issues/1102 which was fixed with b9d3bd0e5f705572e60c7760a196bb3cbf3fa7fd (so the PR needs a recent enough base branch to succeed).

let's hope for the best with the [current build](https://travis-ci.org/github/pure-data/pure-data/jobs/731557038).

in any case: it's not an actual compiler build error.

danomatika:
Up to date with 0.51-3. Still ready to go.

danomatika:
@Spacechild1 Can you quickly check e6c2ec2c94e48174ef6c68848066128cc9cbf49f. I believe I successfully ported it, but I don't have an easy way to test that it works.

Spacechild1:
The commit looks ok. I spotted some minor things in the refactoring of the child threads. Will make a PR to your branch.

danomatika:
Please do.

danomatika:
Latest commit ports #1227 fix.

Issue #856, opened 2019-12-31
undo could also work with properties window


jmmmp:
The recent undo changes are very helpful. It would be even more interesting if the settings changes made in all properties windows would be accounted for.
Here is what I could find:
- undo works for canvas properties (patch, not cnv) if used with right-click on the background, the entry "undo apply" appears
- it doesn't work for the same dialogue if right-clicked on a subpatch e.g. [pd test], there is no undo entry either in the main patch or in the subpatch
- all GUI properties don't count for undo.
(- undo doesn't work with messages sent to the canvas (clear, add objects, ...) - but I myself don't have a problem with that)

umlaeute:
> all GUI properties don't count for undo.

that's a side-effect of `undo` not undoing the internal state of arbitrary objects.

i don't see any proper solution for this and am inclined to close this as `wontfix`

umlaeute:
> it doesn't work for the same dialogue if right-clicked on a subpatch

this i cannot reproduce. i get an `Undo` action on the subpatch-canvas (just as i get when calling the dialog from the subpatch-canvas itself).

this is how i do it:
- start with a patch that contains an empty `[pd test]` subpatch.
- right-click on `[pd test]` and select `Properties`
- enable the `Graph-on-Parent` flag and click <kbd>Apply</kbd>
  - the `[pd test]` is now a GOP
- open `[pd test]` (right-clicking it and selecting `Open`)
- call `undo` with <kbd>Control</kbd>+<kbd>z</kbd> on the `test` canvas.
  - the `[pd test]` is back to normal (non-GOP)


this is with Pd-0.50.2 on Debian/GNU linux.

jmmmp:
What I meant to say was to click inside the subpatch - it seems that operations in the parent canvas are saved, but in the sub-patches they aren't.

Regarding object settings:
If undoing the internal state of arbitrary objects couldn't be done for technical reasons, I would have no problem to say "there are more important things to do, and it's unsolvable undtil tck/tk improves or [insert issue here]". But I don't see a difference between "patching" and "not-really-serious-patching", specially when:
- all objects have send and receive settings which are vital for a patch to work (it's illogical to consider patching to add a [r xxx] above the object, but not inside it, although both have the same result)
- when preparing a GUI, all the internal "aesthetic" settings are as important as any other. E.g. it's considered "patching" to move an object around so that it doesn't clash with other objects, but it's not patching if the size of that object is improved (which has the same effect in the end, not to clash with other objects)

[...]

PR #858, opened 2020-01-03
undo fixes: UNDO for internal states (targeting GUI objects), UNDO for zoomed motion

### LINKED ISSUE #861: bug when you move objects in zoom mode and undo
- go into zoom mode
- move an object
- do undo
- object ends up in place it wasn't before

### LINKED ISSUE #1249: font undo broken
steps to reproduce:
1. open a simple patch, like this one:
   ~~~
   #N canvas 975 638 450 300 12;
   #X obj 281 151 f;
   ~~~
2. change the font
   - <kbd>Edit</kbd> -> <kbd>Font</kbd>
   - click on `36` and then <kbd>OK</kbd>
 2. notice (the expected result) that:
   - the object is at the same position
   - the object is now bigger
   ~~~
   #N canvas 975 640 450 300 36;
   #X obj 281 151 f;
   ~~~

3. undo what you just did: <kbd>Ctrl</kbd>+<kbd>z</kbd>
4. notice that
   - the object is gone to somewhere way off-screen.
   - it still uses the new font-size
   ~~~
   #N canvas 975 640 450 300 36;
   #X obj 28100 15100 f;
   ~~~

5. undo once more: <kbd>Ctrl</kbd>+<kbd>z</kbd>

6. notice that
  - the object moved again, but is still way off-screen
  - the object is now back to the original font
    ~~~
    #N canvas 975 640 450 300 12;
    #X obj -8048 2672 f;
    ~~~

umlaeute:
# bug fixes

## undo for zoomed motion

this PR fixes #861 

## undo for font-changes

this PR fixes #1249 

# undo for *object*-states

this PR attempts to resolved the issue outlined in #856 by providing a generic `undo` function for *object*-states.

## How it works
if an object would like make an internal state-change `undo`able (e.g. a GUI-object with a properties-dialog), it can use `pd_undo_set_objectstate()` to hook into the UNDO system.

this function takes:
 - handles to the canvas and the object
 - the selector for the object's state-change message
- a message (atom-list) to restore the previous state (aka: "undo")
- a message (atom-list) to restore the target state from the previous state (aka: "redo")

goodies:
- `pd_undo_set_objectstate()` detects whether it is called during undo/redo and will not do anything in this case (so if your object has a single state-change method where you call `pd_undo_set_objectstate()`, you can use this very method to undo/redo the state-changes without worrying about recursive undo...
- if the <kbd>undo</kbd> and <kbd>redo</kbd> messages are identical, it is assumed that the state is not actually changed and no undo-action is recorded (e.g. if you open a properties dialog of the object and just clock <kbd>Apply</kbd> without changing anything)


Spacechild1:
I think you need to store the *index* of the object within the glist, not the pointer to the object itself, because it can easily become stale, e.g. after cut + undo. AFAICT, the positions of objects within the glist are always preserved across undo/redo operations.

umlaeute:
> I think you need to store the _index_ of the object within the glist, not the pointer to the object itself

yes.
i just forgot to push that :-) ab73e5152560872b0ec3faf60f9f2023555816d2

umlaeute:
@Spacechild1, @Lucarda, @danomatika, @porres if you have spare time, please test the new undo-fixes (esp. on non-linux platforms).

Spacechild1:
will test next week.

Lucarda:
I have tested this on Windows64 but it behaves the same as current master.

1- Create a [vradio] (edit now says : `Undo motion`).
2- Open [vradio] properties and make it have 3 cells, hit apply, hit Ok. (edit still says : `Undo motion`).
3- When I hit Undo the object is moved but it still has 3 cells.

Same happens if I create a [cnv] and change its color. (I get the `Undo motion` but the color doesn't change back to the previous state.)

umlaeute:
doh!

so far, i have only fixed iemgui's properties and ordinary object's zoomed-motion, but not iemgui's zoomed-motion.

umlaeute:
double doh! I completely misunderstood the feedback of @Lucarda (thinking that iemgui's zoomed-motion undo was not working; and to add insult to injury, i was even able to reproduce this non-issue :woman_facepalming:)

anyhow, the problem with iemgui's properties not being undoable came from a small but buggy optimisation, that tried to prevent adding a `state`-undo if the *undo*- and *redo*-statechange messages where identical (the bug was, that only the first atom of the statechange messages were compared; leading Pd to discard the undo-action).

should be fixed with 9b5f454.
please try again.

Lucarda:
Tested as of 9bc6d20:

"Undo for zoomed motion" & "undo for object-states" is working as expected. I have tested some [pdgui]'s & [iemgui]'s.

Notably it doesn't work for arrays' (shift-ctrl-a) canvas properties. For example if I change "x range size" to 20, then this can't be undone.

umlaeute:
thanks for noticing.

while implementing undo for *Array Properties*, it turned out (of course!), that arrays are similar to GOP-subpatches. this basically means that whichever changes you do via the array properties, can be undo - but only after *open*ing the array-window (<kbd>LeftClick</kbd>+*Open*), and running <kbd>Undo</kbd> *within* this window.

:warning: while i think this is counterintuitive, it's also consistent with how you interact with *Canvas Properties* in general.
also, creating a workaround for the array-properties is probably just too much work :-(

you have been warned.

Lucarda:
>  ... that arrays are similar to GOP-subpatches ...

Ok then. :)

This PR is working Ok on Windows.

:)

umlaeute:
@millerpuckette i think this is probably ready for inclusion in Pd-0.52

porres:
I wonder what I have to do to make this also work for GUI externals in cyclone and ELSE

umlaeute:
@porres probably best to check the changes done to the gatoms to make their state undoable: https://github.com/pure-data/pure-data/pull/858/commits/8c61484b83aa5f9fe39a8931c37c86ab9546059f



it boils down to:
- consider you have a callback method for receiving the updated GUI-properties from pd-gui (bound to e.g. a `dialog` message)
- in this callback method, the *first* thing you do is to
  - create a message that - when sent to the object - restores the *current* state (let's call it *undo-message*)
  - use `pd_undo_set_objectstate` to add this undo-message *and* the current message to  to the undo-queue for your object
- after that, proceed as normal (apply the changed state)

## example

consider an object `[swatch]` that shows a single colour.
when invoking the properties for this object, the dialog would send back a message `[dialog <s:colour> <s:label>(` (e.g. `[dialog #ff0000 foobar(`) to the `[swatch]` instance.

the normal code looks like:

~~~C
static void swatch_dialog(t_swatch*x, t_symbol*s, int argc, t_atom*argv) {
   x->x_color = atom_getsymbolarg(0, argc, argv);
   x->x_label = atom_getsymbolarg(1, argc, argv);
   // redraw the object
}
~~~

to make this undoeable, you would change this to the following (with `x->x_parentcanvas` being the canvas where the object resides in):

~~~C
static void swatch_dialog(t_swatch*x, t_symbol*s, int argc, t_atom*argv) {
   t_atom undomsg[2];
   SETSYMBOL(undomsg+0, x->x_color);
   SETSYMBOL(undomsg+1, x->x_label);
   pd_undo_set_objectstate(x->x_parentcanvas, (t_pd*)x, gensym("dialog"), 2, undomsg, argc, argv);

   x->x_color = atom_getsymbolarg(0, argc, argv);
   x->x_label = atom_getsymbolarg(1, argc, argv);
   // redraw the object
}
~~~

umlaeute:
btw, this is documented (better than most things) in https://github.com/pure-data/pure-data/blob/98d26fa161a20cb4287e7ed6c5da040216432666/src/m_pd.h#L904-L913

PR #869, opened 2020-01-15
Kbdnav

HenriAugusto:
This is a prototype branch that allows you to fully navigate and connect your patches with the keyboard. Here is what's new:

- Navigate through objects in/outlets and connections with the keyboard
- New objects are placed accordingly to selected in/outlet
- Enter/exit editing the text of an object
- "goto" - select any object in the patch
- Canvas console - used to send messages to the current canvas (see uses below)
- Create connections with the keyboard
   - _magnetic connector_: create a connection in ascending order of distance
   - _digit connector_: press a key from 0 to 9 to connect to any of the 10 closest in/outlets
   - canvas console - send "connect" messages to the patch
      - pd fills the message for you with the arguments for the selected object
      - the gui automatically changes _connect_ to _connect_with_undo_
- Simulate a click in an object with the keyboard
   - used to click gui objects like [bng] and [tgl]
   - used to open subpatches
- Show object indexes (useful for the goto and sending connect msgs)
- Help patch in the docs
   - `doc/7.stuff/keyboard.navigation/keyboard.patching-help.pd`

I've tried to keep it simple and lightweight. Most of the navigation is done with ctrl+arrow keys.

Testing is needed in all platforms.

keyboard navigation demo


HenriAugusto:
Currently build tests are failing. Seems to be a problem with the `#define HAVE_KEYBOARDNAV` conditional compilation which is giving linking problems. How can it be fixed?

porres:
I like the idea of showing index of objects cause it is also useful for dynamic patching (knowing the index to make connections)

porres:
- control + arrows not really working on macOS
- I don't think that emulating a bang is that useful for GUIs, like shift+return on a bang

HenriAugusto:
<kbd>ctrl</kbd>+<kbd>arrows</kbd> were being capture by the OS so now <kbd>Cmd</kbd> is the kbdnav modifier on OSX.

Still we should change <kbd>Cmd</kbd>+<kbd>Tab</kbd> for displaying obj indexes.

I'm thinking <kbd>Cmd</kbd>+<kbd>i</kbd> on OSX and <kbd>Ctrl</kbd>+<kbd>i</kbd> on Win.

src/g_editor.c
```diff
@@ -1899,6 +1921,20 @@ static void canvas_rightclick(t_canvas *x, int xpos, int ypos, t_gobj *y)
 
 /* ----  editors -- perhaps this and "vis" should go to g_editor.c ------- */
 
+#ifdef HAVE_KEYBOARDNAV
+static void initialize_kbdnav(t_kbdnav *x){
```
Spacechild1: this could go to `g_keyboardnav.c`

src/g_editor.c
```diff
@@ -11,10 +11,29 @@
...
+#include <math.h>
+#include <stdlib.h>
+#include "g_kbdnav.h"
 #ifdef _MSC_VER  /* This is only for Microsoft's compiler, not cygwin, e.g. */
 #define snprintf _snprintf
 #endif
 
+typedef struct _editor_private {
+#ifdef HAVE_KEYBOARDNAV
+    struct _kbdnav kbdnav; /*keyboard navigation*/
```
Spacechild1: since you don't directly access the members of kbdnav, this could be `struct _kbdnav * kbdnav`. This would help encapsulation because you don't expose the internals of your structure.

src/g_editor.c
```diff
@@ -1920,6 +1962,12 @@ static void editor_free(t_editor *x, t_glist *y)
     binbuf_free(x->e_deleted);
     if (x->e_clock)
         clock_free(x->e_clock);
+    t_editor_private *private = x->e_privatedata;
+    freebytes(private, sizeof(*private));
+#ifdef HAVE_KEYBOARDNAV
+        t_kbdnav *kbdnav = &(private->kbdnav);
+        freebytes((void *)kbdnav->kn_linetraverser, sizeof(*kbdnav->kn_linetraverser));
```
Spacechild1: could be function `kbdnav_free`

src/g_editor.c
```diff
@@ -1909,6 +1945,12 @@ static t_editor *editor_new(t_glist *owner)
     sprintf(buf, ".x%lx", (t_int)owner);
     x->e_guiconnect = guiconnect_new(&owner->gl_pd, gensym(buf));
     x->e_clock = 0;
+    t_editor_private *private = 0;
+    private = getbytes(sizeof(*private));
+    x->e_privatedata = private;
+#ifdef HAVE_KEYBOARDNAV
+    initialize_kbdnav(&private->kbdnav);
```
Spacechild1: If `kbdnav` was a pointer (see https://github.com/pure-data/pure-data/pull/869/files#r367295268), this would have to be something like `kbdnav_new `which also allocates the memory for the `kbdnav` struct itself.

src/g_text.c
```diff
@@ -214,7 +255,33 @@ void canvas_obj(t_glist *gl, t_symbol *s, int argc, t_atom *argv)
         pd_vmess(&gl->gl_pd, gensym("editmode"), "i", 1);
         canvas_objtext(gl, xpix, ypix, 0, 1, b);
         if (connectme)
+        {
+#ifdef HAVE_KEYBOARDNAV
```
Spacechild1: could be a function in `g_kbdnav.c`

src/g_text.c
```diff
@@ -498,7 +565,33 @@ void canvas_msg(t_glist *gl, t_symbol *s, int argc, t_atom *argv)
         glist_select(gl, &x->m_text.te_g);
         gobj_activate(&x->m_text.te_g, gl, 1);
         if (connectme)
+        {
+#ifdef HAVE_KEYBOARDNAV
```
Spacechild1: see https://github.com/pure-data/pure-data/pull/869/files#r367300632

src/g_text.c
```diff
@@ -978,7 +1071,33 @@ void canvas_atom(t_glist *gl, t_atomtype type,
         glist_noselect(gl);
         glist_select(gl, &x->a_text.te_g);
         if (connectme)
+        {
+#ifdef HAVE_KEYBOARDNAV
```
Spacechild1: see https://github.com/pure-data/pure-data/pull/869/files#r367300632

src/g_text.c
```diff
@@ -157,10 +163,45 @@ static void canvas_howputnew(t_canvas *x, int *connectp, int *xpixp, int *ypixp,
         for (g = x->gl_list, nobj = 0; g; g = g->g_next, nobj++)
             if (g == selected)
         {
+#ifdef HAVE_KEYBOARDNAV
```
Spacechild1: could be a function in g_kbdnav.c which returns 1 or 0. Then you can do:

```
#ifdef HAVE_KEYBOARDNAV
if (!kbdnav_trytodowhatever())
#else 
{            
    gobj_getrect(g, x, &x1, &y1, &x2, &y2);
    indx = nobj;
    *xpixp = x1 / x->gl_zoom;
    *ypixp = y2  / x->gl_zoom + 5.5;    /* 5 pixels down, rounded */
}
#endif

File: src/Makefile.am [hunk 9]
```diff
@@ -66,6 +66,7 @@ pd_SOURCES_core = \
     g_hdial.c \
     g_hslider.c \
     g_io.c \
+    g_kbdnav.c \
```
umlaeute: as mentioned before: `g_kbdnav.h` must be added to `noinst_HEADERS`.
And you must include the compilation of `g_kbdnav.c` in `src/maefile.*`

src/g_editor.c
umlaeute: :+1: I agree with @Spacechild1

src/g_bang.c
```diff
@@ -75,6 +77,9 @@ void bng_draw_new(t_bng *x, t_glist *glist)
              (strcmp(x->x_gui.x_lab->s_name, "empty") ? x->x_gui.x_lab->s_name : ""),
              x->x_gui.x_font, x->x_gui.x_fontsize * IEMGUI_ZOOM(x), sys_fontweight,
              x->x_gui.x_lcol, x);
+#ifdef HAVE_KEYBOARDNAV
```
umlaeute: the `HAVE_KEYBOARDNAV` is totally pointless, as it only clutters the code and adds more compile-paths, making it much more difficult to find errors.

if you this is an attempt to keep binary-compatibility, then it didn't work
- you've already broken the (private) ABI by adding a new member to the `t_editor` struct. But now that this is done, you can add whatever you want to `t_editor_private` **without** having to worry about ABI.
- even if `HAVE_KEYBOARDNAV` had an impact on the ABI, this would be bad design. we don't want to ABI to change depending on the compile-time options one choses (but luckily the define doesn't change the ABI anyhow)
- what's the point of a compile-time option that cannot be changed? if you want it settable, you should also add an option to the autotools system (but please don't)

so: please get rid of all the `#ifdef HAVE_KEYBOARDNAV` protectors.

src/g_canvas.c
```diff
@@ -687,6 +693,11 @@ void canvas_map(t_canvas *x, t_floatarg f)
             if (x->gl_isgraph && x->gl_goprect)
                 canvas_drawredrect(x, 1);
             sys_vgui("pdtk_canvas_getscroll .x%lx.c\n", x);
+#ifdef HAVE_KEYBOARDNAV
+            t_kbdnav *kbdnav = canvas_get_kbdnav(x);
```
umlaeute: make sure to conform to C89 (sic!) and declare your variables at the beginning of the code-block.

src/g_editor.c
```diff
@@ -2580,6 +2651,69 @@ static void canvas_doclick(t_canvas *x, int xpos, int ypos, int which,
...
+        return;
+    }
+    canvas_redraw(x);
+    if( !glist_isselected(x, gobj) )
+    {
+        t_editor *editor = x->gl_editor;
+        if(editor)
+        {
+            t_editor_private *private = editor->e_privatedata;
+            t_kbdnav *kbdnav = &private->kbdnav;
```
umlaeute: *always use `canvas_get_kbdnav()` instead of directly accessing the è_privatedata`.
(there are multiple occurences of this anti-pattern in the code)

src/g_editor.c
```diff
@@ -4831,6 +4987,12 @@ void g_editor_setup(void)
         A_GIMME, A_NULL);
     class_addmethod(canvas_class, (t_method)canvas_motion, gensym("motion"),
         A_FLOAT, A_FLOAT, A_FLOAT, A_NULL);
+#ifdef HAVE_KEYBOARDNAV
+    class_addmethod(canvas_class, (t_method)canvas_goto, gensym("goto"),
+        A_FLOAT, A_NULL);
+    class_addmethod(canvas_class, (t_method)canvas_objindexes, gensym("objindexes"),
+        A_FLOAT, A_NULL);
```
umlaeute: the plural of "index" is "indices".
also, the name is misleading (it doesn't get nor set the object-indices). a better name is probably "showindices" (iirc, it also shows the indices of non-objects, like commens, messageboxes,...)

src/g_editor.c
```diff
@@ -4909,6 +5076,12 @@ void canvas_editor_for_class(t_class *c)
         A_GIMME, A_NULL);
     class_addmethod(c, (t_method)canvas_motion, gensym("motion"),
         A_FLOAT, A_FLOAT, A_FLOAT, A_DEFFLOAT, A_NULL);
+#ifdef HAVE_KEYBOARDNAV
+    class_addmethod(c, (t_method)canvas_goto, gensym("goto"),
+        A_FLOAT, A_NULL);
+    class_addmethod(canvas_class, (t_method)canvas_objindexes, gensym("objindexes"),
```
umlaeute: again: "objindexes" -> "showindices"

src/g_editor.c
```diff
@@ -4894,6 +5056,11 @@ void g_editor_setup(void)
 
     class_addmethod(canvas_class, (t_method)canvas_disconnect,
         gensym("disconnect"), A_FLOAT, A_FLOAT, A_FLOAT, A_FLOAT, A_NULL);
+#ifdef HAVE_KEYBOARDNAV
+/* --------------- to be used with the console ----------------------- */
+    class_addmethod(canvas_class, (t_method)canvas_connect_with_undo,
+        gensym("connect_with_undo"), A_FLOAT, A_FLOAT, A_FLOAT, A_FLOAT, A_NULL);
```
umlaeute: i'm not sure whether an undoable canvas-message should be exposed to the user (even if you mask it).
sending messages to the canvas is low-level and in general not undoable. introducing it here, will only make people ask why they can undo when sending the message from the canvas-prompt but not when sending it directly.

src/g_kbdnav.h
```diff
@@ -0,0 +1,165 @@
...
+#define KN_CONN_SELECTED 2 /* navigating through an in/outlet's connections */
+#define KN_WAITING_NUMBER 3 /* waiting for a number from 0-9 when using the "digit connector" */
+
+#define KBDNAV_SEL_PAD_X 4
+#define KBDNAV_SEL_PAD_Y 3
+
+#define IO_INLET 0
+#define IO_OUTLET 1
+
+/* FORWARD DECLARATIONS */
```
umlaeute: a header is not really the place to do "forward declarations" (or rather: it is the place to do them; so why do you say so?)

src/g_kbdnav.h
```diff
@@ -0,0 +1,165 @@
...
+
+/* stuff from g_editor.c */
+
+EXTERN void canvas_connect_with_undo(t_canvas *x, t_float index1, t_float outno,
```
umlaeute: these functions shouldn't be declared as `EXTERN`: this is a private API and we don't want externals to use it.

src/g_kbdnav.h
```diff
@@ -0,0 +1,165 @@
...
+                                         t_float index2, t_float inno);
+EXTERN void canvas_disconnect_with_undo(t_canvas *x, t_float index1, t_float outno,
+                                         t_float index2, t_float inno);
+EXTERN void canvas_reselect(t_canvas *x);
+EXTERN void canvas_clearline(t_canvas *x);
+
+
+/* g_editor.c */
```
umlaeute: these generally useful functions should probably just go into `g_canvas.h` (and be marked as `EXTERN`).

(the `kbdnav_*`-functions should stay here and be kept private)

tcl/pd_bindings.tcl
```diff
@@ -87,7 +87,9 @@ proc ::pd_bindings::global_bindings {} {
     bind all <$::modifier-Shift-Key-I> {menu_send %W hradio}
     bind all <$::modifier-Shift-Key-L> {menu_clear_console}
     bind all <$::modifier-Shift-Key-M> {menu_message_dialog}
+    bind all <quotedbl> {::entry_canvasmsg::openclose_canvasmsg $::focused_window}
```
umlaeute: note that the `"` (double-quote) is found on very different places on various keyboards.
On my German keyboard, it's definitely **not** on <kbd>Shift</kbd>+<kbd>'</kbd>

tcl/pd_bindings.tcl
```diff
@@ -268,13 +271,6 @@ proc ::pd_bindings::patch_bindings {mytoplevel} {
...
     }
 
-    # <Tab> key to cycle through selection
-    bind $tkcanvas <KeyPress-Tab>        "::pd_bindings::canvas_cycle %W  1 %K %A 0 %k"
-    bind $tkcanvas <Shift-Tab>           "::pd_bindings::canvas_cycle %W -1 %K %A 1 %k"
-    # on X11, <Shift-Tab> is a different key by the name 'ISO_Left_Tab'...
-    # other systems (at least aqua) do not like this name, so we 'catch' any errors
-    catch {bind $tkcanvas <KeyPress-ISO_Left_Tab> "::pd_bindings::canvas_cycle %W -1 %K %A 1 %k" } stderr
-
     # window protocol bindings
```
umlaeute: why are they removed?

doc/7.stuff/keyboard.navigation/keyboard.patching-help.pd
```diff
@@ -0,0 +1,236 @@
...
+#X text 33 37 You can now enter/exit the object's text editing but
+using mod+return.;
+#X restore 57 403 pd editing_objects;
+#N canvas 194 97 435 301 the_canvas_message_console 0;
+#X text 32 64 It allows to send messages to the current patch window.
+;
+#X text 30 147 This is the same as the following:;
+#X obj 43 211 s \$0-thisCanvas;
+#X obj 42 248 namecanvas \$0-thisCanvas;
+#X text 32 26 Press Ctrl + ' (single quote) to open the canvas message
```
umlaeute: see my comment below (in `pd_bindings.tcl`): many keyboard layouts will not have doublequote on <kbd>Shift</kbd>+<kbd>'</kbd>!

Spacechild1:
@HenriAugusto if you need help with the reorganization of the code, just tell me!

HenriAugusto:
I do need directions! haha

As discussed in the pd-dev mailing lists it would be ideal for the kbdnav functionality to be somehow optional (as it doesn't make sense for the pd-lib for ex, etc). 

Currently the code uses `#ifdef HAVE_KEYBOARDNAV` for that purpose which is bad. How should it be made optional?

doc/7.stuff/keyboard.navigation/keyboard.patching-help.pd
HenriAugusto: True! 

I was thinking maybe <kbd>Ctrl/Cmd</kbd> + <kbd>Shift</kbd> + <kbd>some letter</kbd> would be better. Do you have a suggestion?

src/g_editor.c
HenriAugusto: I did that so connections made with kbdnav through the canvas console could be undone normally.

Select an outlet and press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>Down</kbd>. An entry field will be pre-filled with the relevant information and relevant part you should replace is already selected so you can just type.



I think it works great for connection distant objects but i couldn't figure out a way of doing this without exposing the method in some way, since the GUI is sending the msg to the core.

What can be done?

tcl/pd_bindings.tcl
HenriAugusto: Because with the "goto" functionality you can go directly to any object you want in the patch.

umlaeute: so? with your patches you don't need the mouse any more, should we remove that as well ;-) ?
(just to make sure: the answer to this rhetoric question should be "no").

as a sidenote: if you want to remove these bindings (and i obviously don't agree that you should) for de-crufting purposes, you should also remove:
- the `canvas_cycle` tcl-proc
- the `cycleselect` method of the canvas
- the `canvas_cycleselect` function

doc/7.stuff/keyboard.navigation/keyboard.patching-help.pd
umlaeute: well, you could also just document that one should press `"` (double-quote) to get that behaviour, as this is what you are binding to.

alternatively, you could re-use the <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>m</kbd> shortcut, for sending messages:
- when on the Pd-console window, this sends a generic message into Pd-land
- when on a normal canvas, this could send a message to the canvas itself.

but on re-thinking this, i actually think that this feature should be included at all.
i think the audience is pretty small, and the terminal (which will mostly generate messages like `canvas: no method for 'foo'`) will be confusing to most people.

you can easily provide it as a gui-plugin, so power-users can selectively install it.


(i guess the main reason to include it is `goto`: instead you could <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>o</kbd> open a specialised "goto" bar, that might even have a number dial, to limit the possibilities to valid indices)

tcl/pd_bindings.tcl
umlaeute: an i find `goto` + trying out all those indices considerably more complicated that just <kbd>Tab</kbd>ing through the patch.

umlaeute:
some general comments on the usability (as opposed to the implementation) for discussion:

## :cool: 

it's really nice.

## `goto`

why isn't the message `select`?

the implementation (`canvas_goto()`) is a bit more than i would expect from a pure `select`, but that is probably only due to the fact that the GUI (console) doesn't do any range-checking.
if you trade the generic canvas-console for a specific *select dialog*, you could do the range-checking there (e.g. even using a pulldown or a spinbox)

also, i don't understand the shortcut <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>o</kbd>.
It is very similar to "Open File..." (<kbd>Ctrl</kbd>+<kbd>o</kbd>), and as a mnemonic it doesn't ring a bell either. <kbd>Ctrl</kbd>+<kbd>g</kbd> is probably better suited.


## canvas-prompt

i like the canvas-prompt as a power user.

however, for the ordinary use-case of selecting (`goto`) and connecting (`connect`) objects, i find this *very* clumsy.

i find rewriting some of those messages (`connect` -> `connect_with_undo`) outrageous (this only leads to questions like "i discovered a bug: if i send a `connect` message to the canvas, it will sometimes add *undo* and sometimes not")

#### cycling
i think i would prefer a cycling paradigm instead:

- for `goto`:
  - add a short-cut to select a single object (e.g. the one nearest to the mouse pinter)
  - use <kbd>Tab</kbd> to cycle through the objects

- for `connect`:
  - when you have selected an iolet, hit <kbd>Tab</kbd> to cycle through the objects that the given iolet can connect to
  - once such an object is selected, use <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>⇆</kbd> to select a compatible iolet
  - use <kbd>Esc</kbd> to go back from iolet selection to object-selection

#### allow mult-digit numbers
alternatively, you could just allow multi-digit numbers instead of single digits.
probably like:
- press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>⇅</kbd> (again) to enter the *number connector*
  - (only) the nearest 10 *compatible* iolets are displaying their IDs
  - press <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>⇅</kbd> again(!) to enter the "extended number connector", where *all* compatible iolets display their IDs
- type a number (e.g. `2`)
   - all iolets whose IDs match (start with `2`, e.g. `2`, `20` and `21`) will still be displayed, the rest will vanish
- keep typing to narrow down the selection to a single iolet
  - use <kbd>Backspace</kbd> to widen the selection again
- once only a single iolet is selected, hit <kbd>Enter</kbd> to do the selection

"cycling" and "multi-digit numbers" could both be implemented (they are not exclusive)

## general issues

- the *magnetic connector* and the *digit selector* allow illegal connections (signal outlets to non-signal inlets)

- there seem to be quite a lot of bindings to get the object indices
  - <kbd>Ctrl</kbd>+<kbd>i</kbd>
  - <kbd>Ctrl</kbd>+<kbd>Tab</kbd>
  - <kbd>Ctrl</kbd>+<kbd>o</kbd>
  - <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>o</kbd> (also opens the canvas-console and prefills it with `goto`)


## issues in the documentation

### dynamic patching
dynamic patching is not officially embraced by Pd. until it is, you probably should remove mentions of it from the documentation.

### message fan-outs
please avoid message-fanouts in the help patches (i know that this is my personal crusade; i also know that miller uses fanouts in some of his help patches; but nevertheless fanouts *are* the single most common cause for bugs, so we really shouldn't encourage people to use them)

- remove blatant displays of fan-outs from the top-level canvas
- prevent users from creating msg-fanouts with shortcuts ("magnetic connector")

src/g_editor.c
HenriAugusto: I've just moved some implementations to `g_kbdnav.c` so i can get everything working that way.

After that i'll work in making this a pointer.

HenriAugusto:
I'm glad you liked it. Its been very useful for me. Not only some task are faster  and/or easier but the object placement according to the selected in/outlet helps a lot in keepign the patches very clean.

## "goto"

"select" seems appropriate indeed! (in fact it is what the code does heh)

> if you trade the generic canvas-console for a specific _select dialog_, you could do the range-checking there (e.g. even using a pulldown or a spinbox)

I've tried a dialog (like the "message dialog") but felt it would either make the user reach for the mouse for interacting with this UI or impose more key presses on the way.
 
> also, i don't understand the shortcut Ctrl+Shift+o.
> It is very similar to "Open File..." (Ctrl+o), and as a mnemonic it doesn't ring a bell either. Ctrl+g is probably better suited.

<kbd>Ctrl</kbd>+<kbd>g</kbd> was the first thing that came to mind but i realized it was already in use by _Find Next_. Thought it would be better to not change anything like that without discussion so i've just went with something for prototyping. 

Also <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>g</kbd> was taken for the _put graph_ shortcut

Worth noting that If we rename it for "select" then <kbd>Ctrl</kbd>+<kbd>s</kbd> and <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>s</kbd> are both taken.

## canvas-prompt

Why it felt clumsy?

> i find rewriting some of those messages (`connect` -> `connect_with_undo`) outrageous (this only leads to questions like "i discovered a bug: if i send a `connect` message to the canvas, it will sometimes add _undo_ and sometimes not")

To be honest i don't remember why i went with that approach. We can stop this "message replacement" and have the gui just pre-fill the prompt with _"connect_with_undo"_.

#### cycling

The main concept was allowing direct operations. If you are in obj index 10 and you want to go to an inlet on object index 113 you will have **a lot** of cycling to do. 

#### allow mult-digit numbers

I like this one better. I've initially thought of this but went with the canvas console because i figured it would be faster to use and simpler. 

After the canvas console is open (and pre-filled for you) you can look straight to the obj you want and do the connection, no matter how much objects you have on your patch.

_On a side note_: i've once posted some gifs on facebook just to see how people were interested in the prototype and a guy sent an inbox saying he teaches PD in some workshops for people with disabilities and that he tought kbd navigation could provide great accessibilty for them. I've tried to keep this in mind while making decisions so i usually go for the fewer keystrokes (trying to not sacrifice usability, of course)

## general issues
> * the _magnetic connector_ and the _digit selector_ allow illegal connections (signal outlets to non-signal inlets)

Oopsies!

> * there seem to be quite a lot of bindings to get the object indices

Indeed. I imagine we could use only <kbd>Ctrl</kbd>+<kbd>i</kbd>.
Displayng the indices when triggering `goto` is intentional as you need the indices.


## issues in the documentation
### dynamic patching

okay! I'll remove the mention

### message fan-outs
 
> remove blatant displays of fan-outs from the top-level canvas

okay! (there is also some possible to-be-made fan outs in the digit connector example)

> prevent users from creating msg-fanouts with shortcuts ("magnetic connector")

I wonder if it isn't better to not enforce a restriction like this. The user might create a fan-out like one of this. Left one is very common for debugging (and lots of users leave those atoms in their patches) while the right one is okay because the order doesn't really matter in that case.

image

umlaeute:
### fan-out

> > prevent users from creating msg-fanouts with shortcuts ("magnetic connector")
> 
> I wonder if it isn't better to not enforce a restriction like this. The user might create a fan-out like one of these. Left one is very common for debugging (and lots of users leave those atoms in their patches) while the right one is okay because the order doesn't really matter in that case.

imo, the problem is not enforcing the restriction but making bad practice the *default* and encouraged behaviour.
right now, you only have to hold your finger on the <kbd>g</kbd> key (while an outlet is selected), and you get more fan-outs than you can ever stomach.
you *could* only connect to cold inlets (as your second example shows), but going down that route quickly makes the system "too intelligent".

as for your first example (quickly adding a numberbox for debugging), i don't think this is something where "magnetic connections" are required at all.
with a normal (unpatched) Pd, i usually just select the upstream object and hit <kbd>Ctrl</kbd>+<kbd>3</kbd>, to create an auto-connected numberbox for debugging.
sure, it does have a dreaded fan-out, but it's quite sandboxed: if i keep hitting <kbd>Ctrl</kbd>+<kbd>3</kbd> no additional fan-outs are created; and i always get a numberbox for the *right* object (and with your selectable outlets, i always get a numberbox for the *right* outlet!)

i haven't used the keyboard patching in practice much, but i figure you do, so:
*How often do you really use **repeated** "magnetic connections" for a msg-outlet in practice?*

in my tests i find that it usually does the wrong thing (it seems that the first connection is usually ok, as the "nearest" inlet comes quite natural; but all the other next-nearest 

btw, there's another small bug with the *magnetic connector*: if i open the help-subpatch, select the `[noise~]` object, go into keyboard-navigation mode and hit <kbd>g</kbd> **8** times, i get an error on the Pd-console:
> magnetic_connector 3 0 0 0 (noise~->text) connection failed

umlaeute:
## "goto"

> I've tried a dialog (like the "message dialog") but felt it would either make the user reach for the mouse for interacting with this UI or impose more key presses on the way.

ah no, i really meant an embedded dialog, like you get when hitting <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>g</kbd>. but more restricted, as in: only allowing to enter an index, rather than a full-fledged message

goto10

so this wouldn't add any additional keystrokes (rather less)

umlaeute:
> ## canvas-prompt
> 
>> however, for the ordinary use-case of selecting (`goto`) and connecting (`connect`) objects, i find this very clumsy.
>
> Why it felt clumsy?

because i have to leave the domain of patching (selecting/connecting objects) and enter the completely different domain of dynamic patching.

> 
> > i find rewriting some of those messages (`connect` -> `connect_with_undo`) outrageous (this only leads to questions like "i discovered a bug: if i send a `connect` message to the canvas, it will sometimes add _undo_ and sometimes not")
> 
> To be honest i don't remember why i went with that approach. We can stop this "message replacement" and have the gui just pre-fill the prompt with _"connect_with_undo"_.

or use a limited dialog, that doesn't expose the actual message to the canvas at all.

> #### cycling
> 
> The main concept was allowing direct operations. If you are in obj index 10 and you want to go to an inlet on object index 113 you will have **a lot** of cycling to do.

just keeping the <kbd>Tab</kbd>-key down you will get to the desired object pretty quickly.

> #### allow mult-digit numbers
> 
> I like this one better. I've initially thought of this but went with the canvas console because i figured it would be faster to use and simpler.

how so?
it's more complicated to implement, but for the end-user i don't think that an extra <kbd>Enter</kbd> is that much of extra work to do.

> After the canvas console is open (and pre-filled for you) you can look straight to the obj you want and do the connection, no matter how much objects you have on your patch.

hmm, i don't really see much regression to what you have now.

the extra <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>⇅</kbd> is only needed if you want to see all iolet-indices.

> 
> _On a side note_: i've once posted some gifs on facebook just to see how people were interested in the prototype and a guy sent an inbox saying he teaches PD in some workshops for people with disabilities and that he thought kbd navigation could provide great accessibility for them. I've tried to keep this in mind while making decisions so i usually go for the fewer keystrokes (trying to not sacrifice usability, of course)

totally.

but exposing a complicated message (even prefilled) makes it much easier to make things go badly wrong if, for example, you made a typing error and in the processes of deleting the typo you also delete part of the message.
this seems to add more keystrokes if you have some (e.g. motoric) disabilities and tend to mistype.

umlaeute:
## documentation

it just occured to me, that you probably should put the help-patch into `doc/7.stuff/patching/`, so there's a broader scope (of the containing folder), that could possibly document other patching techniques as well (re-iterating the ordinary patching paradigm, documenting "intelligent patching",...).

danomatika:
Gotta say I'm glad to see where this is going. I know I will be using it and I'm really happy this will bring full-scale accessibility to the patching itself.

Now, I only need to roll this into a custom curses pd GUI for console-only patching....

danomatika:
>you probably should put the help-patch into `doc/7.stuff/patching/`

Yes. If the documentation patch gets to large, definitely split it into topic patches within a subfolder.

src/g_kbdnav.c
```diff
@@ -0,0 +1,2183 @@
...
+        //https://wiki.tcl-lang.org/page/canvas+text
+        sys_vgui(".x%lx.c create rect [.x%lx.c bbox objindex%d]  -fill white -outline blue -tag [list objindex_bkg%d obj_index_number_bkg]\n",
+                x,
+                x,
+                objindex,
+                objindex);
+        sys_vgui(".x%lx.c lower objindex_bkg%d objindex%d\n",
+                x,
+                objindex,
+                objindex);
```
umlaeute: this should probably go into generic tcl/tk proc for adding tooltips.
something like

~~~tcl
proc tooltip_show {mytoplevel id posx posy text {color blue} {outline none}} {
  # ...
}
proc tooltip_hide {mytoplevel id} {
 # ...
}
~~~

(with `id` being just some identifier (probably without dots...) that allows us to identify the tooltip when we want to remove it later...)

danomatika: Agree.

src/g_kbdnav.c
```diff
@@ -0,0 +1,2183 @@
...
+    if( !kbdnav ) return 1;
+
+    /* set our modifier state */
+#if defined(__APPLE__)
+    if( !strcmp(gotkeysym->s_name, "Meta_L") )
+#else
+    /* TCL reports Right Control keyup events as Control_L */
+    /* https://wiki.tcl-lang.org/page/Modifier+Keys*/
+    if( ( !strcmp(gotkeysym->s_name, "Control_L") || !strcmp(gotkeysym->s_name, "Control_R")) )
+#endif
```
umlaeute: i think it would be better to do the actual keybindings on the Tcl/Tk side.
e.g. rather than checking whether the pd-core received a string like "Control_L", "Control_R" or "Meta_L" (btw, what happened to "Meta_R"?) you should instead create bindings for for `<$::modifier-Key-Up>` and the like.
this will also automatically handle any platform differences in the "Meta"-key

once the bindings are handled in Tcl/Tk, add  one (or sevaral) methods to the canvas-class that explicitly deal with keyboard-patching.
you could also give the GUI a notion of whether the system is currently in keyboard-patching mode, in which case ordinary keystrokes that cannot properly be bound (like <kbd>g</kbd>) might get re-directed.

umlaeute:
just found a really annoying bug: with the current keyboard-patching it is quite impossible to type double-quotes.
e.g. you can no longer create a comment like the following random sample from doc/:
> `You can use "qlist" or "textfile" objects for sequencing.`

HenriAugusto:
I'm updating the first post with a TODO list.

## fan-out

I see what you mean. Maybe we could have only one behavior for the magnetic connector which is the one currently assignet to <kbd>shift</kbd>+<kbd>g</kbd>. When triggered for the 1st time it creates a connection to the nearest in/outlet. When triggering again it will switch the connection to the 2nd closest one and so on. In about 1 year it is really what i was using.

## goto and canvas-prompt connections

> but exposing a complicated message (even prefilled) makes it much easier to make things go badly wrong if, for example, you made a typing error and in the processes of deleting the typo you also delete part of the message.
> this seems to add more keystrokes if you have some (e.g. motoric) disabilities and tend to mistype.

I hadn't tought about it this way. You're right! A dedicated "connect anything" embedded dialog would be way better.

I picture it like this. A "connect" label and 4 number boxes. Besides typing the number directly one could use up/down to change them (possibly with modifiers to change the increment) plus use left/right or tab to quickly navigate through them.

doc/7.stuff/keyboard.navigation/keyboard.patching-help.pd
HenriAugusto: I agree with you and I'm inclined to go with the "not include the canvas console at all and release as gui-plugin" route. Maybe somebody may comment on the issue but if that doesn't happen I'll remove it in favor of the dedicated embedded dialogs for selecting and connecting.

src/g_bang.c
HenriAugusto: I'm gonna resolve this as i've put this in the TODO in the first post.

src/g_canvas.c
HenriAugusto: okay!

src/g_editor.c
HenriAugusto: fixed in b768c3b1b916e3dce649a30f3ca9209d630fdc18

src/g_editor.c
HenriAugusto: fixed b768c3b1b916e3dce649a30f3ca9209d630fdc18

src/g_editor.c
HenriAugusto: okay b768c3b1b916e3dce649a30f3ca9209d630fdc18

src/g_kbdnav.c
HenriAugusto: Yes.

While rewriting the kbd nav code to use PD's class/method paradigm maybe all the kbdnav bindings could be handler gui side and call dedicated methods.

Maybe something like

**Gui:**
```TCL
proc ::pd_bindings::kbdnav_bindings {}{
   [...]
   bind all <$::modifier-Key-Left> { kbdnav_left %W }
   [...]
}
```
**Core:**

```C
void g_canvas_setup(void)
{
   // [...]
   /* -------------------------- keyboard navigation ----------------------------- */
   class_addmethod(canvas_class, (t_method)kbdnav_leftkey, gensym("kbdnav_left"),
                       A_GIMME, A_NULL);
   // [...]
}
```
```

File: src/g_kbdnav.c (hunk 22 above)
Comment (HenriAugusto) [reply src/g_kbdnav.c]: okay

File: src/g_kbdnav.h (hunk 17 above)
Comment (HenriAugusto) [reply src/g_kbdnav.h]: Okay. 

I recall that i tought i would need to expose this function so the `connect_with_undo` message sent from the gui could work. 

But now that i think about it i wonder: i actually can send a `connect_with_undo` from the guibut only doing [this class_addmethod() call](https://github.com/pure-data/pure-data/pull/869/files#r368252839) without removing the "static" from the function?

Of course as you said in the comment it still would expose 'connect_with_undo' to the user so one shouldn't do it. I'm just asking to better understand how PD (and C) works.

Comment (HenriAugusto) [reply src/g_kbdnav.h]: Anyway, when implementing a dedicated embbeded "connect" dialog i would need to somehow make the gui ask for the core to make a connection. How should it be done?

Just make the gui call a (non-exposed) function - lets say, for the sake of clarity "receive_connection_request_from_gui()" -that in turn would call `canvas_connect_with_undo`?

Also, this function would need to be located in `g_editor.c` (since `canvas_connect_with_undo` would be static), right?

File: src/g_kbdnav.h (hunk 18 above)
Comment (HenriAugusto) [reply src/g_kbdnav.h]: ok 8c4bcbed8c7400d3c213b87c00b611e6e770c49b

**Comment by umlaeute** [comment]
>
>Maybe something like
>
>**Gui:**
>```TCL
>proc ::pd_bindings::kbdnav_bindings {}{
>   [...]
>   bind all <$::modifier-Key-Left> { kbdnav_left %W }


well. i think it would be easier to implement and read (and faster in general) if the actual direction was an argument rather than encoded in the methodname.

after all, there has to be a reason why we don't use a different binding for each key (think "key_a_up",...)

**Comment by umlaeute** [comment]
>
>Of course as you said in the comment it still would expose
>'connect_with_undo' to the user so one shouldn't do it. I'm just asking
>to better understand how PD (and C) works.

adding a method to the canvas-class (and using it) is not "exposing it to the user".
opening a canvasmsg dialog with this message prefilled is!

so I don't think there's a harm in having such a method (unless we can think of an even better way to have undoable connections triggered from the gui...)

**Comment by umlaeute** [comment]
you wrote:
>
>Maybe something like
>
>**Gui:**
>```TCL
>proc ::pd_bindings::kbdnav_bindings {}{
>   [...]
>   bind all <$::modifier-Key-Left> { kbdnav_left %W }


well. i think it would be easier to implement and read (and faster in general) if the actual direction was an argument rather than encoded in the methodname.

after all, there has to be a reason why we don't use a different binding for each key (think `key_a_up`,...)

**Comment by umlaeute** [comment]
>
>Of course as you said in the comment it still would expose
>'connect_with_undo' to the user so one shouldn't do it. I'm just asking
>to better understand how PD (and C) works.

adding a method to the canvas-class (and using it) is not "exposing it to the user".
opening a canvasmsg dialog with this message prefilled is!

so I don't think there's a harm in having such a method (unless we can think of an even better way to have undoable connections triggered from the gui...)

File: src/g_editor.c [hunk 24]
```diff
@@ -2655,7 +2655,7 @@ int canvas_isconnected (t_canvas *x, t_text *ob1, int n1,
     return (0);
 }
 
-static int canconnect(t_canvas*x, t_object*src, int nout, t_object*sink, int nin)
+int canconnect(t_canvas*x, t_object*src, int nout, t_object*sink, int nin)
```
umlaeute: if you make this non-static, you should probably change the name to `canvas_canconnect()` to avoid disambiguities.

HenriAugusto:
### Regarding the _"goto"_ feature and indices validation:

We can opt for not enforcing a valid index on the spinbox (less complex) or make the gui request the number of objects from the canvas (more complex) to enforce a valid index range.

What you think? If opting for the request, how should the request and delivery be implemented?

HenriAugusto:
## Kbdnav and the class/method paradigm

@umlaeute just to make sure i got it when you said

> btw, an even nicer way to design the API would be to use Pd's class/method paradigm :-)

This is what you meant?

**GUI:**
```TCL
# pd_bindings.tcl
proc kbdnav_bind { mytoplevel } {
	set kbdnav_mod "Control"
	if {$windowingsystem eq "aqua"} {
		set kbdnav_mod "Cmd"
	}
	bind all <$kbdnav_mod-Up> "pdsend $mytoplevel kbdnav_up"
	bind all <$kbdnav_mod-Down> "pdsend $mytoplevel kbdnav_down"
	

	bind all <$kbdnav_mod-Shift-Up> "pdsend $mytoplevel kbdnav_connect_up"
	bind all <$kbdnav_mod-Shift-Down> "pdsend $mytoplevel kbdnav_connect_down"
	# [...]
}
```

**Core:**
```C
#ifdef HAVE_KEYBOARDNAV
void kbdnav_register( t_canvas *x)
{
	// [...]
	class_addmethod(canvas_class, (t_method)kbdnav_up, gensym("kbdnav_up"), A_NULL);
	class_addmethod(canvas_class, (t_method)kbdnav_down, gensym("kbdnav_down"), A_NULL);
	class_addmethod(canvas_class, (t_method)kbdnav_connect_up, gensym("kbdnav_connect_up"), A_NULL);
	class_addmethod(canvas_class, (t_method)kbdnav_connect_down, gensym("kbdnav_connect_down"), A_NULL);
	// [...]
}
#else
void kbdnav_register( t_canvas *x ){
   //nothing
 }
#endif

void kbdnav_down(){
	// inlet selected: go to an outlet
	// outlet selected: go through connections on that outlet
	// connection selected: go to the inlet (source) of that connection
}

void kbdnav_up() { //similar }

void kbdnav_connect_down(){
	// outlet is selected: enter digit connector
	// digit connecting: open "connect" dialog
}

void kbdnav_connect_up(){ //similar }
```

## Kbdnav outisde g_kbdnav.c

There are places which contain kbdnav code that i don't know how to implement separetedly without code duplication. An example is `canvas_howputnew()` in _g_text.c_.

There could be two versions of the function but there would be a lot of redundant code. Also the functions that call `canvas_howputnew()` would need to be somehow aware of the kbdnav to know which version to call. 

Something similar goes for `canvas_obj()`, `canvas_msg()`, etc.

As a matter of fact there is already somewhat similar code as in `glist_drawiofor()` and `kbdnav_glist_drawiofor()`

Also there is a "redirect" that i think should be more elegant. How to do it in the new, nicer, class/method way?

```C
void glist_drawiofor(t_glist *glist, t_object *ob, int firsttime,
    char *tag, int x1, int y1, int x2, int y2)
{
#ifdef HAVE_KEYBOARDNAV
    kbdnav_glist_drawiofor(glist, ob, firsttime, tag, x1, y1, x2, y2);
    return;
#endif
	// [...]
```

I also wanted to know how to deal with in/outlet highlighting in the iemgui objects. Currently it is like that:

```C
static void hslider_draw_new(t_hslider *x, t_glist *glist)
{
    // [...]
#ifdef HAVE_KEYBOARDNAV
    kbdnav_hslider_draw_io_selection(x, glist, xpos, ypos, lmargin, IEMGUI_ZOOM(x), iow, ioh);
#endif
}
```

umlaeute:
> `pdtk_goto_receive_n_of_objs`

the functionality (to let the GUI know how many objects there are), is unrelated to `keyboardnav_goto` (the latter just happens to need this knowledge).
i think the proper way would be, to send a message to the GUI whenever the object-count changes, so it's always up-to-date.
the `proc` should thus have a generic name and probably live in the `::pwindow::` namespace (and btw, the `pdtk_` prefix is there for historical reasons; don't add it to new, namespaced procs)


> `bind all <$kbdnav_mod-Up> "pdsend $mytoplevel kbdnav_up"`

no. you don't want to bind a separate method for each key.
instead you want to have a few `keyboard-navigation` methods, and use arguments:
e.g. `bind all <$kbdnav_mod-Up> "pdsend $mytoplevel kbdnav up"`

i think i already mentioned this in https://github.com/pure-data/pure-data/pull/869/#issuecomment-580129807

HenriAugusto:
Yes, you did. My bad.

While asking was really focusing on how to separate the kbdnav stuff from the rest. I first imagined creating a separate class to deal with kbdnav but then realize it would be better to use bindings on `canvas`. Since you didn't comment on that i assume it's good. I still have questions about separating kbdnav stuff so i want to feel confident about the direction i'm taking before continuing.

> i think @danomatika 's idea to make this optional as not intended to clutter the code, but instead create hooks to the patching system that would be filled with your enhancements (or not, if it was not required).
>
>that would reduce the use of HAVE_KEYBOARDNAV (should actually be USE_KEYBOARDNAVIGATION) to one or two (1-2!) occurences, rather than putting them all over the place.

On my [last comment](https://github.com/pure-data/pure-data/pull/869#issuecomment-595340037), on the section **Kbdnav outisde g_kbdnav.c** i mentioned some code that i don't know how to separate from the rest. Can those problems be also solved with the class/method paradigm? How? Should i add a canvas method for stuff like `glist_drawiofor()` and `kbdnav_glist_drawiofor()`? And if yes, how would the code know which one to execute? 

> i think the proper way would be, to send a message to the GUI whenever the object-count changes, so it's always up-to-date.

I wonder if it wouldn't be better to leave that to a separate PR then. A proper system for the core to regularly send updates to the GUI probably may have other implications and potential uses.

HenriAugusto:
**Note:** At the current state I've changed **Find Again** to <kbd>Shift</kbd>+<kbd>accel</kbd>+<kbd>F</kbd> and the **Go to** dialog is assigned to <kbd>accel</kbd>+<kbd>G</kbd>

HenriAugusto:
Now the GUI keeps track of which patch is keyboard-navigating and we can have context-sensitive hotkeys. I've included an entry in the _Edit_ menu to enable/disable keyboard navigation.

Some _key_ changes (pun intended)
- To toggle object indices visibility: <kbd>mod</kbd>+<kbd>i</kbd>
- Magnetic connector:
   - <kbd>c</kbd>+ non fan-out
   - <kbd>Shift</kbd>+<kbd>c</kbd>: fan-out

src/g_editor.c
HenriAugusto: Resolving now that we have the "Connect Objects" dedicated dialog

src/g_kbdnav.h
HenriAugusto: fixed e998b865392d8c169b9b784f05b0d2c8f1eb6ea7

tcl/pd_bindings.tcl
HenriAugusto: The idea with "Go To" is that you don't have to try the indices. You just look at the object, see it's index, type it and hit return.

<kbd>Tab</kbd>-traversal is (pratically) unpredictable because the object order can change. A simple, straight"chain" of objects might have completely "spaced" indices depending on the order you created the objects, the copy-pastes you did, and etc. 

Especially on big patches it can be quite a chore to keep tabbing until you get to the object you want.  Even now that you can see the indices and know how many times you need to press the <kbd>Tab</kbd>. 

What if you must press it 45 times? And holding <kbd>Tab</kbd> is no good either. While holding it did it triggered 30 times? 40 ? 60? I can't see how's that more complicated than

- press <kbd>mod</kbd>+<kbd>g</kbd>
   - all object indices are automatically visible
- look at the object you wanna go and type it's index
-<kbd>Return</kbd>

Also the "Go To" is originally intended for "jumps" since when you're on a "chain" you can just <kbd>mod</kbd>+<kbd>arrow</kbd> your way around with ease.

umlaeute: well, this is mostly bike-shedding. but here you go:

> What if you must press it 45 times? And holding Tab is no good either. While holding it did it triggered 30 times? 40 ? 60?

so what if? why would it matter if tab triggered 30 times, 40 or 60? i really don't care as it eventually gets me there.
and *that* is very simple to check with <kbd>Tab</kbd>-cycling. i just keep the key pressed until the object highlights. then i let go. since i was to slow, i back-cycle 3 times. done.
no symbolic processing involved (as in: relate numbers to objects).
to be fair, it might help to have a more contrasting colour that blue/black.

otoh with `goto`, if you have a fairly large patch, having numbers splattered all over the patch won't help readability either: so you have random numbers spilled all over the objects which makes it hard to read both the objects (beneath the numbers) and the numbers themselves. :stuck_out_tongue:

HenriAugusto:
Did a cleanup and update of the first post with the current state of things.

Recently i just realized that the <kbd>shift</kbd>+<kbd>digits</kbd> are [system/locale dependent](https://stackoverflow.com/questions/61806427/keyboard-independent-way-of-detecting-shiftdigits). I'm inclined to change it to <kbd>mod</kbd>+<kbd>digits</kbd>.

PR #870, opened 2020-01-18
implement a stack for class_loadsym

### LINKED ISSUE #850: bug in [jmmmp/matrixctrl] - only one object per pd session allowed
This is the first time I see something like this: the object [jmmmp/matrixctrl] can only be created once in a pd session. After the first time, any other instance will appear as text (no gui); if something is sent to it, it behaves as [zexy/lister] - including opening its help patch when requested. (Btw, lister isn't used in this abstraction)

I installed many pd versions, and found out that the last version where this works is pd-0.46-7 - in windows 7.

Is it possible to track down what changed? This object uses zexy, iemmatrix and list-abs.

(matrixctrl is a pd abstraction emulating the same object in max/msp, to control mtx_*~)

umlaeute:
This is an attempt to properly fix https://github.com/pure-data/pure-data/issues/850

The problem described in #850 is, that abstractions might unwillingly get aliased to other objectclasses.

There's an alternative/complementary fix for the problem in  0dfe3721c5b1b2f4e7dbf0b82ceb02411ff307c2, but the alternative mostly fixes the symptoms, whereas this PR attempts to fix the root of the issue.

### prerequisites
When Pd attempts to instantiate a (yet unknown) object `[foo/bar]`, it will really load a library `bar` in the directory `foo`. This library will then register an objectclass `bar`, but not ~~`foo/bar`~~ (as the path is a runtime-detail and not really known during buildtime). Since the user will most likely want the `foo/bar` object to really be an instance of `bar` (as found in `foo/`), Pd automatically adds a convenience creator `foo/bar` for objectclass `bar`.
It does that by comparing the registered classname (`bar`) with the requested objectname (`foo/bar`), and if the latter ends with the former then it adds the alias:

https://github.com/pure-data/pure-data/blob/628a8d3464a45f25ea3bbd4fbd76655f28ce825e/src/m_class.c#L458-L469

### abstract problem
This automatic objectname aliasing can sometimes clash with the user intentions.
That is: if the mechanism is triggered by some arbirtary objectclass that happens to match the last part of the objectname that is currently being loaded.

This occurs, if the registration of a new objectclass via `sys_load_lib` (e.g. in `[declare]`) is happening while an object is instantiated and this object's name (as stored in `class_loadsym`) ends with the literal name of the newly registered objectclass). Then the object's name will be made an alias to the newly registered objectclass.

afaict, the problem only happens when libraries are loaded via `declare`, since the alternative way of loading libraries (by just instantiating an object of the same name as the library), will reset the `class_loadsym` variable - which will prevent the object from being affected.

### example problem
Assume a simple abstraction `hyperfiddle~`, that adds some logic around our beloved `fiddle~`.
A trivial (and useless) implementation of `hyperfiddle~` might just load the `fiddle~` library (and therefore the object):

~~~puredata
#N canvas 6 61 450 300 12;
#X declare -lib fiddle~;
#X obj 89 74 declare -lib fiddle~;
~~~

The first time, the `[hyperfiddle~]` object is instantiated, it loads the abstraction (without inlets&outlets,...)
However, the second time the `[hyperfiddle~]` object is instantiated, it will have inlets and outlets, cannot be opened any more and will behave just like `[fiddle~]` itself. In fact, this and all further instances of `[hyperfiddle~]` will in fact **be** `fiddle~`-instances.

This is what's happening:

1. The user creates a `[superfiddle~]` object
1. Pd notices that it doesn't know an objectclass *hyperfiddle~*, and runs into pd_objectmaker's `new_anything`
2. it  stores the name *hyperfiddle~* into `class_loadsym`
3. and starts a library-search that eventually loads the "hyperfiddle~.pd" abstraction
4. while loading the abstraction, Pd encounters the `declare -lib fiddle~` statement
5. this will search and load the *fiddle~* library, which registers a new *fiddle~* objectclass
6. while registering the *fiddle~* objectclass, Pd notices that *hyperfiddle~* ends with *fiddle~*, so it assumes that the former is a long-form of the latter, and automatically registers an alias "hyperfiddle~" for the fiddle~-class :exclamation:
7. the instantiation of the `[hyperfiddle~]` object finishes
8. The user creates *another* `[superfiddle~]` object
9. Pd notices that it does know the *hyperfiddle~* objectclass and creates a `[fiddle~]`-instance :boom: 

### make sure, the long-form is a path
the most trivial fix is to add a check whether the supposed long-form classname ends with a path-delimiter. So `[foobar]` will no longer be considered a long-form of `[bar]`, whereas `[foo/bar]` will.

i never implemented this fix, but expect there might be some confusion if the user actually uses `[hyper/fiddle~]` as the abstraction name.

### make sure the long-form isn't an object-class already
this actually fixes another problem (automatic aliases shouldn't be created if they would overwrite existing classes), but as a side-effect solves the problem at hand:

since 0.47, abstractions are loaded like libraries, so each abstraction gets its own method in `pd_objectmaker`, early in the object loading process. when we reach the automatic long-form aliasing code, the containing abstraction's name is already taken

This is implemented in 0dfe3721c5b1b2f4e7dbf0b82ceb02411ff307c2 (included in the `update/0.50` branch)

### use a stack for storing the currently loaded classname
the real problem seems to be, that the loading/auto-aliasing mechanism doesn't handle *nested* classes. In the example above, we load both an abstraction (*hyperfiddle~*) and a library (*fiddle~*), and Pd applies (the auto-aliasing) rules meant for the uplevel abstraction while loading the nested library.
this can be solved with pushing the "currently loaded classname" to a *stack* rather than storing it in a scalar.

This PR implements such a stack (and also includes the safety net to not overwrite existing classnames)


millerpuckette:
I've never understoof  the whole "loader" framework, so I'll just take this on faith :)

PR #874, opened 2020-01-23
change error posts for readsf~ and writesf~

porres:
- I just added a tilde where it was missing and changed 'dsp' error message to 'readsf~' because I think it makes it clearer which object the error is coming from.

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -2124,7 +2124,7 @@ static t_int *readsf_perform(t_int *w)
             int xfersize;
             if (x->x_fileerror)
             {
-                pd_error(x, "dsp: %s: %s", x->x_filename,
+                pd_error(x, "readsf~: %s: %s", x->x_filename,
                     (x->x_fileerror == EIO ?
                         "unknown or bad header format" :
                             strerror(x->x_fileerror)));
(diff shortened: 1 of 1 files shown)


danomatika:
Canyon check if this is fixed in #855?

porres:
> Can you check if this is fixed in #855? I'm pretty sure I did this as I updated some of the error posts in that branch.

yeah, seems you did the same changes! 

I could also suggest to make the following errors more explicit by also printing the name of the object

https://github.com/pure-data/pure-data/blob/update/soundfile/src/d_soundfile.c#L1345

https://github.com/pure-data/pure-data/blob/update/soundfile/src/d_soundfile.c#L1527

https://github.com/pure-data/pure-data/blob/update/soundfile/src/d_soundfile.c#L2155

https://github.com/pure-data/pure-data/blob/update/soundfile/src/d_soundfile.c#L2560

what do you think?

danomatika:
In think in these cases, it's ok as the error is thrown in response to a specific message which is quoted in the error post, ie. open, read, or write, and all posts use `pd_error()` so they findable to highlight the object.

clone has a similar error post but it's in response to a problem with the creation args, so it uses the name of the object which in that context makes sense.

PR #875, opened 2020-01-25
Python-inspired slice notation for dollar and dollsym atoms

myQwil:
Retrieve entire lists or subsets of lists using a syntax very similar to Python's slice notation. The syntax is $<start>:<stop>:<step>. Some examples include:

- $:      the full list
- $::-1  the full list backwards
- $-1    the last item in a list
- $2:    the full list, minus the first item
- $:4    the first 3 items in a list
- $3:6  list items 3, 4, and 5
- $::2   the full list with every other item skipped

Included is a patch with examples.
[dollar-slice.txt](https://github.com/pure-data/pure-data/files/4112709/dollar-slice.txt)


umlaeute:
supersedes #503

umlaeute:
there probably should be some protection for backwards compatibility.

eg sending a `[list foo bar(` to `$1:2` will give you
- `foo:2` with pd-vanilla, and
- `foo:bar` with your patch

you could instead use the newly-allowed curlies as in `${2:3}`

myQwil:
That sounds good to me, although curlies are still getting dropped on my build of pd.

myQwil:
I wasn't aware of the $@ notation. I'm looking at pd-extended's source code and the method is more efficient, so I'll try to work in some of the same ideas and maybe find a happy middle.

porres:
yeah, this is certainly a breakage for backwards compatibility, and I don't know, maybe the best way to go is to manage this with list-management objects. Something like a new method for [list], for instance [list build]... just ideas.

Spacechild1:
@porres 
> maybe the best way to go is to manage this with list-management objects

How? In the context of Pd messages, you can simply use `[list split]`, but I think the point of @myQwil was to introduce a convenience notation.

Although the examples only show usage in Pd messages, I think it could be also very useful for forwarding slices of creation arguments to objects, e.g.` [foo ${1:4}] -> [bar ${4:}]`, because this is something you can't fully achieve otherwise.

> yeah, this is certainly a breakage for backwards compatibility

What about @umlaeute's suggestion to use curly braces?

porres:
> What about @umlaeute's suggestion to use curly braces?

that looks perfect because we never used curlies before, and I still can't use them as well with the latest build

myQwil:
With curly braces, you could probably just do away with the dollar symbol altogether, but I don't know if there's a different usage intended for them.

Speaking of backwards compatibility, what effect would curlies in a patch have on older versions of Pd?

umlaeute:
curlies might get dangerous, as attempting to "print" (e.g. display an object's name) unmatched pairs in Pd<<0.50 might lead to an unresponsive GUI.

as others have pointed out, we still cannot type curlies in Pd, making my suggestion somewhat bogus. i was confusing my attempts to allow printing of curlies with the ability to input curlies.

so, for the problem at hand, i now revise my suggestion to use an even more python-like notation using brackets, e.g. `$[::2]`. this at least should work out of the box :-)

> I don't know if there's a different usage intended for them.

i always imagined curlies to be used for notating nested-list. as in:

~~~
[list i do not like {pizza spinacci} at all(
|
[trigger list list]
|                 [list length]
|                 |7\
[$5(
|
[list length]
|
[2\
~~~

(that's a quick draft; i'd rather have an explicit operation to flatten a nested list; `[$5(` feels wrong)

umlaeute:
thanks for the brackets.

i find the one-based indices a bit odd (as compared to python's list notation), bit i guess this is for consistency with Pd's traditional `$arg` numbering.

### an observation with out of range slices

~~~pd
[list a b c(
|
[foo $[:10] bar(
|
[print]
~~~

- This will print an error message `$10: argument number out of range`, which i wouldn't have expected (python slices silently ignore missing elements, so the above example would just produce `[foo a b c bar(` without an error

- more importantly, the actual output of the above, is `[foo\ $10 bar(` which is most decidedly not what i would have expected. if the slice notation is just a shorthand for naming each element explicitely, it should instead expand to `[foo $1 $2 $3 $4 $5 $6 $7 $8 $9 $10 bar(` with `$1..$3` being expanded to `a b c` and the rest either literal `\$n`-args or just set to `0` (as Pd currently does)

myQwil:
I agree, I'm gonna see about getting it to print as much of the list as possible and then ignore what's out of bounds, as opposed to throwing an error message, and I'll also see about getting it to retain any suffixes and prefixes like a regular dollsym when only the start index is specified.

As far as `$[:].wav` goes. Do you want it to append .wav to the last list item, like `[a b c.wav(`? The earlier version used to do that, but it was moreso a byproduct of piggy-backing off of the binbuf_expanddollsym function. I can look into that too though, if you think that would be a good feature to have.

umlaeute:
i think `$[:].wav` doesn't make much sense. i'm pretty sure it shouldn't only append to the last atom (resp. to the first if you use a prefix).
if you want to add a suffix to the last element you should instead of `$[:-1] $[-1].wav`

so i'd say: simply forbid the use of anything that expands to a list (as opposed to single-atom indexing) in dollsyms.

(i quickly discarded the idea of appending prefix/suffix to *all* elements of the slice, e.g. `./$[:].wav` could but probably should not expand to ~~`./a.wav ./b.wav ./c.wav`~~; use list-comprehensions for that :-))

myQwil:
Okay, I think everything we've talked about thus far has been addressed. I threw in the multiple item append/prepend idea right at the end just to try it out. I know you said "could but should not"  and if you still feel that way after trying it out, I can always roll back the commit.

myQwil:
I've added one last feature. You can now use a custom separator to string together the list items.
It basically works by adding a 3rd colon, followed by the separator character. If nothing comes after the 3rd colon, then it collapses all the items into 1 symbol.
Here's an example:
```
[ list a b c d wav (
|
[ list $[:-1::/] .$-1 (   ==>  "a/b/c/d  .wav"
|
[ $[:::] (  ==>  "a/b/c/d.wav"
|
[print]
```
This would ultimately print `a/b/c/d.wav`.
First, it would get to the 2nd list message and produce a list of 2 items: `a/b/c/d` and `.wav`.
Then, the `$[:::]` message would collapse those two symbols into a single symbol.
If we wanted to, we could have also added dollar prefixes and suffixes.
For example, if the 2nd message was instead `[ list $1$[:-1::/]$[-2] .$-1 (`, the final result would have been `aad/abd/acd/add.wav`
Hopefully, I've gotten everything to format properly but if you notice any odd results, be sure to let me know.

umlaeute:
i think this is getting feature creep.

a generic slice-notation is useful, as it allows to pass sub-lists as arguments to objects within abstractions, something we cannot do right now. e.g.:
~~~
[clone myabstraction $1 $1 $[2:]]
~~~

however, the latest additions are mostly targeted at messages (as opposed to object-arguments), and messages can already be mangled with objects (e.g. the `[list]` family). as such these additions are much less useful.

i would suggest to split this into two PRs, one that only adds the basic list-slice notation, and the other one (on top of the first one) that gives all the extras.
afaict, the first one has a far bigger chance of getting included.

myQwil:
Okay, that sounds good. I should have another pull request up in the near future that'll essentially just be this one in its current state, and then this pull request will be refined to have the original slice concept.

umlaeute:
i'd think it would be better if the "PR with the simple behaviour" only contained commits to get that simple behaviour (rather than have commits that build the entire functionality, and then some extra commits to remove them again).
similarily, the "PR with the full blown behaviour" should build on top of the "PR with the simple behaviour", adding the extra features.

myQwil:
I'll see if I can give the timeline a more natural progression. Some of the changes I made were applicable to both versions, which is why I was hesitant to rollback the commits like I said I would do, but I suppose I can just put those changes in at an earlier part of the commit timeline.

src/m_binbuf.c
```diff
@@ -506,7 +506,7 @@ static int binbuf_expanddollsym(const char *s, char *buf, t_atom *dollar0,
     char c = *cs;
 
     *buf=0;
-    while (c && (c>='0' && c<='9' || c=='-'))
```
umlaeute: the check whether 'c!=0' is to detect a terminating NULL-byte, rather than to prevent null-ptr access.

myQwil: Okay, but I'm a bit confused because if c = 0 then doesn't c>='0' already have that scenario covered?

umlaeute:
sure.
the `c=0` check just makes it explicit (and will probably be optimised away by the compiler anyhow...)

myQwil:
Okay, the commit's been taken out. I should've asked about it before making the change.

PR #876, opened 2020-01-25
Search for hex-formatted filenames when hexmunge is true

myQwil:
While the setup methods for special-character objects are hex-munged, the raw object name still gets passed to the method **sys_trytoopenone**, which means that it searches for a filename with the special characters still intact. Here, the hex-formatted name is passed instead, with the assumption that the filename is also going to be hex-formatted.

A couple of extra conditions for special cases have been added. One allows for a single tilde by itself to be seen as an object name rather than part of a path name, and another allows for a forward slash to be seen as part of an object name if it's the last character in the name.


myQwil:
The changes have been reduced significantly. There are no longer any additional changes for special niche cases.
If an object's name contains special characters, the loader will look for a file with the special characters hex-munged. If no file is found, it will then do a normal search for a file with the special characters still intact.

umlaeute:
iirc (not having checked my old code nor your PR), this is about the functionality the original hexloader patch suggested.
when it was merged, the patch was reduced to only mangle the setup function (and not the filenames)

myQwil:
Yes. Searching for an un-mangled filename works most of the time. Windows/NTFS has a set of forbidden characters that can't be used in filenames, so checking for a mangled version could circumvent that, and the check would only occur if special characters were present in the object name.

Issue #882, opened 2020-01-31
Feature: A hotkey to hide all cable connections


muziker:
Can we have a hot key that will hide all cable connections? That way it's easier to focus on just the ones to connect with.

umlaeute:
how about: automatically graying out connections during connecting (but not hiding completely)

muziker:
yes that would be great

jmmmp:
this could also be solved as part of a patch modding plugin, where each user could choose its own color scheme etc.

PR #887, opened 2020-02-06
background cords while doing connections

umlaeute:
with this PR, existing connections are greyed out while creating a new connection.
once the new connection is done, the other connections are restored to their blackness.

this is to allow the user to concentrate on the connection at hand (and not be confused by all those other spaghettis).

Closes: https://github.com/pure-data/pure-data/issues/882

arguably, people should structure their patches better so they need less connections on a single screen...


porres:
> arguably, people should structure their patches better so they need less connections on a single screen...

I'm with that idea, this enables people to be clumsy, but then, people do crazy spaghettis even not being enabled... 

I'm travelling around and won't have the time to test this soon, but I'd love to

umlaeute:

PR #890, opened 2020-02-09
Python slice notation with extra features

myQwil:
This version of the slice notation has some additional features including:
- Allowing for prefixes and suffixes to be added to each of a slice's items
- Allowing slices to use a custom separator character that combines all of the slice's items into one symbol. This is done by specifying the character after a third colon. If nothing comes after the third colon, the items will be concatenated together.

Some examples, given that we start with the list `[list a b c d (`:
- Sending the list to `[foo$[:]bar (` would print `fooabar foobbar foocbar foodbar`
- Sending the list to `[$[:::/] (` would print `a/b/c/d`

Included is a patch with a couple more examples
[dollar-slice-extra.txt](https://github.com/pure-data/pure-data/files/4176558/dollar-slice-extra.txt)


danomatika:
I think function names like `isnumbr()`, `isqur()`, `binbuf_dlrchr()`, etc are unnecessarily terse and don't match the existing function naming, ie. `canvas_getdollarzero()`.

myQwil:
I suppose I could change the dollar checker names to `isnumber` `issquare` and `isdollar`. I was a bit concerned about `isnumber` already being the name of an apple-specific c++ function, and even though this isn't c++, I still didn't want the function names to be associated with any other commonly used functions. After I changed `isnumber` to `isnumbr`, I figured that the other two functions should follow a similar condensed style, so each had a vowel taken out.

I didn't prepend these functions with `binbuf_` because after making `isdollr` an external function and using it outside of m_binbuf.c, I saw them as being separate from binbuf, more akin maybe to a function like `ftom`. It's still convenient to define them in m_binbuf.c because some binbuf functions use `isnumbr`.

As far as the *chr functions go (`dlrchr`, `sqrchr`, `slicechr`), those were named in reference to `strchr`, although I can definitely see `dlrchr` and `sqrchr` being renamed to `dollrchr` and `squarchr` respectively, since that would make it so they're clearly referencing the dollar checkers that they use.

danomatika:
> I suppose I could change the dollar checker names to `isnumber` `issquare` and `isdollar`. I was a bit concerned about `isnumber` already being the name of an apple-specific c++ function, and even though this isn't c++, I still didn't want the function names to be associated with any other commonly used functions. After I changed `isnumber` to `isnumbr`, I figured that the other two functions should follow a similar condensed style, so each had a vowel taken out.

I agree that we want to avoid name clashes which is why I also suggest having more of a prefix or longer form name. In general, Pd's (read "Miller's") naming style seems to be "prefix_verbsdescribingactiontogether". I try to follow this when I can to at least keep the codebase more or less consistent.

One option for these, as they work specifically with strings/characters is to use a prefix ala `string_isnumber()` or `char_isnumber()`. For things that are simply static helpers used within an implementation file, it's probably less of a problem so `isnumberchar()` is probably fine.

> I didn't prepend these functions with `binbuf_` because after making `isdollr` an external function and using it outside of m_binbuf.c, I saw them as being separate from binbuf, more akin maybe to a function like `ftom`. It's still convenient to define them in m_binbuf.c because some binbuf functions use `isnumbr`.

It follows then that this could be a macro similar to SETSYMBOL or simply named to reflect what it does more specifically, ie. `symbol_isdollarstr()` or `isdollarstr()`.

> As far as the *chr functions go (`dlrchr`, `sqrchr`, `slicechr`), those were named in reference to `strchr`, although I can definitely see `dlrchr` and `sqrchr` being renamed to `dollrchr` and `squarchr` respectively, since that would make it so they're clearly referencing the dollar checkers that they use.

I think my point is that we don't need to be unnecessarily terse to match C/Posix function naming. The Pd naming is already more verbose than that.

danomatika:
One other point: I would double check that the approach for `isnumbr()` work for all number representations, ie. scientific notation or number starting with a period (although I'm not 100% sure about this). As Pd largely uses C printf style, it might be better to let C do the checking for you via sscanf, etc.

myQwil:
Okay, so for the is* functions, we could have:
```
isintegerstr()
issquarestr()
isdollarstr()
```
`isnumber` could be a bit misleading as it implies any real number and we just want integers.

and for the *chr functions, I was thinking of going with this:
```
binbuf_dollarchar()
binbuf_squarechar()
binbuf_slicechar()
```

myQwil:
I ended up using a slightly different naming convention for the is* functions. Rather than putting the "str" at the end, it's at the beginning, and "integer" was shortened to "int", so the names are `strisint` `strissquare` and `strisdollar`, including an additional function `strisdigit`

PR #893, opened 2020-02-11
cache bytespersample/bytesperframe/bigendian in local vars

umlaeute:
using local variables is faster than accessing the struct members.

however, the real reason for this PR is to minimize the diff with the double-precision branch
(merging the two branches `update/soundfile` and `double-precision` still requires user-interaction, but after this PR has been accepted the conflicts are down to two 4-line chunks containing only variable definitions that should be trivial enough to check)

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -320,51 +320,55 @@ int open_soundfile_via_canvas(t_canvas *canvas, const char *filename,
 static void soundfile_xferin_sample(const t_soundfile_info *info, int nvecs,
     t_sample **vecs, size_t framesread, unsigned char *buf, size_t nframes)
 {
-    int nchannels = (info->i_nchannels < nvecs ? info->i_nchannels : nvecs), i;
+    int bigendian = info->i_bigendian;
+    int sfchannels = info->i_nchannels;
+    int nchannels = (sfchannels < nvecs ? sfchannels : nvecs), i;
+    int bytespersample = info->i_bytespersample;
+    int bytesperframe = info->i_bytesperframe;
     size_t j;
     unsigned char *sp, *sp2;
     t_sample *fp;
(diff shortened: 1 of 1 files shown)


Spacechild1:
> using local variables is faster than accessing the struct members.

With the strict aliasing rules, the compiler can easily "prove" that in the given code nothing is ever changing those struct fields, so it will most likely cache them in registers or the stack anyway.

But it makes the code easier to read :-)

danomatika:
Eh, I don't see the need. It just saves you some find/replace during the merge.

I explicitly removed most of these kinds of variables as they are passed in via the struct. In the case of the sample xfer functions, the structs are already cached from the read/write objects in their child threads before these are called.

>But it makes the code easier to read

Really? `bytespersample` vs `info->i_bytespersample`?

umlaeute:
> > But it makes the code easier to read
> 
> Really? `bytespersample` vs `info->i_bytespersample`?

yes. very much.

~~~diff
-                    j < nframes; j++, sp2 += info->i_bytesperframe, wp++)
+                    j < nframes; j++, sp2 += bytesperframe, wp++)
~~~


> Eh, I don't see the need. It just saves you some find/replace during the merge.

hmm well, the idea was to do the trivial "find/replace" *before* the merge to minimize the work *then*.
as in:
- miller starts the "merge-season", while all godly Europeans are fast asleep
 - first, he accepts the *double-precision* branch (because it's less invasive)
 - then he turns attention towards the *update/soundfile* branch
- the MR now yields 13 conflicts
   - none of them is especially scary, but you still have to wade through each of them
- so he defers the merge to *later*
- Europe awakes
- you merge `master` into `update/soundfile`, fixing all conflicts
- but merge-season is over for now

but anyhow, it was just a suggestion.
feel free to close the PR and delete the branch.

danomatika:
We can make a "vetted develop" branch that is pre-prepared with both branches merged.

Issue #892, opened 2020-02-11
Also ship with DejaVu Sans Mono Italic versions


porres:
Pd comes with DejaVu Sans Mono, but it only has normal and bold fonts and it misses italic and italic+bold fonts.

I know Pd doesn't use italic itself, but the thing is I'm working on cyclone's comment and using DejaVu Sans Mono as the default font, so it looks like Pd's comment by default, and then cyclone's comment will now be able to load italic versions of the font, that is if it exists, right now nothing happens cause we don't have it so I'm requesting this feature.

What do you think?

Thanks

porres:
I did a simple test and just included DejaVuSansMono-BoldOblique.ttf and DejaVuSansMono-Oblique.ttf into "font" in Pd's source, compliled it and, voilá, it worked fine for macOS. Wasn't really expecting it'd be that easy, but it still seems it has to be a little bit more elaborate than that. Anyway, will try and see and then maybe even send a PR.

umlaeute:
why can't you just ship the fonts with cyclone?

porres:
I had thought of that and tried it, but sending this or any other font inside the cyclone folder doesn't really work for cyclone/comment (it'd be cool if it did though). Then I tried just adding _DejaVuSansMono-BoldOblique.ttf_ and _DejaVuSansMono-Oblique.ttf_ into "resources/font" in the mac application and it worked. Then also added to the source and compiled, which ended up copying the files to the same folder and also worked then.

Anyway, DVSM  is a dependency for the puredata package in Linux, and it comes with italic, why not also ship then with Pd for macOS? What would be the opposition? Once we're already shipping it (but only half of it).

porres:
I see that in Preferences => Path it says it should also load "fonts", I guess it's cool if you could ship fonts with your patch, so that pd finds it either in the same folder where the patch is, and then via the search paths and etc

porres:
> I see that in Preferences => Path it says it should also load "fonts", I guess it's cool if you could ship fonts with your patch, so that pd finds it either in the same folder where the patch is, and then via the search paths and etc

what I mean is that this is not actually true and not happening right now

umlaeute:
that's probably only there because externals such as Gem search for fonts in the `path` (i don't think that it was *me* who put that in, but the history is lost).

porres:
> externals such as Gem search for fonts in the path

does it also search for them in the patch's directory?

umlaeute:
[...]

PR #899, opened 2020-02-26
Add DejaVu Sans Mono Italic font variants

porres:
This PR closes https://github.com/pure-data/pure-data/issues/892

I'm only adding the italic variants of DejaVu Sans Mono, which is shipped with Pd.

I know Pd doesn't use it itself, but I'm using it for cyclone's comment, and my reasoning is that if Pd already makes the effort to ship this font, there's no reason not to ship it in its entirety.


danomatika:
>I know Pd doesn't use it itself, but I'm using it for cyclone's comment, and my reasoning is that if Pd already makes the effort to ship this font, there's no reason not to ship it in its entirety.

As you've stated, the reason not to ship it is that Pd doesn't use it. However, these aren't very large files so I don't see a big reason *not* to.

Lucarda:
@porres

To be able to use the fonts on Windows include the fonts file names here:

https://github.com/pure-data/pure-data/blob/a12b41a86d046d900843cbaba89592e0f5de4060/tcl/pd-gui.tcl#L408

danomatika:
Add @Lucarda's suggestion and I would be fine with this PR.

porres:
@Lucarda - done with https://github.com/pure-data/pure-data/pull/899/commits/5ed47a64e20f067db3a77fda4b8f15a91b7798ee you may also test the new cyclone from the repository and help file of comment ;)

danomatika:
@millerpuckette This PR would be goo to add for 0.51, if possible. It's small and fleshes out the DVSM font support.

Lucarda:
Works Ok on Windows.

pdfontloader does it job:

pdfontloader-loading

and [cyclone/comment] using DejaVu bold+italic:

cyclone-comment

PR #905, opened 2020-03-07
Fix macros for MSC

### LINKED ISSUE #904: Unallowed definition of dllimport function rint() for Visual Studio 20xx >= Visual Studio 2013
The function *rint() is available from Visual Studio 2013 (_MSC_VER 1800)*. See  [comment](https://github.com/jlblancoc/suitesparse-metis-for-windows/issues/30#issuecomment-258822447) for similar issue in another project.

etk70182:
Fix macros for Visual Studio version >= Visual Studio 2013. This fixes #904

### DIFF
--- a/src/x_vexp_fun.c
+++ b/src/x_vexp_fun.c
@@ -537,13 +537,14 @@ ex_toint(t_expr *e, long int argc, struct ex_ex *argv, struct ex_ex *optr)
                 FUNC_EVAL_UNARY(left, toint, (int), optr, 0);
         }
 
-#ifdef _MSC_VER
-/* rint is now advertised as part of the microsoft SDK but my MSVC still
-doesn't find it - so here it is again. */
+#ifdef __MSC__
+#if (_MSC_VER < 1800)
+/* rint is not available for Visual Studio Version < Visual Studio 2013 */
 static double rint(double x)
 {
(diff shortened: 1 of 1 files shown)


src/x_vexp_fun.c
```diff
@@ -537,14 +537,15 @@ ex_toint(t_expr *e, long int argc, struct ex_ex *argv, struct ex_ex *optr)
                 FUNC_EVAL_UNARY(left, toint, (int), optr, 0);
         }
 
-#ifdef _MSC_VER
-/* rint is now advertised as part of the microsoft SDK but my MSVC still
-doesn't find it - so here it is again. */
+#ifdef __MSC__
```
umlaeute: hmm, i cannot find the `__MSC__` macro on https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros (which i think is a canonical list).
probably this should just be replaced with `#ifdef _MSC_VER`

Spacechild1: > probably this should just be replaced with `#ifdef _MSC_VER`

Definitely!

etk70182: Good point! Thanks you.

umlaeute:
i squashed to two commits and pushed them to the `update/0.50` branch.

Issue #903, opened 2020-03-06
incoming UDP packets truncated to 1000 bytes


umlaeute:
given a patch:
~~~
[netreceive -u -b 9000]
|
[list length]
|
[print]
~~~

resp.
~~~pd
#N canvas 5 51 450 300 12;
#X obj 102 75 netreceive -u -b 9000;
#X obj 102 100 list length;
#X obj 102 125 print;
#X connect 0 0 1 0;
#X connect 1 0 2 0;
~~~

if i feed this patch 2048 bytes using the following command, it will only receive `1000`.
~~~sh
$ dd if=/dev/zero bs=2048 count=1 | nc -u localhost 9000
~~~

if i instead replace `[netreceive -u -b 9000]` with iemnet's `[udpreceive 9000]`, i receive the 2048 bytes just fine.

this might be a regression triggered by #715

Spacechild1:
Can't reproduce here (on Windows). Are you sure that you're using current master? I actually *increased* the buffer for binary netreceive from MAXPDSTRING (=1000) to INBUFSIZE (=4096).

With current master I can send/receive UDP packets up to 4096 bytes. Above that, Pd will segfault (which is another issue, I'll push a fix).

umlaeute:
i was using Pd-0.50-2 (not `master`), and can confirm that with master the limit is 4096 bytes.
it doesn't crash here if i send larger packages (and `valgrind` only complains about a few lost bytes but no other memory problems), the data is just truncated at 4096 bytes.

however: a UDP package has a possible maximum length of 65535 bytes. it there any reason to not allow this maximum size?


the original issue (limit=1000) is therefore already fixed. however, since the follow-up issue (limit=4096) is almost identical, i'll leave this ticket open and re-title it.

Spacechild1:
> it there any reason to not allow this maximum size?

I think it just makes the implementation easier. Having a 65KB buffer on the stack doesn't feel right. There are of course other solutions:
a) query the size of the packet in advance with `socket_bytes_available` + `alloca`
b) allocate a single buffer on the heap in `_instanceinter`, just like `i_inbinbuf`

On the other hand, the limit for binary UDP packets has been 1000 bytes like forever and nobody seemed to notice, let alone complain. Now the limit is 4 times as high. Of course the *theoretical* limit for UDP packets is 65535 bytes, but in practice it's not recommended to send UDP packets larger than the network MTU. I'm not against supporting larger UDP packets, but maybe wait till someone really needs it?

[...]

PR #910, opened 2020-03-19
Increase socket receive buffer sizes

Spacechild1:
This PR increases the buffer sizes of all socket receivers to 65536, so we can receive large UDP packets. It is based on the current update branch to avoid merge conflicts with recent changes in the networking code.

One thing I'm not happy about is that in `socketreceiver_doread` we have a huge buffer on the stack. I think it won't do much harm, though, because messages are passed only after the function returns.

This implements https://github.com/pure-data/pure-data/issues/903

I verified that I can now send and receive UDP messages up to approx. 65355 bytes with `[netreceive -b]` and `[netreceive -b -u]`.

Note that this only makes sense via localhost. Actual networks will fragment and might even drop such large packets.

### DIFF
--- a/src/x_net.c
+++ b/src/x_net.c
@@ -22,8 +22,6 @@
 /* print addrinfo lists for debugging */
 /* #define PRINT_ADDRINFO */
 
-#define INBUFSIZE 65536
-
 static PERTHREAD unsigned char *netreceive_buf;
 
 /* ----------------------------- helpers ------------------------- */
@@ -132,10 +130,10 @@ static void netsend_readbin(t_netsend *x, int fd)
     while (1)
     {
         if (x->x_protocol == SOCK_DGRAM)
-            ret = (int)recvfrom(fd, inbuf, INBUFSIZE, 0,
+            ret = (int)recvfrom(fd, inbuf, NET_MAXBUFSIZE, 0,
                 (struct sockaddr *)&fromaddr, &fromaddrlen);
         else
-            ret = (int)recv(fd, inbuf, INBUFSIZE, 0);
+            ret = (int)recv(fd, inbuf, NET_MAXBUFSIZE, 0);
         if (ret <= 0)
         {
             if (ret < 0)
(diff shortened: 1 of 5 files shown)


danomatika:
Maybe share the BUFSIZE #define by declaring it in `s_net.h`? I'm never a fan of redundnant defines, there should be only 1 when possible.

Spacechild1:
Good idea! Did you close by accident? :-D

src/x_net.c
```diff
@@ -918,16 +920,25 @@ static void netreceive_free(t_netreceive *x)
...
         (t_newmethod)netreceive_new, (t_method)netreceive_free,
         sizeof(t_netreceive), 0, A_GIMME, 0);
+    class_setfreefn(netreceive_class, netreceive_class_cleanup);
     class_addmethod(netreceive_class, (t_method)netreceive_listen,
         gensym("listen"), A_GIMME, 0);
     class_addmethod(netreceive_class, (t_method)netreceive_send,
         gensym("send"), A_GIMME, 0);
     class_addlist(netreceive_class, (t_method)netreceive_send);
+
+    netreceive_buf = malloc(INBUFSIZE);
```
umlaeute: i have no clue about `PERTHREAD`, but is this guaranteed to allocate the buffer for each instance?

shouldn't there be a `(netreceive_buf==NULL)` check in `netsend_readbin()` instead

Spacechild1: you're totally right! d'oh... Actually, I have to do the same thing as in d_fft.c. Thanks for catching that!

Spacechild1: Just pushed a fix. Note that we still leak memory if `netsend_readbin()` is called from a different thread than `netreceive_setup()`. This can only happen with libpd, but not in "regular" Pd. The same problem also exists for `ooura_init` / `ooura_term`. 

In both cases, I'm asking myself whether it might be better to allocate the buffers directly somewhere in the Pd instance... e.g. `_instanceiter` for netreceive and `_instanceugen` for fft.

danomatika:
It's pretty possible for libpd_pollgui() to be called in other threads, so it would be good if there is some way to protect `netrecieve_readbin()`. This is also possible via multiple instance support which could be running in multiple background threads...

Spacechild1:
This should work (it just won't free the memory). It's basically the same as the ooura FFT hack. But now I really think this should be a buffer in `_instanceiter`. And ooura FFT would need to be fixed as well...

Spacechild1:
Ok, moved the buffer to `_instanceiter`. Should now work with multiinstance libpd without leaking memory.

Spacechild1:
CI fails because of timeout on macOS Travis build...

Spacechild1:
@umlaeute Is there a reason why you closed this?

umlaeute:
not that I remember

Spacechild1:
ok, the target branch has been deleted... will open a new PR, as GitHub doesn't allow me to change the target branch.

umlaeute:
actually it should allow to change the target branch (but not from my mobile).

Spacechild1:
this works with an *open* PR, but not with a closed *PR*. This means I can't reopen this PR because the target branch is gone, but I can't change the target branch because this PR is closed. Catch 22.

Anyway, I'm using the opportunity to squash the commits and add a small improvement.

PR #918, opened 2020-03-28
Feature/knob

### LINKED ISSUE #911: feature: add a knob GUI

Ant1r:
closes #911
demanded by the crowd, here is a proposal to include moonlib/mknob in vanilla


porres:
Hi there, as mentioned in https://github.com/pure-data/pure-data/issues/911 - here are some remarks/requests and issues I have.

- change object name to just "knob":
   I know it is a silly thing, but I don't see the point of having the "m" before it. Was this named like this to mean "moonlib" "knob"? Anyway, more than just getting what we have from [mknob] and just including it as it is, I propose a few changes so that we take this opportunity to think about a knob GUI for Vanilla. It could still be pretty clear that the object is based on moonlib's [mknob], and is authored by you, as a modification of iemgui's slider, but I think it is good that we make a distinction that would also allow us to change something (or some things) that would end up making the objects incompatible. And in this way, you can still provide [mknob] in a library for people that use it as it is for backwards compatibility. 

   I have more suggestions about design changes, but before I take some time to better think about them and list them, I wanted to make this remark on the object name. Let me know what you think. And now, more importantly, let me point out a few bugs/issues:

- Edit mode outline not working:   
     Help file says note: "when selected, mknob shows its inlet, outlet and a rectangular outline, in order to help connecting. When edit mode is switched off, they disapear after the mkob is hovered by the mouse or the knob value is changed." But what I experience is that when I get out of the edit mode, the outline and inlet/outlet still persists. And perhaps I could actually propose here a design change. I think it'd make more sense if the outline and inlet/outlet would show when you get on the edit mode, regardless if the object is selected or not. This is what already happens, for instance, with comment boxes, where it shows you where the end of the line is.

- Hide inlet/outlet when you have send/receive name   

     As usual with [iemguis], when we set a send name, it hides the outlet, or inlet gets hidden with a receive name. I don't see this happening in this object and I think it should for consistency. 

- inlet/outlet size  
     It seems the inlet/outlet height is smaller than it should.
Anyway, like I said, these are early remarks, I have more on the way.

Thanks

porres:
- by the way, help file needs to be rewritten, and I'm working on that too

porres:
So here are some thoughts on the design that I think are more important.

- Infinite knob:

I really like that this object allows us to have a so called "infinite" knob option, namely "angular" control.But I think it has an overly complicated number of options. We have three, a non angular motion and two angular motions. I propose we just have a single angular motion option.
Now, the full angular motion option has this differentiated design where the "start" of the knob meets the "end" of the knob, and I actually like this option, so the other "angular" one seems weird to me and it makes a rather weird infinite knob where it jumps from one point to the other. But I see that maybe it is good because it has the same design of the non angular motion, where the start of the knob or the "bottom" value is quite appart from the end. Now, this design on its own is good cause it gives us a visual cue where the start and end of the knob is, but I have another suggestion for that.
So, I propose we just have two options, an angular and a non angular option (the "full" one). Moreover, I also propose that the non angular motion has the same "full" range, where the start point of the knob is the same as the end of it. 

- Visual cue:
Now, the issue here could be that we may lose the visual cue for the non angular motion, and for that I propose another visual cue, where we have a thicker line, in the same color of "front", showing us then if the knob is fully down or fully up.

- Sensibility value and message:

There's this mouse sensibility value that I don't think is very helpful, it's kind of unnecessary and other GUIs in Pd, like the slider, are just fine with a hardcoded single value for that. This would also introduce an inconsistency to the other GUI objects. I just think it makes sense to have a hard coded reasonable value and make use of the "press shift" option for fine tuning.
Moreover, I don't see the point of having this parameter as a second argument to the "size" message. It really seems just like a separate parameter. But one way or another, I'm just proposing we get rid of this sensiblity parameter...
Nonetheless, this parameter still controls if we have a full angular motion or not, so we could kinda "keep" it just for that, but as a separate message, that is like a toggle "1" for full angular (or "infinite") and "0" for not. In the properties window this could also be replaced by a toggle check box.

Ant1r:
Hey @porres , i'll try to answer one point after the other:

- **change name:**
I agree that reusing **`mknob`** is problematic regarding backward compatibility of moonlib/mknob.
Moreover I've just checked that, using present feature/mknob branch, once you create a first `[moonlib/mknob]`, every subsequently created `[mknob]` will be moonlib's version (aka definitive overloading). 
For the very same reason, I think using **`knob`** is a problem for the future of flatgui/knob.
Considering that almost nobody actually sees the real name of the object, we could use virtually any name. So why not using something like **`pknob`** (for "Pd knob") , or **`vknob`** ("vanilla knob")? If anybody has a better idea I'm all OK.

- **edit mode outline not working:**
I think it works as explained, at least it does on my machines; you say that
>  when I get out of the edit mode, the outline and inlet/outlet still persists

but the help clearly says that the outline and iolets will be removed only _"after the mkob is hovered by the mouse or the knob value is changed"_ ; so you don't seems to report an actual bug.
The current behavior certainly isn't ideal, but the design change you're proposing is not feasible, because of certain Pd's limitations. For a deeper discussion about that, refer to e.g https://lists.puredata.info/pipermail/pd-list/2018-09/123359.html

- **hide inlet/outlet when you have send/receive name:**
I definitely don't like the "floating" iolets so I didn't want to see them when not in edit mode. But I could change my mind, and draw both iolets **and** outline when send or receive is not defined. So it would look weird in this case.

- **help file needs to be rewritten, "I'm working on that too":**
with pleasure! You could send PRs to my [fork](https://github.com/Ant1r/pure-data/tree/feature/mknob) (is this the right way?).

Ant1r:
(continuation of my answer to @porres)

- **inlet/outlet size:**
well I'm using reguar IEM_GUI_IOHEIGHT, but I agree it's visually thinner. I'll try to understand why.

- **infinite / visual clue /sensibility:**
you certainly understand that the whole mknob behavior is a compromise, and things are not as flexible as we'd want to (be sure I've been asking myself these questions for a while...).
For e.g your "**thin bar separation**" proposal, linked to a 360° rotation, doesn't distinguish between fully-off and fully-on cases. So I think, for non-angular motion, it's really better to keep 270° rotation as a default.

But I would agree to simplify the **angular mode cases**, maybe the "270° rotation + angular control" is unneeded, and could be removed (BTW this would result in a few nice code simplifications, by removing the `height=-1` case).

As for the **sensibility**, I'm not sure we can easily remove this setting. Consider very small or very large knobs, what should be the right "full-scale mouse distance" that would work for all? 100px ? 15px ? 500 px?... I agree in most cases the default 100px is working well, but for extreme sizes it's less obvious.

OTOH replacing the "mouse" setting by a **toggle check box** in the properties dialog, and/or adding a second one, is far from being trivial. AFAIK it would mean rewriting the whole Tcl/Tk interface related to properties management, specifically for this knob. I think it's better to re-use existing code when possible (even it's more possible now than in the moonlib context).

Sorry to reject most of your points Alex ;-)
But please let's continue until we agree on a long-term solution.

porres:
> change name:
> I think using knob is a problem for the future of flatgui/knob.

Yeah, I wasn't caring about name clash with flatgui, but I can see the concern.

> Considering that almost nobody actually sees the real name of the object, we could use virtually any name. So why not using something like pknob (for "Pd knob") , or vknob ("vanilla knob")? If anybody has a better idea I'm all OK.

Yeah, people don't see it and they'll get it from a shortcut or the put menu... so name is mostly important for people who will create the object via dynamic patching or something... I guess I care a lot about names, and I just thought 'knob' was the best one to the cost of name clashing. Well, this is not the main debate I guess, and I'm just glad you agree we can name it something else.

As for my name suggestions, I think I have better ones, maybe something like "dialknob", and "dknob" for short as an alias (I guess all iemguis have shortname aliases). Since I'm considering this as a new addition to the "iemgui library", another option could be "iemknob".

porres:
> edit mode outline not working:
but the help clearly says that the outline and iolets will be removed only "after the mkob is hovered by the mouse or the knob value is changed"

oh, I get it now...

> The current behavior certainly isn't ideal, but the design change you're proposing is not feasible, because of certain Pd's limitations. For a deeper discussion about that, refer to e.g https://lists.puredata.info/pipermail/pd-list/2018-09/123359.html

I'll check and look closely, but I think there's gotta be a way, or at least a good way. In fact, I'm actually already doing that for an external of mine, see "else/pic". It will show an outline and in/outlets when you get in and out of edit mode. I guess I'm just doing a dirty hack for that, I know... but it seems to work fine and I was also supposed to ask to the dev list if there was a better way. And what makes me think there's a better way to do it is because of the way Pd's comment work. They seem to know when the edit mode is on or not and visually react to it.

porres:
> hide inlet/outlet when you have send/receive name:
> I definitely don't like the "floating" iolets so I didn't want to see them when not in edit mode. But I could change my mind, and draw both iolets and outline when send or receive is not defined. So it would not look weird in this case.

I wasn't really saying inlets and outlets should be visible when not in edit mode, and only hidden if there was a send/receive name... I was going with you in the idea of only showing them in edit mode, but also hiding them in this case only for the sake of consistency with other iemguis.

But on second thought, maybe it would be just simpler to not have an outline at all and just always show the inlet/outlet and hide them when there's a send/receive name. By the way, this is how "flatgui/knob" works. And maybe that's in fact a simpler and more consistent approach, because inlets and outlets always bug the hell out of me in iemguis and I'm ALWAYS just giving the objects dummy s/r names just so they get hidden.

porres:
> infinite / visual clue /sensibility:
> you certainly understand that the whole mknob behavior is a compromise, and things are not as flexible as we'd want to (be sure I've been asking myself these questions for a while...).
> For e.g your "thin bar separation" proposal, linked to a 360° rotation, doesn't distinguish between fully-off and fully-on cases. So I think, for non-angular motion, it's really better to keep 270° rotation as a default.

I'm not sure you understood my proposal. I proposed that the GUI would display the position by also "drawing" in the circular line. Let me try and be clearer about it. In fact, this is how "flatgui/knob" works as well. For instance, if you have a "green" as the front color, you'll have something like this 

Screen Shot 2020-03-30 at 00 11 00

So you will know if you have a full-on case. The thing is that this is not apparent at all if the front color is black. And for that I just proposed we had a thicker black circular line.

So yeah, this way, we can have both the angular and the non angular as a full circle, and I just have a very big problem with the angular mode at 270º because it doesn't really work well - the value jumps in a rather big leap from one point to the other and you lose angular resolution.

porres:
> As for the sensibility, I'm not sure we can easily remove this setting. Consider very small or very large knobs, what should be the right "full-scale mouse distance" that would work for all? 100px ? 15px ? 500 px?... I agree in most cases the default 100px is working well, but for extreme sizes it's less obvious.

I didn't try yet, but I'll take your point. Nonetheless, it would still make a lot of sense to have this parameter as a separate message other than the 2nd argument of the size message, right?

> OTOH replacing the "mouse" setting by a toggle check box in the properties dialog, and/or adding a second one, is far from being trivial. 

I was kinda assuming that after I wrote, so yeah, I guess I get it... 

Not the greatest concern, again, but I just suggest "0" would set it to a full angular motion, and that this could be set as a separate method.

As you can see, in the object I was working on, I was trying to get the best ideas of both knob objects available and still think of other improvements.

Cheers

porres:
>> help file needs to be rewritten, "I'm working on that too":
> with pleasure! You could send PRs to my fork (is this the right way?).

let's work these other issues first, we can do this for last, I'm glad to do it, I've been working hard - by the way - on dealing and revising basically every Pd help/documentation file

Ant1r:
I finally had some inspiration from @porres  's comments ;-)

I introduced `full` (e.g 360° rotation) and `sensitivity` variables and added corresponding input messages, so I think things are clearer now about different modes. If `sensitivity==0` then the mouse control is angular (-1 value is no more used). `full` rotation can now be used for either mode.

I adapted the properties dialog accordingly, and updated the help patch.

I didn't change the way iolets and outline are drawn though. I like it as it is, but it could be changed in a way they are always drawn when receive or send is not defined, if preferred.

porres:
> help file needs to be rewritten, "I'm working on that too":

here are a few of quick things you could change yourself right away: 
- Delete "To create a new knob, just add an object (Ctrl+1) and type: [knb]" - this is not how other iemguis objects are documented and how people will create this is mostly just by using the put menu or shortcut. The object name is for advanced users and is specified in the label of the canvas in green as the other objects.

- Delete "Just an adaptation of musil's slider.", it's unnecessary information, maybe more pertinent when this was an external.

- Delete [meta], cause Pd vanilla's documentation doesn't have those.

porres:
Ok, I see an issue now with full 360º angular control I hadn't noticed before.

The value of zero is a bit tilted to the left, while a full maximum value is not tilted at all, forming a perfect 90º angle to the horizontal line. I would expect a perfect 90º angle to correspond to 0 and not the maximum value. As for the maximum value, I'd expect it to be a bit tilted to the right as the last possible value.

porres:
> I finally had some inspiration from @porres 's comments ;-)

great, thanks, I like where this is going :) 

As for the full angular control and not being able to see wether it is at full on or full off, I guess that my last comment points to the idea that we can actually perceive, if we look closely enough, that the knob line is a bit tilted or not as a visual cue. Nonetheless, my idea of drawing a thicker line, in the same fashion as [knob] works, seems like a better visual cue, what do you think?

And besides this, just to keep a track and record, from my previous comments, we'd still have to: 
- rethink how the object visually behaves when going in and out of edit mode, 
- hide inlet/outlet with s/r names & 
- check/fix how inlet/outlet is displayed.

Ant1r:
- help file: OK I'll do what you propose
- 360° zero-position: yep I'll see what I can do
- thicker line arc: OK I'll see if possible and looks good
- last 3 points (io/sr/edit): I said, I'm not sure I want to change it. But I'll think about it.

jaffasplaffa:
I have not used the older knobs in PD so much yet, I tried them shortly and was a bit confused. I dont remember which one I tried. 

But I just read what you guys have been talking about here, it all sounds great :)

Ant1r:
So I've implemented a "style" feature, refer to #919. The present PR will be closed, excepting if someone strongly objects against this idea.

porres:
If the new PR has all that this has and more, maybe we can close this alright.

I just think that we aere adding lots of complexity and design options though, and we could maybe just have a single one that suits all needs.

megrimm:
hello,

Thanks for all the work! This is really great!

Just a few thoughts. 

1) for consistency sake, the inlet/outlet/box SHOULD be visible in performance mode AND edit mode. this continues visual and functional consistency with other GUI objects (vslider, hslider, etc.)

2) inlet/outlet boxes should only be hidden when a send/receive symbol is added to the object properties.

3) box should ALWAYS be visible even in performance mode.

4) keep colors/fancy stuff and simple and stripped down as possible meaning all unnecessary fluff that does not add to functionality.

5) we should stick to one mode to minimize complexity like 360 mode.

This does look great and the work is fantastic! I fear though, to get it into vanilla it really needs to be as simplistic as possible. once it is in vanilla slight tweaks, added functions, etc can slowly be introduced. To maximize the chance to get into vanilla, which is what we all want here, I fear the object "as-is" is just too complex.

cheers
m

porres:
maybe we should discuss this in the new PR, but I wanna say I agree with @megrimm 

I see a strong inconsistency issue here with how the new knob object is handling the connections, inlets, outline. And it is not a matter of taste I guess, the consistency factor should prevail as to not have an "odd one out".

I was proposing a better way to show and hide outline + inlet/outlet. The way it is seems complicated (select / unselect and send a value to knob). But coming to think of it, having connections being made to thin air in Pd is not common and it'd be very weird to have it in vanilla.

So it would be quite simpler and more straightforward to always have an outline, and inlet/outlet can be hidden as all the other iemguis. This makes sense.

I also fear that a too complex object might get resistance from @millerpuckette in getting in. And it is true we can try and add things later with more discussion. The 270º mode doesn't sit well with angular control, and now we have a way of differentiating a full of to a full on in 360º mode, so if using just one is simpler and better, 360º seems to be the choice.

So yeah, I guess I agree 100% with Mark. 

I'll take this discussion to the new PR though

porres:
so, let's close this in favor of https://github.com/pure-data/pure-data/pull/919 ?

Ant1r:
yes, I'm closing in favor of #919.

PR #919, opened 2020-03-31
Feature/knobstyle

### LINKED ISSUE #911: feature: add a knob GUI

Ant1r:
closes #911 
this is an additional step to #918 , implementing 3 different styles for the proposed knob GUI.
Capture du 2020-03-31 16-00-52

@porres (and anybody else of course!) tell me if you like the styles, and if you'd see other ones that would be good to have.


umlaeute:
please don't use a number for the style in the user-facing dialog.
instead use some fancy names and a pull-down menu (or a radiobutton) to select them.

umlaeute:
i just saw that the style-ids are exposed via the message-API as well; i guess you should change that to symbols too.

i don't really see anything that would mandate numeric styles (e.g. bitfield fuddling).

other GUI-objects in Pd have plenty of problems because they attempted to cram things into numbers (see colors in iemguis and scalars). let's not get there again.

umlaeute:
some more remarks:

### styles
it might also be nice to have a simple `arc` style (like the `wiper+arc` but without the wiper; but not like `filled arc`)

### sensitivity
`sensitivity` is probably bad naming, as it isn't really clear why "sensitivity=0" would give me angular control. also, what if i want **horizontal** scrolling, rather than vertical?

### arc-range
- i would like to be able to have the starting point to be on top of the knob. or on a side. or somewhere else really.
- i would like to be able to define the allowed ark, e.g. going from -90° to +90° rather than from -180° to +180° (as in 360° mode) or from -135° to +135° (as in 270° mode).
  or something else, really

the two options 360° and 270° might be nice, but there are many more interesting layouts:
C

so instead of being able to select between two options (270° vs 360°) i'd rather want to be able to select a startpoint and and endpoint of the arc. (e.g. `[arc -135 135(`)

### +- knob

- for a pan-pot, i would like to be able to have a knob where i can drag the arc in the positive and in the negative direction. that is: the arc can extend to either side of the pointer.
AB

one idea to implement this is to us the "0" value (if upper and lower limit are of opposite sign) as "no arc", the lower-limit value as startpoint (as defined in the "arc-range" proposal) and the upper-limit as the endpoint (ib.)

porres:
I see an issue with style 3 (wiper+arc), the arc doesn't really fit into the circle and seems to have a small offset - this is more apparent in smaller knob sizes. Check below

Screen Shot 2020-03-31 at 12 11 05

porres:
> it might also be nice to have a simple arc style (like the wiper+arc but without the wiper; but not like filled arc)

I agree! I also had the impression the wiper doesn't look good with the arc, so I think it'd look better if we had just a straight line and the arc.

porres:
> sensitivity is probably bad naming, as it isn't really clear why "sensitivity=0" would give me angular control.

I had proposed angular control as a check box, or a toggle. And now I guess it could be something like the [lin] parameter, which is kind of a toggle, that you can click and change the state. The problem is that a full angular control (for an infinite knob) makes the sensitivity parameter pointless, but I guess these can be taken apart even so... if well documented.

While we're at it. I tried bigger and smaller knobs and didn't really feel the need to set a sensitivity parameter, so I still wonder if it is necessary.

porres:
@umlaeute , being able to set a mid point like you proposed so we can implement a pan knob is a great/awesome idea! Thanks for that! 

I have another crazy idea (and I'm the one saying we shouldn't get crazy in adding complexity). How about those knobs that have like only 8 points or so, to select presets (you know, like selecting waveforms and stuff). This would be the same as a radio button I guess, so not the only way to have something like this - but kinda nice...

porres:
> While we're at it. I tried bigger and smaller knobs and didn't really feel the need to set a sensitivity parameter, so I still wonder if it is necessary.

One way or another. I see a bug in the sensitivity parameter. If I change it via messages, the knob position changes! And it should remain in the same spot!

porres:
I'll just bring to this PR other discussions we were having in the last one.

I love that we are checking many options and discussing what's important to have, but maybe it'd be good to remove some of the complexity and stick to fewer options and simpler design. There's always a place for a super fancy knobs with countless visual and other features as an external I guess.

@megrimm and I agree we should have an outline and inlet/outlet always, for the sake of simplicity and consistency. It's just weird to allow a connection to the knob being made to thin air. And Inlet/outlet can be hidden with s/r names as usual.

Another bug is that the height of inlet/outlet is smaller in knob than in other objects.

umlaeute:
> How about those knobs that have like only 8 points or so,

i guess this translates to: have a knob where you can only select steps and not continuous values?
if so, this can easily be done with an abstraction.

[iknob.zip](https://github.com/pure-data/pure-data/files/4411044/iknob.zip)

### regarding feature creep

afaict i haven't really asked for additional settings, but rather for making existing settings more powerful, resp. for re-designing them to make them more powerful.

- setting the origin: the proposal is to overload the semantics of the output range
- `[full 1|0(` -> `[arc <start> <stop>(`
     i think it's easier to understand the semantics of setting the span of the full arc that selecting a "full mode" (what is "non-full" anyhow)

porres:
> i guess this translates to: have a knob where you can only select steps and not continuous values?
> if so, this can easily be done with an abstraction.

**Nice!!!**

> afaict i haven't really asked for additional settings, but rather for making existing settings more powerful

True, I was the one asking for less, settling with just a single full 360º mode. But I guess we can get this setting and allow the user to set a non full mode with an angle span of choosing, plus starting point (such as "top" for panning). So yeah, I can now see the advantage of having more settings for a more versatile knob.

And here's a new idea! 
 - For a non full range, if we're gonna make it customizable, we could have a tiny line as a mark for start and end points.

> @porres (and anybody else of course!) tell me if you like the styles

Just so it is clear, I didn't like the "style 2" (filled arc) at all - that just doesn't look like a knob to me! Looks more like a pizza diagram, and if you have a full on knob, it just looks like a filled circle. But I did like how it had only a single line for the knob point and not a "wiper" (don't know what "wiper" is, but I'm assuming it refers to the more triangular shape rather than a line). This is mostly for the 3rd style ("filled arc"), as I felt it didn't look great and could look better that way. 

In this context, as far as settings go, I was settling just for one idea (the filled arc one), but I can see how one could choose not to have the arc to implement something like a "select knob" (a knob where you can only select steps and not continuous values). So this would require at least two modes.

If a straight line feels to me like a better design for the filled arc, I don't have a problem with "wiper", it does look nice indeed. One could have just a straight line and have a parameter for "arc filling" instead. It would make it simpler and the code cleaner - but maybe these two rather different design styles are fine if the wiper design is already too cherished and stablished.

>  It's just weird to allow a connection to the knob being made to thin air. 

I'm also thinking that a bang GUI is a nice example. All we actually needed was a round button, but there's an outline box so we can make connections. This would make then the knob GUI quite similar in shape to a bang, so I'd also suggest that the color outside the circle would the same as bacground's color.

porres:
> we could have a tiny line as a mark for start and end points.

something like this (which is for the 270º)

Screen Shot 2020-03-31 at 17 46 21

Ant1r:
### sensitivity
I agree it isn't the right name, because when its value increases the actual sensitivity in fact decreases... So, if this setting is kept, it will have to be renamed, suggestions are welcome.
I personally don't dislike the "zero singularity" for angular control. But it could be replaced by a "control linear/angular" toggle.
As for horizontal scrolling, it's already working right now; maybe it's not totally intuitive (the greatest of absolute deltax and deltay wins)...

### arc-range / +-arc
I'm all OK for this. This needs changes in the tcltk interface though, as we need 2 parameters (start-angle / end-angle) instead of 1 (full).

### when to draw outline and inlet/outlet
I would really prefer to keep the ability to hide the outline when iolets are not there. There aren't so many ways to simply draw a circle in Pd... However outline should continue to be drawn when in edit mode, to help the positionning (it isn't easy to align circles or equally space them).
Although I agree that iolets and outline must always be drawn (I mean also in performance mode) when receive or send aren't defined.

### styles
Maybe we could split the style setting in two ones: arc setting and wiper setting (btw: I found the word "wiper" in the wikipedia page for "potentiometer"; sorry for electronics distortion... help finding a better name! "cursor"?)
Thus is would make easier to find the settings names, e.g 
Arc : none / thin / bold / half / full
Wiper : none / triangle / thin / bold

### ticks
Why not adding a "marks" additional boolean parameter. In case of a +-knob it could additionally show the zero-value-position.
But again: it would be good if it stays optional, to keep being able to draw a simple circle.

Well... all of this indeed adds some complexity... Properties dialog would have to be tweaked quite a bit.
I count potential 4 more parameters ("full" and "style" would be discarded):
- linear/angular (bool)
- start-angle (int)
- end-angle (int)
- arc-style (symbol)
- wiper-style (symbol)
- ticks (bool)

My opinion is that it's good to have a more configurable widget; it also can be useful for graphic composition. And after the widget is released, you have far less freedom for adding new features, for compatibility reasons.

umlaeute:
> ### sensitivity
> 
> I agree it isn't the right name, because when its value increases the actual sensitivity in fact decreases... So, if this setting is kept, it will have to be renamed, suggestions are welcome.

no suggestion here. just a note that my `arc` proposal probably needs a better name too :-)

> I personally don't dislike the "zero singularity" for angular control. But it could be replaced by a "control linear/angular" toggle.

i also don't have a problem with the singularity in the origin. i don't understand the 2nd sentence though.

> ### when to draw outline and inlet/outlet
> 
> I would really prefer to keep the ability to hide the outline when iolets are not there.

i tend to agree: *iff* both iolets are not there (because there are send/receive names specified), the outline might go away.

> ### styles

> Arc : none / thin / bold / half / full

how about allowing the userto specify the thickness rather than prividing presets?

| value | effect | example |
| --- | --- | --- |
| 0 | no arc | none
| 1 | thin arc ("1 pixel") | thin
| 5 | thicker arc ("5 pixels") | thick
| <size/2> | full arc | full
| -5 | thick  ("5 pixels") inverted arc | inverted

> Wiper : none / triangle / thin / bold
> ### ticks
> 
> Why not adding a "marks" additional boolean parameter. In case of a +-knob it could additionally show the zero-value-position.

how about making it an integer, rather than a boolean parameter? so you can specify the spacing?

|  value | effect |
| ------- | ------ |
| 0 | no ticks
| 1 | zero mark (or lower limit?)
| 2 | lower/upper limit
| 5 | lower/upper limit and 3 ticks between them

this does **not** allow to set a tick on lower limit, zero and upper limit if the layout is not symmetric.

> But again: it would be good if it stays optional, to keep being able to draw a simple circle.

:+1:

Ant1r:
> height of inlet/outlet is smaller in knob than in other objects.

I already said I'm using the regular IEM_GUI_IOHEIGHT constant; I don't see what you're describing here, e.g here is part of a hsl over a hradio over a knb :
Capture du 2020-04-01 12-01-55


> the arc doesn't really fit into the circle and seems to have a small offset - this is more apparent in smaller knob sizes. Check [screenshot] below

i can't see that either in my machine (Ubuntu 16.04). Can Tk drawing be inconsistent across OSes sometimes? 
Capture du 2020-04-01 11-55-22

> specify the arc thickness rather than providing presets
> making "ticks" an integer, rather than a boolean parameter

I agree, I had thought of these, too.
Maybe "ticks=3" could be treated as a special case, in case of a +-knob: middle tick would be drawn at the zero-value position.

umlaeute:
> > height of inlet/outlet is smaller in knob than in other objects.
> 
> I already said I'm using the regular IEM_GUI_IOHEIGHT constant; I don't see what you're describing here, e.g here is part of a hsl over a hradio over a knb :

same for me: [knb] has the same size of iolets than other iemgui objects (and all of them have a different size than ordinary objects/...)
i agree that [knb] should be consistent with other iemguis.
and that they probably should be fixed to match the size of normal objects, but that is to be fixed separately (outside this PR).

> I agree, I had thought of these, too.
> Maybe "ticks=3" could be treated as a special case, in case of a +-knob: middle tick would be drawn at the zero-value position.

probably. and if people complain, we can bugfix :-)

megrimm:
>> so I think it'd look better if we had just a straight line

why not just a straight line instead of a triangle? wouldn't this be cleaner when turned?

in my opinion pd adheres to bauhaus aesthetics: less is more

porres:
> i agree that [knb] should be consistent with other iemguis.
> and that they probably should be fixed to match the size of normal objects, but that is to be fixed separately (outside this PR).

I see, I didn't realize it was an issue for all guis, it did seem more apparent with knob. Anyway, a new issue just for that: https://github.com/pure-data/pure-data/issues/920

porres:
> i can't see that either in my machine (Ubuntu 16.04). Can Tk drawing be inconsistent across OSes sometimes?

No idea, but maybe? Since you can see in the image I posted that that's how it look on macOS

umlaeute:
> No idea, but maybe? Since you can see in the image I posted that that's how it look on macOS

i don't see no image you posted (if you post one, please don't forget to include reference images of ordinary objects *and* iemguis)

porres:
> i don't see no image you posted

@umlaeute - I posted this image, to show how the arc doesn't fit well into the circle (it's more noticeable in smaller knob sizes).

78042964-23e15c00-7349-11ea-97dd-029ad8eacdef

umlaeute:
ah sorry, i thought that was about the iolet-size.
if you don't mind i'm doing to delete my post, your reply to it and this reply to you, to keep the history clean (it's already too big anyhow).

umlaeute:
@Ant1r should we discuss issues with `knb` in multiple tickets in https://github.com/Ant1r/pure-data/ ?
(you would need to enable the issue-tracker there...)

Ant1r:
>  should we discuss issues with knb in multiple tickets in https://github.com/Ant1r/pure-data/ ?

I agree, absolutely!
Sorry I didn't know I had to allow issue tracker... Should be OK now. So feel free to open specific issue in my fork.

Ant1r:
BTW I'm on the way implementing other settings we spoke about. I'll be always possible to remove some of them later.

Capture du 2020-04-01 18-44-08

porres:
>  feel free to open specific issue in my fork.

done :)

Ant1r:
I'm leaving for today; WIP is there:
https://github.com/Ant1r/pure-data/tree/knobstyle-tmp

porres:
> WIP is there:
> https://github.com/Ant1r/pure-data/tree/knobstyle-tmp

I tried setting it to angular and angles from -180 to 180, but the angular control isn't really working well like this. At one point, it restarts from the beginning

Ant1r:
Well, here's where we are!

Sorry for the advocates of simplification... I kept most ideas @porres and @umlaeute  gave me.

Only "ticks" is not yet done; since I'm not sure I'll do it very soon, maybe I could hide the properties dialog "ticks" entry for the moment. Anyway the setting is fully implemented.

Capture du 2020-04-02 18-30-31
feedback are welcome!

porres:
How do you set the -/+ knob? I don't get it...

porres:
this need to change in the help file now ;)

"note: when selected, knb shows its inlet, outlet and a rectangular outline, in order to help connecting. When edit mode is switched off, they disapear after the mkob is hovered by the mouse or the knob value is changed."

porres:
>  I don't get it...

I got it now... the value of zero is always the start. I thought we'd have a new parameter to choose where the mid point is, but this makes sense. Other than that, we'd need yet another toggle to set "0" to start point or not.

Now, this is because one could actually want to have an output range from -1 to 1, but a behaviour that the arc line should start from the low point and up to the maximum point. As it is, not much of a big deal, as the user can do it by rescaling it afterwards, and I don't see much of a problem with that. That is to say I'm fine if things end up like this - what do you say?

porres:
> this need to change in the help file now ;)

I actually see that what;s described in the text keeps happening. But I don't think it should now anymore. If we're hiding inlet/outlet plus the outline with the s/r names, then we don't need to draw the outline when we select the object. 

I guess this makes the code cleaner and then the documentation should describe the new behaviour

porres:
> feedback are welcome!

- Ok, I'm an advocate for simplicity, but I don't really personally have that much of a problem with having these many options. My main concern is that I actually have this notion things need to be as simple/clean as possible so it has a chance to get into Pd Vanilla... 

- So if there's no real opposition, then fine... I don't really care :) and if we need to simplify things, I'd go for a single "wiper" style of a line. Triangle looks nice, but it doesn't look nice with the arc.

- I don't really get the design idea of having "none".

- The name "wiper" seems weird, I don't know what it means, maybe we can find a better one if we keep more options?

- I like that you can set an arc line width, from none, to fully filled in both positive and negative. I could keep it. But if I were to simplify it, I'd go for a toggle to have an arc or not and have the width in pixels dependant on the gui size.

- Like I previously said, I could add a parameter for the mid point of the arc line, so we wouldn't need to rely on "0".

- I find a bit confusing that the 0º angle is a horizontal line the top of the circle. That's more like a 90º angle for me... but I guess this is easier and maybe we just make it a bit clearer in the documentation.

porres:
> I don't really get the design idea of having "none".

The problem with it is that it seems it's only useful if you have an arc, and then when you have a value of "0" or at the start point, we just have no cue whatsoever where the start point of the knob is, we just have plain circle.

Getting a plain circle might be nice and a workaround for those who want to draw just a circle, but you can do that also if you just have the same color for front and background... 

Instead of a "none", perhaps we could have a style that uses a line with the arc width, be it positive (starting from the circle bounds), and negative (starting from the middle of the circle).

@umlaeute , what exactly got me a thumbs down or confusement? :)

porres:
I just have to say I couldn't be happier, thanks a lot, this looks great, I guess there's nothing else to worry ;)

Ant1r:
cool! I'll rewrite a simpler PR soon.

Well, there's still a problem (which I'm going to fix now): the 'init' value is wrong when the patch is saved while zoomed-in! And it's also the case with iemsliders... I'll propose a fix for them too.

There's another fix to sliders I'll propose soon: when the patch is zoomed in, and in the case send=receive, it's currently hard to drag the slider (only 2-pixels moves are working).

porres:
> in the case send=receive, it's currently hard to drag the slider

wouldn't that generate a stack overflow?

btw, as long as we're here, how do you feel about this? :)  https://github.com/pure-data/pure-data/issues/932

Spacechild1:
> btw, as long as we're here, how do you feel about this? :) #932

let's not discuss another issue here. instead, go to the issue and ping to person you want to get feedback from.

porres:
that ping thing never works for me... I just see it when I eventually read the thing on github, why is that?

Spacechild1:
https://help.github.com/en/github/receiving-notifications-about-activity-on-github/about-notifications

https://help.github.com/en/github/writing-on-github/basic-writing-and-formatting-syntax#mentioning-people-and-teams

Ant1r:
Ticks are implemented!
Capture du 2020-04-14 16-10-07

porres:
> Ticks are implemented

looks good! So, this is finally done with, right?

Ant1r:
yeah! 
this PR is superseded by #938
I'm closing it.

PR #922, opened 2020-04-03
g_editor: Don't enter object-edit mode after dragging an object

jamshark70:
There is one behavior in the Pd interface that drives me actually insane -- I mean, occasionally shouting/screaming at the computer when it's requiring extra steps just to move an object around.

And that is: When you drag a single object on the canvas, and release the mouse button, it enters object-edit mode (to edit the text in the object).

From a UI design perspective, this is... baffling. You're either editing the object (click) or moving it (drag). When you're editing it, you're not moving it; when you're moving it, there's no justification to assume that the user means to edit. I can't think of any UI design principle that would call for these use cases to be intermixed.

The fix appears to be simple: A flag to detect whether the mouse has moved or not already exists! It's just not used in this one place.

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2917,11 +2917,13 @@ void canvas_mouseup(t_canvas *x,
         canvas_doconnect(x, xpos, ypos, mod, 1);
     else if (x->gl_editor->e_onmotion == MA_REGION)
         canvas_doregion(x, xpos, ypos, 1);
-    else if (x->gl_editor->e_onmotion == MA_MOVE ||
-        x->gl_editor->e_onmotion == MA_RESIZE)
+    else if ((x->gl_editor->e_onmotion == MA_MOVE ||
+        x->gl_editor->e_onmotion == MA_RESIZE) && !x->gl_editor->e_lastmoved)
     {
-            /* after motion or resizing, if there's only one text item
-               selected, activate the text */
+            /* if there's only one text item selected *and* the mouse hasn't moved,
+               activate the text -- i.e., standard click/drag behavior:
+	       - Single click, no motion: enter object for editing.
+	       - Click-drag with motion: move object, keep it selected */
         if (x->gl_editor->e_selection &&
             !(x->gl_editor->e_selection->sel_next))
         {
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2383,8 +2383,8 @@ static void canvas_doclick(t_canvas *x, int xpos, int ypos, int which,
         else
         {
             int noutlet;
-                /* resize?  only for "true" text boxes or canvases*/
-            if (ob && !x->gl_editor->e_selection &&
+                /* resize?  only for "true" text boxes or canvases */
+            if (ob &&
                 (ob->te_pd->c_wb == &text_widgetbehavior ||
                     pd_checkglist(&ob->te_pd)) &&
                         xpos >= x2-4 && ypos < y2-4)


Spacechild1:
Hmmm... I never had a problem with that behavior. The "extra steps" consist of single mouse click.

Anyway, you might also want to adapt the comment to the new behavior. To be consistent, text selection shouldn't happen for resizing, either. So I guess the the if-clause could be:

```
else if (!x->gl_editor->e_lastmoved &&
    (x->gl_editor->e_onmotion == MA_MOVE || x->gl_editor->e_onmotion == MA_RESIZE))
```

umlaeute:
> Hmmm... I never had a problem with that behavior.

i even had to start Pd to check whether the problem  is actually reproducible.
that is: i never even noticed  :-)

porres:
> i even had to start Pd to check whether the problem is actually reproducible.
> that is: i never even noticed :-)

that makes two of us

jamshark70:
> To be consistent, text selection shouldn't happen for resizing, either. So I guess the the if-clause could be:

Sure -- I thought of that later, and tried it. There's a little quirk with resizing: the resize cursor doesn't appear after resizing.

1. Ctrl-2. Type something.
2. Put the mouse pointer on the right-hand edge. Cursor changes to "<-->".
3. Drag to resize.
4. After letting go of the mouse, you get the "hand" cursor (in Linux, anyway). Putting the mouse pointer on the right-hand edge again does not change the cursor, and dragging at this point does a move rather than resize.

This is probably related to the fact that you can't resize an object that is rectangle-selected. I know that the object is selected after resizing -- with my change, operating on a single object:

Action | Selected after? | Display color | Can move by arrows
-------|-----------|---------------|-------------------
Move | Yes | Blue (probably+) | Yes
Resize | Yes | Black (why?) | Yes

+ If you resize, then move, the object is black. If you deselect everything, then move one object, it turns blue.

I guess I should poke around and try to figure out why that's the case. It makes sense not to be able to resize if the selection consists of multiple objects -- but if you've rectangle-selected one, it doesn't make sense to restrict the resize operation.

(FWIW, resizing is a lot less commonly done and moving -- I didn't even notice that you can manually resize message boxes.)

> I never had a problem with that behavior. The "extra steps" consist of single mouse click.

and

> i never even noticed :-)

Sure, fair enough. I might have overstated my reaction -- I only yell at Pd when I'm preparing class material and I have not enough time to do it slowly and carefully. Under time pressure, it's much better if the interface works with me, not against me.

But it does remind me of an analogy I picked up from Project Runway, where one designer wanted to use human hair in hi collection and Tim Gunn said, "Make sure you're not living in the monkey house... when you first go in the monkey house, you think '*phew*, it really stinks in here!' But after you stay in the monkey house for a while, you start thinking, maybe it's not so bad."

A user interface's effectiveness or elegance shouldn't be measured only by the perspective of experienced users -- new users' reactions matter too. There's a big difference between an interface that you can learn to tolerate and an interface that just does what you expect.

Spacechild1:
> A user interface's effectiveness or elegance shouldn't be measured only by the perspective of experienced users

That's certainly true. But I'm curious: what is the actual problem with the current behavior other than aesthetic reasons? The object being selected doesn't interfere with most (any?) other actions, because keyboard shortcuts don't affect the text and other actions require a mouse click anyway...
On the contrary, it actually saves you a mouse click for auto-patching after moving.

jamshark70:
> what is the actual problem with the current behavior other than aesthetic*) reasons?

Two:

- I very often position an object by dragging it into its rough position, and then nudging it into its final place using arrows. By "very often" I mean dozens of times an hour. Common workflows should be streamlined as much as possible. An extra click doesn't matter if it's something you rarely do. If it has to be done for 80% of objects you create, the extra clicks add up.

- As a matter of principle: editing is editing, positioning is positioning. Positioning is not editing! Pd's behavior here is not found in any other point-click interface that I know... because it fundamentally makes sense to to keep editing and moving/selecting distinct. (After this change, Pd's behavior is more like that of, say, PowerPoint or LibreOffice Draw -- click inside a text box to edit the text; drag the box to move it and keep it selected. Well, everybody hates PowerPoint, but that's not because of its click-drag behavior.)

> The object being selected doesn't interfere with most (any?) other actions... On the contrary, it actually saves you a mouse click for auto-patching after moving.

The issue is not that the object is selected after moving -- I would expect the object to remain selected in that case! The issue is that the object goes into text-edit mode. This *does* interfere with many keyboard shortcuts (arrows, copy/paste etc.).

Auto-patching using the proposed code change: If I move an object and release the mouse (selected, but not editing text) and then do ctrl-1, the new object box *is* auto-patched. So, nothing saved or lost here.

Spacechild1:
Ok, you have good points! Would appreciate if you continue working on this problem!

jamshark70:
OK, I think I got it. Also updated the comment.

`canvas_setcursor(x, CURSOR_EDITMODE_RESIZE);` evaluates based on number of conditions. One of those conditions was `!x->gl_editor->e_selection` -- the list of selected objects must be null, or resize is not available.

So I tested removing that condition, and it seems to be fine: I can resize an object, and resize it again without any intervening mouse clicks or keystrokes. (The message box still doesn't show the blue color after resize, but that's cosmetic so it "doesn't count," as stated yesterday. I think this is because the selection is created, destroyed, and created again when resizing, which might desynchronize the display.)

I haven't traced through all of the logic, so I can't be completely sure this doesn't break something. I haven't been able to crash the editor with either of these changes, though.

Today, I have to do some course preparation with Pd -- I'll use my patched version and see if it holds up under more extended usage.

jamshark70:
Oh, the last change (resize) is premature -- it affects the first branch in a series so it blocks some other actions. Will look a bit later.

jamshark70:
After another week and a half of using this patched version, I can report no unexpected behavior. As far as I can see, it's fine.

danomatika:
This makes sense. I also never noticed it but I've also taught people to always "click out of the box" after creation in workshops. :P

jamshark70:
> I also never noticed it but I've also taught people to always "click out of the box" after creation...

It's still necessary to do that, as (I believe) there's no key binding to "finish object creation." I didn't look into that. This is about click-drag behavior (I never understood why click-drag triggered object editing).

jamshark70:
BTW: I've been using this patch on my Linux machine for about a year and a half now (April 2020 - Sept. 2021), with *zero* weird behavior. So I'm pretty well satisfied that it's a valid change and that it doesn't crash or break other stuff.

Review?

danomatika:
@umlaeute @Spacechild1 @millerpuckette I will take a look at this.

danomatika:
@jamshark70 This is something that I frankly never notice, probably because I trained myself out of it years ago. Looking again I agree that it doesn't really make sense.

@Spacechild1 There have been edits to g_editor.c in `master` and I can't find the right location for the first change in the diff. Do you know where it might be now?

danomatika:
@Spacechild1 Nevermind, I think I have it. The improved behavior is working and I can try updating this branch or make a new one.

danomatika:
#1484 is an updated version of this PR tracking the current master.

jamshark70:
Sorry to necrobump this, but I'm puzzled: This PR was merged, and it shows up in g_editor.c's commit history, but the code changes are not in the eventual file (e.g., search g_editor.c for "lastmoved" -- there are only two references, neither of which is `else if ((x->gl_editor->e_onmotion == MA_MOVE || x->gl_editor->e_onmotion == MA_RESIZE) && !x->gl_editor->e_lastmoved)`).

I rebuilt Pd locally the other day and this interface change is gone. I can hack it locally but something is amiss here.

Spacechild1:
Hmmm... something went wrong here. I'll update this PR and merge it into develop.

Spacechild1:
Ok, I think what happend is that your PR got merged, but afterwards someone else made some other changes to the editor logic and your improvements got lost by accident.

Can you reboot the PR based on current master? Also maybe add a short comment explaining the reasoning, so that they won't accidentally get removed again in the future :-)

I can then merge your PR into develop.

Spacechild1:
Ah, looks like Miller reverted it on purpose: https://github.com/pure-data/pure-data/commit/0ac62063b578f451678db95f53882041ac653099

jamshark70:
> Ah, looks like Miller reverted it on purpose: https://github.com/pure-data/pure-data/commit/0ac62063b578f451678db95f53882041ac653099

I spotted that in the g_editor.c comments later.

So there are two competing wishes for the interface:

1. Maintain compatibility with ctrl-D - move - type.
2. Or, mimic global UI standards wherein dragging an object *only* moves and does nothing else.

I can see how some users may rely on ctrl-D - move - type a lot. I have... well... never used that pattern.

If no. 1 is chosen, then *all* users are forced to accept that way, whether they want to or not. Same for no. 2. So there should be a good reason for choosing one or the other.

I would argue that there is no such good reason, on either side -- I cannot objectively justify my preference, and neither can MSP or anyone else. So the best solution is to let the user decide = user preference.

I'd reiterate that the current behavior makes Pd frankly unusable for me. The only way I can tolerate it is if I patch it independently in my local build.

Spacechild1:
> So the best solution is to let the user decide = user preference.

+1

danomatika:
I could see this being a toggle in the Edit menu in the short term and perhaps a checkbox in the unified settings dialog long term.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
@millerpuckette I think this was reverted during the 0.53 release process. Is there a reason? It seems like an annoyance for many and a simple fix.

millerpuckette:
It's useful for some and anathema for others.. (for example those who like to duplicate a box, move and edit the dup).  If indeed that's the only reason to jump into edit mode a possible compromise would be to only activate the text if the dragged object had just been created before clicking on it?

sebshader:
or maybe there could be a keybinding to edit currently selected objects? since people who want to edit after duplicating need to use the keyboard at that point anyways

jamshark70:
> It's useful for some and anathema for others

To put it more positively: some users rely heavily on "dup - move - edit"; other users (myself) rely heavily on "new object - drag - arrow-key fine positioning."

I'm assuming that neither of these workflows would be considered invalid. (I'll admit that when I submitted the PR, I hadn't considered dup-move-edit -- but after seeing the reference to it in the code comment, I had to admit that of course it's legit.)

> If indeed that's the only reason to jump into edit mode a possible compromise would be to only activate the text if the dragged object had just been created before clicking on it?

I'm not sure that would solve the problem. The typical case for me is to create an object, then use the mouse to get it close to the best position, then use arrow keys to push it into a location that minimizes diagonal lines.

An automated approach would need to distinguish between an object that had just been "Put" into the canvas, and one that had been duplicated or pasted.

Or... hmm... a possibly interesting heuristic might be: If the object has been pasted and dragged without editing, then we might assume that the user would want to edit. If the object is being dragged after typing into it, then assume that the user doesn't want to edit further (because "new - drag - arrow" is really "new - type text - drag - arrow," while "dup - move - edit" does the typing *after* the drag operation).

> or maybe there could just be a keybinding to edit currently selected objects?

Or, a key binding to enter/exit object-edit mode (like F2 in LibreOffice). The thing that bothers me about auto-edit-after-drag is that the only way to exit object editing is to click outside of the object. (Why does "Return" / "Enter" not *enter* the text?) If I could create - drag - hit a convenient key - use arrow keys, that would be acceptable.

Spacechild1:
Often there is just no One Right Way™ to do things. IMO, instead of making assumptions and adding complicated logic, the better solution is simply giving the user a choice!

danomatika:
I agree, however the first reaction from both sides was "mine is the One Right Way™" so getting to this point is good. :)

jamshark70:
> however the first reaction from both sides was "mine is the One Right Way™"

FWIW I had dropped the "one right way" idea [two months ago](https://github.com/pure-data/pure-data/pull/922#issuecomment-1146617627) and proposed a user preference at that time.

To sum up, I'd be ok with a check-mark item in the edit menu, or a hotkey to enter/exit object-edit mode (could be shift-return, ctrl-return, escape, or other).

PR #924, opened 2020-04-04
Fix TCL error when closing the message dialog

### LINKED ISSUE #923: Pressing Esc in message dialog gives TCL error
1. Open Dialog Message
2. Press Esc

`Error: bad window path name ".message.f.entry"`

The problem seems to occur when [this line](https://github.com/pure-data/pure-data/blob/e2ef8158e83900734d263a3fda343a47ed0c9e3a/tcl/pd_bindings.tcl#L118) in pd_bindings.tcl is called.

HenriAugusto:
Fixes #923

The problem is caused because the .message window gets destroyed *before* [this binding](https://github.com/pure-data/pure-data/blob/e2ef8158e83900734d263a3fda343a47ed0c9e3a/tcl/pd_bindings.tcl#L118) is executed.


umlaeute:
merged into `update/0.50`

PR #933, opened 2020-04-13
Sliders fixes

### LINKED ISSUE #1042: [slider] value isn't preserved accross zooming
Currently, sliders (either vertical or horizontal) stores their value in a zoom-dependent form.

As a result:
- an incorrectly doubled value is saved to the file when the patch is zoomed in.
- copy/zoom/paste operation doesn't preserve the value, example:

Capture du 2020-05-29 11-30-23

#933 fixes that bug.

### LINKED ISSUE #1043: [slider] buggy shift-mouse behaviour when zoomed-in and send=receive
When a slider (either vertical or horizontal) is configured with identical send and receive symbols, and when the patch is zoomed-in, fine dragging (shift+mouse motion) is buggy:

it works as expected when the motion is downward, but it doesn't when the motion is upward and slow:

slider-bug-1

#933 fixes that.

### LINKED ISSUE #1121: [slider] knob position isn't updated after editing size/range/linlog
After editing a h/vslider, either by sending `size`, `range`, `lin`, `log` messages, or by editing the dialog properties, the "knob" position isn't in sync anymore with the actual slider value.

test_slider.pd

Ant1r:
here are some fixes with hslider and vslider:
- `init` value was wrongly saved when the patch is zoomed-in (so this fixes #1042 )
- mouse control was wrong in the case (fixes  #1043 ):
  - the patch is zoomed-in
  - `send` and `receive` symbol are equal
- the knob position wasn't updated after changing size/range/linlog (fixes #1121 )


src/g_vslider.c
```diff
@@ -243,7 +243,7 @@ static void vslider_save(t_gobj *z, t_binbuf *b)
                 x->x_gui.x_ldx, x->x_gui.x_ldy,
                 iem_fstyletoint(&x->x_gui.x_fsf), x->x_gui.x_fontsize,
                 bflcol[0], bflcol[1], bflcol[2],
-                x->x_val, x->x_steady);
+                x->x_val/IEMGUI_ZOOM(x), x->x_steady);
```
umlaeute, line 246: urgh. this looks awful.
isn't `x_val` meant to be the **real*** value that is controlled by te slider? as such, the zoom-scaling should go into the functions that interface with the GUI, not with saving/restoring.

Ant1r: unfortunately not... see for example:
https://github.com/pure-data/pure-data/blob/e2ef8158e83900734d263a3fda343a47ed0c9e3a/src/g_hslider.c#L344-L348

Ant1r: maybe I should have exposed the problem in the issue tracker. Just try, you'll see that init value isn't correctly saved/restored across zooming (even using copy/paste for example).

umlaeute:
i don't doubt that there is indeed a bug.
it just seems to me that you are fixing it in the wrong place.

Ant1r:
Well yeah, it's a quick fix I found while writing the knob. 
But I'm afraid the real fix needs large rewriting of h/vsliders (and knob), simplifying that pos/val/fval mess...

src/g_vslider.c
Ant1r: OK I'm currently studying this simplification.

Ant1r:
@umlaeute I've rewritten this bug fix, by cleaning the code, making internal representation of position independent from zoom factor.
And the bugs (init value / drag while zoomed) magically disappear :-)
BTW do you think I should create specific issues for reporting those 2 bugs, or is it just unnecessary noise?

src/g_vslider.c
Ant1r: done

src/g_all_guis.h
```diff
@@ -216,7 +215,6 @@ typedef struct _vslider
 {
     t_iemgui x_gui;
     int      x_pos;
-    int      x_val;
```
umlaeute, line 219: i wonder why you kept `x_pos` rather than `x_val`.
isn't the latter what we actually want (the current *value* of the slider, not its position)

src/g_hslider.c
```diff
@@ -36,7 +36,7 @@ static void hslider_draw_update(t_gobj *client, t_glist *glist)
     t_hslider *x = (t_hslider *)client;
     if (glist_isvisible(glist))
     {
-        int r = text_xpix(&x->x_gui.x_obj, glist) + ((x->x_val + 50)/100);
+        int r = text_xpix(&x->x_gui.x_obj, glist) + ((x->x_pos + 50)/100) * IEMGUI_ZOOM(x);
```
umlaeute, line 39: the `((x->x_pos + 50)/100) * IEMGUI_ZOOM(x)` should probably go into a (static) helper-function or at least a macro. it's repeated over and over and over.

src/g_hslider.c
```diff
@@ -396,27 +392,22 @@ static void hslider_dialog(t_hslider *x, t_symbol *s, int argc, t_atom *argv)
...
 
     if(x->x_gui.x_fsf.x_finemoved)
-        x->x_pos += (int)dx;
+        x->x_pos += (int)(dx);
     else
-        x->x_pos += 100 * (int)dx;
-    x->x_val = x->x_pos;
-    if(x->x_val > (100*x->x_gui.x_w - 100))
+        x->x_pos += (int)(100 * dx / IEMGUI_ZOOM(x));
+    if(x->x_pos > (100 * (x->x_gui.x_w / IEMGUI_ZOOM(x)) - 100))
```
umlaeute, line 401: that looks bogus. why is the position not scaled by the zoom-factor when we <kbd>Shift</kbd>-drag?

actually i think that the entire `if`/`else` clause should be:

~~~C
x->x_pos += (int)(((x->x_gui.x_fsf.x_finemoved)?100.:1.) * dx / IEMGUI_ZOOM(x));
~~~

src/g_all_guis.h
Ant1r: Well I think it's more really its position, [0;1] normalized. 
The "value" involves scaling and optional log transformation, which makes it seem to me further away.

Ant1r: (sorry it's not normalized to [0;1] but to [0; width*100] I think. [0;1] is what I've chosen for knob)

src/g_hslider.c
Ant1r: > why is the position not scaled by the zoom-factor when we Shift-drag?

because half-pixel mouse steps would be rounded to 0, so you couldn't move the slider, excecpting if you moved the mouse quickly (so you have more than 2-pixels steps).

src/g_all_guis.h
Ant1r: anyway, maybe you're right: changing then name `x_pos` to `x_val` could make the code more readable, as currently there's a clash between `x->x_pos` and `xpos = text_xpix(...)`.
should I?

src/g_hslider.c
Ant1r: done (and some others) by macros.
see https://github.com/pure-data/pure-data/pull/933/commits/b39bfd260d100b3e06d9a77e11be34c202a4e688

src/g_hslider.c
umlaeute: hmm.
if you make `x_pos` a `t_float` then that wouldn't e an issue.

otoh, i now see that taking IEMGUI_ZOOM into account, will give finer precision of the slider when zoomed in.
i'm not sure whether this is actually desired.

e.g. a default slider will have integer increments. with your patch applied, it will suddenly have `0.5`-step increment in zoomed-mode which might make for some surprise.

Ant1r: > make `x_pos` a `t_float`

yes, that's what I did with the [knob].
but it has some float-rounding side effects when you want to have integer increments, and anyway it wouldn't remove the effect you see (finer precision when zoomed).

I would see a way to solve that: we could keep track of the actual pointer position; this would require adding some local global variables. So we could [div] the movement by the zoom factor before applying the motion method.

But I'm not sure it's worth it? 
My first attempt hadn't that drawback. Too bad I lost it with force-pushing ;-) (maybe I should never do that)

Ant1r: OK, please excuse my raving, the solution was finally rather simple, see https://github.com/pure-data/pure-data/pull/933/commits/6873180b88d3247a786865510a0850babf306350.

I think the behavior is now the same than before.

danomatika:
Is this something to submit for 0.51-1 or does it need more time?

Ant1r:
Well there's actually still one or two things to fix:

 - I agree with IOhannes that `x_val` is a better name than `x_pos` (see: https://github.com/pure-data/pure-data/pull/933#discussion_r415122607 ); I've just done the replacement (though not yet pushed it) so this should be ready quickly.

 - I've just noticed another bug: the position of the "knob" isn't updated after editing the size of the slider (either by sending a `[size(` message or through the properties dialog). It's already the case on current master (afaict it has always been the case). I'm on the way to fix it.

Ant1r:
please wait, something's still wrong with updating the knob position after receiving "size" message...

src/g_all_guis.h
Ant1r: done with https://github.com/pure-data/pure-data/pull/933/commits/31fad0f4a1b430b3b82d5444f5b05174a95e35b5

Ant1r:
Okay, I think it's OK now!
I performed several tests (see attached patch if you want to try) and all seems to behave normally.

[test_slider.pd.zip](https://github.com/pure-data/pure-data/files/5074971/test_slider.pd.zip)

Ant1r:
I brought this branch up to date from master.

porres:
> Okay, I think it's OK now!

cool, seems good to me as well!

> Is this something to submit for 0.51-1 or does it need more time?

now that this ship has sailed, I hope all gui related fixes that I've been organising (at https://github.com/pure-data/pure-data/issues/947) are considered, so a good GUI update can be part of the 0.52 release cycle.

Ant1r:
updated to master.
@millerpuckette could be good to merge ;-)

umlaeute:
i have included the fix for #1121 in my iemgui-updates branch #1633.

if you update this PR (or reboot it on top of #1633) i will happily merge in the zoom-fixes as well...

Ant1r:
Yes, I think I can reboot it. I'll do it asap.

Ant1r:
@umlaeute #1644 is ready. And better than this one I think (dragging didn't behave exactly as it should).

porres:
close this one then?

Ant1r:
well, this one is against master branch, so I first thought we could wait merge-iemgui to be merged before closing; but I always can re-open (and amend...) it, if needed.

PR #829, opened 2019-12-10
Documentation updates

porres:
This PR includes some documentation fixes and updates from newly added features

PR #937, opened 2020-04-14
finally save iemguis colors as hex symbols

Ant1r:
IMO it's time to give up retro-compatibility to Pd <= 0.47 by default, so that iemguis colors can be saved using full resolution hex-symbols.

e.g: instead of saving vslider like this:
`#X obj 96 97 vsl 15 128 0 127 0 0 empty empty test 0 -9 0 10 -258113 -257985 -159808 0 1;
`
let's use:
`#X obj 96 97 vsl 15 128 0 127 0 0 empty empty test 0 -9 0 10 #fc0400 #f8fc00 #9c00fc 0 1;`

That was planned for 0.48 ;-)

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -270,9 +270,7 @@ void iemgui_all_sym2dollararg(t_iemgui *iemgui, t_symbol **srlsym)
 }
 
 static t_symbol* color2symbol(int col) {
-    const int  compat = (pd_compatibilitylevel < 48)?1:
-        /* FIXXME: for Pd>=0.48, the default compatibility mode should be OFF */
-        1;
+    const int  compat = (pd_compatibilitylevel < 48) ? 1 : 0;
 
     char colname[MAXPDSTRING];
     colname[0] = colname[MAXPDSTRING-1] = 0;


Ant1r:
@porres I kept the original code for minimal diff, but I agree it could be rewritten.

Spacechild1:
Ha, I didn't notice that the mechanism was already in place!

Now I just found out that we can already use hex symbols in the `[color(` message :-o They are just not saved as such (yet). This features needs to be documented in the IEM GUI help patches. @porres Do you want to do this? ;-)

porres:
sure, I can't believe we can do this either

porres:
When was this added by the way? Extended didn't have this, neither does Purr Data, so it's relatively new, right?

porres:
I actually already wanted to make a proper overhaul of the documentation of iemguis for a while. I already did lots of cleaning recently, but this is something that encourages me to do a better redesign, and also accounts for a "updated for 0.51.0" or something

umlaeute:
> Ha, I didn't notice that the mechanism was already in place!

sure. the keyword is "smooth upgrade path".
if you save a patch right now (with this patch applied), Pd-versions that are now old (>=0.47) can already handle these colours.
but back then (when Pd-0.47 was released), saving a patch with hex-colors would have broken on then-not-so-old Pd-versions (e.g. 0.46)

> When was this added by the way? 

try `git bisect`? (please let's not discuss here, *how* that would actually be done)
or simply try [`git blame`](https://github.com/pure-data/pure-data/blame/0.50-2/src/g_all_guis.c#L272)

porres:
> This features needs to be documented in the IEM GUI help patches. @porres Do you want to do this? ;-)

there's just so much going on, with knob, new color settings, that I wanna do this when we know what's merged or not and do it all once ;)

umlaeute:
@porres can i remove all the comments from us both since https://github.com/pure-data/pure-data/pull/937#issuecomment-613545001? (they are pretty irrelevant to this PR)

porres:
go ahead

porres:
>  @porres Do you want to do this? ;-)

Ok, I started working on a new documentation for iemguis, I updated the help file of bang for now.

I actually removed from the documentation the old way of setting collors as floats, as it wasn;t properly documented how it worked and since now I think there's a much better and clearer example on how to set the colors. Please check it out to see if the changes I'm doing are ok, so I can go on and do the same for other iemguis.

I'm also making changes that are mainly aesthetical, but I'm also adding more information.

porres:
> Ok, I started working on a new documentation for iemguis, I updated the help file of bang for now.

I opened a new issue for this https://github.com/pure-data/pure-data/issues/1019

And I just made a new commit with another revision model, again only for bang - see https://github.com/pure-data/pure-data/pull/829/commits/f42c28ba2e91324d1234dfea886bedfe4a59aac1

> I actually removed from the documentation the old way of setting collors as floats, as it wasn;t properly documented how it worked and since now I think there's a much better and clearer example on how to set the colors.

I decided to add a new example with this old way of doing things as well in my new revision... I'll finish updating the documentation in the next few days.

Issue #920, opened 2020-04-01
inlet/outlet, font and object size inconsistency for iemguis 


porres:
The inlet/outlet for iemguis is a bit smaller 

EDIT: general height of the iemguis could also match atom boxes and objects

EDIT 2: Default font size could also be the same as Pd's default (12)

danomatika:
I actually implemented this when I did the gui overhaul a couple of years ago, but changed it back. The main reason was that larger IOs cover up too much of the actual UI behind them.

If you want to try it, change `IEM_GUI_IOHEIGHT` to 3 in `src/g_all_guis.h`.

porres:
I'm giving it a try, when checking the new knob GUI, it really felt unnatural, and then I realized how the others were also "small", but I haven't tested yet with the new size...

Ant1r:
I agree with @porres . 
`IEM_GUI_IOHEIGHT = 3` doesn't hurt my eyes...

danomatika:
Don't get me wrong, I'm not against the idea.

At the time, it seems like too much a change to push through due to all of the other updates I did, so I turned it back off but kept the define for easy switching. :)

Ideally, that line would be:

```c
#define IEM_GUI_IOHEIGHT IOHEIGHT
```

danomatika:
For comparison's sake, here are screenshots from @Ant1r's knob PR.

IEM_GUI_IOHEIGHT = 2 (current, gem guis have slimmer IOs):

<img width="911" alt="Screen Shot 2020-04-17 at 5 45 17 PM" src="https://user-images.githubusercontent.com/480637/79587902-7fdf0b00-80d3-11ea-8fe1-453f854df0be.png">

IEM_GUI_IOHEIGHT = 3 (iem guis matching objects & atoms):

<img width="911" alt="Screen Shot 2020-04-17 at 5 46 12 PM" src="https://user-images.githubusercontent.com/480637/79587988-9dac7000-80d3-11ea-9405-d65a83ec2eea.png">

danomatika:
One aspect, for instance, is that the radio's default size now has the selection rect touching the IOs when height = 3. I think that was one of the bigger reasons I held off before.

porres:
> Ideally, that line would be:
> #define IEM_GUI_IOHEIGHT IOHEIGHT

done with https://github.com/pure-data/pure-data/pull/946/commits/e3864ac74a8875529f5a4f9718cf970cef8e70c1

[...]

PR #938, opened 2020-04-14
Feature/knb

### LINKED ISSUE #911: feature: add a knob GUI

Ant1r:
closes #911
Here is a proposal to add a new knob GUI, associated to "Shift+Ctrl+K" key accelerator.

Originally imported from moonlib/mknob, it has now gained several features, such as:
- optional arc drawing
- configurable angular area
- optional ticks
- removable cursor
- +/- mode

Capture du 2020-04-14 18-16-43

for those wanting to read the long discussions and the preparatory commits (that have been compacted here for clarity), see: #919  #918 
and [knob] related issues in https://github.com/Ant1r/pure-data


danomatika:
Do your dialog changes take into account translations?

Ant1r:
I'm not sure what you mean exactly; these changes are little to do with text that are displayed to the user. Only "stead on click" labels and, of course, knob related labels are impacted, but I tried to keep it consistent with the localization system.

But maybe what you mean is that many of the po/ files are touched (rather needlessly), and that it makes merging of the PR more difficult.  If this is the case, maybe I should rewrite the PR without touching to po/ files, excepting template.pot and fr.po?

danomatika:
The last comment was written quickly on a phone, here is the elaboration:

> I'm not sure what you mean exactly; these changes are little to do with text that are displayed to the user. Only "stead on click" labels and, of course, knob related labels are impacted, but I tried to keep it consistent with the localization system.

That was my main question: did you make sure that updates & additions were picked up by gettext? Sounds like you have, good. I would hate to have new stuff be introduced without it.

When I overhauled the dialogs some years ago, I tried to make sure existing objects which used them such as `mknob` still worked. As you noticed, I also made it easier to extend with custom text labels, so it's cool to see that being used now.

> But maybe what you mean is that many of the po/ files are touched (rather needlessly), and that it makes merging of the PR more difficult. If this is the case, maybe I should rewrite the PR without touching to po/ files, excepting template.pot and fr.po?

Two things:

1. This is a work flow issue. I deal with the same thing as well and I suggest: don't check in changes to the `/po` folder. git add the subfolders you are working with, ie. `/src`, `/tcl`, etc and someone else will probably routinely do a translation update with changes to /po folder in the course of a release. So far I and @umlaeute have done so. If you have a specific change, ie. to /po/fr.po, then indeed add the template & fr files ... although at that point, you might as well update the rest fo the pos anyway.

2. `<begin rant>` The translation files are *not* touched "rather needlessly" but *on purpose*. This is the whole point of the automated system: to pick up changes & additions to the localized text in the project. In fact, when I overhauled this system (also some years ago), I found that it had been imported form Pd-extended and then *taken out of the build system*. This means that ad Pd has been updated the translations quickly grew stale and there wasn't much impetus to update them.`<end rant>`

I'm pretty sure it was taken out because of this "rather needlessly" thinking, although there was some truth to it: timestamps and file change lines were constantly being adopted even though the relevant text was not. I reworded the gettext calls to avoid doing this and only update files when the actual text changes which cuts down on changes to /.po folder by quite a bit.

tcl/dialog_iemgui_knob.tcl
```diff
@@ -0,0 +1,175 @@
+# For information on usage and redistribution, and for a DISCLAIMER OF ALL
+# WARRANTIES, see the file, "LICENSE.txt," in this distribution.
+# Copyright (c) 1997-2009 Miller Puckette.
```
danomatika, line 3: Wouldn't this be your copyright? You are the author here. If not, you could adopt a more general line such as "Copyright (c) 2020 Pure-Data team." etc.

tcl/dialog_iemgui.tcl
```diff
@@ -12,6 +12,24 @@ namespace eval ::dialog_iemgui:: {
     namespace export pdtk_iemgui_dialog
 }
 
+source [file join [file dirname [info script]] "dialog_iemgui_knob.tcl"]
+
+proc procExists p {
```
danomatika, line 17: Can this be snake case the match the rest of the API?

tcl/dialog_iemgui.tcl
```diff
@@ -12,6 +12,24 @@ namespace eval ::dialog_iemgui:: {
...
+    return uplevel 1 [expr {[llength [info procs $p]] > 0}]
+}
+
+proc ::dialog_iemgui::callGUIproc {mytoplevel p args} {
```
danomatika, line 21: This could also be snake case.

tcl/dialog_iemgui_knob.tcl
Ant1r: yep i'm going to update this.

src/g_knb.c
```diff
@@ -0,0 +1,1030 @@
+/* Copyright (c) 1997-1999 Miller Puckette.
```
danomatika, line 1: I don't think this copyright line needs to be here, just the one below.

tcl/dialog_iemgui.tcl
Ant1r: sorry I don't understand "snake case"... could you help me?

Ant1r: OK! you mean using `proc_exists`

tcl/dialog_iemgui.tcl
Ant1r: ...and `::dialog_iemgui::call_gui_proc`
I'm gonna change these.

danomatika:
Something to check: when I updated the GUIs for full zoom support, I also looked into some layering issues.

For instance, setting a send or receive would take away the IO boxes and clearing them would bring the IOs back, however now they would be drawn *over* the labels as they are created after. The fix for this was to m[anually raise the label after creating the IO](https://github.com/pure-data/pure-data/blob/master/src/g_bang.c#L150).

I think there may be a few other details like this and I suggest double-checking with the current gem GUI code, ie. `src/g_bang.c`, so that this object also adopts the same relevant patterns.

I'm sorry, but I cannot do a more in-depth check on this right now, but I can look later perhaps.

Ant1r:
> manually raise the label after creating the IO

OK I understand, i'll check this. I tested most related zoom issues (BTW I found some bugs in sliders, see https://github.com/pure-data/pure-data/pull/933) but not this one.

danomatika:
Can you also add a knob to the sizing test patch: `docs/7.stuff/tools/sizingtest.pd`

I added this mainly so we all have something to compare with when talking about canvas GUI stuff.

danomatika:
I would also suggest calling the source file "g_knob.c" as none of the other GUI implementations use the short object name, ie. "g_bang.c" instead of "g_bng.c".

danomatika:
I'm testing the PR and I will do some small commits to fix some clang warnings, etc. I will also check the dialog layout (I spent *too much* time on this the last time, but I'd like things to be consistent).

Ant1r:
Thanks, however I'm about to force push the changes you proposed, so maybe you could wait I'm finished (say half an hour)?

danomatika:
I pulled your branch for testing and pushed to a temp `feature/knb` branch here. The warning fix is pretty small: https://github.com/pure-data/pure-data/commit/365a789c9accc3a320a7322f44e4df7bc5fe01f6

danomatika:
Another nit, if you have time, would be to make the formatting better match "Pd's overall code style". To me, this is basically the parts written by Miller, which is most of the code base.

It's down to things like indenting leading braces:
```c
void some function() 
{
...
```

To a space after `if`, `for`, `while` etc:

```c
if (!something)
{
    ...
```

To an indented comment block:

```c
    /* a comment about this part */
for (i = 0; i < foo; i++)
{
...
}
```

And so forth. I tried to follow this when I refactored the sound file code in another PR, [for example](https://github.com/pure-data/pure-data/pull/855/files#diff-9a23baa4f154c66c4e58b2201f6ae822).

tcl/dialog_iemgui_knob.tcl
Ant1r: done

tcl/dialog_iemgui.tcl
Ant1r: done

tcl/dialog_iemgui.tcl
Ant1r: done

src/g_knb.c
Ant1r: done

Ant1r:
OK I've force-pushed a new version, taking your comments into account:

- snake case for new Tcl procs
- updated / simplified copyrights in new files
- more consistent code style
- knb.c -> knob.c
- don't commit po/*.po excepting french one
- raise label after creating inlet/outlet
- add knb to sizingtest
- abs() -> fabs()

porres:
You don't seem to be using "MKNOB_TANGLE" in the code anymore, so you can remove it.

As for default size, I wish it was bigger, specially if iolet size change - see https://github.com/pure-data/pure-data/issues/920 (I was going for 42, but anything above 36 seems good).

porres:
I don't get why you have "MKNOB_DEFAULTH", knob should have only one size for both width and height. It seems you also have an argument just for that, but it is pointless

Ant1r:
MKNOB_DEFAULTH was really MKNOB_DEFAULT_SENSITIVITY (widget 'height' is reused for 'sensitivity').
I've fixed it, removed useless MKNOB_DEFAULTH.
I also change default size to 35, and default label fontsize to 12 ;-)

danomatika:
Ok cool. I will check this and adjust the dialog layout a bit.

danomatika:
In looking at `src/g_knob.c`, the formatting looks good except, to me, using if & else one liners followed by additional lines is a bit confusing, ie. I think this

```c
if (h < 5) h = 5;
x->x_gui.x_h = h;
```

is harder to read than this:

```c
if (h < 5)
    h = 5;
x->x_gui.x_h = h;
```

You save a line, but my original read through had me thinking the `x->x_gui.x_h = h;` was affected by the if statement.

Ant1r:
OK, I'll add new lines/indentation.

src/g_knob.c
```diff
@@ -0,0 +1,1048 @@
...
+#include <io.h>
+#else
+#include <unistd.h>
+#endif
+
+#ifndef M_PI
+    #define M_PI 3.14159265358979323846
+#endif
+
+#define MKNOB_DEFAULT_SENSITIVITY 128
```
danomatika, line 29: I think these can probably drop the "MKNOB_" prefix as they are only used internally or maybe becomes simply "KNOB_". I notice the other item guis have similar local defines, but without prefixes.

Ant1r: urgh... right! fixing...

Ant1r: i removed the prefixes.

danomatika:
Wouldn't "Arc width" conceptually be the "Arc height"? I suppose it could go either way, but maybe something like "Arc size" make more sense? I'm not sure.

danomatika:
Ok, I made a new temp `feature/knb` branch which has a couple commits to the dialog:

* parameter label & entry alignment, this is not 100% perfect but good enough for now
* added macOS-specfic entry bindings

The alignment update basically adds an empty expanding label on the left side of each parameter line to take up space so the lines match up vertically, relatively. Here's a screenshot with debug colors to show the dummy labels:

Screen Shot 2020-04-19 at 1 35 47 PM

porres:
> Wouldn't "Arc width" conceptually be the "Arc height"? I suppose it could go either way, but maybe something like "Arc size" make more sense? I'm not sure.

"Arc thickness"?

danomatika:
"size" is shorter and also matches the overall "Size" of the control.

Ant1r:
I think that "thickness" is more precise in this case, isn't it?
But English is not my mother tongue...

danomatika:
It is, but then it’s a wider word and might mess up the alignment. ;)

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Ant1r:
anyway the alignment is ruined when texts are translated...
so I used a `grid` instead of `pack`, and I chose "Arc thickness".

also I moved "Arc" color example to the right, to match the color radio button.
 
how is it now?
Capture du 2020-04-22 00-31-20

Ant1r:
I've just tested the grid part of the dialog with wish8.4, and it seems it works well:
Capture du 2020-04-22 16-16-57

Ant1r:
(BTW: is there is a way, under linux, to force Pd to use an older wish?)

danomatika:
There is the "--with-wish" configure option to specific the name of the wish command:

    ./configure --with-with=wish8.4

umlaeute:
or just launch the Pd-GUI manually (and it let it launch the Pd-core).

~~~sh
wish8.4 tcl/pd-gui.tcl
~~~

danomatika:
You can also invoke `pd-gui.tcl` with wish directly, ala

    wish8.4 tcl/pd-gui.tcl

This requires Pd to be built and the symlink set in the `bin` directory. Another option is to invoke the installed pd-gui.tcl which is probably in the `/usr/local/lib/pd/tcl` folder.

Ant1r:
Thanks! `wish8.4 tcl/pd-gui.tcl` works well.

thanks to that I could fix some minor 8.4 incompatibilities, while keeping the grid layout.

danomatika:
Opening the Properties dialog is giving me the following:

~~~
(Tcl) UNHANDLED ERROR: bad window path name ".gfxstub7ff040e2da70.para.knbangle.start.ent"
    while executing
"bind $mytoplevel.para.knbangle.start.ent <KeyPress-Return> "::dialog_iemgui::apply_and_rebind_return $mytoplevel""
    (procedure "::dialog_iemgui::create_properties_knb" line 72)
    invoked from within
"::dialog_iemgui::create_properties_knb .gfxstub7ff040e2da70 0 #000000 0 -135 135"
    ("eval" body line 1)
    invoked from within
"eval $gui_proc $mytoplevel [eval concat $args]"
    (procedure "::dialog_iemgui::call_gui_proc" line 9)
    invoked from within
"::dialog_iemgui::call_gui_proc $mytoplevel ::dialog_iemgui::create_properties_ $args"
    (procedure "pdtk_iemgui_dialog" line 393)
    invoked from within
"pdtk_iemgui_dialog .gfxstub7ff040e2da70 |knb|             --------dimensions(pix)(pix):-------- 35 12 size: 128 0 sensitivity:             -----------..."
    ("uplevel" body line 1)
    invoked from within
"uplevel #0 $docmds"
~~~

danomatika:
Fix is in `feature/knb` branch.

src/g_knob.c
```diff
@@ -0,0 +1,1085 @@
...
+    if ((!x->x_gui.x_fsf.x_snd_able) || (!x->x_gui.x_fsf.x_rcv_able))
+    {
+        if (!x->x_outline_visible)
+            sys_vgui(".x%lx.c create rectangle %d %d %d %d -width %d -tags %lxOUTLINE\n",
+                canvas, xpos, ypos,
+                xpos + x->x_gui.x_w, ypos + x->x_gui.x_w,
+                IEMGUI_ZOOM(x),
+                x);
+        x->x_outline_visible = 1;
+
```
danomatika, line 350: Extra line. Remove?

danomatika:
I checked through the source and it looks good. The labels and ordering look good.

danomatika:
The only thing I can think of now is that the edge of the wiper at the center doesn't stay perpendicular as it rotates. At certain angles, it bends a little "wonky" but I imagine that's more of a problem of drawing angles lines at small resolutions, so there isn't much to do about it.

Screen Shot 2020-04-22 at 11 15 40 PM

versus

Screen Shot 2020-04-22 at 11 15 52 PM

Ant1r:
Ok, I think it might be better with a thick line instead of a polygon (I kept the polygon at first to easily switch between 'triangle' and 'line', but it's useless now). I'll try soon.

danomatika:
Also, I don't know if you are aware, but I implemented a version of mknob for PdParty on iOS. Once this is all finalized, I can also make an update.

Screen Shot 2020-04-23 at 2 46 07 AM

Ant1r:
ok I've switched to a thick line, @danomatika please report (IMO it's smth you can only see on over-sampled displays, like retina; actually maybe only specific to MacOS).

yeah, cool if you can add [knob] for PdParty (if [knob] get merged)! 
Maybe I could do the samefor PdDroidParty, even if I haven't had a chance to use it for a while.

porres:
here's something I'm noticing: full positive arc thickness value still keeps the knob circle outline. But a full negative one superseeds it and takes over (I think it should be like the positive one and keep the outline)

porres:
> Also, I don't know if you are aware, but I implemented a version of mknob for PdParty on iOS

that can't be data structures, right? How did you do it? Is that an external?

Ant1r:
>  full positive arc thickness value still keeps the knob circle outline. But a full negative one superseeds it and takes over

yep, it's on purpose, so you can draw every case, including pure colored circle (without outline). I think it can be handful in some cases.

porres:
> yep, it's on purpose, so you can draw every case, including pure colored circle (without outline). I think it can be handful in some cases.

I don't understand why does this have to be in the negative case, why not both? In the positive you could also reach and superseed the outline, but if you don't want it, go to maximum -1

it's confusing I guess to have that discrepancy, and it makes it more complicated to explain/document it

porres:
I guess I understand... negative goes from center to the outline

porres:
Well, now I see an actual bug. Minimum and maximum values are drawn the same for full angular control. You should have the arc on full spin for the maximum value, but it's wrapping to the beggining. So you cannot differentiate from full off to full on.

Ant1r:
Ok ... Yet another rendering difference between osx and Linux. I'll try to limit max angle to something like 359.99 degrees.

danomatika:
>ok I've switched to a thick line, @danomatika please report (IMO it's smth you can only see on over-sampled displays, like retina; actually maybe only specific to MacOS).

Looks better. I still see the center edge warping *slightly* at certain angles but it's much less pronounced and probably unavoidable as you say.

>>Also, I don't know if you are aware, but I implemented a version of mknob for PdParty on iOS
>that can't be data structures, right? How did you do it? Is that an external?

PdParty uses custom GUI implementations using CoreGraphics. I have basically *emulated* the ones built into Pd and use the send and/or receives to communicate between the GUI layer and the Pd patch. In the future, I would like to abstract the GUI messaging from the Pd core so I can drop my emulations and simply draw to a CoreGraphics canvas directly.

porres:
>  Minimum and maximum values are drawn the same for full angular control.

it's for linear control as well

Ant1r:
> Minimum and maximum values are drawn the same for full angular control.

ok thanks alex, this is fixed now.

btw I restored something that iemsliders do as well, which is to have a small security-margin at the boundaries, to help setting the gui to minimum or maximum.

umlaeute:
> Well, now I see an actual bug. Minimum and maximum values are drawn
> the same for full angular control. You should have the arc on full
> spin for the maximum value, but it's wrapping to the beggining. So you
> cannot differentiate from full off to full on.

hmm i haven't really checked, but isn't the "buggy" behaviour you describe desired for "endless rotary" controllers?

Ant1r:
No, the fact is there was this bug (fixed now): 
when a knob angular range is 360° (like -180:180), if you send max value to the knob, then the arc is displayed empty ( just like if you send min value).

porres:
> fixed now

confirmed, fixed! 

this is finally ready, huh? :)

Ant1r:
> this is finally ready, huh? :)

hey, you never know ;-)

umlaeute:
> when a knob angular range is 360° (like -180:180), if you send max value to the knob, then the arc is displayed empty ( just like if you send min value).

so the problem was mainly when setting the values via messages (rather than dragging the rotary)?

thanks for clarifying.

porres:
> so the problem was mainly when setting the values via messages (rather than dragging the rotary)?

for linear control the problem would still occur for 360º rotation, if you dragged it to the maximum, the arc disappeared

porres:
so, this now has conflicts and maybe should also adopt the changes being made in https://github.com/pure-data/pure-data/pull/1633

Ant1r:
> so, this now has conflicts and maybe should also adopt the changes being made in #1633

sure! however I would first need to know if this PR is likely to be adopted before it get outdated again ;-) 

@millerpuckette do you think this knob could find its place into Pd? Or is there something you would want to change?

danomatika:
My $0.02: I would love to have `[knb]` integrated to Pd. It fills a needed space for certain GUIs and it's already modeled on the existing IEM GUI code pattern.

Spacechild1:
I also think that `[knb]` is badly needed! Knobs are such a ubiquitous UI element in electronic music - and for good reasons: they can offer the same precision as sliders while taking up very little space.

danomatika:
Also note: I have `[knb]` implemented in PdParty. :)

danomatika:
> I also think that `[knb]` is badly needed! Knobs are such a ubiquitous UI element in electronic music - and for good reasons: they can offer the same precision as sliders while taking up very little space.

Hah ja. I need it for my minimoog implementation...

porres:
of course I can only say knobs are a key GUI element issue as well and I need these to implement a matrix grid of knobs :) 

moreover, with the work being done in refactoring GUIs by IOhannes now, it seems like a good thing to concentrate for 0.53: fixing lots of iemguis stuff and including knob as a combo ;)

Ant1r:
closing this one in favor of #1660.

PR #939, opened 2020-04-14
Fix methods memory management with multi-instance

pierreguillot:
This PR is related to https://github.com/libpd/libpd/issues/271.

With multi-instance support (PDINSTANCE=1), the memory of the methods of the `t_class` is freed twice, https://github.com/pure-data/pure-data/blob/master/src/m_class.c#L215 and https://github.com/pure-data/pure-data/blob/master/src/m_class.c#L527. 
The solution is to check the validity of the pointer before and to initialize the pointer to NULL after.


umlaeute:
merged into `update/0.50`

Issue #941, opened 2020-04-14
request add RGB colors to iemguis


porres:
In my idea, you could have separate methods for front/back/label, that takes lists of three floats (ints, actually, from 0 to 255).

It wouldn't be too hard to include these methods. Even I can achieve such task :)

umlaeute:
why wouldn't you use the the hex-notation instead?

which extra functionality does your request give you?

porres:
> which extra functionality does your request give you?

I think using RGB is more convenient and you can have a patch with 3 sliders setting the color of each colored element individually (also more convenient). Whereas converting from RGB to hex requires some work and you always have to set the 3 colors at the same time. 

It's like in the same way that hex-notation doesn't add any functionality, it'd be only about convenience

Spacechild1:
> setting the color of each colored element individually

you can use `[makefilename %0.2x]` for each color component

> Whereas converting from RGB to hex requires some work and you always have to set the 3 colors at the same time.

not really, just use `[pack s s s] -> [symbol #$1$2$3(`. in the end, it's not much different than using 3 floats. just make an abstraction and use it in the documentation, so people get the idea.

Ant1r:
I think Alex (also) speaks about setting individually colors: back, front, label.
It seems to me it's quite easily feasible in hacking` iemgui_color()`: https://github.com/pure-data/pure-data/blob/e2ef8158e83900734d263a3fda343a47ed0c9e3a/src/g_all_guis.c#L522-L536

We could change it to allow :
`[ color label #RRGGBB (`
or:
`[ color background 125 200 10 (`

porres:
> not really, just use [pack s s s] -> [symbol #$1$2$3(. in the end, it's not much different than using 3 floats. just make an abstraction and use it in the documentation, so people get the idea.

I was actually doing that already, and using [makefilename] ;) yeah, it's not that much different in the end...

> I think Alex (also) speaks about setting individually colors: back, front, label.

I am! So it's not like I was asking to change the message format of 'color' to allow RGB.

I was thinking about new special methods, like 'back_color 127 100 50', 'front_color <list>', label_color <list>', but whatever. This is the main idea I guess, and the idea that this would allow RGB just feels more natural than choosing to stick with hex-notation.

Ant1r:
try #942 :-)

PR #942, opened 2020-04-14
iemguis color method: allow setting individual colors

Ant1r:
allow to send to a gui: `[ color name {col} (` ; closes https://github.com/pure-data/pure-data/issues/941

color names are: "background" "front" and "label"

color can be expressed as:
- hex color: `#RRGGBB`
- float triplet in [0;255] range: `red green blue`
- one-float Pd color scheme index (e.g: 13 = red)
- legacy one-float 18-bit style
Capture du 2020-04-15 00-03-49

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -363,6 +363,14 @@ int iemgui_compatible_colorarg(int index, int argc, t_atom* argv)
     return iemgui_getcolorarg(index, argc, argv);
 }
 
+int iemgui_color_single(int ac, t_atom *av)
+{
+    if(ac < 3) return iemgui_compatible_colorarg(0, ac, av);
+    return (((int)atom_getfloatarg(0, ac, av) & 255) << 16)
+        + (((int)atom_getfloatarg(1, ac, av) & 255) << 8)
+        + ((int)atom_getfloatarg(2, ac, av) & 255);
+}
+
 void iemgui_all_dollar2raute(t_symbol **srlsym)
 {
     srlsym[0] = iemgui_dollar2raute(srlsym[0]);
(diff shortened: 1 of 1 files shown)


Spacechild1:
> * one-float Pd color scheme index (e.g: 13 = red)
> * legacy one-float 18-bit style

aren't those the same? ;-)

Ant1r:
> aren't those the same? ;-)

I don't think so, for example:
- pure red (0xFF0000) is -1.67117e+07 in 18-bits (try in `hslider_help/[pd edit]`)
- palette red is 13, and is actually 16525352 = 0xFC2828

Spacechild1:
Ok, after all those years I now realize that a *positive* float color value actually is the index in the color palette in the properties dialog... :-/ I've always wondered what's the logic behind the positive values and why the bitpacked color is negative... *huge facepalm* It's actually not documented at all...

Ant1r:
well anyway maybe the 18-bit format will loose its utility after the hex format (or float triplet) is better documented, and is used for saving the patches...

porres:
> Ok, after all those years I now realize that a positive float color value actually is the index in the color palette in the properties dialog... :-/ I've always wondered what's the logic behind the positive values and why the bitpacked color is negative... huge facepalm ... Actually, it's not documented at all...

Me neither, but now that I'm working on a reivision of it, I also had this revelation.

> legacy one-float 18-bit style

so that's "negative" values?

Ant1r:
@millerpuckette maybe we can merge this too? It's a little change I think, and could be useful sometimes.

porres:
yeah, there are many simple fixes and improvements on GUIs that have been sitting here for a while, I hope this and the slider fixes get an attention to get it and, if so, maybe some other simple ones that I can search for.

porres:
well, obviously this one also needs some updates, would you take a look at this one also @Ant1r ? hope it's not too much work :) I am planning to start a new GUI related PR and can bring this one in anyway

Ant1r:
This PR is now up to date.

Issue #417, opened 2018-08-30
Labels with "[#]" makes PD non-responsive


HenriAugusto:
Replicating the issue:

1. use a messagebox to set the label for a gui object to anything that contains an **[#]**
2. right-click and try to open the [hslider]'s properties

![pd label bug](https://user-images.githubusercontent.com/11790621/44881460-1738e580-ac86-11e8-8c42-42bd1d794557.png)

PD becomes non-responsive.

danomatika:
Have you tried this with the Pd 0.49 test builds? I think this is fixed already.

I'm try this with 0.49-0test3 and it seems to be working fine, although the right bracket is sometimes dropped when entering.

HenriAugusto:
I've just did some testing and actually the problem appears to be when use using anything inside square brackets.

Just created a Canvas and set it's label to [test]. Nothing will happend. But when i right-click and try to go to properties i get TCL error on the console

~~~
(Tcl) INVALID COMMAND NAME: invalid command name "test"
    while executing
"test"
    ("uplevel" body line 2)
    invoked from within
"uplevel #0 $docmds"
~~~

Probably because TCL interprets stuff inside square brackets as commands. But I've noticed that even if i tried to set the label to \[test\] i would still get the error when going to the properties

HenriAugusto:
Also if a create a canvas and set it's label to **"["** the next time i try to open it's properties then pd becomes unresponsive.

porres:
@HenriAugusto , did you notice "#" becomes "$"? It's a separate issue, I guess, but it's something that happens...

electrickery:
With 0.50.2 there is no problem with a # in the iem_gui labels. 
Yes, it is mutated into a $. This might be related to a hack which makes it possible to use $ in messages, at the cost of not being able to use #. 
The non-responsiveness doesn't occur (Xubuntu 18.04 64-bit).

porres:
possible duplicate? => https://github.com/pure-data/pure-data/issues/880

HenriAugusto:
I figure we should escape square brackets to prevent [command evaluation](https://www.tcl.tk/man/tcl8.5/tutorial/Tcl5.html) when sending the message do the GUI.

umlaeute:
@HenriAugusto problems with square brackets are tracked in #880, which is **not** a duplicate of this ticket. (and #880 also outlines how to do it)

porres:
[...]

PR #948, opened 2020-04-18
Permissive iemgui labels

### LINKED ISSUE #880: weird characters like "[" (as label) hang iemgui properties
It seems that this arrangement causes problems:

a canvas, with a label far outside (to the top-right) but with a really large font size - or is it to do with the use of '[' as a label?
Right-click on the canvas object handle and select properties, pd freezes completely and you cannot do anything, and have to kill Pd.

Pd-0.50.2, Linux, Debian 9, 64 bit, intel
largeFontCanvasBug

[largeFontCanvasBug.pd.zip](https://github.com/pure-data/pure-data/files/4128587/largeFontCanvasBug.pd.zip)

HenriAugusto:
This allows the use of `[` and `]` in iemgui's objects labels.

I've tested the labels and checked if the send/receives are working as expected.

_(Tested on Win 10)_

image

Fixes #880

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -12,6 +12,7 @@
 #include <ctype.h>
 #include "m_pd.h"
 #include "g_canvas.h"
+#include "s_stuff.h"
 
 #include "g_all_guis.h"
 #include <math.h>
@@ -186,21 +187,17 @@ t_symbol *iemgui_raute2dollar(t_symbol *s)
 t_symbol *iemgui_put_in_braces(t_symbol *s)
 {
     const char *s1;
-    char buf[MAXPDSTRING+1], *s2;
+    char buf[MAXPDSTRING], *s2;
     int i = 0;
     if (strlen(s->s_name) >= MAXPDSTRING)
         return (s);
     for (s1 = s->s_name, s2 = buf; ; s1++, s2++, i++)
     {
         if (i == 0)
-        {
-            *s2 = '{';
-            s2++;
-        }
+            *(s2++) = '{';
         if (!(*s2 = *s1))
         {
-            *s2 = '}';
-            s2++;
+            *(s2++) = '}';
             *s2 = '\0';
             break;
         }
@@ -477,10 +474,14 @@ void iemgui_label(void *x, t_iemgui *iemgui, t_symbol *s)
     iemgui->x_lab_unexpanded = iemgui_raute2dollar(s);
     iemgui->x_lab = canvas_realizedollar(iemgui->x_glist, iemgui->x_lab_unexpanded);
 
+    char lab_escaped[MAXPDSTRING];
+    lab_escaped[MAXPDSTRING-1] = 0;
+    pdgui_strnescape( lab_escaped, MAXPDSTRING, iemgui->x_lab->s_name, strlen(iemgui->x_lab->s_name) );
+
     if(glist_isvisible(iemgui->x_glist) && iemgui->x_lab != old)
-        sys_vgui(".x%lx.c itemconfigure %lxLABEL -text {%s} \n",
+        sys_vgui(".x%lx.c itemconfigure %lxLABEL -text [::pdtk_text::unescape \"%s \"] \n",
                  glist_getcanvas(iemgui->x_glist), x,
-                 strcmp(s->s_name, "empty")?iemgui->x_lab->s_name:"");
+                 strcmp(s->s_name, "empty")?lab_escaped:"");
 }
 
 void iemgui_label_pos(void *x, t_iemgui *iemgui, t_symbol *s, int ac, t_atom *av)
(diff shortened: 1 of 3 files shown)


Ant1r:
great! with this, you can still open the properties dialog, which was hanging Pd before.

porres:
I tested and it does fix the usage of brackets!

I also see that this fixes https://github.com/pure-data/pure-data/issues/417

porres:
hmmm, I thought the deal was to just put brackets in the tcl code, which would be much simpler than put them in the C code and send them to tcl...

HenriAugusto:
> I also see that this fixes #417

Same for me.

HenriAugusto:
> hmmm, I thought the deal was to just put brackets in the tcl code, which would be much simpler than put them in the C code and send them to tcl...

Indeed. Maybe it would better to add brackets in the `*_properties` function for each iemgui object.

```TCL
static void hslider_properties(t_gobj *z, t_glist *owner)
{
    [...]
    sprintf(buf, "pdtk_iemgui_dialog %%s |hsl| \
            --------dimensions(pix)(pix):-------- %d %d width: %d %d height: \
            -----------output-range:----------- %g left: %g right: %g \
            %d lin log %d %d empty %d \
            {%s} {%s} \
            {%s} %d %d \
            [...]);
    gfxstub_new(&x->x_gui.x_obj.ob_pd, x, buf);
}```

I'll update and force-push,

**Comment by porres** [comment]
> Maybe it would better to add brackets in the *_properties function for each iemgui object.

yeah, I was thinking about doing something like today in my other PR ;) https://github.com/pure-data/pure-data/pull/946 but I hadn't really started checking on the idea or thought about it at all...

**Comment by Ant1r** [comment]
I think I like the present C-side unified solution, better than manually editing the string for every object. But certainly it's a matter of taste...

**Comment by umlaeute** [comment]
i agree with @Ant1r that the current solution is preferable than adding the curlies in all invocations of `<iemgui>_properties`.

this makes it much more robust.

File: src/g_all_guis.c [hunk 1]
```diff
@@ -183,6 +183,31 @@ t_symbol *iemgui_raute2dollar(t_symbol *s)
...
+    int i = 0;
+    if (strlen(s->s_name) >= MAXPDSTRING)
+        return (s);
+    for (s1 = s->s_name, s2 = buf; ; s1++, s2++, i++)
+    {
+        if (i == 0)
+        {
+            *s2 = '{';
+            s2++;
+        }
```
umlaeute, line 199: please remove the trailing whitespace.

src/g_canvas.h
```diff
@@ -630,6 +630,7 @@ EXTERN void guiconnect_notarget(t_guiconnect *x, double timedelay);
 /* ------------- IEMGUI routines used in other g_ files ---------------- */
 EXTERN t_symbol *iemgui_raute2dollar(t_symbol *s);
 EXTERN t_symbol *iemgui_dollar2raute(t_symbol *s);
+EXTERN t_symbol *iemgui_put_in_braces(t_symbol *s);
```
umlaeute, line 633: no, please don't export this new function.
as @Ant1r pointed out in https://github.com/pure-data/pure-data/pull/953#issuecomment-616194527 there's already trouble that the `raute2dollar` and `dollar2raute`  functions are exported. (they should be deprecated and retired asap)

src/g_all_guis.c
umlaeute, line 199: please remove the trailing whitespace.

(side note, git has a template pre-commit hook that takes care of this; you probably should enable it in your local clone of the repository...)

HenriAugusto: ok. thanks for the tip

porres:
Just mentioning this also fixes other issues besides being able to use "[". For instance, it is necessary for https://github.com/pure-data/pure-data/pull/953

Moreover, I was detecting other similar issues, if you set a label with some characters allowed by Pd via escaping with backlash, you get a tcl error when opening the properties, but this PR fixes it.

See example of that issue below

79516014-7f396c80-8020-11ea-8c51-d6595f21dae5

HenriAugusto:
> while you are there, please also allow labels that have curlies (while you cannot type curlies in Pd, it's easy enough to create them with `[makefilename %c]`, and Pd supports displaying curlies now.
> use `pdgui_strnescape` for this.

Ok this feels very silly but i can't seem to get it working. 
The `pdgui_strnescape()` adds a `'\\'` char before the braces. But when TCL receive the string it receive the backward slashes . 

image


```C
void iemgui_label(void *x, t_iemgui *iemgui, t_symbol *s)
{
    t_symbol *old;
        /* tb: fix for empty label { */
        if (s == gensym(""))
                s = gensym("empty");
        /* tb } */
    old = iemgui->x_lab;
    iemgui->x_lab_unexpanded = iemgui_raute2dollar(s);
    iemgui->x_lab = canvas_realizedollar(iemgui->x_glist, iemgui->x_lab_unexpanded);

    char lab_escaped[MAXPDSTRING];
    lab_escaped[MAXPDSTRING-1] = 0; //is this line necessary?
    pdgui_strnescape( &lab_escaped[0], MAXPDSTRING, iemgui->x_lab->s_name, strlen(iemgui->x_lab->s_name) );
    if(glist_isvisible(iemgui->x_glist) && iemgui->x_lab != old)
        sys_vgui(".x%lx.c itemconfigure %lxLABEL -text {%s} \n",
                 glist_getcanvas(iemgui->x_glist), x,
                 strcmp(s->s_name, "empty")?lab_escaped:"");
}
```

**PS:** I've clicked re-review changes accidentally

umlaeute:
>  But when TCL receive the string it receive the backward slashes.

there's an unescape proc somewhere in Pd-GUI.

/me looks

ah yes: `::pdtk_text::unescape` (and it requires your label to end with a space (` `) that will be stripped away. don't ask)

HenriAugusto:
Ok that was a little bit trickier than i thought.

Users still won't be able to use backward slashes in the labels but i guess it's okay.

umlaeute:

PR #953, opened 2020-04-18
gui properties less chars substitution/escaping

### LINKED ISSUE #558: '$/#' substitutions on iemguis and atom boxes
EDITED title and this comment to summarize discussion below:

Because Pd didn't use to allow ' $' not in the beginning of symbols, iemguis use '#' to represent ' $',  even though '$' doesn't get replaced in the beginning of a symbol.

Moreover, '#' gets replaced by '$'. So we have issues like not being able to set a label as "C#" (you get 'C$' instead).

Paralel issue: escaping "$" with "\" doesn't really work inside properties.

----------

Edit 2: atom boxes also seem to substitute "#" for "$" (i.e. a label of "C#" becomes "C$" in a number atom box)

### LINKED ISSUE #960: spaces in iemguis / atom boxes become underline
Nowadays we can use spaces with backlashes in symbols in Pd. We can use it to set a label, for instance. See picture below.

Screen Shot 2020-04-20 at 00 18 32

Currently, if we do it and try to open the properties, we get a tcl error, but there's already a PR that fixes the error, see https://github.com/pure-data/pure-data/pull/948

So, anyway, the problem is that if you open the properties (with that PR being accepted), you still have a problem because if you change properties and close it, the space will get replaced by an underline, cause that's what happens these days if you add spaces to a label in the properties window.

Ant1r:
closes  #558 , fixes #960
If the `$` is prefixed with a `\` before the properties dialog returns the message to Pd, then `binbuf_eval() `no longer tries to interpret the `$`.

Hence $/# substitution is no longer needed.

This PR needs 723f70a from https://github.com/pure-data/pure-data/pull/948, for the strings to be protected by curly braces before being sent to Tcl.


porres:
Awesome, I was also working on this and realized it needed 723f70a but everyone is on fire in this repository (yeah!)

well, I still see "#" becoming "$" for labels, so let me have a better look, I think I know what was missing.

HenriAugusto:
In #948, @porres gave a suggestion for a cleaner/easier approach.

It basically consists of putting the braces only when constructing the string to be sent to the GUI. ([link](https://github.com/HenriAugusto/pure-data/commit/feba1c54e96a06e58dd30fa576037080c42f9800))

Although i wonder if a function to put braces around a symbol might be useful in other places (@umlaeute ?)

Let me know if i should force-push #948 with feba1c5

porres:
>

porres:
> Let me know if i should force-push #948 with feba1c5

you missed g_mycanvas

HenriAugusto:
and g_array! updated 3a578fbc80dfd1256ae4b1d5240fe4b1097065a8

porres:
> and g_array! updated 3a578fb

cool, now we have all iemguis and the array... I'll test this.

Ant1r:
I think you're speaking of PR #948 ?

porres:
> I think you're speaking of PR #948 ?

yeah :) 

anyway, I'm just about to check why "#" still gets substituted

porres:
> I'm just about to check why "#" still gets substituted

There we go, we still have this...

https://github.com/Ant1r/pure-data/blob/iem_no_dollar_subsitution/src/g_all_guis.c#L154

By the wat, being able to print "#" is what I was mostly concerned about :)

porres:
> anyway, I'm just about to check why "#" still gets substituted

Ok, not really looking deeply yet, but we still have these functions and the one that converts "#" to "$" must still be in action...

https://github.com/Ant1r/pure-data/blob/iem_no_dollar_subsitution/src/g_all_guis.c#L154

By the wat, being able to print "#" is what I was mostly concerned about :)

Ant1r:
OK! You mean # is substituted when setting the label (or send or receive BTW) via a message, like:
`[label #foo (`
you're right. I'm fixing that.

Ant1r:
done; actually the `# -> $` substitution was still here, also when receiving the properties dialog's payload .
I also removed the substitution for the garray.

BTW I was about to completely remove `dollar2raute` and `raute2dollar` functions, but as they are declared as public I couldn't (some externals may use them).

porres:
> BTW I was about to completely remove dollar2raute and raute2dollar functions, but as they are declared as public I couldn't (some externals may use them).

good point, I was removing them in my attempts :)

umlaeute:
as mentioned in #948, i somewhat prefer injecting a single function at a single place rather than having to touch all `<iemgui>_properties` functions.
as i've said elsewhere, the only *proper* solution would require a complete refactoring of the iemgui code, that follows the DRY pattern.

porres:
> Although i wonder if a function to put braces around a symbol might be useful in other places 

I can't see where this would be needed.

> i somewhat prefer injecting a single function at a single place rather than having to touch all <iemgui>_properties functions. As i've said elsewhere, the only proper solution would require a complete refactoring of the iemgui code, that follows the DRY pattern.

Still not sure what this means, but I understand that the new function "iemgui_put_in_braces" is not a "complete refactoring", so it's not like it is needed. And it is not like we're including a new function that's being repeated all of the time in all <iemgui>_properties functions, it's just that all of them had a bug (they could only receive limited symbols), and all had to be fixed.

Adding a new function that's only necessary for fixing a bug (and not used anywhere else) instead of just fixing the bugs seems hacky and convoluted code to me.

but as long as the bugs are fixed, I don't care :)

porres:
Perhaps we can build on this PR and deal with yet another annoying and now unnecessary substitution, see https://github.com/pure-data/pure-data/issues/960

umlaeute:
> Still not sure what this means, 

what? "DRY"? "refactoring"? google's your friend :-)

> but I understand that the new function "iemgui_put_in_braces" is not a "complete refactoring", so it's not like it is needed.

"it" being what?

> Adding a new function that's only necessary for fixing a bug (and not used anywhere else) instead of just fixing the bugs seems hacky and convoluted code to me.

i think it's the other way round.

Ant1r:
last commits (dbf5459) now fixes #960

HenriAugusto:
I've merged locally and tested this PR with the latest #948 commit and everything seems to be working (Win10)

HenriAugusto:
Hey! Just pointing out this PR still doesn't have c8ab6a17bd8da51e11434c5e35ae1be4dddbd9fd from #948.

(Just in case, for any reason, #953 get merged but #948 don't.)

umlaeute:

Issue #411, opened 2018-08-12
[fexpr~] segfaults


mxa:
3 objects needed:

~~~
[fexpr~ if ($x1[0] > xmax-$0, xmax-$0 = $x1[0], xmax-$0=xmax-$0*decay-$0)]
[v decay-$0]
[v xmax-$0]
~~~

then switch DSP on = segfault
**probably a wrong syntax, but segfault should not happen never the less.**

porres:
it's a syntax error, check the help file of [expr]

2018-08-12 14:22 GMT-03:00 mxa <notifications@github.com>:

mxa:
For the record: correct syntax is `_$0` instead of `-$0`.
Still, segfaults should be caught and handled gracefully.

umlaeute:
as of Pd-0.50, this `[fexpr~]` fails to create with:

~~~
Bad left value: 
=
-
xmax 
1003 
$x1
0 
~~~

so i guess the issue is fixed.

PR #956, opened 2020-04-19
ex_if: add NULL check to prevent segfault

Spacechild1:
fixes crashes like in https://github.com/pure-data/pure-data/issues/411


umlaeute:
included in `develop` (PR #1246 )

Issue #952, opened 2020-04-18
Improve linux docs     and audio-device userfriendlyness


user706:
Hi,

I'm using Debian (lastest version: Debian buster), and I've had no sound for 3 hours now.
Finally I've got ALSA work.

I'll tell you how. But please incorporate this in your documentation and improve it.

Want happy users? 
->  Mimick my setup and see how you fare...
Then improve the documentation accordingly.

## Setup

* debian buster, 
* a hdmi monitor with integrated sound (monitors with speakers built in).

Then try your pd! (without jack)

Nothing works. No sound, nothing

## Getting audio to work

Here's what I need to do:
```bash
pd -alsa -alsaadd hdmi
# then go to Media->Audio Settings->Output Devices: hdmi
# Media->Test Audio and MIDI:      TEST TONES works!
#                                                     MIDI output does not work. No sound. Why?
```
But I cannot adjust the volume. That is to say: Every Linux on the planet is using PulseAudio today, and the volume adjuster has no influence.

So how did I figure out that I needed `-alsaadd hdmi` (i.e. `hdmi` parameter) ?

Bloody hard!

By running `pasuspender -- audacity` 
(`pasuspender` to temporarily disable PulseAudio the argument after `--`) 
and trying various Playback Devices... till I found that `hdmi` works and gives me sound.

pd is not helpful here - but it should be:
```bash
./pd -listdev
# results in
audio input devices:
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)
audio output devices:
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)               ## no hdmi
API number 1

no midi input devices found              ## no midi? why?
no midi output devices found
```

That pd is unable to list `hdmi` automatically (but requires me to spell out `-alsaadd hdmi -alsamidi`), is a defect. Please improve that!

```bash
pd -listdev -alsaadd hdmi -alsamidi          # spoonfeeding pd
# results in
audio input devices:
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)
3. hdmi
audio output devices:
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)
3. hdmi
API number 1

MIDI input devices:
1. ALSA MIDI device #1
MIDI output devices:
1. ALSA MIDI device #1
```

Here's my `aplay -L` list: [link](https://justpaste.it/1o4p6)

[...]

Issue #846, opened 2019-12-21
Pulseaudio API


hjek:
**Feature request:** PulseAudio input/output for Pd.

Currently, the Pd ALSA API doesn't work unless PulseAudio is suspended,

    ALSA input error (snd_pcm_open): Device or resource busy
    ALSA output error (snd_pcm_open): Device or resource busy

I'm aware [that there are workarounds](https://puredata.info/docs/JackRoutingMultichannelAndBrowserAudio/) but it would still be convenient if Pd could play nicely with PulseAudio out-of-the-box.

claudeha:
you can use pulse via Pd's portaudio implementation

umlaeute:
closing this as PulseAudio  is already supported (as @claudeha mentions), and explicitely requesting `ALSA` is just this: you tell Pd to use `ALSA` instead of PulseAudio or JACK; and if the (low-level) ALSA-device is grabbed, you just cannot grab it.

hjek:
> you can use pulse via Pd's portaudio implementation

OK. PortAudio is not showing any inputs or outputs here but I guess that's a separate issue.

user706:
Pulseaudio is NOT supported.

> you can use pulse via Pd's portaudio implementation
> closing this as PulseAudio is already supported (as @claudeha mentions)

NO. There is no `pulse/` header in the [portaudio repository](http://www.portaudio.com/usinggit.html).

It's confirmed by comments [here](https://portaudio.music.columbia.narkive.com/doCvUSqY/pulseaudio-vs-portaudio):

* _"Portaudio has an Alsa implementation on Linux, but no PulseAudio implementation"_
* Ross Bencina himself confirms: _"It is true that there is no PulseAudio-based implementation of PortAudio in our repository."_


What would one require to use pulseaudio? Typically the pulseaudio client libs.
See examples here:
* openal: https://github.com/kcat/openal-soft/blob/master/alc/backends/pulseaudio.cpp 
(see [backends](https://github.com/kcat/openal-soft/tree/master/alc/backends))
* libsdl: https://github.com/dhewg/libsdl/tree/master/src/audio/pulseaudio 
(see [parentdir](https://github.com/dhewg/libsdl/tree/master/src/audio))


This issue, is a very valid feature-request. 
I'd recommend: **reopen this issue**.
(There's no-one forcing you to implement it. But it is not correct to have this feature-request closed with an argument that pulseaudio is supported via portaudio. That's plain wrong.)


[...]

PR #963, opened 2020-04-20
initial PulseAudio support

umlaeute:
PulseAudio is a widespread audio framework for linux.
even though (i think that) *jackd* and plain *ALSA* are preferably used in a ((semi-)professional) environment, we cannot deny that virtually all desktop environments come with a pre-configured PulseAudio daemon running - and people just expect an audio application like Pd to "just work"™ 

Closes: #952 
Closes: #846 

the PR also does a bit of cleanup in the audio-backend side:
- remove *AudioUnit* backend (it was an empty stub; nobody ever used it)
- remove *Enlightened Sound Daemon* (it's been dead for a decade upstream; it's been disabled in Pd's build-system for eons)
- generalize the `-jackname` cmdline flag to `-audioname` and made it usable by other audio backends as well (notably: PulseAudio :-))
   - as a side-effect, the actual client-name used by jack (wihch might be different from the requested one, in the case of name clashes) is no longer saved .


Ant1r:
that's great!

it works quite well here.

well... it adds about 200ms latency, it seems that the "delay" setting in the audio dialog has no effect, I get noticeable dropouts, but... I can finally watch videos while making sound with Pd easily ;-)

umlaeute:
right, the `delay` doesn't do anything for now.

i personally don't worry much about the large default delay (the only reason to use PulseAudio is to be able to watch youtube while patching...).

however, the dropouts are a bit disturbing.

user706:
Hi, thanks for working on this.

I get a lot of crackle sound artifacts: in fact the whole time, whenever I have youtube sound running at the same time.

So this has to be fixed first.

PS: 
I do not get crackle sounds with
```bash
./bin/pd -alsa   -alsaadd pulse   -audioadddev pulse   -alsamidi
```

user706:
> I get a lot of crackle sound artifacts: in fact the whole time, whenever I have youtube sound running at the same time.

The easiest way to recognize it, is when running it together with this video...
https://www.youtube.com/watch?v=xGXYFJmvIvk

(Again... `./bin/pd -alsa   -alsaadd pulse   -audioadddev pulse   -alsamidi` does not give me those artifacts)

user706:
I also get a kind of beep sound sometimes, when running 
* that youtube video 
and the 
* Test audio and MIDI (440 Hz sine)

**if** I ramp up the slider to 100 dB (***) (and jump it down and up a bit; while possibly pausing and continuing the youtube video). I wonder if someone else also gets this?

(***) Just put your normal volume lower, to prevent losing your hearing! :)

user706:
`top` reports ridiculously high CPU usage for process `pulseaudio`, when trying the implementation here (but also when running `./bin/pd -alsa -alsaadd pulse -audioadddev pulse -alsamidi`).

That does not happen when pd is killed and I'm just running youtube; not even when running 3 youtube videos simultaneously (it's still very low on cpu). Where is that high CPU usage coming from??? (probably eats a lot CPU cycles to generate those popping noises and artifacts :) !)

user706:
By the way:
I get no artifacts and less CPU usage if I put this

https://github.com/pure-data/pure-data/blob/9ae3f4f7cb2db4b1f82ff65cb38d95666cfc13fd/src/s_audio_pulse.c#L77

back to NULL.

src/s_audio_pulse.c
```diff
@@ -0,0 +1,171 @@
...
+        t_sample * fp = STUFF->st_soundin;
+        if(pa_simple_read(pulse_source->client, pulse_source->buffer, pulse_source->bufsize, &err) < 0)
+            error("Pulse Audio: %s", pa_strerror(err));
+        else {
+            const int channels = STUFF->st_inchannels;
+            int chan, frame;
+            float*buf = pulse_source->buffer;
+            for(chan = 0; chan<channels; chan++)
+                for(frame = 0; frame<DEFDACBLKSIZE; frame++)
+                    *fp++ = (t_sample)buf[chan + frame * channels];
```
user706, line 130: `float*buf = pulse_source->buffer;` can be const: `const float*buf = pulse_source->buffer;`

Please optimize the fragment as shown here:

```c
            int chan;
            const float*buf = pulse_source->buffer;
            if (channels == 1 && (sizeof(*fp) == sizeof(*buf)))
                memcpy(fp, buf, DEFDACBLKSIZE * sizeof(*fp));
            else {
                t_sample*lim;
                for(chan = 0; chan<channels; chan++) {
                    buf = pulse_source->buffer + chan;
                    lim = fp + DEFDACBLKSIZE;
                    for (; fp < lim; buf += channels) {
                        *fp++ = (t_sample)*buf;
                    }
                }
            }
```

or if you want to be "fancy" (but less readable), then you can use loop unrolling with duff's device:

```c
            int chan;
            const float*buf = pulse_source->buffer;
            if (channels == 1 && (sizeof(*fp) == sizeof(*buf)))
                memcpy(fp, buf, DEFDACBLKSIZE * sizeof(*fp));
            else {
                t_sample*lim;
                for(chan = 0; chan<channels; chan++) {
                    buf = pulse_source->buffer + chan;
                    lim = fp + DEFDACBLKSIZE;
                    // optimization: Duff's device
                    switch (DEFDACBLKSIZE % 4) {
                        case 0:
                            do {
                                *fp++ = (t_sample)*buf; buf += channels;
                        case 3:
                                *fp++ = (t_sample)*buf; buf += channels;
                        case 2:
                                *fp++ = (t_sample)*buf; buf += channels;
                        case 1:
                                *fp++ = (t_sample)*buf; buf += channels;
                            } while (fp < lim);
                    }
                }
            }
```

src/s_audio_pulse.c
```diff
@@ -0,0 +1,171 @@
...
+        int err = 0;
+        t_sample * fp = STUFF->st_soundout;
+        if (fp) {
+            int chan, frame;
+            float*buf = pulse_sink->buffer;
+            const int channels = STUFF->st_outchannels;
+            for(chan = 0; chan<channels; chan++)
+                for(frame = 0; frame<DEFDACBLKSIZE; frame++)
+                    buf[chan + frame * channels] = (float)*fp++;
+        }
```
user706, line 144: `t_sample * fp = STUFF->st_soundout;` can be const: `const t_sample * fp = STUFF->st_soundout;`

Please optimize the fragment as shown here:

```c
        const t_sample * fp = STUFF->st_soundout;
        if (fp) {
            const int channels = STUFF->st_outchannels;
            float*buf = pulse_sink->buffer;
            if (channels == 1 && (sizeof(*fp) == sizeof(*buf)))
                memcpy(buf, fp, DEFDACBLKSIZE * sizeof(*fp));
            else {
                int chan;
                const t_sample*lim;
                for(chan = 0; chan<channels; chan++) {
                    buf = pulse_sink->buffer + chan;
                    lim = fp + DEFDACBLKSIZE;
                    for(; fp < lim; buf += channels)
                        *buf = (float)*fp++;
                }
            }
        }
```
or if you want to be "fancy" (but less readable), then you can use loop unrolling with duff's device:
```c
        const t_sample * fp = STUFF->st_soundout;
        if (fp) {
            const int channels = STUFF->st_outchannels;
            float*buf = pulse_sink->buffer;
            if (channels == 1 && (sizeof(*fp) == sizeof(*buf)))
                memcpy(buf, fp, DEFDACBLKSIZE * sizeof(*fp));
            else {
                int chan;
                const t_sample*lim;
                for(chan = 0; chan<channels; chan++) {
                    buf = pulse_sink->buffer + chan;
                    lim = fp + DEFDACBLKSIZE;
                    // optimization: Duff's device
                    switch (DEFDACBLKSIZE % 4) {
                        case 0:
                            do {
                                *buf = (float)*fp++; buf += channels;
                        case 3:
                                *buf = (float)*fp++; buf += channels;
                        case 2:
                                *buf = (float)*fp++; buf += channels;
                        case 1:
                                *buf = (float)*fp++; buf += channels;
                            } while (fp < lim);
                    }
                }
            }
        }
```

umlaeute: same as above

umlaeute:
so adding some real numbers for CPU usage on my system.

## system

- Debian/GNU linux (sid)
- pulseaudio (13.0-5) using *jackd2* (1.9.12~dfsg-2) as backend
- Pd started with `testtone.pd`
- browser has a single window showing https://www.youtube.com/watch?v=xGXYFJmvIvk
- browser is **either** of `firefox` (firefox-esr_68.7.0esr-1) **or** `chromium` (81.0.4044.92-1)

|process| CPU | notes |
| ---- | --- | --- |
| chromium | 20-30% |  *that's* a CPU hog |
| Pd-GUI | ~60% | *that's* a CPU hog |
| pd -jack | ~1.5% | 
| pd -pulseaudio | ~8% |  `buff_attr=NULL` 
| pd -pulseaudio | ~8% | `buff_attr` with blocksize=64
| pd -pulseaudio | ~3% | `buff_attr` with blocksize=2048

i have no idea why Pd-GUI takes so much resources (but it's somewhat consistent with other wish-applications, like `gitk`)

src/s_audio_pulse.c
Spacechild1: @user706 To expand: Using Duff's device with a *constant* loop length is entirely pointless. The compiler already nows the loop length at compile time and will unroll appropiately. Also, the compiler will automatically use `memcpy` (or proper SIMD instructions) whenever it can, there's no need to put a special check for length=1, in fact you've just added another branch.

Duff's device and similar optimization techniques were invented in the 80s. C/C++ compilers have gone a *long* way since the last 30-40 years, so take any optimization guide you read with a grain of salt. Test it yourself by compiling with full optimization levels and look at the assembly (godbolt.org is you're friend). We should aim to write readable and maintainable code in the first place and only resort to tricks when we actually notice a bottleneck or can't trust the compiler to do the right thing. I always recommend the following quite entertaining video to get a glimpse at the capabilities of modern compilers: https://www.youtube.com/watch?v=w0sz5WbS5AM

Spacechild1:
> i have no idea why Pd-GUI takes so much resources

:-O do those numbers refer to the CPU cost relative to a single core or all four cores, i.e. does 100% mean 1 core utilized or all 4 cores utilized? 

As a reference, on my very similar Lenovo Thinkpad (i3-4100-M, 2.5 GHz, 2 cores / 4 hardware-threads, 16 GB RAM, Windows 7), running `teststone.pd` takes 1.4% and the Pd-GUI (without mouse interaction) takes 0.2%. Eitherway, this sounds like a very serious Tk bug on Linux...

umlaeute:
the values are normalized per CPU (i think); so 100% would just eat a single CPU.

in any case, the values are all scaled to the same reference, so that doesn't matter much.
the point is, that i don't see any CPU hog at all (in the pulse implementation)

Spacechild1:
so the Pd-GUI will *only* eat over half a core? I'll check on my Linux machine... Sorry for the digression.

src/s_audio_pulse.c
user706: > either this optimisation fixes the "ridiculously high CPU usage"

No this is completely unrelated to that.

The main point here is just a very minor point.

And that point is:
I personally would not use the following, in the innermost loop
```c
                for(frame = 0; frame<DEFDACBLKSIZE; frame++)
                    buf[chan + frame * channels] = (float)*fp++;
```
but instead
```
                    for(; fp < lim; buf += channels)
                        *buf = (float)*fp++;
```
for the chance that it might be slightly more efficient (while preserving readability).
This is spelling out invariants more easily for the human reader (and for dumb compilers).

(Forget duff!)


But yes, compilers have come a long way. One would need to benchmark this.
Not much use, at guessing which one is faster: [link](ttps://godbolt.org/e?readOnly=true#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAM1QDsCBlZAQwBtMQBGAFlICsupVs1qhkAUgBMAISnTSAZ0ztkBPHUqZa6AMKpWAVwC2tEJNJb0AGTy1MAOWMAjTMRABmAAykADqgWE6rR6hiZmvv6BdDZ2jkYubl6KypiqQQwEzMQEIcam5koqanQZWQQxDs6uHt4Kmdm5YQX15baV8dVeAJSKqAbEyBwA5FLuWFRtANQAItgAYtMAgjrSVgDSDACSAFrYkwBs3OKei8eLAG6oeOiTwKjUrKjMBJMAVE4GVJykk1SPz28qD5vpNbC9kAgRHZWApOF0zuIAOyyE6TNGTNC0Oqg%2BgYyG0aEKSbidzTPFQ5Swkkoxbo37/F7vT50lkkskfL7Us50zHYv5PRlAuls37ArknbnosHk2g/KjEZhGTDi2nomjECAQkTE0mTTzUmUknRagmUg0muRyLrE5GSlm/EjUBVKnVk/XuaS/Z3K9w6WYLZarDY7bAG%2BWK5UyK12%2B3ojniACs0hNxJkXojbxlhMTZJFDwFXVeQMtUY9MaR0wRiMrEpOl2ut1QAH08EYfMRUOdMOh8wCmV85QzAcCftKTYS4VWaTy6Nix/jCa6s5TOCq6fy%2BxzY2iRRzV2XUejeS8N4KfMLdUD9zSY/ORGuj7OTwzXqxWw%2B0erNfil%2B7PRbfXHM0PUNKMZGtJFp1jLdd0%2BThU3/fEPxZN8jCXIU5BmeYlhWdYtl2ZD10dMNzyNSZUINGDSzJICYXhQ9tzRfsl17AhC2LMDZAPVUd2rKtpiGHpWBAIYEyGUhTCGTxxNQESdBLT0FD6AZMFTdxOHEggROkroegAaxABNvGEoZuHEyTpNIWShnEhQQG8LSpME0g4FgJA0DbPB2DIChNVQTzvJAYAAE53FICZWAIVw7IgJxtPEpxbCyABPESNNIDylXoAB5WhWFSpzSCwIwRGAdh4qKvBiFSNQuzswrMAAD1SAworS8SwWUCq3ycBViGSvQsHa0gCGIVt2p6Gh6CYNgOB4fhBGEUQUAUoQ8CcOzIB6VAfGKLERNslI0g0CBLEaUxvksCo4gSQQ/ACPbzruyI9uuqo3G%2BQoavSFons%2Bo69tKbI3o6D7FF%2B/Q8kEOoyhB264V6fpBi4ISRLEiSKusxqAA59gAWkOW5kGQSZgoAOncSYIFwQgSDUkE9AC1x6eteTOM0%2BLdNIAyjKEESzIxwrrNs%2ByRs51GhkkczMYOsWnK5rtiACDRuCAA%3D)

What does the compiler make of it? One would need to benchmark etc.

user706:
> so adding some real numbers for CPU usage on my system.

Thanks.
Here are my numbers:

**Running:** Debian buster

* **`pd -pulseaudio` with blocksize=64**  
  
CPU | Process
--- | ---
69,5% | pulseaudio                                                                                                          
38,4% | pd      

`-verbose` with 3 times checking and unchecking the DSP box:
```
input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (512 bytes)
PulseAudio latency 'In':	1 ms (512 bytes)

input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (512 bytes)
PulseAudio latency 'In':	662 ms (512 bytes)

input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (512 bytes)
PulseAudio latency 'In':	582 ms (512 bytes)
```

* **`pd -pulseaudio` with blocksize=2048**  
  
CPU | Process
--- | ---
36,4% | pulseaudio                                                                                                          
25,2% | pd      

`-verbose` with 3 times checking and unchecking the DSP box:
```
input channels = 2, output channels = 2
input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (16384 bytes)
PulseAudio latency 'In':	1 ms (16384 bytes)

input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (16384 bytes)
PulseAudio latency 'In':	824 ms (16384 bytes)

input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (16384 bytes)
PulseAudio latency 'In':	682 ms (16384 bytes)
```

* **`pd -pulseaudio` with buff_attr=NULL (and blocksize default of 64)**  
  
CPU | Process
--- | ---
6,0% | pulseaudio                                                                                                          
5,3% | pd      

`-verbose` with 3 times checking and unchecking the DSP box:
```
input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (512 bytes)
PulseAudio latency 'In':	1 ms (512 bytes)

input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (512 bytes)
PulseAudio latency 'In':	674 ms (512 bytes)

input channels = 2, output channels = 2
PulseAudio latency 'Out':	0 ms (512 bytes)
PulseAudio latency 'In':	1 ms (512 bytes)
```

* **`pd -jack`**

CPU | Process
--- | ---
1,7% | pd
0,3% | jackd

* **`chromium`** running  https://www.youtube.com/watch?v=xGXYFJmvIvk

CPU | Process
--- | ---
4,3% | chromium                                                                                                            
6,0% | chromium                                                                                                            
3,0% | pulseaudio                                                                                                          
1,3% | Xorg                                                                                                                
1,3% | chromium                                                                                                            
1,0% | chromium                                                                                                              

* **`firefox`** running  https://www.youtube.com/watch?v=xGXYFJmvIvk

CPU | Process
--- | ---
50,5% | Web Content                                                                                                         
18,3% |firefox-esr                                                                                                         
2,7%  | pulseaudio                                                                                                          
2,0%  | Xorg

user706:
PS: What is Pd-GUI in your CPU-load table???

If I do `ps -e | grep -i pd`
I only get
```
16803 pts/1    00:00:00 pd
16805 pts/1    00:00:00 pd-watchdog
```

Thanks.

src/s_audio_pulse.c
user706: Let me reiterate: this has no recognizable (!) influence on the exteme CPU loads that I am seeing. (Still I believe mine will be slightly faster in the inner loop. But again: This is very low priority; and not related to the extreme CPU loads I'm experiencing. and also the crackling noise artifacts).

Spacechild1: > for the chance that it might be slightly more efficient

Why do you think so? Don't guess, look at the assembly and benchmark!

> This is spelling out invariants more easily for the human reader (and for dumb compilers).

Ok, readability is certainly debatable (personally, I still find the first version clearer), but the compiler doesn't care. It perfectly understands what the author of the first version wants to do.

Now let's compile the whole thing with "-O3" and see:

https://godbolt.org/z/BhmtbF

GCC 9.3:

The inner loops are 100% equivalent. The difference is that in your version GCC decides to add an extra branch for the special case of `channels1 == 1`.

Clang 10.0:

Clang decides to do some loop unrolling, but the amount depends on the target architecture. E.g., adding "-march=haswell" will make Clang unroll more aggressively.

---

What's the takeaway? Even different compilers don't fully agree on the optimal solution (there's basically a tradeoff between reduced branching VS increased code size), so how can you? Compilers have deep knowledge of the different CPU architectures, so you can trust them that they do something reasonable with such simple code.

When writing code, express your *intent*, don't try to outsmart the compiler. In 2020 there is no need for any kind of pointer arithmetic tricks*), just use plain array subscripts, the compiler knows how to do the right thing. Only suggest micro-optimizations if you've looked at the assembly for various compilers and either proved that the result is indeed faster. Most potential for optimizations is in algorithms and data structure layout, not in how to write a for-loop.

Finally, the code in question runs *once* during each DSP tick, it is certainly not the bottleneck of your application and any attempt to "optimize" it is just wasted time.

*) you see them a lot in Pd, but don't forget that Pd was written in the 90s.

Spacechild1: > Let me reiterate: this has no recognizable (!) influence on the exteme CPU loads that I am seeing. 

Of course it doesn't, nobody ever claimed that...

user706: Nice. Thanks.

umlaeute:
indeed i forgot the `pulseaudio` process:

| process | CPU | notes |
| --- | --- | --- |
| pulseaudio | 5.5% | buff_attr with blocksize=64 (default) |
| pulseaudio | 6%| buff_attr with blocksize=2048|
| pulseaudio | 6-8% | `buff_attr==NULL`|

the "pd-gui" is the GUI of Pd, which runs is really a tcl/tk application (that is `wish`)

user706:
> indeed i forgot the `pulseaudio` process:
> process 	CPU 	notes
> pulseaudio 	5.5% 	buff_attr with blocksize=64 (default)
> pulseaudio 	6% 	buff_attr with blocksize=2048
> pulseaudio 	6-8% 	`buff_attr==NULL`
> 

Amazing. Why is yours so low?
I configured with: `./configure --prefix=$HOME/pd --disable-oss --enable-jack --enable-portmidi`
(but that will not be the difference)
My `pulseaudio --version` on my debian buster reports: `pulseaudio 12.2` (which one can also easily see here: https://packages.debian.org/buster/pulseaudio )

umlaeute:
i don't really know, but:
- i use *jack* as the backend for pulseaudio
- Pd's samplerate matches that of jack

it must be something on my side (as @Ant1r seems to have an experience similar to yours).
i haven't got any distorted sound (since i figured out that pulseaudio expects the data to be interleaved :-))

user706:
Also: why is your `pd` CPU-usage so low? (I get ridiculous [38,4%](https://github.com/pure-data/pure-data/pull/963#issuecomment-617068072), but you get [8%](https://github.com/pure-data/pure-data/pull/963#issuecomment-617010770))

user706:
>     * i use _jack_ as the backend for pulseaudio

Ah. Ok thanks. 

But: What exactly does this mean?
How can you set jack as backend for pulseaudio? Could you please explain.

Thanks.

umlaeute:
something like this:

~~~su
sudo apt install pulseaudio-module-jack
jackd
pactl load-module module-jack-sink
pactl load-module module-jack-source
~~~

might require some kills of pulseaudio to free the soundcard for jack (which i start via `qjackctl` rather than directly)

umlaeute:
so i stopped jack and switched to the native (ALSA) backend of Pulse

### `buff_attr==NULL`
| process | CPU | note
| --- | --- | --- |
| pulseaudio | ~20% |  |
| alsa-sink-ALC29 | ~9% | (child-process of pulseaudio) |
| pd | 5% | |

###  `buff_attr` with bufsize=64
| process | CPU | note
| --- | --- | --- |
| pulseaudio | ~67-90%  | :skull:  |
| alsa-sink-ALC29 | ~30% | (child-process of pulseaudio) |
| pd | 25% |crackling|

### `buff_attr` with bufsize=2048 
| process | CPU  | note
| --- | --- | --- |
| pulseaudio | ~40% |   |
| alsa-sink-ALC29 | ~15%| (child-process of pulseaudio) |
| pd | 15% | (almost?) no crackling |

umlaeute:
@Spacechild1 wrote:
> Eitherway, this sounds like a very serious Tk bug on Linux...

i realized just now that i was using the experimental Tcl/Tk-8.7 (unreleased at time of writing) for running the Pd-GUI.
downgrading to Tcl/Tk-8.6 (or upgrading from 8.7.0~a1-6 to 8.7.0~a3-1) seems to have fixed the performance hog of the Pd-GUI.

Spacechild1:
Portaudio has a PR for pulse audio support: https://github.com/PortAudio/portaudio/pull/336. Let's hope it gets merged soon.

Although I very much appreciate the effort of *this* PR, I think it would be better to use the portaudio implementation instead of rolling our own (unless there are compelling reasons to do so).

illuusio:
As an main author of https://github.com/PortAudio/portaudio/pull/336 it needs testing in real applications and bug reports from them. I try to keep up-to-date binary package here to test: https://download.opensuse.org/repositories/home:/illuusio:/portaudio/. If some distro is missing let met know.

umlaeute:
closing this.

- PortAudio will eventually gain support for PulseAudio (if they manage to make a release before the sun turns into a white dwarf)
- afaict, PulseAudio is a dead horse anyhow, and has been replaced by PipeWire as the standard desktop audio framework. and Pd's JACK backend integrates very nicely with PW (I'm using it all the time).

PR #464, opened 2018-09-13
trackable print

Ant1r:
This allows to ctrl-click on the messages in the Pd console, to find to corresponding [print] object (e.g to remove [print]s scattered all over the patch...).

It needed to add a new public function "logpoststring()" in m_pd.h, that just prints the string (associated to the object address) without a newline.

PR #965, opened 2020-04-20
improve find and error tracking

### LINKED ISSUE #1044: [clone] Menu find and error tracking cannot access clone instances
Clones should behave like any other abstractions, however the Find menu and the error tracking system (through Ctrl-clicking an error in the Pd console) cannot currently enter into clone instances.

This makes clones more difficult to debug.

### LINKED ISSUE #1045: [msg] error coming from messages are not traceable
when a `message` triggers an error, this one isn't traceable:
- menu `Find / Find Last Error` returns "no findable error yet"
- control-clicking the error in the Pd console does nothing

Capture du 2020-05-29 15-42-42

### LINKED ISSUE #1052: [find error] cannot open GOP windows
When an error originates from a normal subcanvas (or abstraction), right-clicking the error in the Pd console or calling `Find / Find Last Error` opens the subcanvas window (or brings it back to front).

But if the subcanvas (or the abstraction) is a graph-on-parent, then the window isn't opened. Only the first non-GOP parent window is opened, or brought to foreground.

Capture du 2020-06-04 00-19-49

### LINKED ISSUE #1193: [open parent window] cannot open parent if it's a GOP
it opens the parent's parent instead.
so e.g if you have multiple instances of the same GOP, you cannot know which one contains a currently visible subpatch (or abstraction).
gop_find_parent-1

Ant1r:
The 'find' and 'error tracking' (ctrl-click on the error message) features are really useful for debugging.
However some cases are not covered yet:
- errors coming from message boxes are not traceable (#1045)
- errors coming from a clone instance are not traceable (#1044)
- 'find' doesn't look into clone instances (#1044)
- find_error cannot open the window containing the emitting object if the canvas is a GOP (it opens its parent window instead) (#1052)
 - menu 'Parent Window' cannot open the parent's window if it's a GOP. (#1193)

This PR fixes #1044, fixes #1045, fixes #1052, fixes #1193 .

(in the same vein, see PR #464 which proposes to track messages sent by [print])


Ant1r:
hey @millerpuckette , would it be possible to consider this PR for next release?

I'm a little tired not being able to retrieve the exact location of some my bugs (yeah, I confess I _occasionally_ make mistakes while working on my projects ;-), although I know I've already fixed the cases where this is not possible.

Many thanks!

Ant1r:
I've brought this branch up to date.
I also added allowing "Parent Window" menu to open the parent's window even when it's a GOP (even though this could considered a bit OT, I preferred to add it to the current branch/PR for simplicity).

Issue #967, opened 2020-04-23
with alsa: fails to list many devices


user706:
alsa does not show many of my devices.

`pd -alsa -listdev` shows only
```
audio input devices:
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)
audio output devices:
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)
API number 1

MIDI input devices:
1. ALSA MIDI device #1
MIDI output devices:
1. ALSA MIDI device #1
priority 94 scheduling failed.
```

But `aplay -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq` shows:
```
default
null
jack
pulse
sysdefault
front
surround21
surround40
surround41
surround50
surround51
surround71
iec958
hdmi
dmix
dsnoop
hw
plughw
usbstream
```

and `arecord -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq`
shows
```
default
null
jack
pulse
sysdefault
front
dmix
dsnoop
hw
plughw
usbstream
```

This is a miserable situation.
I currently launch pd as follows:
```
pcms="$(aplay -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq)"
top_pcm="$(echo "$pcms" | head -1)"
alsaadder="$(echo $pcms | xargs printf "-alsaadd %s ")"
./bin/pd -alsa $alsaadder -audioadddev $top_pcm -alsamidi
```

umlaeute:
Pd only lists hardware devices.

afaict, this is intentional (all those virtual devices are really not going to make you happy when dealing with a system like Pd).

i'm inclined to close this, but...

PR #968, opened 2020-04-23
improve alsa device detection

user706:
This is an attempt to improve  #967

It now makes `pd -alsa -listdev` list
```
audio input devices:
1. default
2. null
3. jack
4. pulse
5. sysdefault
6. front
7. dmix
8. dsnoop
9. hw
10. plughw
11. usbstream
audio output devices:
1. default
2. null
3. jack
4. pulse
5. sysdefault
6. front
7. surround21
8. surround40
9. surround41
10. surround50
11. surround51
12. surround71
13. iec958
14. hdmi
15. dmix
16. dsnoop
17. hw
18. plughw
19. usbstream
API number 1

no MIDI input devices found
no MIDI output devices found
priority 94 scheduling failed.
```
i.e. the same as 

`aplay -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq`
and
`arecord -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq`

### DIFF
--- a/src/s_audio_alsa.c
+++ b/src/s_audio_alsa.c
@@ -25,6 +25,7 @@
 #include <sched.h>
 #include <sys/mman.h>
 #include "s_audio_alsa.h"
+#include "s_alsa_pcm_list.h"
 #include <endian.h>
 
 /* Defines */
@@ -37,7 +38,8 @@
 #endif
 
 static void alsa_checkiosync(void);
-static void alsa_numbertoname(int iodev, char *devname, int nchar);
+static void alsa_out_numbertoname(int devno, char *devname, int nchar);
+static void alsa_in_numbertoname(int devno, char *devname, int nchar);
 static int alsa_jittermax;
 #define ALSA_DEFJITTERMAX 5
 
@@ -290,7 +292,7 @@ int alsa_open_audio(int naudioindev, int *audioindev, int nchindev,
 
     for (iodev = 0; iodev < naudioindev; iodev++)
     {
-        alsa_numbertoname(audioindev[iodev], devname, 512);
+        alsa_in_numbertoname(audioindev[iodev], devname, 512);
         err = snd_pcm_open(&alsa_indev[alsa_nindev].a_handle, devname,
             SND_PCM_STREAM_CAPTURE, SND_PCM_NONBLOCK);
         check_error(err, 0, "snd_pcm_open");
@@ -304,7 +306,7 @@ int alsa_open_audio(int naudioindev, int *audioindev, int nchindev,
     }
     for (iodev = 0; iodev < naudiooutdev; iodev++)
     {
-        alsa_numbertoname(audiooutdev[iodev], devname, 512);
+        alsa_out_numbertoname(audiooutdev[iodev], devname, 512);
         err = snd_pcm_open(&alsa_outdev[alsa_noutdev].a_handle, devname,
             SND_PCM_STREAM_PLAYBACK, SND_PCM_NONBLOCK);
         check_error(err, 1, "snd_pcm_open");
@@ -568,7 +570,7 @@ int alsa_send_dacs(void)
             sys_log_error(ERR_DATALATE);
             if (result == -EPIPE)
             {
-                result = snd_pcm_prepare(alsa_indev[iodev].a_handle);
+                result = snd_pcm_prepare(alsa_outdev[iodev].a_handle);
                 if (result < 0)
                     fprintf(stderr, "read reset error %d\n", result);
             }
@@ -894,8 +896,17 @@ static void alsa_checkiosync(void)
 #endif
 }
 
+typedef struct {
+    int n;
+    char **names;
+} t_detected_names;
+t_detected_names detected_in_names = {0, NULL};
+t_detected_names detected_out_names = {0, NULL};
+
 static int alsa_nnames = 0;
 static const char **alsa_names = 0;
+
+static void alsa_numbertoname(const t_detected_names *detected_names, int devno, char *devname, int nchar);
 
 void alsa_adddev(char *name)
 {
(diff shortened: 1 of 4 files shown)


src/s_audio_alsa.c
```diff
@@ -951,9 +951,20 @@ void add_to_list(t_detected_names *detected_names, const char *name)
     strcpy(detected_names->names[n], name);
 }
 
-static int get_devs(char *devlist, int maxndev, int devdescsize, t_detected_names *detected_names, alsa_pcm_list_dir dir)
+void free_list(t_detected_names *detected_names)
```
umlaeute, line 954: why is this not static?

umlaeute:
also this fails when building with the legacy build-system `src/makefile.gnu`.

user706:
There is a change that I might look at that. (I'm just checking Portaudio's implementation etc.)

> it seems quite a few devices are missing anyhow.
> here i get:
> 
> ```
> $ ./src/pd -alsa -listdev
> audio input devices:
> 1. default
[...]
> 20. dsnoop
> audio output devices:
> 1. default
[...]
> 20. surround40
> API number 1
> ```
> 
> but:
> 
> ```shell
> $ aplay -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq | grep -c .
> 30
> $ arecord -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq | grep -c .
> 23
> ```

Please can you list: what are the **additional** names that `aplay` and `arecord` lists? Thanks.

user706:
> Please can you list: what are the **additional** names that `aplay` and `arecord` lists? Thanks.

And how does that compare to the list given by portaudio? `./bin/pd -pa -listdev`

umlaeute:
missing input devices

~~~
hw                                   
plughw                               
usbstream   
~~~

missing output devices

~~~
surround41
surround50
surround51
surround71
hdmi
dmix
dsnoop
hw
plughw
usbstream
~~~

but that's pretty irrelevant, as the limiting factor seems to be `MAXNDEV`.

here's the output with (system-provided) PortAudio (even though from a user POV i don't care about pa, as i only ever use jack and ALSA)

~~~sh
$ pd -pa -listdev
audio input devices:
1. (0) jack
2. (0) pulse
3. (0) default
4. (2) system
5. (2) PulseAudio JACK Sink
audio output devices:
1. (0) HDA Intel PCH: HDMI 0 (hw:0,3)
2. (0) HDA Intel PCH: HDMI 1 (hw:0,7)
3. (0) HDA Intel PCH: HDMI 2 (hw:0,8)
4. (0) HDA Intel PCH: HDMI 3 (hw:0,9)
5. (0) HDA Intel PCH: HDMI 4 (hw:0,10)
6. (0) hdmi
7. (0) jack
8. (0) pulse
9. (0) default
10. (2) system
11. (2) PulseAudio JACK Source
~~~

user706:
>     * i no longer can select my soundcard

Currently alsa in pd only handles the following:
```bash
aplay -l | grep '^card' | cut -d: -f1 | cut -d' ' -f2 | uniq | xargs -I {} printf "    hw:%d\nplughw:%d\n" {} {}
```

I never get sound with that.

Should it not rather handle
```bash
aplay -l | grep '^card' | sed -E 's/^card ([^:]+):.*device ([^:]+):.*/\1,\2/g' | xargs -I {} printf "    hw:%s\nplughw:%s\n" {} {}
```
(if you really want `hw` and `plughw`)
?

user706:
> ```shell
> aplay -l | grep '^card' | cut -d: -f1 | cut -d' ' -f2 | uniq | xargs -I {} printf "    hw:%d\nplughw:%d\n" {} {}
> ```
> 
> ([ref](https://github.com/pure-data/pure-data/blob/80929213b540af1ba12c2c8abd7a58d0c1eb2204/src/s_audio_alsa.c#L919))
> I never get sound with that.

i.e. **no** sound with
```
speaker-test -c2 -D 'hw:0' -t sine -f 440 -l 1
```

> 
> Should it not rather handle
> 
> ```shell
> aplay -l | grep '^card' | sed -E 's/^card ([^:]+):.*device ([^:]+):.*/\1,\2/g' | xargs -I {} printf "    hw:%s\nplughw:%s\n" {} {}
> ```

I do get sound with 
```
speaker-test -c2 -D 'hw:0,3' -t sine -f 440 -l 1
```

umlaeute:
> `aplay -l | grep '^card' | sed -E 's/^card ([^:]+):.*device ([^:]+):.*/\1,\2/g' | xargs -I {} printf "    hw:%s\nplughw:%s\n" {} {}`

i think so.

user706:
Over here... https://alsa-project.org/wiki/DeviceNames#hw
they state about `hw:0,3` etc.:
_This is always available if the hardware supports playback at all, but the semantics are not defined, so **avoid using this** for other things than mono output. This should also be fine if you only target specific hardware that you know for sure how it behaves when using the "hw" device name._

umlaeute:
i think this "undefined semantics" are to be seen in the light of the other device names that go like (quoting the page that speaks of "undefined semantics"):

~~~
front
rear
center_lfe
side
surround40
surround41
surround50
surround51
surround71
~~~

the picture seems pretty clear to me: the "semantics" they talk about are playback loudspeaker layout.
this is not how Pd is designed: there is no `[dac~ center]` (and thank goodness there isn't and we can access raw channels without an arbitrary layout assigned to them; if this wasn't the case my university would need to ditch Pd immediately for any *serious* work)

so i think that the *semantics* this page is talking about apply to playback software of "produced" content (e.g. for giving you a nice experience when watching "Avatar").
it is meaningless for a low-level computer music environment like Pd: we really want a punch of "mono streams" to send our audio to, ideally with no further processing done by the "soundsystem" (this is what we use Pd for)

user706:
Extended the pull-request

`./bin/pd -alsa -audioadddev 'hw:0,3 -noadc'` 
now works and gives me sound.


> the limiting factor seems to be `MAXNDEV`.

Thanks, yes. Changed this to 40.


> this breaks backward compatibility.
> 
>     * i no longer can select my soundcard

Need to break compatibility, because (as an example) I do not get sound with the defaults that alsa device detection gives me: `hw:0` ([ref](https://github.com/pure-data/pure-data/blob/80929213b540af1ba12c2c8abd7a58d0c1eb2204/src/s_audio_alsa.c#L920)). 
I need `hw:0,3`


`./bin/pd -alsa -listdev` 
now shows me
```
audio input devices:
1. default
2. null
3. jack
4. pulse
5. sysdefault
6. front
7. dmix
8. dsnoop
9. hw
10. plughw
11. usbstream
12. hw:0,0
13. hw:0,2
audio output devices:
1. default
2. null
3. jack
4. pulse
5. sysdefault
6. front
7. surround21
8. surround40
9. surround41
10. surround50
11. surround51
12. surround71
13. iec958
14. hdmi
15. dmix
16. dsnoop
17. hw
18. plughw
19. usbstream
20. hw:0,0
21. hw:0,1
22. hw:0,3
23. hw:0,7
24. hw:0,8
25. hw:0,9
API number 1

no MIDI input devices found
no MIDI output devices found
priority 94 scheduling failed.
```

By the way: what is this `priority 94 scheduling failed.` stuff, I see there?

umlaeute:
> Need to break compatibility, because (as an example) I do not get sound with the defaults that alsa device detection gives me: `hw:0` ([ref](https://github.com/pure-data/pure-data/blob/80929213b540af1ba12c2c8abd7a58d0c1eb2204/src/s_audio_alsa.c#L920)).
> I need `hw:0,3`

so what again is the reasoning to switch from "devices" to "PCMs" (in aplay(1) lingo)?
the `aplay -l` list seems to list exactly the devices we want to see (including the "PCH [HDA Intel PCH], device 3: HDMI 0 [HDMI 0]" you want), whereas the `aplay -L` mostly gives noise (with the exception of `default` and maybe some virtual devices defined explicitely by the user in their `~/.soundrc`)

so why not simplify the PR (or make a new one), to list *all* devices (`hw:x,y`) not just the base-device (`hw:x`).
as an enhancement one could probably add `default`.

i cannot think of a scenario where i wanted `surround71`.

also the `hw` entries (without further ado) are pretty undefined. if i have multiple soundcards in my computer (and in desktops i practically *always* do; in laptops mostly when performing), they all add a `hw` entry.

user706:
Why do it like this?

You answer it yourself: To automatically have devices such as `default`, `pulse` etc. which give a new user like myself the works-out-of-the-box feeling.

What reason would one want, to reduce the list? It autodetects things and the user can choose what he wants (one always has the option of ignoring things like `surround71`, right?).

By the way: this mimics portaudio's device detection -> that also gives you a long list.

umlaeute:
imo the device-list really has become much less intelligible.

on my system (desktop-pc, on-board sound and external soundcard) `-listdev` shows me:

~~~
1. HDA Intel PCH (hardware)
2. HDA Intel PCH (plug-in)
3. Scarlett 2i2 USB (hardware)
4. Scarlett 2i2 USB (plug-in)
5. HDA NVidia (hardware)
6. HDA NVidia (plug-in)
~~~

which i can decode as the name of some soundcard (e.g. "HDA Intel PCH") and some weird option "hardware" or "plugin".
the soundcard-name is a nice identifier, as is usually know which (external) device I plugin into my system. if i want to use my external soundcard, i use the "Scarlett 2i2 USB" device.

with your patch applied i get this instead:

~~~
1. default
2. surround21
3. surround40
4. surround41
5. surround50
6. surround51
7. surround71
8. null
9. lavrate
10. samplerate
11. speexrate
12. jack
13. oss
14. pulse
15. upmix
16. vdownmix
17. sysdefault
18. front
19. surround21
20. surround40
21. surround41
22. surround50
23. surround51
24. surround71
25. dmix
26. dsnoop
27. hw
28. plughw
29. usbstream
30. sysdefault
31. front
32. surround21
33. surround40
34. surround41
35. surround50
36. surround51
37. surround71
38. dmix
39. dsnoop
40. hw
41. hw:0,0
42. hw:1,0
43. hw:2,3
44. hw:2,7
45. hw:2,8
46. hw:2,9
~~~

only very few of those options actually make *any* sense to me:
- `default` looks like a reasonable default
- `null` is probably a dummy sink
- `jack`, `pulse`, `oss` are most likely backends to other audio-infrastructures
the rest is mostly gibberish.
i know that those `hw:..` devices somehow relate to real (hardward) devices, but i have no clue whether my external USB interface is `hw`, or `hw:2,3` or whatever.
and why do i have two (2) `hw` interfaces, and three (3!) `surround71` choices even though i know for a fact that my external soundcard has only 2 output jacks, and while my onboard soundcard might have multiple outputs, only two are connected and while my monitor is connected via DisplayPort it doesn't have speakers at all.

umlaeute:
and to answer your question:

> Why do it like this?

what i really want is to access low-level ALSA devices rather than high-level PCM devices for a simple reason: **low latency**

imo, this is more important than making everything beginner ready.

> By the way: this mimics PortAudio's device detection 

:tada:  so now you get the **same** list from both PortAudio and ALSA. that sounds like a net-loss to me (as you lose the list from ALSA, and gain what you already have).

i understand that this might sound overly arrogant, which is not my intention.
it *is* important to help people to get Pd (cum audio) running without spending several hours to get their soundcard right, or having to start coding C and fix the audio backend (and then discuss with blokes like me).
however, it is *also* important to keep the more "professional" aspect (allowing people to tune their system for optimum performance).
i don't think we can have both simultaneously (as in: the default setting works for newbies and gives incredibly low latency), but i think we can have both side-by-side (as in: get newbies up and running immediately, and achieve low latency with "just a few clicks").

umlaeute:
side-note: on my desktop machine, the (system-provided) PA backend lists the following devices:

~~~
1. (0) HDA Intel PCH: CX20632 Analog (hw:0,0)
2. (0) HDA NVidia: HDMI 0 (hw:2,3)
3. (0) HDA NVidia: HDMI 1 (hw:2,7)
4. (0) HDA NVidia: HDMI 2 (hw:2,8)
5. (0) sysdefault
6. (0) front
7. (0) surround40
8. (0) surround51
9. (0) surround71
10. (0) lavrate
11. (0) samplerate
12. (0) speexrate
13. (0) jack
14. (0) pulse
15. (0) default
16. (2) system
17. (2) PulseAudio JACK Source
~~~

so there are much less devices than with your patch applied (for starters there are no duplicates; and for whatever reasons there are less `surround` options in total), and the names are much nicer.

i cannot really test whether any of these work (the machine being in my office, and i'm at home), and while i still don't know what the `samplerate` output is supposed to be, at least i can identify a number of actually useful devices.

Spacechild1:
Strongly agree with https://github.com/pure-data/pure-data/pull/968#issuecomment-619808859.

If portaudio already gives you a "user-friendly" PCM device list, isn't this already sufficient?

user706:
Ok.
How do you like it now?

`./bin/pd -alsa -listdev` now lists
```
audio input devices:
1. default
2. jack
3. pulse
4. hw:0,0 (PCH : ALC887-VD Analog)
5. hw:0,2 (PCH : ALC887-VD Alt Analog)
audio output devices:
1. default
2. jack
3. pulse
4. hw:0,0 (PCH : ALC887-VD Analog)
5. hw:0,1 (PCH : ALC887-VD Digital)
6. hw:0,3 (PCH : HDMI 0)
7. hw:0,7 (PCH : HDMI 1)
8. hw:0,8 (PCH : HDMI 2)
9. hw:0,9 (PCH : HDMI 3)
API number 1
```

user706:
> with your patch applied i get this instead:
> 
> ```
> [...]
> 7. surround71
> [...]
> 24. surround71
> [...]
> 37. surround71
> [....]
> ```

>       and why do i have two (2) `hw` interfaces, and three (3!) `surround71` choices

Just out of interest (and to mitigate the possibility of a bug in my code):

Does
```
aplay -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq
```
give you the same duplication?

Thanks

umlaeute:
yes, kinda (`MAXNDEV` is still too small on that system, so Pd misses a few PCMs)

~~~sh
$ aplay -L | grep '^[[:graph:]]' | cut -d: -f1 | uniq
default
surround21
surround40
surround41
surround50
surround51
surround71
null
lavrate
samplerate
speexrate
jack
oss
pulse
upmix
vdownmix
sysdefault
front
surround21
surround40
surround41
surround50
surround51
surround71
dmix
dsnoop
hw
plughw
usbstream
sysdefault
front
surround21
surround40
surround41
surround50
surround51
surround71
dmix
dsnoop
hw
plughw
usbstream
hdmi
dmix
dsnoop
hw
plughw
usbstream
~~~

PR #970, opened 2020-04-25
fixes for NULL-pointer access, unfreed resources,... discovered by coverity-scan

umlaeute:
running static code-analysis, a few bugs were discovered, as in:
- post-format specifying more arguments then were specified
- unfreed resources for early-exit due to errors
- undefined behaviour (modulo-by-0)
- NULL-pointer checks

this PR fixes these (the relevant coverity-IDs are given in the commits).


user706:
Can one see the coverity logs somewhere?

I've tried finding something but only see some graphs 
here: https://scan.coverity.com/projects/pure-data-pure-data
and seemingly nothing useful here: https://travis-ci.org/github/pure-data/pure-data/jobs/679358960#L1187

Thanks

umlaeute:
the coverity-scan logs are only available to project members (for the coverity-scan project).
afaict, coverity-scan does not provide a way to make them available to the general public (given that the issues exposed there are generally security sensitive).

i wouldn't mind adding anybody to the coverity-project, but the user-experience is steadily degrading, at least with the browsers i use (mostly firefox; if need be chromium), and *right now* both browsers give me a blank page when i try to manage the project members :-(

Issue #976, opened 2020-05-05
[bng] "intrrpt" and "hold" should be renamed.


porres:
If the idea is to call this "interrupt", we could just do that and not hide two vowels, but I'd like to open a discussion to maybe rename this to something else.

It's not intuitive to use "interrupt". The intuition could be something like interrupting the hold time if it got activated before the hold time (for like 10ms) and reflash, but that's not it... In fact the logic is a bit complicated, and hold time is not really the time it will stay on if it is not "interrupted".

The "Interrupt" time is actually the **minimum** time the bang stays activated! While "hold" is the **maximum** time. This period will vary in between depending on the last time it was banged and it can be half the time it was last activated if the the last time it was activated is less than twice the hold time...

I've been doing some work in the documentation and I was hoping to explain this behaviour but it seemed too complicated to explain simply and I'm not sure how to make it simple and straight forward, specially with this non intuitive/unproper names.

Anyway, I propose renaming to "minimum" and "maximum" flash time. It helps understanding and documenting. It saves us from explaining that these parameters are not what they seem to be...

porres:
By the way, I already have an update of the [bng] help file here https://github.com/pure-data/pure-data/commit/978606e16d404eb99e935bec3b424f97b5c3673c

I did not document the "intrrpt" and "hold" time yet cause I just couldn't understand it untill now...

porres:
By the way, I'm already updating [bng]'s help file as part of https://github.com/pure-data/pure-data/pull/829 - the commit https://github.com/pure-data/pure-data/commit/978606e16d404eb99e935bec3b424f97b5c3673c was mainly to document the possibility of setting colors as hexadecimals as proposed in https://github.com/pure-data/pure-data/pull/937#issuecomment-613522728

I did try to document the "intrrpt" and "hold" in https://github.com/pure-data/pure-data/pull/829/commits/2080df7d4896c408e543f07358b3345fe647b157

porres:
The PR https://github.com/pure-data/pure-data/pull/978 presents a solution to this.

If the other PR I'm working on that's updating the documentation of iemguis (see https://github.com/pure-data/pure-data/pull/829 ) I will also update it to rename the parameters. So far, this is only a minimal renaming change.

porres:
[...]

PR #978, opened 2020-05-05
[bng] - rename "intrrpt" and "hold" parameters in properties

porres:
This closes https://github.com/pure-data/pure-data/issues/976 and presents an alternative that seems to be a more proper.

"intrrpt" is renamed to "min" and "hold" is renamed to "max".


porres:
applied with https://github.com/pure-data/pure-data/commit/3c6ec04d9994d462c09cbbb108d1d23a62818434

Issue #983, opened 2020-05-06
request: make [symbol] convert float to symbol.


porres:
These days [float] can convert from symbol to float. I suggest we can make [symbol] easily convert a float to a symbol.

On a paralel note, other more complex conversions from other messages like lists and anythings with more than one item to symbol would also be good and desirable for vanilla, but you can still do it in vanilla with some work... see patch image below.

<img width="598" alt="Screen Shot 2020-05-05 at 23 28 51" src="https://user-images.githubusercontent.com/5360142/81133546-53771b80-8f28-11ea-877c-6765333b53ef.png">

Maybe I could add this example to the documentation

porres:
at least a "see also: makefilename" is welcome in the help file...

porres:
> see patch image below.

the patch uses [makefilename %g] to convert from float to symbol, which would not be required if [symbol] can convert floats (which makes the patch simpler). Anyway, I will upadtes now the documentation to include such an example.

porres:
additionally, we could also make it load a float as an argument, which now simply gives us an empty symbol

porres:
> additionally, we could also make it load a float as an argument, which now simply gives us an empty symbol

As it turns out, this is impossible right now. The PR https://github.com/pure-data/pure-data/pull/987 closes this issue and also https://github.com/pure-data/pure-data/issues/986

PR #984, opened 2020-05-06
make [symbol] convert floats

porres:
allows [symbol] to accept floats in the main left inlet and convert it to a symbol https://github.com/pure-data/pure-data/issues/983

closes https://github.com/pure-data/pure-data/issues/983


nicolasdanet:
Note that the danger is to make very easy for an user to flood the symbol table.

porres:
> Note that the danger is to make very easy for an user to flood the symbol table.

sorry, I don't know what you mean, can you ellaborate?

nicolasdanet:
Each generated symbol is contained in the symbols hash-table. There is a very huge number of floating-points. It is possible thus to add a huge number of entries in the table. For instance at the output of a [snapshot~] object. It could result to slow down the hash-table lookup (and so ALL Pure Data). Normally it doesn't really matter, but in that case perhaps.

porres:
I don't see nothing of that in the code of the symbol object in x_connective.c

where is this hash-table?

How is this not a problem if I'm using [makefilename %g]?

Do you have any suggestion for the code to prevent this? Have you checked it?

thanks

nicolasdanet:
Look at the gensym code < https://github.com/pure-data/pure-data/blob/master/src/m_class.c#L838 >.

It can also happen with [makefilename] (and probably in lots of other cases). But it is a theoretical problem that could occurs only after several millions of symbols (at least in Max/MSP, and IIRC). I guess that an user is not supposed to generate such number of file names. 

In that case i can not see how to avoid it properly.

porres:
> It can also happen with [makefilename] (and probably in lots of other cases)

then this is not specific to this PR, thus it is a parallel issue

> In that case i can not see how to avoid it properly.

it would probably require something completely unrelated to solve this parallel issue. I'd start by opening an issue just for that. And since this doesn't really affect this PR, can I ask you to delete your comments? (Again, if you want to discuss this, you can open a new issue).

thanks

porres:
friends, I need help with this one... https://github.com/pure-data/pure-data/pull/984/commits/b4a6a5fdfc983ab97ab0ba386bca87c1b5c744dd @umlaeute  / @Spacechild1 ?

thanks

porres:
Ok, regarding my attempt to load a float as an argument, this is the deal, just by switching to A_GIMME alone, what happens is that when I load the [symbol] object, it converts this to "list" and loads the [list] object instead. So I can do something like [symbol append 1 2 3] and it's the same as [list append 1 2 3]. Funny that [symbol 1 2 3] will give "list : unknown function".

Anyway, the reason why this happens tottaly baffles me... I have NO IDEA... so I need help, I can't see how I can prevent this.

Spacechild1:
This is one of edge cases with pd_objectmaker where the object name itself is a pd atom type. the solution is to manually assign the resulting instance to `pd_this->pd_newest`, see `pdfloat_new` in `x_connective.c` for an example an explanation. A similar thing happens in `list_new` in `x_list.c`.

porres:
> This is one of edge cases with pd_objectmaker where the object name itself is a pd atom type. The solution is to manually assign the resulting instance to pd_this->pd_newest. See pdfloat_new in x_connective.c for an example and explanation. A similar thing happens in list_new in x_list.c.

@Spacechild1 , the problem is that we already have "pd_this->pd_newest" in the pdsymbol_new function, and only by switching to A_GIMME alone (and keeping everything else as is) makes this happen.

-------------------------

For instance, this doesn't work and causes the problem, and I don't even get "new" printed.

static void *pdsymbol_new(t_pd *dummy, t_symbol *s)
{
    post("new");
    t_pdsymbol *x = (t_pdsymbol *)pd_new(pdsymbol_class);
    x->x_s = s;
    outlet_new(&x->x_obj, &s_symbol);
    symbolinlet_new(&x->x_obj, &x->x_s);
    pd_this->pd_newest = &x->x_obj.ob_pd;
    return (x);
}

void pdsymbol_setup(void)
{
    pdsymbol_class = class_new(gensym("symbol"), (t_newmethod)pdsymbol_new, 0,
        sizeof(t_pdsymbol), 0, A_GIMME, 0);
    class_addbang(pdsymbol_class, pdsymbol_bang);
    class_addfloat(pdsymbol_class, pdsymbol_float);
    class_addlist(pdsymbol_class, pdsymbol_list);
    class_addsymbol(pdsymbol_class, pdsymbol_symbol);
    class_addanything(pdsymbol_class, pdsymbol_anything);
}

Spacechild1:
The problem is that with an `A_GIMME` you effectively send a list to pd_objectmaker which will create a `[list]` object for you... for the gritty details see `pd_typedmess`. I think the solution for now is: don't do it.

porres:
> yeah, I was afraid this would require some surgery under the hood...

porres:
> don't do it

I reverted the last commit. So far, this converts floats to symbols and allows the conversion of lists as requested in https://github.com/pure-data/pure-data/issues/986

It's weird we cannot load the float argument, but I guess it is not "that" weird... I can leave a note in the documentation and we could live with it.

porres:
Ok, I now updated the help file and it seems this PR is ready to go, and closes both https://github.com/pure-data/pure-data/issues/983 and https://github.com/pure-data/pure-data/issues/986

There's no need anymore to mark this as "draft" anymore, I suppose.

porres:
Well, we had a reasonably big discussion here and I made more commits than I could have. I think I'll reboot this PR.

Issue #986, opened 2020-05-06
symbol has an unused "pdsymbol_list" method


porres:
this is unused

https://github.com/pure-data/pure-data/blob/master/src/x_connective.c#L168

it seems there was an intent to do all that, so I wonder if we should use it and allow "list a b c" to convert to "symbol a" and more...

umlaeute:
i'm not sure what you think the issue is here.

porres:
I think that [symbol] had the intention of converting lists to symbols ("list a b c" would give you "symbol a"), but this might have been accidentally left off. Currently, what happens is that we get "symbol list" instead.

If that was not an accident, then I guess I could suggest that we could actually do what this unused code does.

porres:
well, I changed the title and I think this now is best described as "feature request" rather than "subject:code-cleanup"

porres:
and please note that I'm using it in this new PR which also adds the ability to convert floats to symbols (and hence the first float of a list that starts with a float as well) - see https://github.com/porres/pure-data/commit/e476e7c0b5e0af9ebb60fb89e21afb7bcfce2b05

umlaeute:
first, i'm not entirely sure whether we want `[list a b c(` to be converted to `[symbol a(` (i know that this is what the comment is suggesting; but that doesn't automatically make it a good idea; otoh `[t s]` already does this conversion)

however: my point is that the issue you raise here is that there's an unused function in the source code, which in itself is not a problem (nor a feature).

if you want to request a feature to "use the 1st atom of a list as the symbol iff it is a symbol", then you should say so, and move the rambling about the precedent (but unused) function into the description.

porres:
changed title and description once again

Spacechild1:
> first, i'm not entirely sure whether we want [list a b c( to be converted to [symbol a(

I'm entirely sure we want this :-) Actually, we had a discussion about this in some other github issue, where you at least agreed with me that the current behavior is useless. Let's just hope nobody is relying on the current behavior. I mean, if someone wants the symbol `[symbol list(`, they should type `[symbol list(` and not rely on `[symbol]` to convert any list to `[list symbol(`...

I think the list method is the intended behavior and I suspect Miller just forgot to add to the class.

porres:
> I think the list method is the intended behavior and I suspect Miller just forgot to add to the class.

My first idea in raising the issue was to say this was intended but accidentally left out, from the comments in the code it's quite clear there's an idea to not just get "symbol list".

[...]

PR #987, opened 2020-05-06
allow [symbol] to convert float and lists

porres:
this is a reboot of https://github.com/pure-data/pure-data/pull/984

This closes https://github.com/pure-data/pure-data/issues/983 and https://github.com/pure-data/pure-data/issues/986

The idea is to allow [symbol] to convert floats to symbol because it'd be useful and because [float] can do the opposite conversion (hence, consistency purposes).

This PR also allows lists to be converted, which seems it was intentional but left out of the code accidentally (see https://github.com/pure-data/pure-data/pull/984 )

I also included in the help an example on how to convert full lists and anythings (all items) to a symbol.

### DIFF
--- a/doc/5.reference/symbol-help.pd
+++ b/doc/5.reference/symbol-help.pd
@@ -1,26 +1,100 @@
-#N canvas 531 220 512 412 12;
-#X msg 44 107 bang;
-#X text 134 244 creation argument initializes the value;
-#X text 271 372 updated for Pd version 0.45;
-#X text 89 104 output the value;
-#X text 153 130 set and output the value;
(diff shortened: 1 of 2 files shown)


porres:
The only thing I couldn't do is make [symbol] load a float as an argument. This is not possible yet (see https://github.com/pure-data/pure-data/pull/984#issuecomment-624896480 for reference), but it's not a deal breaker at all, and the documentation clearly specifies you need to load a symbol atom as an argument.

porres:
> I also included in the help an example on how to convert full lists and anythings (all items) to a symbol and also included "see also: makefilename"

will redo this if it gets accepted

PR #989, opened 2020-05-07
Draw array name when creating a garray

### LINKED ISSUE #979: array name doesn't show when first inserted via the put menu
you have to click and drag so it can appear, or close and reopen the patch window.

sebshader:
fixes #979


danomatika:
Looks like a good idea to me.

umlaeute (approved):
seems innocent enough

Issue #992, opened 2020-05-10
nbx: setting width via a message doesn't update the number box


porres:
When you change the width via properties, the number box gets updated so you display the right amount of digits according to the box's width. This doesn't happen via messages, causing things like this.

<img width="250" alt="Screen Shot 2020-05-10 at 12 52 40" src="https://user-images.githubusercontent.com/5360142/81503982-26769000-92bd-11ea-9923-2da2b13205cf.png">

Issue #633, opened 2019-05-12
nbx clips incoming values


porres:
this is inconsistent behaviour to sliders, radios and atom boxes.

<img width="340" alt="Screen Shot 2019-05-12 at 03 55 19" src="https://user-images.githubusercontent.com/5360142/57578856-f5753c80-7469-11e9-9cf8-d0bc8aefa2b7.png">

An atom number box will only clip values if it has a min/max range and if you're clicking and dragging. If you type any other number into it or if it receives values outside that range, then it doesn't clip.

For consistency, I think [nbx] should behave the same way

This issue is a duplicate of https://sourceforge.net/p/pure-data/bugs/1224/ - created by Dan. 

Moreover, the behavior for many IEM GUIs changed in Pd 0.46, before that, all objects did clip. This has changed, but not for [nbx] - it seems it got accidentally left out.

I see no way out other than also changing this and creating a compatibility flag for this one too, for the sake of consistency.

porres:
This issue is a duplicate of https://sourceforge.net/p/pure-data/bugs/1224/

porres:
https://github.com/pure-data/pure-data/pull/631 fixes it

porres:
The behavior for many IEM GUIs changed in Pd 0.46, before that, all objects did clip. This has changed, but not for [nbx]. 

I see no way out other than also changing this and creating a compatibility flag for this one too, for the sake of consistency.

millerpuckette:
I haven't touched this since I want to just freeze nbx and later drop it from the menu (and just keep it around to keep old patches running).

porres:
this has been fixed in master

PR #999, opened 2020-05-14
[nbx] fixes

### LINKED ISSUE #629: no "inf" and "nan" for [nbx]
As part of the iemgui updates that we were talking about here, I'd like to point out these issues. First is that [nbx] will not print "inf". I actually know of a hacky workaround way for it to be able to print 'inf', but it's too ugly. Anyway, check below.

Screen Shot 2019-05-11 at 17 49 19

Unlike the atom box, the code of numbox creates min/max values here:

https://github.com/pure-data/pure-data/blob/master/src/g_numbox.c#L777

with the atom box, by default, you get a range from 0 to 0 and this means it's unrestricted. So it can go up to inf or -inf... but with nbx, a range of 0-0 clips it to this single value of 0... and since it does not understand if you type "-inf" or "inf" as an actual infinite float, we're pretty much tied up in setting this via the interface. My suggestion is to make it simply behave like the atom box and fix this inconsistency.

--------------

Now, regarding nans, this is an old issue I opened here, that I closed and now I'm reopening as part of this new issue.

In Pd 0.46, there was some bug introduced where NaNs were not displayed either in the atom box or [nbx], this was fixed for atom boxes in 0.47, but not [nbx]. This was probably introduced here, where the comparison would filter nans: https://github.com/pure-data/pure-data/commit/cd737749b51e03ba1bdc6d8b66b890d89823d158

I just tried a quite old version of Pd (0.43) and we can see here how this used to work fine

Screen Shot 2019-05-11 at 18 15 04

This was also brought up in the Pd dev list https://lists.puredata.info/pipermail/pd-dev/2017-07/021185.html , where it mentioned a simple revert didn't seem to work, so something else must be going on...

both inf and nan display were really important for an object like [cyclone/number~]

porres:
- The first 2 commits (https://github.com/pure-data/pure-data/pull/999/commits/5bf0995914d9190efe073e09454c5a96cf50204f & https://github.com/pure-data/pure-data/pull/999/commits/c57c41ac7286af6331b891d4c8133984a043db32) closes #629

for the nan, I used a 'isnan' check - gatom does that too. What I didn't do is have a check if the last value was also a nan, but maybe Miller can do that in the way it's considered best? I wasn't sure.

refarding inf/-inf, I also got inspired on how gatom handles this as well.

- third commit (https://github.com/pure-data/pure-data/pull/999/commits/2c32db9327f90edeb7eb66d450ba51426da6618e) closes https://github.com/pure-data/pure-data/issues/633

- fourth commit (https://github.com/pure-data/pure-data/pull/999/commits/25fae17c31c2b76d0e1d07aa2bed9263b4632233) closes https://github.com/pure-data/pure-data/issues/992

---------------

If this PR gets merged, I'll update the documentation as part of https://github.com/pure-data/pure-data/pull/829

### DIFF
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -648,6 +648,7 @@ static void my_numbox_size(t_my_numbox *x, t_symbol *s, int ac, t_atom *av)
     }
     my_numbox_calc_fontwidth(x);
     iemgui_size((void *)x, &x->x_gui);
+    (*x->x_gui.x_draw)(x, x->x_gui.x_glist, IEM_GUI_DRAW_MODE_UPDATE);
 }
 
 static void my_numbox_delta(t_my_numbox *x, t_symbol *s, int ac, t_atom *av)
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -14,12 +14,6 @@
 #include "g_all_guis.h"
 #include <math.h>
 
-#ifdef _WIN32
-#include <io.h>
-#else
-#include <unistd.h>
-#endif
-
 #ifdef _MSC_VER
 #include <float.h>
 #define isnan _isnan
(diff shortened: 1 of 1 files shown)


porres:
closed this as I have a new PR for these and other fixes --> https://github.com/pure-data/pure-data/pull/1645

PR #1006, opened 2020-05-15
add arguments and flags to bob~

### LINKED ISSUE #398: suggestion, add arguments to [bob~]
I wish bob could take freq and q arguments (like [bp~]) as well as flags for oversampling and saturation

porres:
this closes  #398

documentation needs to be updated afterwards in my docs branch


porres:
tried to solve a conflict by just replacing the help file, so no changes to the code were made, and now I get this build error that I don't understand...

Ant1r:
the error means that `error(...)` must be replaced by something like `pd_error(x, ...)`.

porres:
> the error means that error(...) must be replaced by something like pd_error(x, ...).

oh, thanks, on it :)

PR #1037, opened 2020-05-26
Documentation issues

porres:
Improvements, fixes and updates to the documentation to document existing features


millerpuckette:
Shouldn't the VU meter go all the way to the top when the signal clips?  That's why I use -88 and not -100 - I think of the top 12dB as headroom before the signal gets trashed.  (Otherwise, why show that range of values at all?)  But I might not be understanding how standard VU metering is done.

Except for my doubts about the VU metering I think this is all on target.

millerpuckette:
(also, I intended for '50' on the gain control to map to unit gain, so that we can have some headroom.  This should somehow be made clearer in teh visual design of the out1~ object I guess.)

porres:
> Shouldn't the VU meter go all the way to the top when the signal clips? 

Nope, [vu] comes with a scale in dBFS, where 0 dB is the maximum (clipping) value, and what corresponds to 100 dB coming out of env~. So anything above gets red and tells you your signal is "dangerous" - that is, it'll clip.

Depending on what your doing in your patch, you can generate signals at levels higher than desired, so this is why [vu] has this danger/red zone above 0 dbFS.

For example, when you have a [noise~] object into it, you can see a peak level detection at about 0 dB, but if you maybe unintentionally have 2 white noise sources coming in, then it goes higher than 0 dB 

Screen Shot 2020-06-03 at 01 56 50

millerpuckette:
This is a choice imposed by the vu~ object, not by the meter, which, itself, merely displays numbers from -100 to +12 with anything above 0 showing some red.  These numbers are intended as dB but aren't attached to any particular scaling (such as 0dB <--> 1 (peak amplitude), or 12dB <-->1, or
whatever.  My experience with VU meters in real situations is that, in analog days, red implies the possiblity of soft distortion but the hard clip point is at the top or even above the top of the scale.  I don't know what common practice is for digital ones, but I'm pretty sure that in at least some DAWs, red doesn't mean "the signal got clipped" but that "you're in danger of clipping".

Spacechild1:
Hi Miller, I think Alex is right about VUs in the digital domain. In a DAW, the VU meter would typically show anything above 0dBFS as red, indicating that the output would be clipped.

porres:
Yeah, don't worry, I'm totally sure they are intended as dBFS, and the headroom is just so you know you're doing something wrong.

Cool, I just have to do another update, as I was using [out1~] in the help file of [vu] as well :)

porres:
> I was using [out1~] in the help file of [vu] as well :)

help file of slop~ was also still trying to load out1~ but giving an error instead.

this new PR fixes it https://github.com/pure-data/pure-data/pull/1051

PR #1038, opened 2020-05-27
fix reading NULL pref strings on some macOS versions

danomatika:
This is a small bugfix PR which fixes some preference strings being read as NULL preference on macOS by using CFStringGetCString instead of CFStringGetCStringPtr.

The [Apple docs](https://developer.apple.com/documentation/corefoundation/1542133-cfstringgetcstringptr?language=objc) mention that CFStringGetCStringPtr may return NULL even if the string value is valid. It then mentions to use CFStringGetCString as a backup. In this case, we need to have our own copy so it makes sense to simply use CFStringGetCString anyway.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -9,8 +9,8 @@ extern "C" {
 #endif
 
 #define PD_MAJOR_VERSION 0
-#define PD_MINOR_VERSION 50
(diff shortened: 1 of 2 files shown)


danomatika:
@umlaeute Probably.

PR #1039, opened 2020-05-27
Resize several objects together (bis)

pierreguillot:
This PR is an update of https://github.com/pure-data/pure-data/pull/99 that has not really been merged (see the comments). 
Differences from the previous PR:
- Add multi-object support for undo_apply (1st commit)
- Rewrite the commit history (2nd and 3rd commit)


umlaeute:
hmm, i wonder: the undo chain can use reply a sequence of undo-actions (enclosed by `SEQUENCE_START` and `SEQUENCE_END`) as a single step.
this is meant to build complex undos into a single undoable action.

why can't you use that?

pierreguillot:
I'm not familiar with the undo-redo functions. Do you suggest to do something like this?
```
SEQUENCE_START
for (sel = x->gl_editor->e_selection; sel; sel = sel->sel_next)
{
   y = sel->sel_what;
   canvas_undo_add(x, UNDO_APPLY, "resize", canvas_undo_set_apply(x, glist_getindex(x, y)));
}
SEQUENCE_END
```
Or do you think of something else?

I changed the canvas_undo_set_apply() because the function deselects all the objects so the resizing of multiple objects doesn't work anymore.  More the function contains this comment:

>       /* deselect all objects (if we are editing one while multiple are
>       * selected, upon undoing this will recreate other selected objects,
>       * effectively resulting in unwanted duplicates)
>       * LATER: consider allowing concurrent editing of multiple objects
>       */

umlaeute:
> Do you suggest to do something like this?

yes, something like this.
typically the invocation is:
~~~C
canvas_undo_add(x, UNDO_SEQUENCE_START, "resize objects", 0);
/* here come multiple undo-steps that will be called as a single-sequence */
// ...
canvas_undo_add(x, UNDO_SEQUENCE_END, "resize objects", 0);
~~~


> I changed the canvas_undo_set_apply() because the function deselects all the objects so the resizing of multiple objects doesn't work anymore

this is of course an issue that needs to be solved.

however, there are actually two issues, and they should be solved separately:
1. undo and selections
it's a general problem of undo that the selection goes missing whenever you undo/redo something.
i'm not totally sure how to best solve this yet.
*and* i haven't looked at your proposed solution, so i cannot say whether its the/a "correct" solution.

2. your code depending on the selection
the quick fix for your problem is to create a copy of the list of selected object before it gets deleted.

pierreguillot:
> and i haven't looked at your proposed solution, so i cannot say whether its the/a "correct" solution.

I think that canvas_undo_set_apply() is only used when resizing objects, so my solution only works for this situation. I didn't mean to fix a global "undo & selection" problem".

> the quick fix for your problem is to create a copy of the list of selected object before it gets deleted.

Ok, I will try that and rebase the branch

pierreguillot:
I updated the PR based on your suggestion. The Travis CI failed but it seems to be unrelated to these changes (the problem is related to brew on the CI). Let me know if it's okay or if I should change or imrpove something.

pierreguillot:
I rebased the branch on the master branch, as the CI has been fixed on it, everything should be ok now.

PR #1047, opened 2020-05-29
Use pdfontloader for wish86 on Windows 64bit builds.

Lucarda:
This PR includes a working `pdfontloader.dll` compiled for wish86 and Windows 64bit.
I added some lines to `build-msw-64.sh` so that it gets extracted from the zip file to the `bin` directory.

Untested because I'm not sure how to use `build-msw-64.sh` (my linux machines don't have cross-compiler yet).


millerpuckette:
seems to compile and build the pd directory fine with the new fontloader.  merging....

Lucarda:
@millerpuckette 

cool :)

there is one other thing regarding "fonts": On all your win64bit builds the file `DejaVuSansMono-Bold.ttf` is missing on the distribution/font folder. The file is in the repository so I think is just not present on your build machine.

The file is really needed because if the file is not there TK forces the not-bold font to display as bold and it gets horrible rendered. 

I'll be checking for this in the next test release.

PR #1048, opened 2020-05-29
explicitly sign bundled frameworks when creating mac app

danomatika:
This is a *possible* fix for code signing issues when building the macOS .app bundle.

Following this [Stack Overflow post](https://stackoverflow.com/questions/7697508/how-do-you-codesign-framework-bundles-for-the-mac-app-store), apparently bundled frameworks must be explicitly signed as the --deep flag will not actually handle this on it's own. Go figure.

Please test.


millerpuckette:
Gotta get ready to teach - I'll follow up on this as soon as I can.

PR #1049, opened 2020-06-01
Add option for native tk bezier cables

sebshader:
I used -smooth option for canvas create line, and am doing some math in 2 new tcl procs for creating the lines and setting the coordinates, and there's a new global variable ::curve_cords for turning it off and on (off by default)

This is adding more computation in tcl.. though not much more if straight cables are opted for I think.

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -13,6 +13,7 @@ to be different but are now unified except for some fossilized names.) */
 #include "s_utf8.h"
 #include "g_canvas.h"
 #include <string.h>
+#include <math.h>
 #include "g_all_guis.h"
 #include "g_undo.h"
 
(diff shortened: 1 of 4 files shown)


Spacechild1:
Can you also make them wiggle, so we are compatible with recent Max? :-p (sorry, couldn't resist)

sebshader:
> Can you also make them wiggle, so we are compatible with recent Max? :-p (sorry, couldn't resist)

i'll definitely leave that part for the tcl plugins to do ;-)

porres:
> they don't look that great anyways, kind of "aliased" maybe, also I wasn't sure how high to go with the -splinesteps option & just set it to 16 (edit: 36)

No, they don't look good :/ but even if they did, this is something I'm not really fond of anyway (but I'm not the one making the decisions).

But I don't get if this should be curved out of the box for Pd, and I strongly oppose to that. Maybe some setting for that... or this should maybe just open the doors for people to write plugins for such things, as apparently some people like it for some reason. Ok, reason must be that it makes it "easier" avoid messy patches with cords crossing objects and stuff, improving readability, but I think that's just not true :(

sebshader:
@porres I don't personally use them either, I just saw that there are native tk splines and thought "why not?". Maybe the people that like pd-l2ork style cables would like these more..

I do think that it would be cool to have gui hooks to be able to modify how the cords are drawn with a tcl plugin, one of the issues with that is that the logic for when a cable is being hovered over is all done on the c side (in pd) tho.

porres:
I can't trust people who (worry about them) likes curvy lines so much that it's the selling point to not use Pd Vanilla... :) but allowing plugins is fine, and maybe segmenting cords are not always that bad of an idea/feature

danomatika:
I agree with moving the implementation-specific stuff to Tcl. I would personally just move the straight line stuff to the TCL side, then write a plugin which implements the splines as an override. The beliers may look better but they may be slow on some systems, especially embedded. To me, one of Pd's strengths is being able to run it almost everywhere...

PR #1051, opened 2020-06-03
replace out1~ with subpatches

porres:
out1~ has been removed, but it was only removed from the peak-meter example. It was still used in the help file of slop~ (for the compander-limiter example) as well as the help file of vu.

This change just includes a subpatch so people can listen to the examples and fixes the "can't load out1~" error.


millerpuckette:
Hmm, I thought I had cleaned out all those references... will look again.

porres:
> will look again.

I looked deeply, it was only still used in slop~ and vu help files

Issue #1058, opened 2020-06-08
(feature) add argument flag to pass args to Pd from terminal


mariobuoninfante:
Hi,

I was thinking about having a flag that allows to pass arguments to Pd via terminal.
On the mailing list the following flag was suggested:

> pd mypatch.pd -a 23

and then in _mypatch.pd_ have [$1] equal to 23

Spacechild1:
Note that we can open more than one patch, which could look like this:

`pd patch1.pd -a "1 2 3" patch2.pd -a "foo bar"`

danomatika:
How about "-args" to better match Pd's longer names?

Spacechild1:
sounds good

danomatika:
OTOH if we follow your example, the args are supplied as a single message. In this case "-args" might be somewhat misleading, aka implying they don't need to be quoted.

mariobuoninfante:
I mean _args_ is also what we use for [pdcontrol] so I'd say let's stick with that and just make it clear that they must be quoted.

danomatika:
Yeah, it could still work with the appropriate context and could be explained as "sending an args *messge*" then. I just point out that this would work slightly differently than similar flags in other programs, ie. the macOS `open` command takes an "--args" flag which works similarly to "--": pass everything after as arguments to the invoked program.

mariobuoninfante:
yap, I think as long as it's in the command line help (pd -h), I think it's fine.
People will be looking at that when doing this things

Spacechild1:
In the meantime, here's another proposal I made on pd-dev:

`pd "patch1.pd: 1 2 3" "patch2.pd: foo bar"`

or

`pd -open "patch1.pd: 1 2 3" -open "patch2.pd: foo bar"`

I kind of like this because we don't have to change the argument parsing code. Instead we just have to add a little bit of logic to `openit` in `s_main.c`.

Note that colons are reserved/forbidden characters on Windows and macOS, but not on Linux. This is not a problem for Pd patch names (which shouldn't contain colons for portability), but Linux users might have colons in their directories.

Personally, I would prefer semicolons over colons because they are already used by Pd as a message delimiter and must not appear in creation arguments (unless they are escaped by a backslash). Semicolons are allowed for filenames in every major OS, but they are pretty much discouraged (they can easily lead to arbitrary command execution on the shell).

[...]

PR #1064, opened 2020-06-10
add "-open-with-args" command line option

Spacechild1:
This allows to open patches with creation arguments.

Syntax: `-open-with-args <patch> <arguments>`

Multiple arguments have to be enclosed in quotation marks.

Example:

`pd -open-with-args foo.pd "bar baz 1 2 3"`

Simple test patch: [foo.zip](https://github.com/pure-data/pure-data/files/4758116/foo.zip)

Closes https://github.com/pure-data/pure-data/issues/1058

---

Of course, you can also open multiple patches:

`pd -open-with-args patch1.pd "1 2 3" -open-with-args patch2.pd "foo bar baz"`

---

Or you can load several patches with the same arguments:

`pd -open-with-args "patch1.pd:patch2.pd" "1 2 3"`

NOTE: the colon seperator is *not* a new feature. (On Windows you have to use a semicolon, instead.)

### DIFF
--- a/src/s_main.c
+++ b/src/s_main.c
@@ -60,7 +60,15 @@ t_symbol *sys_flags;    /* more command-line flags */
 
 const char *sys_guicmd;
 t_symbol *sys_libdir;
-static t_namelist *sys_openlist;
+
+typedef struct _patchlist
+{
+    struct _patchlist *pl_next;
+    char *pl_file;
+    char *pl_args;
+} t_patchlist;
+
+static t_patchlist *sys_openlist;
 static t_namelist *sys_messagelist;
 static int sys_version;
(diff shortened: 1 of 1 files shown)


Spacechild1:
> do you mean multiple arguments?

yes, I was too mention that you obviously don't need it for a single argument.

> also i think i prefer -open-with-args

I don't really care about the name of the flag. I tried to keep it short, but I can totally change it to "-open-with-args" or "open-with-creation-arguments" :-)

umlaeute (approved):

Spacechild1:
@millerpuckette Any plans to include this? It's a small, but IMO very useful feature.

Issue #1065, opened 2020-06-10
sys_doflags() doesn't respect quotation marks


Spacechild1:
`sys_doflags` parses the startup flags saved in the Pd settings. Unfortunately, it doesn't respect quotation marks, which means that something like the following will fail:

`-font-face "Courier New"`

Spacechild1:
yes

PR #1066, opened 2020-06-10
fix flag parsing in sys_doflags

Spacechild1:
Currently, the parser in `sys_doflags` simply splits tokens based on whitespace and doesn't respect quotation marks. This means we can't store command line flags like the following:

`-font-face "Courier New"`

Solution:

* use `CommandLineToArgvw` on Windows
* use `wordexp` for Unix

Tested on Windows, Linux and macOS. Fixes https://github.com/pure-data/pure-data/issues/1065


Spacechild1:
@umlaeute I think this PR has the simpler argument parsing code (and doesn't reinvent the wheel). Your PR has better Tcl escaping, though. Should I port the latter to this PR - or vice versa?

Spacechild1:
closing in favor of #741

PR #1070, opened 2020-06-12
use standard Win32/Posix function to parse command line string

Spacechild1:
* Win32: `CommandLineToArgvW`
* Unix: `wordexp`

### DIFF
--- a/src/makefile.msvc
+++ b/src/makefile.msvc
@@ -57,6 +57,7 @@ PDLIB = /NODEFAULTLIB:libcmt /NODEFAULTLIB:libcpmt /NODEFAULTLIB:oldnames \
     /NODEFAULTLIB:libc /NODEFAULTLIB:uuid /NODEFAULTLIB:ole32 \
     $(EXTRA_LIBPATH) \
     kernel32.lib \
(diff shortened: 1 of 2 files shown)


danomatika:
Good idea. Are there platforms where we would still need the old implementation? Maybe it's not bad to keep it around.

Spacechild1:
I did some more research and it turns that `wordexp` is deprecated on iOS (https://github.com/lagadic/visp/issues/37) and not implemented on OpenBSD (https://github.com/SpiderLabs/ModSecurity/issues/1841).

My intention was to avoid a handwritten argument parser in favor of robuts library functions, but it seems we have to keep the old implementation around anyway... I'm tempting to close this...

danomatika:
Yeah, it’s actually a relatively non standard thing, even though it’s a part of Posix. I use a bare C++ parser library in many programs which mimics the GNU style but is cross platform.

For iOS, it might be easier to drop building the arg processing, however I can see a use where people could theoretically pass them manually somehow.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

PR #1071, opened 2020-06-16
Feature: Adds a way to load pd files from a buffer

yamashi:
This PR contains the basics to load a .pd file from a text buffer instead of going through the file system.
### Motivation
This is the ground work to better use the library in environments that do not provide a standard file system such as game engines.
### Limitations
The current version will only work with pd files that do not have any dependencies as the library will still attempt to load them from disk. I believe fixing this goes deeper than just a front end function and am not sure how to go about doing this.


Spacechild1:
> This is the ground work to better use the library in environments that do not provide a standard file system such as game engines.

As I've written elsewhere, the real foundation for non-standard file system support would be hooks to replace all relevant file system functions, like `open`, `close`, `read`, etc. Then you don't even need to load abstractions from text buffers in the first place. I want to work on this.

I think the PR at hand is useful for programmatically creating Pd patches in C. I've written such functions myself in the past, and it's not 100% straightforward, so I think it's nice to have this in the public API.

src/g_canvas.c
```diff
@@ -2045,19 +2045,19 @@ void pd_doloadbang(void);
 
     /* evaluate a file, which is expected to create a patch, and perform
     post-evaluation cleanup and loadbang */
-t_pd *glob_evalfile(t_pd *ignore, t_symbol *name, t_symbol *dir)
+t_pd* glob_evalfile(t_pd* ignore, t_symbol* name, t_symbol* dir)
```
umlaeute: no.
please keep indentation, spacing,... as they are and adjust your coding style to that of Pd.
if you found a coding-style inconsistency, you might want to fix it - but in a *separate* commit.

src/g_canvas.c
```diff
@@ -2071,6 +2071,32 @@ t_pd *glob_evalfile(t_pd *ignore, t_symbol *name, t_symbol *dir)
     return x;
 }
 
+t_pd* glob_evaltext(const char* data, size_t length, t_symbol* name, t_symbol* dir)
```
umlaeute: that looks wrong.
i think the proper interface would be to pass a *binbuf* to be evaluated, not a raw text.
Pd already provides functions to parse a raw text into a binbuf.

src/m_imp.h
```diff
@@ -93,7 +93,8 @@ void pd_globalunlock(void);
 /* misc */
 #define SYMTABHASHSIZE 1024
 
-EXTERN t_pd *glob_evalfile(t_pd *ignore, t_symbol *name, t_symbol *dir);
+EXTERN t_pd* glob_evalfile(t_pd* ignore, t_symbol* name, t_symbol* dir);
+EXTERN t_pd* glob_evaltext(const char* data, size_t length, t_symbol* name, t_symbol* dir);
```
umlaeute, line 98: why are those functions declared both in `m_imp.h` and `m_pd.h`?

src/x_vexp_fun.c
```diff
@@ -537,14 +537,6 @@ ex_toint(t_expr *e, long int argc, struct ex_ex *argv, struct ex_ex *optr)
...
         }
 
-#ifdef _MSC_VER
-/* rint is now advertised as part of the microsoft SDK but my MSVC still
-doesn't find it - so here it is again. */
-static double rint(double x)
-{
-        return (floor(x + 0.5));
-}
-#endif
```
umlaeute: that's a totally unrelated and just wrong.

src/g_canvas.c
Spacechild1: Pd also provides a function to load a file into a binbuf, so why do we need `glob_evalfile`? The idea is to have a nice high level function to load a patch from a given text buffer. The signature of `glob_evaltext` is certainly not "wrong". I can see how `glob_evalbinbuf` could be more flexible, so why not have both? The former could simply delegate to the latter.

src/m_imp.h
Spacechild1: yes, they should only be exported in a single header. do you think it's ok to put/move them to `m_pd.h`?

src/g_canvas.c
Spacechild1: on the other hand, we could have `glob_evalbinbuf` in Pd and `libpd_opentext` in libpd (complementing `libpd_openfile`).

umlaeute: > The signature of `glob_evaltext` is certainly not "wrong"

sorry; i didn't want to imply that the signature looks technically wrong.

a `glob_evaltext` (or `libpd_opentext`) funtion would look like:

~~~
t_binbuf*b=binbuf_new();
binbuf_text(b, data, length);
glob_evalbinbuf(b, name, dir);
binbuf_free(b);
~~~

i don't see much merit in such a simple wrapper function consisting of four (4) calls to public functions.

src/m_imp.h
umlaeute: i thought having providing a public API was the entire point of the exercise.
since `m_imp.h` is by definition the *private* API, `m_pd.h` looks like the correct place.

src/g_canvas.c
Spacechild1: yes, the wrapper function could go to libpd.

src/m_imp.h
Spacechild1: just asking because libpd uses `m_imp.h`, so it doesn't *need* to be in `m_pd.h`, but for external authors it's of course nicer to have it in the public header.

BTW, the signature of `glob_evalfile` is really ugly, but we can't change it because it's exported, I guess...

umlaeute:
there's a concurrent implementation in #1096

mcclure:
You could imagine **both** this implementation and #1096 being useful under different circumstances. You could even imagine using the two together (IE load a puredata file from a buffer and let it load sound assets from a VFS, though I'm not sure I know why this would be optimal)

Spacechild1:
Loading a Pd patch from a string buffer can be handy for dynamically creating patches in an external. I'm actually doing this in a project and it requires intimate knowledge of the Pd code base to pull it off correctly. Wrapping the whole process in a nice function would make it much easier.

eeropic:
Interesting! I've made a hacky copy/paste   selected items as text to clipboard (tcl gui plugin) as like in Max, for easy patch sharing in forums etc, this would help a lot. 

And like @Spacechild1 said, opens possibilities for dynamic patching

millerpuckette:
I independently tried something that's even simpler... messages to Pd to "start" and "stop" interpreting messages to #N etc as patch-creation messages.  YOu add one line of prolog and one of epilog, then just paste the patch into a text object or a mssage box (or send it over a serial line as I'm doing ATM :)

Issue #1072, opened 2020-06-21
Building with --disable-alsa fails because ALSA libraries are still used


simonvanderveldt:
When building with `--disable-alsa` the build fails when ALSA isn't installed because it still uses ALSA
```
./configure --prefix=/usr --build=x86_64-pc-linux-gnu --host=x86_64-pc-linux-gnu --mandir=/usr/share/man --infodir=/usr/share/info --datadir=/usr/share --sysconfdir=/etc --localstatedir=/var/lib --disable-dependency-tracking --disable-silent-rules --docdir=/usr/share/doc/pure-data-0.50_p2 --htmldir=/usr/share/doc/pure-data-0.50_p2/html --with-sysroot=/ --libdir=/usr/lib64 --disable-portaudio --disable-portmidi --disable-alsa --enable-jack --disable-fftw --disable-oss
```
```
    pd 0.50.2 is now configured

    Platform:             Linux
    Debug build:          no
    Universal build:      no
    Localizations:        yes
    Source directory:     .
    Installation prefix:  /usr

    Compiler:             x86_64-pc-linux-gnu-gcc
    CFLAGS:               -O2 -pipe -ffast-math -funroll-loops -fomit-frame-pointer -O3
    LDFLAGS:              -Wl,-O1 -Wl,--as-needed
    INCLUDES:             
    LIBS:                 -lpthread -ldl  -lrt

    External extension:   pd_linux
    External CFLAGS:      -fPIC
    External LDFLAGS:     -Wl,--export-dynamic -fPIC

    fftw:                 no
    wish(tcl/tk):         wish
    audio APIs:           JACK 
    midi APIs:            NONE!
```
```
[...]

PR #1073, opened 2020-06-21
Disable local-portaudio if portaudio is disabled in general

umlaeute:
right now, the `--disable-portaudio` flag disables PortAudio support, but still compiles the embedded copy of the PortAudio library (it just skips the linking step).

this is suboptimal, as it
- requires more ressources to compile
- is error prone (cf #1072 )

this PR completely disables PortAudio if the user passes `--disable-portaudio`.
the same goes for PortMidi.

Closes: #1072 

this PR also replaces the few remaining uses of bash's `if/else` and `switch/case` constructs to the autotools' `AS_IF` resp. `AS_CASE` macros.


danomatika:
Hmm yeah, it shouldn't build if its disabled. I wonder how I missed that those years ago. :P

umlaeute:
> EDIT: The same may be true for portmidi.

the full commit message of d0faa33 reads:
>  disable local_portaudio, if portaudio is disabled in general (same for portmidi)

:wink:

Issue #1053, opened 2020-06-04
GUIs drawing out of bounds in GOP


porres:
Some GUI objects, like iemguis might draw in the parent patch even if outside the graphed range in a GOP, which is bad.

For instance, a radio button will draw on parent if you change its number of cells via a the number message.

See below.

<img width="254" alt="Screen Shot 2020-06-04 at 13 04 22" src="https://user-images.githubusercontent.com/5360142/83780796-05a71d80-a664-11ea-9709-0c734d8e3e38.png">

check the patch: 
[test-GOP.pd.zip](https://github.com/pure-data/pure-data/files/4731328/test-GOP.pd.zip)

Above, the radio button is not actually on the parent, but the subpatch - so you can't click on it or anything.

When setting sendreceive names to empty in iemguis, inlets/outlets will also get drawn.

A similar issue is if you have a GOP inside a GOP that is also outside the graphed range and you change its size via messages.

porres:
I've fixed this for iemgui radios, but there should be more similar fixes to be done for other objects, I'll look through

HenriAugusto:
Somewhat related: issue #752 and PR #627

porres:
> Somewhat related¹: issue #752 and PR #627

Cool,  didn't know #627 would also take care of that... :) 

that looks like a much better way indeed to deal with the drawing issue of 'gop' abstractions nested in other 'gop' but outta bounds.

I can gladly revert the last commit of my PR ( https://github.com/pure-data/pure-data/pull/1079/commits/d6d2218853a7a5f60201bc68f3ec8ad7549f3651 ) in favor of that one!

porres:
fixed in master

PR #1079, opened 2020-06-28
prevent iemguis from drawing when they're out of bounds in a 'GOP'

porres:
closes https://github.com/pure-data/pure-data/issues/1053

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -393,7 +393,7 @@ void iemgui_send(void *x, t_iemgui *iemgui, t_symbol *s)
     iemgui->x_snd = snd = canvas_realizedollar(iemgui->x_glist, snd);
     iemgui->x_fsf.x_snd_able = sndable;
     iemgui_verify_snd_ne_rcv(iemgui);
-    if(glist_isvisible(iemgui->x_glist))
+    if(glist_isvisible(iemgui->x_glist) && gobj_shouldvis((t_gobj *)x, iemgui->x_glist))
         (*iemgui->x_draw)(x, iemgui->x_glist, IEM_GUI_DRAW_MODE_IO + oldsndrcvable);
 }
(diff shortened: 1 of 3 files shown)


porres:
reverted my last commit in favor of https://github.com/pure-data/pure-data/pull/627 - which feels like a much better way to handle this

umlaeute:
thanks for reverting. i force-pushed the older-revision (so now this PR contains only the wanted commits, and lacks the obsolete commit and it's reversion - which i think is much cleaner).
make sure to update your local branch to the new HEAD, before committing again (lest d6d2218853a7a5f60201bc68f3ec8ad7549f3651 and e1410a327766efb49ab08f8fbd9362bdd5a4b149 creep in again)

porres:
> thanks for reverting. i force-pushed the older-revision (so now this PR contains only the wanted commits

Ok, I wondered if we should do this, because I wanted to keep record that I also wanted to fix that in a different way. But it's alright if we keep the comments.

> make sure to update your local branch to the new HEAD, before committing again

Ok, sure, but this is quite certainly done ;)

umlaeute:
afaict there's still some inconsistencies, e.g. when moving an iemgui out of the GOP-rectangle with `delta` or `pos` messages (which will keep drawing the iemgui even though it should not).

otoh, it might sometimes be useful to be able to spill over the GOP boundaries...

porres:
> otoh, it might sometimes be useful to be able to spill over the GOP boundaries...

like when?

umlaeute:
e.g. creating an abstraction that is the source of an in-canvas particle system (creative coding anybody?)

or more practical: a browser-style alert-box that "blanks" (hides) the surrounding patch while active.

Spacechild1:
someone's bug is always someone else's feature ;-)

PR #1096, opened 2020-07-27
Convert file operations to use "fileops" abstraction layer

mcclure:
This is a largeish feature patch. The basic idea is all OS/standard file operations (`open`, `fopen` etc) have been removed and replaced with calls to function pointers in a global structure named `sys_fileops`.

### Motivation

I am interested in embedding libpd into a video game. In video game world, it is common to use a "**virtual file system**". A popular example of a VFS is [physfs](https://icculus.org/physfs/). Physfs allows you to specify one or more assets sources such as a directory or a .zip file. Physfs then presents a fake filesystem constructed from your assets sources and exposes I/O functions for working with this fake filesystem. Reasons you might want to use a VFS library like Physfs include: Security (Physfs ensures your program doesn't read or write outside the assets directory even by accident); the convenience of storing your assets in a .zip file or appending them to your exe; the possibility of combining several assets folders into one (I've used this before to implement mod support); Android friendliness (APK files are .zip files— I see there is an Android build of libpd, but I'm confused how exactly it loads files since your Android assets are going to be in the APK, not at a file path); and support for exotic game systems that might not have any filesystem at all.

The game I personally want to embed libpd in targets Android, and uses a [small VFS library](lovr.org) which is modeled on physfs. Unless I somehow teach libpd to talk to our VFS library I won't be able to use puredata, because pdlib normally loads everything by path and the .pd files and sound assets I want to load are embedded in the Android APK.

I had a conversation on the [mailing list](https://lists.puredata.info/pipermail/pd-dev/2020-05/022388.html) about this, I was given some hints as to how to load a .pd file without touching the filesystem, but they involved calling undocumented garray functions that I don't think are currently directly exposed in the libpd interface. Dan Wilcox suggested that if I had a proposed API or PR for VFS support I should try this github.

### API

This patch does not implement a Virtual File System. Instead, it creates a set of simple wrappers that allow you to "bring your own VFS", and which default to using the normal filesystem.

The patch consists of a new `s_fileops.c` and `s_fileops.h`. s_fileops.h defines these types:

* `t_fileops`: This is a structure containing a series of function pointers named `open`, `close`, `stat`, `seek`, `read`, `write`, `scanf`, `vscanf`, `printf`, `vprintf`. The interfaces for these functions mirror `open`, `close`, `stat`, `lseek`, `read`, `write`, `fscanf`, `vfscanf`, `fprintf` and `vfprintf` respectively. These turned out to be the functions you need to offer to support all code currently in libpd.

    s_fileops.c comes with two filled-out `t_fileops` structures, `sys_fileops_standard`, in which every function wraps the `FILE *` functions from the C standard library (`fopen`, etc)  and `sys_fileops_null`, where every function immediately fails.

    There is a global extern `t_fileops` named `sys_fileops`. By default (unless you compile with `_PD_FILEOPS_NO_DEFAULT`), it is set to the contents of `sys_fileops_standard`.
* `t_fileops_handle`: This is an opaque handle which is passed into the `t_fileops` functions in lieu of a file descriptor. Its meaning is defined by a particular t_fileops struct. It is an `intptr_t` because the intent is it should be possible for it to be any of a pointer, a filehandle, or an index of some kind.
* `t_fileops_flags`, a bitmask enum that incorporates both the flags passed to `open` and the "whence" value passed to `seek`.
* `t_fileops_stat`, a structure that `stat` writes its values into (currently the only members are two bools, `isdir` and `isdir_known`; puredata doesn't use `stat` for anything except making sure it isn't accidentally opening a directory as a file).

I have converted all file operations in pure-data to use the `sys_fileops` functions instead of whatever they were doing before (for example, calls to `fopen` have been replaced with calls to `sys_fileops.open`). The idea is that when running the pure-data application, `sys_fileops` will remain set to its default values (call the stdlib) and users will not experience anything different; but when using pdlib, you have the option of calling `void sys_fileops_set(t_fileops *fileops);` at startup, and this will overwrite `sys_fileops` with your own custom filesystem implementation.

One note on API consequences: the t_fileops `open` does not return a `t_fileops_handle`; instead, it returns a bool (true for success, false for failure) and the `t_fileops_handle` value is a pointer out argument. This is because there's no universal "invalid" value for `t_fileops_handle` (if a `t_fileops` defines its handles as pointers the "invalid" pointer would be NULL; if it implements them as filehandles, an invalid value might be -1) so it has to return the handle and success/failure separately. This had repercussions throughout pure-data/libpd because there are many functions in pure-data, such as `canvas_open`, which return a filehandle or a `FILE *` and expect the caller to detect failure from an invalid value.

### TODO

This patch "works" in that I can build `samples/c/pdtest` in the libpd repo against patched pure-data and the sample works as expected. However there are a few things that need to be done before this patch can be accepted. (I wanted to get this proof-of-concept posted here and get feedback on the API before moving on these final items.)

* I haven't implemented `sys_fileops_set` yet. I want this function to be kind of "smart" in that if a function isn't included it should fill it in with functions from `sys_fileops_null`, if `read` and `write` are provided but `printf`/`scanf` aren't it would be really cool if it could provide a printf/scanf based on the given `read` and `write` (you might be able to do this with `funopen`? though I'm not sure you can get funopen on Windows…).

* **I have not added `s_fileops.c` to the build system yet.** Instead it says `#include "s_fileops.c"` at the end of s_path.c. This is awful, but I'm not sure I understand the build system(s) here well enough to fix it myself. Is there a list of every file you need to add a new .c file to to get the various pure-data and libpd build scripts to recognize it?

* If this patch is accepted, I would like to also submit sample code to the `libpd` repo containing a `t_fileops` implementation for Physfs (the VFS library people are most likely to use).

* I broke _LARGEFILE64_SOURCE but haven't fixed it.

* I haven't tested on Windows.

### …Questions?

This turned out to be a pretty big patch. I tried to make the API as simple (and therefore hopefully uncontroversial) as possible, and to my eye the patch actually simplifies the libpd code (before, libpd was actually using a mishmash of `open` and `fopen` in different places, occasionally awkwardly switching from one to the other with `fdopen`). However, there are some places I am not quite sure I did the right thing.

Feedback invited on:

* My emulation of the `open` flags using `fopen` turned out to be pretty… awkward. `fopen` with "w" behaves as if O_CREAT is always true, and emulating O_CREAT false on "w" is pretty convoluted. (One option would be to *always* open with `open` and then `fdopen`, as the old code did in some places, but it seems **nice** to be using the C stdlib exclusively here.)

* I think the O_TRUNC behavior is totally wrong. I think the current code in this patch **always** truncates files when writing them. I have not reviewed the uses of `open` in the existing code to see if this breaks anything anywhere.

* I'm not sure my decision to use `true` for success and `false` for failure was a good one. It required introducing `bool` to the `_MSC_VER` block in `m_pd.h` defining all the C99 types MSVC doesn't support, and a weird `tobool` function (because without C99 a simple cast to bool won't coerce to true/false). Some other options here would have been:

    * A number of file operations in the codebase follow the convention of 0 for success and -1 for failure. Using 0 and -1 would make the newly introduced fileops functions more consistent with the existing file functions (as well as `t_fileops::seek`, which uses -1 for failure because its interface is lseek's), and would mean we don't have to introduce `bool` to puredata.

    * Instead of `intptr_t`, `t_fileops_handle` could be `void *` and we could simply define that NULL always means an invalid fileops_handle. This would mean I could back out the signature changes to functions like `canvas_open` (although we'd still be returning a different value, and we'd still have to replace n<0 filehandle checks with !n pointer checks). If someone defining a `t_fileops` wants to use a file descriptor they can malloc the space for one and return that.

* This patch simplifies the Windows compatibility significantly (because we're no longer working with filehandles), which is great, but because I don't completely understand the existing Windows compatibility I'm worried I broke something. In particular, instead of simply calling `fopen`, on `_WIN32` you translate the path to a `wchar_t` and call `_wfopen`. Why?

* There is a directory named `extras/pd~`. The source in this directory calls `open` and other file operations independently. It would be slightly hard to convert this to `t_fileops` because in one place it makes explicit use of the fact stdin is a filehandle. I don't know what this code is or whether it's a problem it isn't using fileops. What is it?

* I implemented this patch against the git revision in the most recent version of `libpd`. I think this corresponds to the `libpd` branch. Github says the patch does not cleanly apply against `master` or `develop`. I don't know what your branching strategy is so I don't know what you need from me here.

* I am not totally sure how to give this a full test. As mentioned I tested with `samples/c/pdtest` in libpd but this doesn't test, for example, loading sound files.

* Because this patch simplified the Windows compatibility and removed some functions, a comment that legitimately appears to date from the 90s saying `/* Bill Gates is a big fat hen */` was removed. This felt like demolishing a historic preservation site

### PS

By the way, after writing `s_fileops.c` and `s_fileops.h` I realized that the API I'd designed for this patch was not puredata specific. At some point I will probably extract the contents of those two files, remove the puredata derived code (currently, `std_open` only) and post a version as a CC0 (ie, not encumbered by the puredata license) standalone library on my own github.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -71,8 +71,16 @@ typedef unsigned __int8   uint8_t;
 typedef unsigned __int16  uint16_t;
 typedef unsigned __int32  uint32_t;
 typedef unsigned __int64  uint64_t;
+typedef unsigned __int8   bool;
+#define false 0
+#define true 1
+#define PRIxPTR "p"
+#define tobool(x) ((x)?true:false)
 #else
-# include <stdint.h>
+#include <stdint.h>
+#include <stdbool.h>
+#include <inttypes.h>
+#define tobool(x) ((bool)(x))
 #endif
 
 /* for FILE, needed by sys_fopen() and sys_fclose() only */
@@ -304,6 +312,10 @@ typedef void (*t_gotfn5)(void *x,
 EXTERN void obj_list(t_object *x, t_symbol *s, int argc, t_atom *argv);
 EXTERN t_pd *pd_newest(void);
 
+/* --------------- fileops -------------------- */
+
+#include "s_fileops.h"
+
 /* --------------- memory management -------------------- */
 EXTERN void *getbytes(size_t nbytes);
 EXTERN void *getzbytes(size_t nbytes);
@@ -443,8 +455,8 @@ EXTERN int sys_zoomfontheight(int fontsize, int zoom, int worstcase);
 EXTERN int sys_fontwidth(int fontsize);
 EXTERN int sys_fontheight(int fontsize);
 EXTERN void canvas_dataproperties(t_glist *x, t_scalar *sc, t_binbuf *b);
-EXTERN int canvas_open(const t_canvas *x, const char *name, const char *ext,
-    char *dirresult, char **nameresult, unsigned int size, int bin);
+EXTERN bool canvas_open(const t_canvas *x, const char *name, const char *ext,
+    char *dirresult, char **nameresult, t_fileops_handle *fd, unsigned int size, int bin);
 
 /* ---------------- widget behaviors ---------------------- */
 
@@ -549,20 +561,11 @@ EXTERN void sys_ouch(void);
 EXTERN int sys_isabsolutepath(const char *dir);
 EXTERN void sys_bashfilename(const char *from, char *to);
 EXTERN void sys_unbashfilename(const char *from, char *to);
-EXTERN int open_via_path(const char *dir, const char *name, const char *ext,
-    char *dirresult, char **nameresult, unsigned int size, int bin);
+EXTERN bool open_via_path(const char *dir, const char *name, const char *ext,
+    char *dirresult, char **nameresult, t_fileops_handle *file, unsigned int size, int bin);
 EXTERN int sched_geteventno(void);
 EXTERN double sys_getrealtime(void);
 EXTERN int (*sys_idlehook)(void);   /* hook to add idle time computation */
-
-/* Win32's open()/fopen() do not handle UTF-8 filenames so we need
- * these internal versions that handle UTF-8 filenames the same across
- * all platforms.  They are recommended for use in external
- * objectclasses as well so they work with Unicode filenames on Windows */
-EXTERN int sys_open(const char *path, int oflag, ...);
-EXTERN int sys_close(int fd);
-EXTERN FILE *sys_fopen(const char *filename, const char *mode);
-EXTERN int sys_fclose(FILE *stream);
 
 /* ------------  threading ------------------- */
 EXTERN void sys_lock(void);
--- a/src/s_fileops.h
+++ b/src/s_fileops.h
@@ -1,0 +1,46 @@
+#pragma once
+
+/* Copyright (c) 2020 Andi McClure.
+* For information on usage and redistribution, and for a DISCLAIMER OF ALL
+* WARRANTIES, see the file, "LICENSE.txt," in this distribution.  */
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdarg.h>
+
+typedef intptr_t t_fileops_handle;
+
+typedef enum {
+	FILEOPS_NONE = 0,
+	FILEOPS_READ = 1,
+	FILEOPS_WRITE = 2,
+	FILEOPS_CREAT = 4,
+	FILEOPS_TRUNC = 8,
+	FILEOPS_SEEK_SET = 0x10,
+	FILEOPS_SEEK_CUR = 0x20,
+	FILEOPS_SEEK_END = 0x40,
+} t_fileops_flags;
+
+typedef struct _fileops_stat {
+	bool isdir_known; // If false, isdir value is undefined
+	bool isdir;
+} t_fileops_stat;
+
+typedef struct _t_fileops /* All function pointers return true on success */
+{
+	bool (*open)(const char *path, t_fileops_flags flags, t_fileops_handle *handle);
+	bool (*close)(t_fileops_handle handle);
+	bool (*stat)(t_fileops_handle handle, t_fileops_stat *stat);
+	int64_t (*seek)(t_fileops_handle handle, int64_t offset, t_fileops_flags flags);
+	ssize_t (*read)(t_fileops_handle handle, void *buf, size_t nbyte);
+	ssize_t (*write)(t_fileops_handle handle, const void *buf, size_t nbyte);
+	ssize_t (*scanf)(t_fileops_handle handle, const char * restrict format, ...);
+	ssize_t (*vscanf)(t_fileops_handle handle, const char * restrict format, va_list ap);
+	ssize_t (*printf)(t_fileops_handle handle, const char * restrict format, ...);
+	ssize_t (*vprintf)(t_fileops_handle handle, const char * restrict format, va_list ap);
+	bool (*flush)(t_fileops_handle handle);
+} t_fileops;
+
+// Also search: fstat, sys_fopen
+
+extern t_fileops sys_fileops;
--- a/extra/bonk~/bonk~.c
+++ b/extra/bonk~/bonk~.c
@@ -991,13 +991,12 @@ static void bonk_read(t_bonk *x, t_symbol *s)
     t_float *fp2;
 
     /* fbar: canvas_open code taken from g_array.c */
-    FILE *fd;
+    t_fileops_handle fd;
     char buf[MAXPDSTRING], *bufptr;
-    int filedesc;
+    t_fileops_handle filedesc;
 
-    if ((filedesc = canvas_open(x->x_canvas,
-            s->s_name, "", buf, &bufptr, MAXPDSTRING, 0)) < 0 
-                || !(fd = fdopen(filedesc, "r")))
+    if (!canvas_open(x->x_canvas,
+            s->s_name, "", buf, &bufptr, &filedesc, MAXPDSTRING, 0))
     {
         post("%s: open failed", s->s_name);
         return;
@@ -1007,7 +1006,7 @@ static void bonk_read(t_bonk *x, t_symbol *s)
     while (1)
     {
         for (i = x->x_nfilters, fp = vec; i--; fp++)
-            if (fscanf(fd, "%f", fp) < 1) goto nomore;
+            if (sys_fileops.scanf(fd, "%f", fp) < 1) goto nomore;
         x->x_template = (t_template *)t_resizebytes(x->x_template,
             ntemplate * sizeof(t_template),
                 (ntemplate + 1) * sizeof(t_template));
@@ -1027,7 +1026,7 @@ nomore:
     }
     post("bonk: read %d templates\n", ntemplate);
     x->x_ntemplate = ntemplate;
-    fclose(fd);
+    sys_fileops.close(fd);
 }
 #endif
 
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -18,7 +18,7 @@ objects use Posix-like threads.  */
 #ifdef _WIN32
 #include <io.h>
 #endif
-#include <fcntl.h>
+#include "s_fileops.h"
 #include <stdio.h>
 #include <string.h>
 #include <errno.h>
@@ -259,7 +259,7 @@ static void outlet_soundfile_info(t_outlet *out, t_soundfile_info *info)
 * header and fill in the properties.
 */
 
-int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
+bool open_soundfile_via_handle(t_fileops_handle fd, t_soundfile_info *p_info, long skipframes)
 {
     int nchannels, bigendian, bytespersamp, samprate, swap;
     int headersize = 0;
@@ -285,7 +285,7 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
             t_datachunk b_datachunk;
             t_comm b_commchunk;
         } buf;
-        long bytesread = read(fd, buf.b_c, READHDRSIZE);
+        long bytesread = sys_fileops.read(fd, buf.b_c, READHDRSIZE);
         int format;
         if (bytesread < 4)
             goto badheader;
@@ -358,10 +358,10 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
                 if (!strncmp(wavechunk->wc_id, "fmt ", 4))
                 {
                     long commblockonset = headersize + 8;
-                    seekout = (long)lseek(fd, commblockonset, SEEK_SET);
+                    seekout = (long)sys_fileops.seek(fd, commblockonset, FILEOPS_SEEK_SET);
                     if (seekout != commblockonset)
                         goto badheader;
-                    if (read(fd, buf.b_c, sizeof(t_fmt)) < (int) sizeof(t_fmt))
+                    if (sys_fileops.read(fd, buf.b_c, sizeof(t_fmt)) < (int) sizeof(t_fmt))
                             goto badheader;
                     nchannels = swap2(buf.b_fmt.f_nchannels, swap);
                     format = swap2(buf.b_fmt.f_nbitspersample, swap);
@@ -374,10 +374,10 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
                     else goto badheader;
                     samprate = swap4(buf.b_fmt.f_samplespersec, swap);
                 }
-                seekout = (long)lseek(fd, seekto, SEEK_SET);
+                seekout = (long)sys_fileops.seek(fd, seekto, SEEK_SET);
                 if (seekout != seekto)
                     goto badheader;
-                if (read(fd, buf.b_c, sizeof(t_wavechunk)) <
+                if (sys_fileops.read(fd, buf.b_c, sizeof(t_wavechunk)) <
                     (int) sizeof(t_wavechunk))
                         goto badheader;
                 /* post("new chunk %c %c %c %c at %d",
@@ -421,10 +421,10 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
                 {
                     long commblockonset = headersize + 8;
                     t_comm *commchunk;
-                    seekout = (long)lseek(fd, commblockonset, SEEK_SET);
+                    seekout = (long)sys_fileops.seek(fd, commblockonset, SEEK_SET);
                     if (seekout != commblockonset)
                         goto badheader;
-                    if (read(fd, buf.b_c, sizeof(t_comm)) <
+                    if (sys_fileops.read(fd, buf.b_c, sizeof(t_comm)) <
                         (int) sizeof(t_comm))
                             goto badheader;
                     commchunk = &buf.b_commchunk;
@@ -437,10 +437,10 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
                     else goto badheader;
                     samprate = readaiffsamprate((char *)commchunk->c_samprate, swap);
                 }
-                seekout = (long)lseek(fd, seekto, SEEK_SET);
+                seekout = (long)sys_fileops.seek(fd, seekto, SEEK_SET);
                 if (seekout != seekto)
                     goto badheader;
-                if (read(fd, buf.b_c, sizeof(t_datachunk)) <
+                if (sys_fileops.read(fd, buf.b_c, sizeof(t_datachunk)) <
                     (int) sizeof(t_datachunk))
                         goto badheader;
                 headersize = (int)seekto;
@@ -450,7 +450,7 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
         }
     }
         /* seek past header and any sample frames to skip */
-    sysrtn = (long)lseek(fd,
+    sysrtn = (long)sys_fileops.seek(fd,
         ((off_t)nchannels) * bytespersamp * skipframes + headersize, 0);
     if (sysrtn != nchannels * bytespersamp * skipframes + headersize)
         return (-1);
@@ -464,46 +464,48 @@ int open_soundfile_via_fd(int fd, t_soundfile_info *p_info, long skipframes)
     p_info->bytespersample = bytespersamp;
     p_info->bytelimit = bytelimit;
     p_info->samplerate = samprate;
-    return (fd);
+    return true;
 badheader:
         /* the header wasn't recognized.  We're threadable here so let's not
         print out the error... */
     errno = EIO;
-    return (-1);
+    return false;
 }
 
     /* open a soundfile, using open_via_path().  This is used by readsf~ in
     a not-perfectly-threadsafe way.  LATER replace with a thread-hardened
     version of open_soundfile_via_canvas() */
-int open_soundfile(const char *dirname, const char *filename,
+bool open_soundfile(const char *dirname, const char *filename, t_fileops_handle *file,
     t_soundfile_info *p_info, long skipframes)
 {
     char buf[OBUFSIZE], *bufptr;
-    int fd, sf_fd;
-    fd = open_via_path(dirname, filename, "", buf, &bufptr, MAXPDSTRING, 1);
-    if (fd < 0)
-        return (-1);
-    sf_fd = open_soundfile_via_fd(fd, p_info, skipframes);
-    if (sf_fd < 0)
-        sys_close(fd);
-    return (sf_fd);
+    t_fileops_handle fd, sf_fd;
+    if (!open_via_path(dirname, filename, "", buf, &bufptr, &fd, MAXPDSTRING, 1))
+        return false;
+    if (!open_soundfile_via_handle(fd, p_info, skipframes)) {
+        sys_fileops.close(fd);
+        return false;
+    }
+    *file = fd;
+    return true;
 }
 
     /* open a soundfile, using open_via_canvas().  This is used by readsf~ in
     a not-perfectly-threadsafe way.  LATER replace with a thread-hardened
     version of open_soundfile_via_canvas() */
-int open_soundfile_via_canvas(t_canvas *canvas, const char *filename,
+bool open_soundfile_via_canvas(t_canvas *canvas, const char *filename, t_fileops_handle *file,
     t_soundfile_info *p_info, long skipframes)
 {
     char buf[OBUFSIZE], *bufptr;
-    int fd, sf_fd;
-    fd = canvas_open(canvas, filename, "", buf, &bufptr, MAXPDSTRING, 1);
-    if (fd < 0)
-        return (-1);
-    sf_fd = open_soundfile_via_fd(fd, p_info, skipframes);
-    if (sf_fd < 0)
-        sys_close(fd);
-    return (sf_fd);
+    t_fileops_handle fd;
+    if (!canvas_open(canvas, filename, "", buf, &bufptr, &fd, MAXPDSTRING, 1))
+        return false;
+    if (open_soundfile_via_handle(fd, p_info, skipframes)) {
+        sys_fileops.close(fd);
+        return false;
+    }
+    *file = fd;
+    return true;
 }
 
 static void soundfile_xferin_sample(int sfchannels, int nvecs, t_sample **vecs,
@@ -828,7 +830,7 @@ usage:
 }
 
 /* p_headersize is a getter, set to NULL if not needed */
-static int create_soundfile(t_canvas *canvas, const char *filename,
+static bool create_soundfile(t_canvas *canvas, const char *filename, t_fileops_handle *file,
     int filetype, long nframes, int bytespersamp, int bigendian, int nchannels,
     int swap, t_float samplerate, int *p_headersize)
 {
@@ -837,7 +839,7 @@ static int create_soundfile(t_canvas *canvas, const char *filename,
     t_wave *wavehdr = (t_wave *)headerbuf;
     t_nextstep *nexthdr = (t_nextstep *)headerbuf;
     t_aiff *aiffhdr = (t_aiff *)headerbuf;
-    int fd, headersize = 0;
+    t_fileops_handle fd; int headersize = 0;
 
     strncpy(filenamebuf, filename, MAXPDSTRING);
     filenamebuf[MAXPDSTRING-10] = 0;
@@ -906,20 +908,21 @@ static int create_soundfile(t_canvas *canvas, const char *filename,
     }
 
     canvas_makefilename(canvas, filenamebuf, buf2, MAXPDSTRING);
-    if ((fd = sys_open(buf2, O_WRONLY | O_CREAT | O_TRUNC, 0666)) < 0)
-        return (-1);
+    if (!sys_fileops.open(buf2, FILEOPS_WRITE | FILEOPS_CREAT | FILEOPS_TRUNC, &fd))
+        return false;
 
-    if (write(fd, headerbuf, headersize) < headersize)
+    if (sys_fileops.write(fd, headerbuf, headersize) < headersize)
     {
-        close (fd);
-        return (-1);
+        sys_fileops.close (fd);
+        return false;
     }
+    *file = fd;
     if(p_headersize)
         *p_headersize = headersize;
-    return (fd);
+    return true;
 }
 
-static void soundfile_finishwrite(void *obj, const char *filename, int fd,
+static void soundfile_finishwrite(void *obj, const char *filename, t_fileops_handle fd,
     int filetype, long nframes, long itemswritten, int bytesperframe, int swap)
 {
     if (itemswritten < nframes)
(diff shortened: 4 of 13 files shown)


mcclure:
@umlaeute Thanks. I'd be happy to make these changes, but unfortunately I am not familiar with this project so I need some help if that's okay. (And if there is a contribution document somewhere I missed, I can read that.)

First, my big question:

> this PR breaks both API and ABI. Pd is very conservative when it comes to dropping APIs (i don't remember this happening in the last 20 years), so you basically can still run binaries from the 1990s with a current version of Pd. let's keep it like this

What is the surface of the ABI/API? That is, which specific signatures/header files do I need to leave unchanged? Are all EXTERN symbols fair game?

If "API" is all externs, then the only way I can think of to leave that unchanged is:

- Leave in `canvas_open`, `open_via_path`, `sys_open_absolute`, `sys_trytoopenone` (and `sys_open` et al), but *add* `canvas_open_handle`, `open_via_path_handle`, `sys_open_absolute_handle` `sys_trytoopenone_handle`. Internally all puredata functions should use the _handle versions, and if an fd needs to be returned by one of those functions it would be converted to an fd at the point of return. (To make this possible I would add a `to_fd` to `t_fileops` which is supported by the std fileops but probably by not other t_fileops implementations.) This would mean that if you ever call `sys_fileops_set`, then until the process terminates the original `canvas_open` would fail and only `canvas_open_handle`, but since legacy software that uses `canvas_open` will never `sys_fileops_set` it will continue to work fine.

Is this acceptable?

> is there any specific reason to base this PR on 0.50-2, while Pd-0.51-0 was released a couple of weeks ago?

I started the patch 3 weeks ago… lol >_>

**What branch should I target?**  Should I build on top of 0.50-2 ? Develop? Master?

> please make your commits atomic (there is no need to modify (e.g.) extra/bonk~/bonk~.c when adding a new API to the core of Pd

Would one patch for extra/ and one patch for everything else be acceptable?

> Pd traditionally uses #include guards rather than #pragma once. i guess we should keep the style
> the CI-builds all fail. please investigate

I have fixed the CI, fixed the #include guards, merged the patch on top of Pd-0.51-0, and changed the PR to target develop. After merging the patch to Pd-0.51-0, I am no longer able to build libpd master from libpd's included `Makefile`. It complains a bunch of networking-related symbols such as `_sys_startgui` and `_sockaddr_is_multicast` are missing. If I add `pure-data/src/s_net.c` to the libpd `Makefile`, libpd can compile again and I can run my test.

I have **not** added the new files to any build scripts **except** libpd Makefile and pure-data src/Makefile.am.

Note: **The Mac CI is still failing, but this is not related to this patch.** The error is

```
Error: libtool 2.4.6_1 is already installed
To upgrade to 2.4.6_2, run `brew upgrade libtool`.
```

This is a problem with your builder configuration, my patch has no influence over this.

mcclure:
(Got the Appveyor/Windows CI working and adjusted the above comment slightly.)

src/m_pd.h
```diff
@@ -71,8 +71,26 @@ typedef unsigned __int8   uint8_t;
...
+#ifndef true
+#define true 1
+#endif
+#ifndef PRIxPTR
+#define PRIxPTR "p"
+#endif
+#ifndef PRIx64
+#define PRIx64 "I64x"
+#endif
+#define tobool(x) ((x)?true:false)
```
Spacechild1, line 88: why do we need bool types? note that `m_pd.h` might be used by C++ code!

src/m_pd.h
```diff
@@ -443,8 +465,8 @@ EXTERN int sys_zoomfontheight(int fontsize, int zoom, int worstcase);
 EXTERN int sys_fontwidth(int fontsize);
 EXTERN int sys_fontheight(int fontsize);
 EXTERN void canvas_dataproperties(t_glist *x, t_scalar *sc, t_binbuf *b);
-EXTERN int canvas_open(const t_canvas *x, const char *name, const char *ext,
-    char *dirresult, char **nameresult, unsigned int size, int bin);
+EXTERN bool canvas_open(const t_canvas *x, const char *name, const char *ext,
```
Spacechild1, line 468: please don't use `bool` return types. in general, don't change existing `EXTERN` functions.

src/m_pd.h
```diff
@@ -545,21 +567,12 @@ EXTERN void logpost(const void *object, const int level, const char *fmt, ...)
...
 EXTERN int (*sys_idlehook)(void);   /* hook to add idle time computation */
 
-/* Win32's open()/fopen() do not handle UTF-8 filenames so we need
- * these internal versions that handle UTF-8 filenames the same across
- * all platforms.  They are recommended for use in external
- * objectclasses as well so they work with Unicode filenames on Windows */
-EXTERN int sys_open(const char *path, int oflag, ...);
-EXTERN int sys_close(int fd);
-EXTERN FILE *sys_fopen(const char *filename, const char *mode);
-EXTERN int sys_fclose(FILE *stream);
```
Spacechild1, line 561: don't remove existing `EXTERN` functions.

src/s_fileops.h
```diff
@@ -0,0 +1,47 @@
...
+
+/* Copyright (c) 2020 Andi McClure.
+* For information on usage and redistribution, and for a DISCLAIMER OF ALL
+* WARRANTIES, see the file, "LICENSE.txt," in this distribution.  */
+
+#include <stdint.h>
+#include <stdbool.h>
+#include <stdarg.h>
+
+typedef intptr_t t_fileops_handle;
```
Spacechild1, line 12: You can't simply change the `int` fd argument to `intptr_t` without breaking the ABI. Why not simply keep the `int` for all functions and treat it as an abstract index? For "real" file systems, it is the file descriptor, and for virtual file systems it would be an array index or hashmap key.

src/s_fileops.h
```diff
@@ -0,0 +1,47 @@
...
+	int64_t (*seek)(t_fileops_handle handle, int64_t offset, t_fileops_flags flags);
+	ssize_t (*read)(t_fileops_handle handle, void *buf, size_t nbyte);
+	ssize_t (*write)(t_fileops_handle handle, const void *buf, size_t nbyte);
+	ssize_t (*scanf)(t_fileops_handle handle, const char *format, ...);
+	ssize_t (*vscanf)(t_fileops_handle handle, const char *format, va_list ap);
+	ssize_t (*printf)(t_fileops_handle handle, const char *format, ...);
+	ssize_t (*vprintf)(t_fileops_handle handle, const char *format, va_list ap);
+	bool (*flush)(t_fileops_handle handle);
+} t_fileops;
+
```
Spacechild1, line 44: This struct is good for setting the hooks in the host app, but for externals I thinks it's better to also have functions like `sys_open`, `sys_close`, `sys_stat`, etc. in `m_pd.h`, which simply call the corresponding function pointer. The reason is backwards compatibility: if we later add more functions and someone tries to load an external which uses a newer function in an older Pd, the external would simply refuse to load - instead of trying to call a non-existing function pointer and crashing.

The `sys_*` functions would be part of the public API, but `t_fileops` would be semi-private (only to be used by host apps).

src/m_pd.h
Spacechild1: I've mentioned in some other place that we could keep using `int` as abstract file handles. We could also keep the convention for `-1` to mean failure. The host app only has to map file resources to positive integers, which shouldn't be a problem. (It is actually quite natural if file resources are kept in an array).

Spacechild1:
Generally, I think this PR is a great idea! I've already planned to implement something like this, but you have been faster :-) Please see my comments. I share @umlaeute's concerns regarding API/ABI breakage. I think most of it can be simply avoided by keeping `int` as the (abstract) file descriptor type and -1 as indicating failure.

Also, only introduce and include things which are 100% necessary for the task at hand. No `bool` type, please :-)

mcclure:
@spacechild1, thanks. The thing about not externally exposing sys_fileops makes sense.

Do you have any thoughts on my comment above? https://github.com/pure-data/pure-data/pull/1096#pullrequestreview-455517463 I don't know what the API/ABI surface is defined as, but if we're trying just to not change EXTERN functions in headers (are some of the headers private? which ones?) then I think we can get away with just restoring the functions I removed and then adding alternate versions of four functions. There actually doesn't seem(?) to be a lot of opportunities for the handles to escape the puredata internals using the public API. (canvas_open can return a filehandle to external code but I'm not sure what the external code is supposed to do with it.)

Using integers for the "handle" and making that integer equal to an FH would absolutely be an option. However, there are some problems with it:
* Minor: `t_fileops` implementations might have to be a little less efficient.
* Kinda minor: Windows doesn't do filehandles natively, and the current code for yanking them out of _open et al is uglyyy
* Major: **Puredata uses fscanf, fprintf a lot and these work on FILE * pointers, not filehandles.** We can convert between FILE *s and filehandles but I am not sure if there are costs associated with doing so. fdopen()/_fdopen()ing and then fclose()ing a FILE * on every single invocation of sys_fileops.scanf sounds… like it might actually have noticeable costs associated with it. (In the code I replaced it dealt with this by a patchwork of arbitrarily fdopening at various times and fclosing at various times, I'm not sure the existing code is ultra efficient either.)

This problem with fscanf/fprintf are why, when I implemented this patch in the first place, I decided to use the FILE * as the underlying handle for std_fileops rather than filehandles.

Spacechild1:
I see the problem with `FILE *`. I will now have a closer look to your PR and think about it.

mcclure:
OK makes sense.

Whatever approach is the same to me. I will wait for your further comments before switching back the function signatures and rebasing the changesets as umlaeute suggested.

Spacechild1:
I think you can switch to `int` handles already, but I would like to wait for @umlaeute's opinion.

umlaeute:
i basically agree with everything @Spacechild1 said ;-)

the file-descriptors returned by `sys_open()` and friends (`canvas_open()`...) are mostly *used* as such (file-descriptors), so they should naturally map to real file-descriptors (if possible).

i agree that the user should never see the `sys_fileops` struct, it should be **private**.
instead of directly manipulating the struct (by the host), use a function to modify it:

~~~
int sys_set_fileops(int (*open)(const char *path, t_fileops_flags flags, int *handle),
                    int (*close)(int handle),
                    int (*stat)(int handle, t_fileops_stat *stat),
                    off64_t (*seek)(int handle, int64_t offset, t_fileops_flags flags),
                    ssize_t (*read)(int handle, void *buf, size_t nbyte),
                    ssize_t (*write)(int handle, const void *buf, size_t nbyte),
                    ssize_t (*vscanf)(int handle, const char *format, va_list ap),
                    ssize_t (*vprintf)(int handle, const char *format, va_list ap),
                    int (*flush)(int handle));
~~~

and the only way to access the new file-handles is through `sys_*` wrappers (`sys_open`,...). (which means that we get a couple of new wrappers).
sidenote: my `sys_set_fileops()` declaration intentionally omits the variadic variants (`printf`,`scanf`) as they cannot be sanely wrapped.

for the sake for forward compatibility i was tempted to suggest `int sys_set_fileops(t_fileops_type type, t_fileops_fun fun);` (as in `sys_set_fileops(OPEN, my_weird_openfunction);`); this might make sense iff there is a slight chance that we ever need to extend the list of functions, but none otherwise.
or provide both stdio and POSIX variants (`sys_set_fileops_stdio(...)` vs `sys_set_fileops_posix(...)`) ... ;-)

umlaeute:
## Q & A
> 
> First, my big question:
> 
> What is the surface of the ABI/API? That is, which specific signatures/header files do I need to leave unchanged? Are all EXTERN symbols fair game?

yep, all `EXTERN`s (and `extern`s)

> > is there any specific reason to base this PR on 0.50-2, while Pd-0.51-0 was released a couple of weeks ago?
> 
> I started the patch 3 weeks ago… lol >_>

ah well, Pd-0.51-0 it was released around 2020-06-06, which is a bit longer than three weeks :-P

> 
> **What branch should I target?** Should I build on top of 0.50-2 ? Develop? Master?

either.
since this is a big structural change, it won't get merged into `develop` directly (as this only contains "trivial" and undisputed changes), the only chance for inclusion is through a direct merge into `master` by @millerpuckette 

> > please make your commits atomic (there is no need to modify (e.g.) extra/bonk~/bonk~.c when adding a new API to the core of Pd
> 
> Would one patch for extra/ and one patch for everything else be acceptable?

rather not.
i think it would be best to split into three stages:
- implementing the abstraction layer (without actually using it)
- using the abstraction layer internally (that is: in `src/`)
- using the abstraction layer in the shipped externals (that is: in `extra/`)

and keep these stages separate even for follow-up commits.

> I have fixed the CI, fixed the #include guards, merged the patch on top of Pd-0.51-0, and changed the PR to target develop. After merging the patch to Pd-0.51-0, I am no longer able to build libpd master from libpd's included `Makefile`. It complains a bunch of networking-related symbols such as `_sys_startgui` and `_sockaddr_is_multicast` are missing. If I add `pure-data/src/s_net.c` to the libpd `Makefile`, libpd can compile again and I can run my test.

bug in libpd's build system. non of our concern here ;-)

> I have **not** added the new files to any build scripts **except** libpd `Makefile`, pure-data `src/Makefile.am` and pure-data `src/Makefile.msvc`.

as said, we are not concerned about the libpd buildsystem (#519 where art thou...).
before this gets accepted, all `src/makefile.*` must be updated accordingly.


> 
> Note: **The Mac CI is still failing, but this is not related to this patch.** 

fixed in `develop` (good reason to use `develop` as the base ;-))

mcclure:
@umlaeute "ah well, Pd-0.51-0 it was released around 2020-06-06, which is a bit longer than three weeks " Not sure if this is helpful to know, but https://github.com/libpd/libpd is still on 50-1. I'm not sure what your "libpd" branch is (in this repo) for but it seems to also be on a version of 50-1.

umlaeute:
> ### TODO
> 
> This patch "works" in that I can build `samples/c/pdtest` in the libpd repo against patched pure-data and the sample works as expected. 

alas, i think that currently the best way would be to directly test with Pd rather than libpd.

>     * I haven't implemented `sys_fileops_set` yet. I want this function to be kind of "smart" in that if a function isn't included it should fill it in with functions from `sys_fileops_null`, if `read` and `write` are provided but `printf`/`scanf` aren't it would be really cool if it could provide a printf/scanf based on the given `read` and `write` (you might be able to do this with `funopen`? though I'm not sure you can get funopen on Windows…).

i wouldn't build too much magic into the system.
it mostly adds complexity, and complexity tends to tumble.

> 
>     * **I have not added `s_fileops.c` to the build system yet.** Instead it says `#include "s_fileops.c"` at the end of s_path.c. This is awful, but I'm not sure I understand the build system(s) here well enough to fix it myself. Is there a list of every file you need to add a new .c file to to get the various pure-data and libpd build scripts to recognize it?

Pure Data is a standard autotool project. add your files to `src/Makefile.am`.
Pure Data is also a standard makefile project. add your files to `src/makefile.*`.
libpd is a standard makefile project. add your files to the libpd `Makefile`

> 
> ### …Questions?
> 
> Feedback invited on:
> 
>     * I'm not sure my decision to use `true` for success and `false` for failure was a good one. It required introducing `bool` to the `_MSC_VER` block in `m_pd.h` defining all the C99 types MSVC doesn't support, and a weird `tobool` function (because without C99 a simple cast to bool won't coerce to true/false). Some other options here would have been:

i think  @Spacechild1 answered this already (with a good plan "no")

>     * This patch simplifies the Windows compatibility significantly (because we're no longer working with filehandles), which is great, but because I don't completely understand the existing Windows compatibility I'm worried I broke something. In particular, instead of simply calling `fopen`, on `_WIN32` you translate the path to a `wchar_t` and call `_wfopen`. Why?

because Pd internally uses UTF-8 for all strings and Microsoft never really understood UTF-8 and instead have UTF-16. So if you want to open files with non-ASCII chars (probably only non-ISO8859 chars, i haven't checked), you must convert to UTF-16.

> 
>     * There is a directory named `extras/pd~`. The source in this directory calls `open` and other file operations independently. It would be slightly hard to convert this to `t_fileops` because in one place it makes explicit use of the fact stdin is a filehandle. I don't know what this code is or whether it's a problem it isn't using fileops. What is it?

it shouldn't use fileops.
the code established a stdin/stdout pipe between to processes. that should be besides VFS.

also, there's something similar going on in `src/s_inter.c` which also needs to bypass the VFS.

umlaeute:
> @umlaeute "ah well, Pd-0.51-0 it was released around 2020-06-06, which is a bit longer than three weeks " Not sure if this is helpful to know, but https://github.com/libpd/libpd is still on 50-1. I'm not sure what your "libpd" branch is (in this repo) for but it seems to also be on a version of 50-1.

unfortunately, `libpd` lags behind `puredata` (being a separate project).

the `libpd` branch in *this* repository is my attempt to merge the two projects (PR #519), but inclusion was somehow delayed and then forgotten (hence it's still where it is)

Spacechild1:
```
int sys_set_fileops(int (*open)(const char *path, t_fileops_flags flags, int *handle),
                    int (*close)(int handle),
                    int (*stat)(int handle, t_fileops_stat *stat),
                    off64_t (*seek)(int handle, int64_t offset, t_fileops_flags flags),
                    ssize_t (*read)(int handle, void *buf, size_t nbyte),
                    ssize_t (*write)(int handle, const void *buf, size_t nbyte),
                    ssize_t (*vscanf)(int handle, const char *format, va_list ap),
                    ssize_t (*vprintf)(int handle, const char *format, va_list ap),
                    int (*flush)(int handle));
```
I don't like this too much because it's unweildy and not extensible.

---

> for the sake for forward compatibility i was tempted to suggest `int sys_set_fileops(t_fileops_type type, t_fileops_fun fun);` (as in sys_set_fileops(OPEN, my_weird_openfunction);); this might make sense iff there is a slight chance that we ever need to extend the list of functions, but none otherwise.

Pro: easy to extend, possible to override only certain functions
Con: not type-safe (the caller has to make sure that they pass a function of the correct type)

---

Another possibility is to keep the `t_fileops` struct as a means to get/set the functions: `sys_set_fileops(t_fileops *ops)` / `sys_get_fileops(t_fileops *ops)`.

Pro: clear, type safe, extensible
Con: when using an older header with a newer Pd binary, the caller's struct could be smaller than the callee's, which means that the one side would read/write garbage. I don't think this is a likely scenario, since fileops are set by host apps which embed Pd (so the Pd headers and binaries should be in sync), but to make it future proof, we could add some padding to the end of the struct, e.g. `void *reserved[64];`, which has to be zeroed out by the caller.

---

Finally, the easiest and fool-proof solution is to simple use a setter/getter function for each fileop, similar to how libpd manages its hooks: https://github.com/libpd/libpd/blob/master/libpd_wrapper/z_hooks.h

Con: perfectly simple and safe (attempting to set/get a non-existent fileop causes a linker error)
Pro: more text

---

I think deciding about the API for setting/getting the fileops could be the very last step, as the code itself is trivial. So let's not worry too much about it for now.

Spacechild1:
> sidenote: my sys_set_fileops() declaration intentionally omits the variadic variants (printf,scanf) as they cannot be sanely wrapped.

Note that we can still supply `size_t sys_printf(int fd, const char *format, ...)` which internally calls the `vprintf` hook. Same for `sys_scanf`.

mcclure:
I'll move on with the refactor as discussed above.

@Spacechild1 "I don't like this [the function] too much because it's unweildy and not extensible."

I think the thing that is good about a function which takes either a struct or a number of args is that it means all the functions get set at once. This means absent functions can have reasonable replacements inserted.

One (admittedly weird but functional) option for an ABI-forward-compatible way to design this function would be to have it take a "number of arguments" parameter as its first argument.

For example 

int sys_set_fileops(int num_args,
                    int (*open)(const char *path, t_fileops_flags flags, int *handle),
                    int (*close)(int handle),
                    int (*stat)(int handle, t_fileops_stat *stat),
                    off64_t (*seek)(int handle, int64_t offset, t_fileops_flags flags),
                    ssize_t (*read)(int handle, void *buf, size_t nbyte),
                    ssize_t (*write)(int handle, const void *buf, size_t nbyte),
                    ssize_t (*vscanf)(int handle, const char *format, va_list ap),
                    ssize_t (*vprintf)(int handle, const char *format, va_list ap),
                    int (*flush)(int handle));

In this case it would be required to pass 10 for num_args.
If a later revision of the library added an 11th argument, it could check num_args and if it gets 10 assume the code was compiled against the old ABI and disregard the 11th argument.
To make this convention less weird, num_args could be a constant supplied in the header, or the invocation of sys_set_fileops could itself be a #define macro that supplies the hardcoded 10 itself.

millerpuckette:
I had to do something like this about 30 years ago to get "FTS" (Faster Than Sound" to run on bare metal.  I simply wrote functions open(), read(), write(), etc. as a library that my (posix-tested) program could call in the os-less environment.  I think this approach would work fine here and wouldn't require messing with Pd sources- it would just be adding a library to the link command before libc (at least, that worked back then).

mcclure:
@millerpuckette Hm. That honestly sounds like an alarming solution to me, it would create complexity on platforms with unusual build environments (my current project targets 3 different OSes… on the Android target I must build with Gradle and I am not sure I could convince Gradle to override which system libraries it includes) and it would seem to preclude the possibility of the VFS being itself implemented on top of open().

umlaeute:
Am 29. Juli 2020 02:04:59 MESZ schrieb Christof Ressi <notifications@github.com>:
> > sidenote: my sys_set_fileops() declaration intentionally omits the
> variadic variants (printf,scanf) as they cannot be sanely wrapped.
> 
> Note that we can still supply `size_t sys_printf(int fd, const char
> *format, ...)` which internally calls the `vprintf` hook. Same for
> `sys_scanf`.

yes of course.
that's why the vprintf variant is still in my solution (but i should have mentioned that explicitely)

umlaeute:
> I had to do something like this about 30 years ago to get "FTS" (Faster Than Sound" to run on bare metal.

oh. i'd love to have Pd on bare metal (iirc @ritsch tried to talk me into that in the late 90es!)

umlaeute:
> ```
> int sys_set_fileops(int (*open)(const char *path, t_fileops_flags flags, int *handle), ...);
> ```
> 
> I don't like this too much because it's unweildy and not extensible.

yes.

> 
> > for the sake for forward compatibility i was tempted to suggest `int sys_set_fileops(t_fileops_type type, t_fileops_fun fun);` (as in sys_set_fileops(OPEN, my_weird_openfunction);); this might make sense iff there is a slight chance that we ever need to extend the list of functions, but none otherwise.
> 
> Pro: easy to extend, possible to override only certain functions
> Con: not type-safe (the caller has to make sure that they pass a function of the correct type)
> 

true.
but then: we already do that a lot in Pd :stuck_out_tongue: 
i think the worst thing about this is that - as @mcclure mentioned - setting multiple functions (which is what you usually do) is not atomic.

> Another possibility is to keep the `t_fileops` struct as a means to get/set the functions: `sys_set_fileops(t_fileops *ops)` / `sys_get_fileops(t_fileops *ops)`.
> 
> Pro: clear, type safe, extensible
> Con: when using an older header with a newer Pd binary, the caller's struct could be smaller than the callee's, which means that the one side would read/write garbage. I don't think this is a likely scenario, since fileops are set by host apps which embed Pd (so the Pd headers and binaries should be in sync), but to make it future proof, we could add some padding to the end of the struct, e.g. `void *reserved[64];`, which has to be zeroed out by the caller.

hmm well. i don't really see much benefit over the *unweildy and not extensible* solution (but it requires more typing).
it also makes it hard to distinguish between functionpointers that have been deliberately set to NULL and those that are NULL because of padding.

> 
> Finally, the easiest and fool-proof solution is to simple use a setter/getter function for each fileop, similar to how libpd manages its hooks: https://github.com/libpd/libpd/blob/master/libpd_wrapper/z_hooks.h
> 
> Con: perfectly simple and safe (attempting to set/get a non-existent fileop causes a linker error)
> Pro: more text

??? afaict there's no setter/getter functions but just global exported variables. which i think is the worst solution.

> ~~~
> int sys_set_fileops(int num_args,
>                     int (*open)(const char *path, t_fileops_flags flags, int *handle),
>                     ...);
> ~~~
> In this case it would be required to pass 10 for num_args.

that might work.
though i'm not entirely sure whether it **actually** works with the calling convention on some more "exotic"  architectures (like `arm`). i seem to remember that there were some related issues (but might as well be misremembering).


> I think deciding about the API for setting/getting the fileops could be the very last step, as the code itself is trivial. So let's not worry too much about it for now.

probably true.

Spacechild1:
> ??? afaict there's no setter/getter functions but just global exported variables. which i think is the worst solution.

@umlaeute d'oh, I posted exactly the wrong link :-( `z_hooks.h` is of course private, I meant to post this: https://github.com/libpd/libpd/blob/3fd9b7fc43fa8cf82e60339e6b6568f61eda0f88/libpd_wrapper/z_libpd.h#L423

Spacechild1:
@millerpuckette you're suggestions could certainly work for bare-metal environments, but the main motivation for this PR is to use libpd with arbitrary custom virtual filesystems, like those used in game engines. Overriding actual system functions is not a real option here, I think.

umlaeute:
> > ??? afaict there's no setter/getter functions but just global exported variables. which i think is the worst solution.
> 
> @umlaeute d'oh, I posted exactly the wrong link :-( `z_hooks.h` is of course private, I meant to post this: https://github.com/libpd/libpd/blob/3fd9b7fc43fa8cf82e60339e6b6568f61eda0f88/libpd_wrapper/z_libpd.h#L423

thanks.
while it does guarantee type-safety, i still think that with the ubiquituous use of `class_addmethod` in Pd a single registration function `sys_set_fileops(t_fileops_type type, t_fileops_fun fun)` would be nicer...

but yes, that's just bike-shedding.



> @millerpuckette you're suggestions could certainly work for bare-metal environments, but the main motivation for this PR is to use libpd with arbitrary custom virtual filesystems, like those used in game engines. Overriding actual system functions is not a real option here, I think.

agreed, but i think that we should keep the possibility to run on bare-metal on the radar for this PR.

> 
> Also it would provide externals with a well defined set of file I/O functions which are guaranteed to work in all sorts of environments. That would be pretty cool.

yep.

we added `sys_close()` mainly because on Windows you get a file-handle leak if you `open()` with one version of libc (e.g. `libcmt`) but call `close()` from another version of libc (e.g. `libcmtd()`). so to allow externals to choose which libc they want to use, we needed to provide both `open()` and `close()` wrappers.

for whatever reason you can call `read()` using one version of libc on file-handle that was `open()`ed with another version of libc, that's why there's no `sys_read()`.
but it always felt a bit brittle to do so.

Spacechild1:
> agreed, but i think that we should keep the possibility to run on bare-metal on the radar for this PR.

I think that bare-metal support is already supported by this PR. If you define `_PD_FILEOPS_NO_DEFAULT`, all fileops are initialized with null functions and Pd doesn't try to link to any standard file I/O functions. You can then set your own appropriate fileops.

umlaeute:
> I think that bare-metal support is already supported by this PR.

most likely.
my note was meant as a reminder that whatever *comes out* of this PR should also make it (more) possible to run on bare-metal . if it already does, then good - just don't let us accidentally drop this feature while changing the PR.

millerpuckette:
Another problem I think this would cause is that externs like bonk~ would be more difficult to compile outside of the Pd context (as MAX externs for example).  One idea that migth be worth considering is simply defining "open" (etc) to differently-named functions - this is already done in lines like
#define snprintf _snprintf
to get Pd to compile on microsoft.  This should be no big deal if one is recompiling the whole thing for a wierd runtime environment anyhow.

Spacechild1:
> One idea that migth be worth considering is simply defining "open" (etc) to differently-named functions

That would work, but we lose a bit flexibility, as externals have be to recompiled as well. With the current approach, the same external will run on any Pd as long as it uses the proposed file API.

> nother problem I think this would cause is that externs like bonk~ would be more difficult to compile outside of the Pd context (as MAX externs for example). 

Or externals like bonk~ could provide fallback definitions:
```
#ifndef PD
/* fallback for fictional `sys_read` function */
#define sys_read read
#endif
```
There's already so much conditional compilation in objects like bonk~ that a couple more lines won't matter too much, I guess.

mcclure:
I think we could think of vfs support as a new feature, and say externals must be updated to get support for this new feature. Doing otherwise would be changing the behavior of an external out from under it. Maybe the external author is using open on purpose, or uses the filehandle *as* a filehandle in which case replacing open could break the code without warning when the vfs is activated. Also you have no guarantee externals are using open and not fopen…

I have been occupied the last week but will be coming back to this patch soon.

mcclure:
(New commits are maintenance, still planning to come back and make the requested changes)

src/m_pd.h
```diff
@@ -61,8 +61,14 @@ extern int pd_compatibilitylevel;   /* e.g., 43 for pd 0.43 compatibility */
 #include <stddef.h>     /* just for size_t -- how lame! */
 #endif
 
-/* Microsoft Visual Studio is not C99, it does not provide stdint.h */
-#ifdef _MSC_VER
+/* Microsoft Visual Studio is not C99. However it started adding C99 headers
+   beginning with VS2015: https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#C */
+#if !defined(_MSC_VER) || _MSC_VER >= 1900
```
mcclure, line 66: So this might be worth discussing. You've been defining your own int types if MSC_VER is found on the assumption MSVC doesn't have stdint.h. However, MSVC has had stdint.h since Visual Studio 2015. Instead of checking for MSC_VER, you could check for the presence of an MSC_VER less than 1900. This would make it easier to include m_pd.h in files that independently include stdint when embedding (which the C++ program I'm using to test this PR does).

Would this be an acceptable change? Should it be a separate PR? (I assume it should at least be its own commit.)

(Let's leave aside for the moment that you probably don't need bool or ssize_t, as already discussed.)

Spacechild1: > Would this be an acceptable change? Should it be a separate PR?

I've noticed this myself and I agree with you, but that could be a seperate PR (probably even on the develop branch). Only make changes that are necessary for task at hand and safe the rest for later. The smaller the change set, the bigger the chance that things get merged :-)

Spacechild1:
I think this PR would be very useful and we should continue working on it.

---

For reference, `libsndfile` provides the following virtual I/O functions: https://github.com/libsndfile/libsndfile/blob/efe3d9b1ee232e2cf15de833fe3f4e05d983c125/include/sndfile.h#L570-L586.

Of course, they lack some functions that *we* would need (like `open` or `close`), but I also noticed that they have a `tell` function, which this PR is missing.

---

Generally, I think the first important step would be to address all the concerns about API/ABI breakage. In particular my suggestion to keep using `int` as the opaque file handle type (see https://github.com/pure-data/pure-data/pull/1096#discussion_r461234476).

I have also suggested to add `sys_*` API functions, which *internally* call the respecting file op, and use those throughout the Pd code base, i.e. `sys_open` instead of `sys_fileops.open`, `sys_write` instead of `sys_fileops.write`, etc. The actual `sys_fileops` structure should be an implementation detail.

PR #1103, opened 2020-07-28
add "set" method to binops

sebshader:
Add "set" method to 2-operations to allow setting a left-inlet value without outputting: +, -, *, ==, !=, >, <, >=, <=, &, |, &&, ||, <<, >>, %, mod, div


umlaeute:
hmm, i don't really see the benefit of:

~~~
|   |
|   [set $1(
| /
[+ 2]
|
~~~

over what already works:

~~~
|  |
[f ]
|
[+ 2]
|
~~~

(the latter having the advantage that you can initialize the "left-hand" value, e.g. with `$1`)

porres:
I agree there's no much benefit

sebshader:
@umlaeute just less memory usage, and fewer necessary objects. (though I suppose a message box counts as an object..)

umlaeute:
> @umlaeute just less memory usage, and fewer necessary objects. (though I suppose a message box counts as an object that also uses memory..)

well, a message box certainly uses memory, and you have to patch it anyhow (so the only thing you gain is really a *single connection*).

more importantly, the `[set(` message performs much worse than an `[f ]` latch.

a quick benchmark shows that using `[f ]` is about **twice** as fast as using `[set(`.

Spacechild1:
I don't really see the benefit, either....

umlaeute:
so i did another benchmark to check the assumption about memory.

patchA:
~~~
[inlet]   [inlet]    [inlet]
|         |          |
|         [set $1(   |
|       /            |
| +---+              |
|/                   |
[+                   ]
|
[outlet]
~~~

vs patchB:
~~~
[inlet]   [inlet]    [inlet]
|         |          |
[float    ]          |
|                    |
[+                   ]
|
[outlet]
~~~

both patches have exactly the same number of connections (5) and objs/msgs (6).

loading 10000 instances of each patch using `[clone patch<x> 10000]`, and running that through`valgrind` 3 times each, i observe the following memory consumption (as reported by valgrind as "total heap allocated"):

| instances | patchA | patchB |
|---|---|---|
| 5000   | 341094953   | 340929926   |
| 10000 | 1279814956 | 1279484929 |

the numbers are consistent across various runs.
it seems that the abstraction using `[set $1(` takes about 32 bytes more than the one with the `[f ]` latch.

to conclude:
- the `[set(` message is slower
- the `[set(` messages uses (minimally) more memory
- both require about approximately the same number of keystrokes to patch.
- the `[set(` message is nowhere as established in the Pd-objects as the typical hot/cold paradigm (i couldn't find *any* object in Pd-vanilla that uses a `[set(` message like proposed)

i really don't see the point.

sebshader:
fair enough, I was going to leave it up for a little in case I somehow came up with better reasons during patching but it does seem wasteful and unnecessary I guess..

the only thing I can possibly think of is maybe using set with a specific value but that case is pretty rare anyways

Issue #552, opened 2019-02-05
"Delete All Preferences..." should delete Pd-core settings as well


umlaeute:
Sometimes the preferences get corrupted (e.g. #551), and the user might want to start fresh.

The <kbd>File</kbd> -> <kbd>Preferences</kbd> -> <kbd>Forget all...</kbd> menu looks like the solution. However, it doesn't seem to remove any settings from the registry.

Also, when running the `Forget all...` on linux, I get the following in the Pd-console:
> no Pd settings to clear

and on the terminal:
> sh: 2: Syntax error: Bad fd number

umlaeute:
closed with #1107

PR #1107, opened 2020-07-29
bugfix: forget preferences

umlaeute:
in the <kbd>File</kbd>-><kbd>Preferences</kbd> menu, there's an item that says `Forget All...` and which is supposed to reset-all preferences.

Only, it doesn't.

- GUI preferences are completely left alone
- on linux, i get an error on the Pd-console
  > "no Pd settings to clear"

  and an error in the terminal
  > "sh: 2: Syntax error: Bad fd number"

  and the `~/.pdsettings` file is left as is (#552)

- on Windows there might be a similar problem (as indicated in #552), though i have't confirmed so far.

this PR fixes at least the GUI-side and the linux side.

- GUI: implement a `::pd_guiprefs::delete_config` proc and uses it when the user requests to "Forget All...[preferences]".
- core/linux: replace the `system()` calls by proper `stat()` and `unlink()` calls.

the deleting of GUI-preferences has only been tested on Linux, but it is implemented for macOS and Windows.
given that a small typo might delete the entire registry, i would highly appreciate if one of the Windows-folks (@Spacechild1 , @Lucarda ) could review and test.

### DIFF
--- a/src/s_file.c
+++ b/src/s_file.c
@@ -841,11 +841,19 @@ void glob_savepreferences(t_pd *dummy, t_symbol *filesym)
 void glob_forgetpreferences(t_pd *dummy)
 {
 #if !defined(_WIN32) && !defined(__APPLE__)
-    if (system("cat ~/.pdsettings >& /dev/null\n"))
+    char user_prefs_file[MAXPDSTRING]; /* user prefs file */
+    const char *homedir = getenv("HOME");
+    struct stat statbuf;
+    snprintf(user_prefs_file, MAXPDSTRING, "%s/.pdsettings",
+        (homedir ? homedir : "."));
+    user_prefs_file[MAXPDSTRING-1] = 0;
+    if (stat(user_prefs_file, &statbuf) != 0) {
         post("no Pd settings to clear");
-    else if (!system("rm ~/.pdsettings\n"))
-        post("removed .pdsettings file");
-    else post("couldn't delete .pdsettings file");
+    } else if (!unlink(user_prefs_file)) {
+        post("removed %s file", user_prefs_file);
+    } else {
+        post("couldn't delete %s file: %s", user_prefs_file, strerror(errno));
+    }
 #endif  /* !defined(_WIN32) && !defined(__APPLE__) */
 #ifdef __APPLE__
     char cmdbuf[MAXPDSTRING];
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -579,18 +579,28 @@ proc pdtk_pd_startup {major minor bugfix test
     set ::done_init 1
 }
 
-##### routine to ask user if OK and, if so, send a message on to Pd ######
-proc pdtk_check {mytoplevel message reply_to_pd default} {
+##### routine to ask user if OK and, if so return '1' (or else '0')
+# (this really should be in some other file)
+proc pdtk_yesnodialog {mytoplevel message default} {
     wm deiconify $mytoplevel
     raise $mytoplevel
     if {$::windowingsystem eq "win32"} {
-        set answer [tk_messageBox -message [_ $message] -type yesno -default $default \
-                        -icon question -title [wm title $mytoplevel]]
-    } else {
-        set answer [tk_messageBox -message [_ $message] -type yesno \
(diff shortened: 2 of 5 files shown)


millerpuckette:
I haven't been touching GUI preferences because I really would like not to have two preference systems running in parallel.  Ideally I'd like to figure out how to get the GUI preferences to run through Pd's own preferences system.

umlaeute:
ah well. i don't really mind.

however, currently the Pd-GUI has a much more dynamic approach to being able to store preferences - e.g. GUI-plugins can use the preference system to store whatever data they like.
we would need some possibility to have such a dynamic behaviour in the core-preferences first (and honestly: things like the mere existence of `nloadlib` don't make me want to work on that :stuck_out_tongue_closed_eyes: )

i don't see that happening in the near future (and would of course love to be suprised), so could we keep the current double-implementation as an intermediary?
the good thing about the GUI-preference system is that it has a rather generic interface which might make it somewhat possible to switch the native backends currently used against a "Pd-core based backend".
(the bad news is, that it would require a synchronous implementation of `::pd_guiprefs::read` with a network connection in between).

btw, merging the two preference systems might be a bit problematic if running Pd-core and Pd-GUI on different systems.

Lucarda:
> if one of the Windows-folks could review and test.

I'm getting this error:

forgetpref1

```
invalid bareword "errorMsg"
in expression "... }
                }] _@_errorMsg";
should be "$errorMsg" or "{errorMsg}" or "errorMsg(...)" or ...
invalid bareword "errorMsg"
in expression "... }
                }] _@_errorMsg";
should be "$errorMsg" or "{errorMsg}" or "errorMsg(...)" or ...
    (parsing expression "[catch {
             ...")
    invoked from within
"if {[catch {
                    if {$key == ""} {
                        registry delete ${domain}
                    } {
                        r..."
    (procedure "::pd_guiprefs::delete_config" line 9)
    invoked from within
"::pd_guiprefs::delete_config"
    (procedure "::pd_menus::forgetpreferences" line 4)
    invoked from within
"::pd_menus::forgetpreferences"
    (menu invoke)
```

umlaeute:
@Lucarda ah those parens, brackets and curlies... should be fixed with 81dbfa4

Lucarda:
All good with 81dbfa4. 
The `Pure-Data` folder gets wiped from `Computer\HKEY_CURRENT_USER\Software\`

Note: The next time you run Pd the registry folder gets created again and Pd ask to create the documents\pd folder. I think this question about creating the folder is needed to set up the registry?
If I select yes nothing changes to the already created dir documents\pd, so nothing bad happens.

danomatika:
>Note: The next time you run Pd the registry folder gets created again and Pd ask to create the documents\pd folder. I think this question about creating the folder is needed to set up the registry?

No. It asks as the preference setting for the docs folder option is empty as all the settings are wiped. This is normal.

umlaeute:
so it should probably ask whether it should *use* the *Documents\Pd* folder.

as in:

> Do you want Pd to use (and eventually create) a documents directory for patches and external libraries?
> Location: ~/Documents/Pd
> You can change or disable this later in the Path preferences.

Lucarda:
> Do you want Pd to use (and eventually create) a documents directory...

That's better. The current dialog question opens doubts for me as in if choose "no" Pd will not use my already created folder and if I choose yes the: will Pd overwrite my dir?

danomatika:
>The current dialog question opens doubts for me as in if choose "no" Pd will not use my already created folder and if I choose yes the: will Pd overwrite my dir?

No. It's just an existence check. If it doesn't exist and you choose "yes" it creates it and remembers the setting. That's all.

umlaeute:
well the actual question asked by Pd is (emphasis mine)
> Do you want Pd to **create** a documents directory for patches and external libraries?

so Pd asks to *create* a directory (which already exists), because it doesn't know whether the user wants to *use* it.

We also could just check whether the directory exists, and if it doesn't use the original message, but if it does ask:
> Do you want Pd to use a documents directory for patches and external libraries?

danomatika:
>We also could just check whether the directory exists, and if it doesn't use the original message, but if it does ask:
>
>Do you want Pd to use a documents directory for patches and external libraries?

Yes. This is what I'm thinking.

PR #1112, opened 2020-08-06
improve message forwarding to signal inlets

Spacechild1:
Don't send private "fwd" message, instead perform a class check and pass to a helper method.

There are three advantages:
1) no special treatment for "fwd" messages to non-signal inlets
2) no extra allocation needed
3) better error messages (e.g. when trying to forward to a signal inlet without "fwd" argument)

Also, we can now forward pointer messages.

Here's a simple test patch: 
[fwd_test.zip](https://github.com/pure-data/pure-data/files/5038064/fwd_test.zip)

(Note: the gpointer bug mentioned in the test patch should be solved with https://github.com/pure-data/pure-data/pull/841)

### DIFF
--- a/src/g_io.c
+++ b/src/g_io.c
@@ -113,14 +113,14 @@ t_int *vinlet_perform(t_int *w)
     return (w+4);
 }
 
-
-static void vinlet_fwd(t_vinlet *x, t_symbol *s, int argc, t_atom *argv)
+int vinlet_fwd(t_vinlet *x, t_symbol *s, int argc, t_atom *argv)
 {
-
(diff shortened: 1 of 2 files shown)


Spacechild1:
Will fix conflicts if there is interest in merging

millerpuckette:
I don't think this is a high priority since the only reason to have the "fwd" capability at all is to make Max-compatible abstractions.

Spacechild1:
Yes, this PR is just an optimization. (One the other hand, it has been sitting around for almost 3 years now... But maybe PRs need to age like good wine :-D)

PR #1115, opened 2020-08-11
fix MSVC compilation

pierreguillot:
1st commit - x_vexp: remove use of _atoldbl instead of strtof for msvc
The expr objects are not working properly when compiling libpd on Windows with MSVC (I didn't try with Pd directly because I never managed to compile it with MSVC but I guess there is the same problem). The object doesn't manage to parse the ascii to float or integer because `atoif()` ([x_vexp.c#L1979](https://github.com/pure-data/pure-data/blob/master/src/x_vexp.c#L1979)) uses `strtof()` that on Windows with MSVC is defined by `#define strtof(a, b) _atoldbl(a, *b)` ([x_vexp.c#L78](https://github.com/pure-data/pure-data/blob/master/src/x_vexp.c#L78)). So it generates an error: `expr: syntax error:...`. I don't know why this method is overwritten for MSVC but using the default system implementation fixes the problem. 
2nd commit - pd~: use #include <stdio.h>
This can be included directly for all OS and compilation system

### DIFF
--- a/extra/pd~/pd~.c
+++ b/extra/pd~/pd~.c
@@ -12,6 +12,7 @@ learning XCode.  */
 #define MSP
 #endif
 
+#include <stdio.h>
 #ifdef _WIN32
 #include <io.h>
 #include <fcntl.h>
@@ -19,7 +20,6 @@ learning XCode.  */
 #include <windows.h>
 typedef int socklen_t;
 #else
-#include <stdio.h>
 #include <string.h>
 #include <unistd.h>
 #include <stdlib.h>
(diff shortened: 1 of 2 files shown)


millerpuckette:
I altered this slightly to keep things working on MSVC 9.0 as well (laterst version I can run :)

pierreguillot:
The changes to fix the compilation for pd~ work well but I still have an issue with the expr objects. Do you think we could fix this also or do you prefer to keep that for later? I can update the branch to keep only the first commit if necessary.

millerpuckette:
Oops, I fixed this (hopefully) but somehow lost the change in a sea of other confusion.  Updating now (slightly
differently - I still need the old substitution for older MSVC versions)

pierreguillot:
It works perfectly. Thanks!

PR #1119, opened 2020-08-13
disable dark mode for GUI on macOS 10.14+

danomatika:
This is a quick PR which disables "dark mode" for the Pd GUI on macOS 10.14+.

Setting the NSRequiresAquaSystemAppearance default to true instructs the system to use the default "light mode" when opening the Pd GUI. This was not a problem with earlier versions of Pd using TK 8.5, but the newer Tk 8.6 is dark mode aware and result in a mismatch of light text on light backgrounds etc as the GUI is using a light-oriented color scheme.

Until we get around to full supporting dark mode, this triage fix will keep Pd legible for now.

### DIFF
--- a/mac/README.txt
+++ b/mac/README.txt
@@ -176,6 +176,7 @@ Some important per-application settings required by the GUI include:
 * NSQuitAlwaysKeepsWindows: false, disables default 10.7+ window state saving
 * ApplePressAndHoldEnabled: false, disables character compose popup,
                                    enables key repeat for all keys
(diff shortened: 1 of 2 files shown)


danomatika:
Anyone can test this on macOS by running the following in Terminal:

     defaults write org.puredata.pd.pd-gui NSRequiresAquaSystemAppearance -bool true

You can remove this laster with:

    defaults delete org.puredata.pd.pd-gui NSRequiresAquaSystemAppearance

PR #1120, opened 2020-08-13
set canvas background color for macOS and newer Tk versions

danomatika:
This small PR sets the canvas background color too white. On most systems, this is automatically white anyway, but newer versions of macOS using Tk 8.6 will use the system default gray.

*Original idea from @umlaeute on the pd-list*

### DIFF
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -117,7 +117,8 @@ proc pdtk_canvas_new {mytoplevel width height geometry editable} {
     canvas $tkcanvas -width $width -height $height \
         -highlightthickness 0 -scrollregion [list 0 0 $width $height] \
         -xscrollcommand "$mytoplevel.xscroll set" \
-        -yscrollcommand "$mytoplevel.yscroll set"
+        -yscrollcommand "$mytoplevel.yscroll set" \
+        -background white
     scrollbar $mytoplevel.xscroll -orient horizontal -command "$tkcanvas xview"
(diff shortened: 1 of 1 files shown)


tcl/pdtk_canvas.tcl
```diff
@@ -117,7 +117,8 @@ proc pdtk_canvas_new {mytoplevel width height geometry editable} {
     canvas $tkcanvas -width $width -height $height \
         -highlightthickness 0 -scrollregion [list 0 0 $width $height] \
         -xscrollcommand "$mytoplevel.xscroll set" \
-        -yscrollcommand "$mytoplevel.yscroll set"
+        -yscrollcommand "$mytoplevel.yscroll set" \
+        -background white
```
umlaeute, line 121: couldn't we just start using variables rather than hardcoding colours (in the spirit of #196)?

danomatika:
My thinking was to do this tiny triage fix now for Pd 0.51.1-test2 as opposed to rekindling something that could end up being a much larger change, leading to a much lower chance of this getting fixed in the meantime. There is no reason to *not* incorporate your suggestion into #196.

sebshader:
@umlaeute speaking of #196, any chance of reviewing/ any idea when it might get merged? I've made all the changes necessary afaict

danomatika:
>@umlaeute speaking of #196, any chance of reviewing/ any idea when it might get merged? I've made all the changes necessary afaict

As Miller is doing a 0.51-1 bug fix release, I imagine it may not be considered until 0.52.

PR #1122, opened 2020-08-14
Increase socket receive buffer sizes

Spacechild1:
This PR increases the buffer sizes of all socket receivers to 65536, so we can receive large UDP packets. It is based on the current update branch to avoid merge conflicts with recent changes in the networking code.

This implements https://github.com/pure-data/pure-data/issues/903

I verified that I can now send and receive UDP messages up to approx. 65355 bytes with `[netreceive -b]` and `[netreceive -b -u]`.

(This only makes sense via localhost. Actual networks will fragment and might even drop such large packets.)

NOTE: This is basically a (squashed) resubmit of https://github.com/pure-data/pure-data/pull/910 because the target branch has deleted in the meantime. 6b98843 has been added as a further improvement.


millerpuckette:
I'll also plan to merge this for 0.51-3 along with the update branch.  Waiting for the new expr code and then it's off to the races.

millerpuckette:
I'm just looking at this now (somehow it slipped off my plate earlier).  I don't like the idea if having a huge memory buffer lying around statically that usually won't get used at all (UDP packets are only guaranteed to be kept together up to
something like 1500 bytes).  So I think the better appproach would be to just use slip for this kind of comm.  (And yes, Pd needs a vanilla slip encoder/decoder, ugh...)

Spacechild1:
> UDP packets are only guaranteed to be kept together up to something like 1500 bytes

I think 1500 bytes is the MTU for Ethernet. Internet routers usually have an even smaller MTU. Packet fragmentation only increases the *risk* of packet loss (becausewhen a single fragment is lost, the whole UDP packet is dropped), but it is very much possible that all fragments are received and the packet is reassembled successfully.

I agree that it's not a good idea to send huge UDP packets (especially over the internet), but it's definitely possible on certain (local) networks. Finally, I can successfully and reliably send/receive UDP messages up to approx. 65355 bytes over local host.

The theoretical size limit for UDP datagrams is 65535 bytes (with 65507 bytes for data on IPv4). I think we should not arbitrarily limit the size just to save some bytes. If the user knows that their network is capable for such UDP traffic, why not allow this?

To put things into perspective: on 64-bit Pd, 65536 bytes corresponds to a garray of 4096 atoms - which is ~10 ms of audio at 44.1 kHz :-)

umlaeute:
> something like 1500 bytes

high throughput GBit networks would usually use jumbo-frames, that have an MTU of 9000.
that's for real "physical" networks.

and as @Spacechild1 mentioned, "virtual" networks (like `loopback`) can have a larger MTU

~~~sh
$ ip addr show lo | grep -i mtu
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
~~~

danomatika:
Can we just make this a settable value via a message? Leave it small as is for IOhannes and people can expand the size as needed (for everyone else). I have, personally, not needed to increase the the size for what I do, so 4096 is not an *unacceptable* size IMO.

umlaeute:
IOhannes wants it big :-) and I don't think anybody called it *unacceptable*.

probably we could turn the `sys_getrecvbuf()` into a general purpose, heap-allocated per-thread buffer, so that the "huge memory buffer" sees more use.

Spacechild1:
> into a general purpose, heap-allocated per-thread buffer

Small nitpick: the receive buffer is per instance.

Anyway, I think 65 kB per Pd instance is peanuts on todays systems, so I don't quite understand the reservations. IMO it's better to waste a little bit of memory than imposing an arbitrary size limit on incoming UDP packets. 

If someone wants to run Pd on a system with very tight memory constraints, we could add a way to override the buffer size at compile time. Having a runtime option seems a bit overkill to me.

Spacechild1:
> Can we just make this a settable value via a creation argument or message?

umlaeute:
> The buffer is per Pd instance and not per [netreceive] instance.

well, this is true for this PR in its current form.
it could of course be altered to use an object-local heap-allocated buffer.

danomatika:
If it's per-instance, then making it settable would probably solve the problem.

Again, I mention not having run into the 4096 byte barrier in my usage, so I feel like it's a reasonable default. We simply print a message when the buffer is too small and people can then increase it as needed. We could also offer a simple option to use "max available" and query the system/socket or simply use the 65 kB value ala:

~~~
[netreceive -u -b -s max]
~~~

Spacechild1:
> it could of course be altered to use an object-local heap-allocated buffer.

The buffer is only needed once per instance. Why should we make it per object? Only so we can justify to make it settable via `[netreceive]`?

>  I also agree with not wanting to simply bump the size of a fixed buffer by default as we also target more limited systems where this would be a waste of memory

But that's exactly what Miller has done with the symbol table: https://github.com/pure-data/pure-data/commit/36eefd23d793969d3c5c160d9986e0300c474bab

He bumped the size, but let people override it at compile time (for memory constrained environments). I don't see why we can't do the same with the net receive buffer...

To put things into perspective: The (empty) symbol table alone takes 131 kB per instance. The net receive buffer would add another 65 kB. Again, both can be overriden at compile time.

danomatika:
If it's changeable, it's probably fine... but this will need to communicated somehow as people possibly start running into issues on embedded systems (theoretically).

Spacechild1:
Looking at my code again, I noticed that I also use `NET_MAXBUFSIZE` in `t_socketreceiver` (= FUDI type netobjects). Now this is a case where it might make sense to go back to the old buffer size and (later?) make it settable per-object in `[netreceive]` and  `[netsend]`.

Spacechild1:
> but this will need to communicated somehow as people possibly start running into issues on embedded systems (theoretically).

I think mentioning it in the release notes should be enought. And maybe also do it retroactively for the new symbol table size ;-)

danomatika:
Yes. I'm thinking in the context of libpd where the receive buffer may not actually be used.

Spacechild1:
BTW, I just realized that the per-object buffer `sr_inbuf` is only needed for FUDI + TCP. FUDI + UDP could just use the new per-instance receive buffer. Will fix this.

danomatika (approved):

Spacechild1:
sqashed to a single commit for clarity.

src/s_inter.c
```diff
@@ -114,6 +112,7 @@ struct _instanceinter
     int i_maxfd;
     int i_guisock;
     t_socketreceiver *i_socketreceiver;
+    unsigned char *i_recvbuf;
```
umlaeute, line 115: why can't this be a fixed-size buffer?

~~~
unsigned char i_recvbuf[NET_MAXPACKETSIZE];
~~~

since `t_pdinstance.pd_inter` is allocated on the heap anyhow, i don't see much use for the additional indirection.

src/s_inter.c
```diff
@@ -393,6 +392,11 @@ void sys_set_priority(int mode)
 
 /* ------------------ receiving incoming messages over sockets ------------- */
 
+unsigned char *sys_getrecvbuf(void)
+{
+    return INTER->i_recvbuf;
+}
+
```
umlaeute, line 399: what i personally don't like so much is, that the size of the buffer is hardcoded somewhere else (in `s_net.h`).
couldn't that be returned along with the buffer (which should make it more robust for the concocted situtation where `s_inter.c` has been compiled with a different value for `NET_MAXPACKETSIZE` than `x_net.c`):

~~~C
unsigned char *sys_getrecvbuf(unsigned int*size)
{
    if(size)*size=NET_MAXPACKETSIZE;
    return INTER->i_recvbuf;
}
~~~

it might be overengineering though.

umlaeute (approved):
cool.
less stack allocated memory.

src/s_inter.c
Spacechild1: good point.

src/s_inter.c
Spacechild1: > which should make it more robust for the concocted situtation where s_inter.c has been compiled with a different value for NET_MAXPACKETSIZE than x_net.c):

In that case the build system would be horribly broken.

> what i personally don't like so much is, that the size of the buffer is hardcoded somewhere else (in s_net.h).

I think for internal use this is no problem.

*But* I'm just asking myself whether we should (semi) export the receive buffer, e.g. in `s_net.h? I guess this could be useful for networking externals. In this case we really have to provide the size, because the external might be compiled against a different version of `s_net.h`. What do you think?

Spacechild1:
@millerpuckette I think I have clean this up a lot and the final set of changes is rather small. Note that it is possible to override `NET_MAXPACKETSIZE` at compile time, e. g. for usage on memory constrained systems.

src/s_inter.c
umlaeute: sag ich ja :+1:

Issue #1123, opened 2020-08-14
request: check for Pd updates


porres:
A user on facebook mentioned how he'd think it'd be useful if Pd could check for existing updates.

How hard would it be?

pierreguillot:
My 2 cents: The GitHub API can be used to get all the tags from this repository and if a tag number is higher to the Pd version, a popup window suggests to access the download page of Miller's website. This can be done manually and/or automatically at the startup of the application. And an option can be used to enable or disable the beta versions by checking if the tag name contains "test" in it.

umlaeute:
urgh.
software that phones home without the user's explicit consent should be strictly prohibited in 2020.
software that calls unrelated corporate data kraken without the explicit consent of the user just because they offer a convenient service should be penalized.

apart from the privacy issues, it's just plain annoying if you get a popup at every startup.

its of course better if the user initiates a check.
but then, we could just add a menu entry that opens up miller's page ( or puredata.info, though the former is canonical), and the user can see for themselves...

pierreguillot:
Of course this can be disable by default and the user can enable the it if he wants and the application should warn the user about the web request (perhaps there is something better than GitHub API?). In fact, it can be an add-on and not the main application and perhaps it can be done with a patch. There is no more than 2 or 3 releases per year so except if you only open Pd every 4 months, you should not get a warning at every startup ;)

pierreguillot:
Of course this can be disable by default and the user can enable the it if he wants and the application should warn the user about the web request (perhaps there is something better than GitHub API?). In fact, it can be an add-on and not the main application and perhaps it can be done with a patch. There is no more than 2 or 3 releases per year so except if you only open Pd every 4 months, you should not get a warning at every startup ;)

pierreguillot:
Of course this can be disable by default and the user can enable the it if he wants and the application should warn the user about the web request (perhaps there is something better than GitHub API?). In fact, it can be an add-on and not the main application and perhaps it can be done with a patch. There is no more than 2 or 3 releases per year so except if you only open Pd every 4 months, you should not get a warning at every startup ;)

pierreguillot:
Of course this can be disable by default and the user can enable the it if he wants and the application should warn the user about the web request (perhaps there is something better than GitHub API?). In fact, it can be an add-on and not the main application and perhaps it can be done with a patch. There is no more than 2 or 3 releases per year so except if you only open Pd every 4 months, you should not get a warning at every startup ;)

[...]

PR #1127, opened 2020-08-18
Add "Check for updates" menu

umlaeute:
this opens a webpage https://pdlatest.puredata.info/ which simply redirects to miller's homepage, where the user can see what the newest version of Pd is.

this is trivial from an implementation point of view, as we don't have to do complicated parsing of the http-response (or setup a page that does that for us).
this should also be OK from a privacy point of view, as it doesn't automatically "call home".

i'm using an indirection, as this makes it easier to redirect to some other page, if this is ever needed.

Closes: https://github.com/pure-data/pure-data/issues/1123


aaaaalbert:
Note that the commit message promises the menu to link to an HTTPS site, but the pull request adds a link to the HTTP site instead.

(Ideally, the HTTP 302 would lead to an HTTPS site as well, but I can't seem to reach https://msp.ucsd.edu/ ).

umlaeute:
hmm, well it's more complicated than that:
- clicking on the link opens http://pdlatest.puredata.info
- this redirects to https://pdlatest.puredata.info
- this in turn redirects to http://msp.ucsd.edu/software.html

one miller (or whoever is in charge for his website) enables `https://` the redirect from https://pdlatest.puredata.info would go to that site.

i'm a big fan of https-everywhere, though lately i started to hand-out `http://`-links rather than `https://` (with all my `http://` redirecting to `https://`) - mostly because i think that the actual security is an implementation detail (as long as it is string enough).

changing the actual link to https://pdlatest.puredata.info would of course be trivial.

aaaaalbert:
I understand your concern. I would personally err on the side of more security, as in: less trivial to hijack etc.

Anyhow, my original comment meant to point out the discrepancy between HTTPS in the commit message and HTTP in the patch. It would fit better if either the patch followed the commit message, or the other way around. (I'm not taking sides :-)

umlaeute:
> It would fit better if either the patch followed the commit message

yes. but that's hard to achieve in retrospect.

aaaaalbert:
Not sure I get it. Add a second commit? Force-push with an amended commit message? Create another PR?

umlaeute:
that seems slightly overkill to me.

aaaaalbert:
Sure, just leave as is! Sorry for the noise.

umlaeute:
at least i changed it in the PR-description ;-)

umlaeute:

PR #1128, opened 2020-08-19
Augment _MSC_VER check on stdint include to check _MSC_VER number

mcclure:
In m_pd.h it checks for `_MSC_VER`, and if defined it inlines a tiny fake stdint.h instead of including stdint.h, noting "Visual Studio is not C99".
However, [since VS2015](https://docs.microsoft.com/en-us/previous-versions/hh409293(v=vs.140)#c-runtime-library) many C99 headers including stdint.h have shipped with MSVC.
So for `_MSC_VER` greater than or equal to 1900 ([Visual Studio 2015 / Visual Studio 14.0](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019)) it is safe to simply include stdint.

This would be a good change because it prevents conflicts when you include both m_pd and stdint.h in a file compiled with modern MSVC, which is something I experienced using libpd while testing #1096.

I have not exhaustively tested this patch with different versions of Visual Studio or anything.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -61,8 +61,10 @@ extern int pd_compatibilitylevel;   /* e.g., 43 for pd 0.43 compatibility */
 #include <stddef.h>     /* just for size_t -- how lame! */
 #endif
 
-/* Microsoft Visual Studio is not C99, it does not provide stdint.h */
-#ifdef _MSC_VER
+/* Microsoft Visual Studio is not C99, but since VS2015 has included most C99 headers:
+   https://docs.microsoft.com/en-us/previous-versions/hh409293(v=vs.140)#c-runtime-library
+   These definitions recreate stdint.h types, but only in pre-2015 Visual Studio: */
+#if defined(_MSC_VER) && _MSC_VER < 1900
 typedef signed __int8     int8_t;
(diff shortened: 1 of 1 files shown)


mcclure:
By the way, as a piece of trivia, the MSVC team recently [claimed in a thread on their issue tracker](https://developercommunity.visualstudio.com/content/idea/387315/add-c11-support.html) that they will fully support C11 in the next version of Visual Studio. Their roadmap document doesn't make that promise yet tho.

Spacechild1:
> that they will fully support C11 in the next version of Visual Studio.

That would be surprising, as they don't even fully support C99 yet... I wouldn't be so optimistic and expect C11 support to happen in the next VS version. In the linked thread they only said: "We have added the feature to our roadmap", which of course is vague enough :-)

mcclure:
"That would be surprising, as they don't even fully support C99 yet... "

Well, I'm cluttering up your issues here, but the trick here is that there is a C99 feature that Microsoft really, REALLY hates and does not want to implement, Variable Length Arrays. But C11 made VLAs optional. So Microsoft will never support C99. They will skip directly from C89 to C11, and their C11 compiler will technically not be able to compile C99.

Issue #1110, opened 2020-07-30
Menus to choose MIDI device with not enough items.


jlochard:
Hi, 

In MIDI preferences, I wanted to choose a midi device but it didn't appear in the list of device. It appears that the menu is limited to 20 items, which is not enough for my setup...

All the best,

Jean

Issue #814, opened 2019-12-06
updating GOP shortly after closing it's parent window generates a Tcl/Tk error


umlaeute:
consider a subpatch, that holds a GOP-subpatch with a numberbox.
- open the subpatch, to see the GOP in all it's glory.
- send a value to the GOP-subpatch to change the numberbox
- and *immediately* (within a few milliseconds) close the subpatch (so the GOP becomes invisible)
- :boom: you get a nasty Tcl/Tk error-message, and the GUI starts to malfunction:

~~~
(Tcl) INVALID COMMAND NAME: invalid command name ".x562c48980850.c"
    while executing
"$tkcanvas itemconfig $tag -text [::pdtk_text::unescape $text]"
    (procedure "pdtk_text_set" line 2)
    invoked from within
"pdtk_text_set .x562c48980850.c .x562c48980850.t562c4897f9e0 {666 }"
    ("uplevel" body line 1)
    invoked from within
"uplevel #0 $docmds"
~~~

umlaeute:
the problem seems to be, that the numberbox (in the GOP) uses a queue to update (for throttling updates), and the queue is flushed after the window is closed.
no sanity checks are performed when the queue gets flushed, so we send data to the closed window, leading to the error.

umlaeute:
possible fixes:
- do sanity checks before sending data (in `rtext_senditup`)
- flush the queue before deleting windows.

umlaeute:
turns out that force-flushing the GUI-queue is not trivially done, so we just do the sanity checks

Spacechild1:
> you get a nasty Tcl/Tk error-message, and the GUI starts to malfunction:

I think this is the most troubling part. A bad Tcl message shouldn't break the whole GUI. I've experienced such situations in the past, but could never reproduce it. This bug is the chance to finally fix it!

umlaeute:
forgot to attach a simple patch to trigger the problem (generates an error; gui is not properly malfunctioning thought ;-))

~~~pure-data
#N canvas 723 361 985 599 12;
#X msg 382 361 vis \$1;
#X msg 382 305 1;
#X obj 382 333 t f;
#X obj 382 387 s pd-\$0-oops;
#N canvas 436 316 284 266 \$0-oops 1;
#N canvas 6 61 450 300 bad_GoP 0;
#X floatatom 104 126 10 0 0 0 - #0-numbox -;
#X coords 0 -1 1 1 100 50 1 100 100;
[...]

PR #1129, opened 2020-08-19
stable development branch

umlaeute:
this is meant as a successor of the update/0.51 and similar PRs (where it was always a bit awkward to have the Pd target-version embedded).

the idea is to have a permanent branch develop that only gets curated, small, no-brainer changes that can be easily merged into master at any time.

PS: if anybody can explain how to create PRs that are kept open after merging (or at least: can be opened again after merging), that would be fantastic. (it's a bit boring to re-open this PR after each release ;-))

PPS: supersedes #1105

---

- GUI-stuff
  - #814 (tcl-error when updating a closed GOP)
  - #1152 (some keyboard shortcuts don't work with <kbd>CapsLock</kbd> on
  - #1158 (Tcl error and freezes when using an out-of-range value for setting the color of iemguis)
- audio stuff
  - #1163 (Pd hangs when quitting with a callback scheduler)
  - #1162 (Pd segfaults if jackd has a blocksize that is smaller than `DEFDACBLKSIZE`)
  - #1110 (not enough audio/midi devices displayed in the menu)
- compilation stuff
  - refined check for MSVC supporting `stdint.h`
  - allow building a fresh-clone with the non-autotools makefiles (#1161 , see also https://git.iem.at/pd/pure-data/-/pipelines for automatic builds on a number of OSs/architectures/build-systems)

### DIFF
--- a/msw/build-wxp-32.sh
+++ b/msw/build-wxp-32.sh
@@ -44,3 +44,5 @@ zip -r /tmp/pd-$pdversion.msw.zip  pd-$pdversion
 
 echo /tmp/pd-$pdversion.msw.zip
 echo /tmp/pd-$pdversion.windows-installer.exe
+
+echo wine `pwd`/pd-$pdversion/bin/wish85.exe `pwd`/pd-$pdversion/tcl/pd-gui.tcl
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -305,7 +305,10 @@ static int iemgui_getcolorarg(int index, int argc, t_atom*argv)
     {
         t_symbol*s=atom_getsymbolarg(index, argc, argv);
         if ('#' == s->s_name[0])
-            return (int)strtol(s->s_name+1, 0, 16);
+        {
+            int col = (int)strtol(s->s_name+1, 0, 16);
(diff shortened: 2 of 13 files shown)


Lucarda:
> PS: if anybody can explain how to create PRs that are kept open after merging (or at least: can be opened again after merging), that would be fantastic. (it's a bit boring to re-open this PR after each release ;-))

I just wanted to delete a merged branch and next to the delete button there`s another to create another "pull request".

Also on my branches I have a button to create a new PR.

Both methods untested.

Lucarda:
PS: I forget to mention that this buttons appears on the web interface.

umlaeute:
well, this is what i'm already doing.
i was wondering how to keep the PR open once its merged, not how to create a new one (like *this*)

millerpuckette:
Unless there's a reason not to, I'll just merge this now and call it 0.51-3test - I have a couple of pd~ and expr fixes that should go out soon, and some of these look like they should too.  (And none of them look dangerous at all :)

umlaeute:
i think it's fine.
(after all, the this PR is supposed to always be mergeable as-is)

PR #1145, opened 2020-09-20
Add entries to Info.plist and pd.entitlements for Xcode SDK > 9

sebshader:
It seems that if using Xcode SDK 10 and later, OSX won't prompt for user permission to use the camera and will just terminate the process unless there are certain lines that indicate those resources will be used in the bundle's Info.plist file. I also added a microphone entry. (Using the camera from the Pd bundle is necessary for ofelia, for example)

https://eclecticlight.co/2018/08/27/mojaves-privacy-protection-is-complex-and-will-crash-innocent-apps/
I wasn't sure what to put for the panel's message

### DIFF
--- a/mac/stuff/Info.plist
+++ b/mac/stuff/Info.plist
@@ -2,6 +2,10 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>NSMicrophoneUsageDescription</key>
+	<string>Pd wants to hear the sounds</string>
+	<key>NSCameraUsageDescription</key>
+	<string>Pd wants to see the sights</string>
 	<key>ATSApplicationFontsPath</key>
 	<string>font</string>
 	<key>CFBundleDevelopmentRegion</key>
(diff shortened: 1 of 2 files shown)


danomatika:
What is the "Xcode SDK 10"? DO you mean Xcode *version* 10 or are you referring to the macOS SDK version, ie 10.10, etc?

danomatika:
>I wasn't sure what to put for the panel's message

Just basic info works as long as something is there, ie. "Pure Data needs access to your camera for video input." and "Pure Data needs access to your microphone for audio input."

Also, which version of macOS are you running into this issue, 10.15? I'm stilling running 10.14 which may explain why I've not personally seen this yet, even though Pd does show the microphone access request dialog without the string explicitly set on my system.

sebshader:
@danomatika I mean the command line tools and libraries that are installed along with Xcode 10, I think (I'm just going off of the article I linked). At any rate, putting these lines in these files fixed an issue I was having with permissions:

a new version of Ophelia was released, and I opened a video capture example on a version of pd I had packaged. It crashed. However, the same examples in vanilla pd did not. After some digging I found that it was a permissions issue, and that when building with newer Xcode versions the app just crashes instead of asking for permission unless the plist file specifies that the app needs microphone/camera access.
I'm on 10.14.6 (Mojave) with Xcode 11.3.1

danomatika:
I'm going to merge this into the develop branch and adjust a few things.

PR #1170, opened 2020-10-01
improve symbol table performance

Spacechild1:
Pd's symbol table has a constant size (1024), which means that performance degrades and approaches O(N) if many symbols are added.

The usual solution is to grow and rehash the symbol table after it has exceeded a certain load factor (`N elements / N buckets`) to keep it O(1).

0.75 is a common value for the maximum load factor. I've found out that Pd already creates ~500 symbols on startup (without any externals), which corresponds to a load factor of ~0.5. Add 250 more symbols and you've already approached the recommended limit.

I did some benchmarks with the following test patch: [symtab-test.zip](https://github.com/pure-data/pure-data/files/5313375/symtab-test.zip)

With this PR, symbol insertion and lookup is consistently faster, but the difference becomes really apparent after repeatedly inserting 100 000 random symbols:
| N | old (ms) | new (ms) |
| - | --------- | --------- |
| 1 | ~200 | ~60     |
| ... |          |             |
| 10 | ~8000 | ~60 |

Of course, rehashing is not free. However, it only happens relatively rarely - in fact, it becomes less rare with increasing table size!
Here's some benchmark for the additional cost of rehashing:

| old size | time (ms) |
| --------| -------- |
| 1024   | 0.042286 |
| 2048   | 0.061581   |
| 4096   | 0.183102   |
| 8192   | 0.413827   |
| 16384 | 1.021430 |
| 32768 | 1.442237 |
| 65536 | 2.822481 |
| 131072 | 9.373512 |
| 262144 | 23.967074
| 524288 | 57.996962 |

You can see that rehashing is pretty much O(N). Rehashing a large symbol table naturally takes some time and might lead to audio dropouts. From the point of realtime safety, this is bad. On the other hand, rehashing occurs very rarely and the rest of the time you get significantly better performance.

Another benchmark:

Time needed to add 10 random symbols with `[makefilename %d]` for specific symbol table sizes:
| table size | old (ms) | new (ms) |
| --------- | -------- | ---------- |
| 100 000 | 0.1         |  0.012      |
| 200 000 | 0.2         |  0.012      |
| 300 000 | 0.3         |   0.012      |
| 400 000 | 0.4         |  0.012      |
| 500 000 | 0.5         |  0.012      |

It is pretty obvious how the new implementation stays O(1), while the old implementation degrades to O(N). Actually, 0.5 ms for 10 random symbols is pretty terrible, that's 1/3 of the time budget for a 64 sample block!

---

One thing I'm not sure about:

Currently, I store the symbol table size and item count in `t_pdinstance`. I don't really like this because they shouldn't be exposed in the public API. I've thought about storing it in `t_instancestuff`, but that would mean a potential cache miss. The Pd instance, on the other hand, is always in the cache (because we already access `pd_symhash`). I guess that's also the reason why the symbol table itself resides in `t_pdinstance`, although it has no business being exposed in `m_pd.h`...

### DIFF
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -23,6 +23,11 @@
 #define snprintf _snprintf
 #endif
 
+    /* initial symbol table size */
+#ifndef SYMTABHASHSIZE
+#define SYMTABHASHSIZE 1024
+#endif
+
 static t_symbol *class_loadsym;     /* name under which an extern is invoked */
 static void pd_defaultfloat(t_pd *x, t_float f);
 static void pd_defaultlist(t_pd *x, t_symbol *s, int argc, t_atom *argv);
@@ -74,9 +79,12 @@ static t_pdinstance *pdinstance_init(t_pdinstance *x)
     x->pd_clock_setlist = 0;
     x->pd_canvaslist = 0;
     x->pd_templatelist = 0;
+        /* init symbol table */
     x->pd_symhash = getbytes(SYMTABHASHSIZE * sizeof(*x->pd_symhash));
     for (i = 0; i < SYMTABHASHSIZE; i++)
         x->pd_symhash[i] = 0;
+    x->pd_symhashsize = SYMTABHASHSIZE;
+    x->pd_symhashcount = 0;
 #ifdef PDINSTANCE
     dogensym("pointer",   &x->pd_s_pointer,  x);
     dogensym("float",     &x->pd_s_float,    x);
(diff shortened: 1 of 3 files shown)


Spacechild1:
Also, we could start with a larger table size, like 4096 or 8192, to avoid the first couple of rehashings. As I've mentioned, Pd already *starts* with more than 500 symbols. What do you think?

millerpuckette:
I've been thinking abut this for years... I think 500 symbols out of about 1500 (which would start to load the table down) is a reasonable fraction but at the same time I think 124 is a bit small for these days.  At the same time I try to keep Pd's footprint down to something that will fit in an Arduino-sized system.  On the whole I'd support a "reasonable" increase in default size and/or make the default symtab size easily reconfigurable at compile time.

I wouldn't want ever to have Pd stop to rehash - that creates a CPU spike.  OTOH one could leave the symtab intact, but once it exceeds a certain fill level just start another bigger one, ad nauseam.

I'd put the  symbol table stuff in a local struct to keep things out of the pd instance structure.

Spacechild1:
> On the whole I'd support a "reasonable" increase in default size and/or make the default symtab size easily reconfigurable at compile time.

Sounds good!

> OTOH one could leave the symtab intact, but once it exceeds a certain fill level just start another bigger one, ad nauseam.

That's not a bad idea, actually... With a more aggressive grow factor this could work (otherwise we still end up with long linked lists). However, there will always be *some* point, where allocating a new symbol table (without rehashing) will cause a CPU spike (a table of 100 000 elements certainly will), but at least it will happen only in extreme cases.

I can make another PR and benchmark against this one.

umlaeute:
> I wouldn't want ever to have Pd stop to rehash - that creates a CPU spike.

the problem (obviously) comes from the possibility, that we run out of run free slots during runtime.
this is similar to resizing an array (e.g. because you load a larger soundfile) during runtime.
the solution for the latter is to simply make sure (at init time) that the arrays are big enough, so they don't need resizing during runtime.

could we add something like that for the symboltable as well?

~~~
[; pd symboltablesize 65536(
~~~

Ant1r:
(should we continue to discuss here? #1172 seems having only advantages over this one)

I agree we can hardly take the risk of rehashing while audio is being performed.

We also could manually ask Pd to rehash if needed, like:
`[; pd rehash(`
I was thinking that rehash-if-needed could be automatically triggered each time DSP is shut off.

Also: the initial symbol table size could also be settable via a new startup flag (e.g` -symboltablesize 65536`); and personally I would be fine with a default initial size of 65536...

Whatever: I think (as a patch writer) it would be great if we could query the current number of symbols, for debugging purpose, e.g with [pdcontrol].

umlaeute:
> We also could manually ask Pd to rehash if needed, like:
> `[; pd rehash(`
> I was thinking that rehash-if-needed could be automatically triggered each time DSP is shut off.

ah well.
but if you know that you are going to need 100000 symbols, that doesn't really help you much.
the symbol generation will become slower and slower over time, and once it is slow enough you must manually trigger a re-hash so that it gets fast again.
if you can specify the hashsize beforehand, you would never get slow in the fist place.

Spacechild1:
Please see https://github.com/pure-data/pure-data/pull/1172 for a solution that doesn't require rehashing. The symbol table expands seemlessly, so there shouldn't ever be a need to manually specifiy the symbol table size, let alone do a manual rehash!

I think that the symbol table is an implementation detail that shouldn't be exposed in the public API. It should work well in all cases without requiring the user to know how it works internally.

> I was thinking that rehash-if-needed could be automatically triggered each time DSP is shut off.

Actually, that brings me to an idea for my other implementation.

I'm closing this PR in favor of https://github.com/pure-data/pure-data/pull/1172

Spacechild1:
> Actually, that brings me to an idea for my other implementation

Done.

PR #1172, opened 2020-10-01
improve symbol table performance (version 2)

Spacechild1:
This is an alternative version of https://github.com/pure-data/pure-data/pull/1170. The performance is only a little bit worse, but not significantly.

Since @millerpuckette had concerns regarding the rehashing, this PR uses a hybrid approach for the case where the (last) table exceeds the max. load factor (`N elements / N buckets`)

* if DSP is on, we *add* a new, larger table. We *don't* rehash to avoid audio dropouts
* if DSP is off, we create a single new, larger table and insert all existing symbols (= rehashing). We also do this whenever DSP is switched off and there is more than 1 symbol table.

Here's a benchmark of the symbol table creation routine:

| table size | time (ms) |
| --------- | ---------- |
| 16384 | 0.023712 |
| 65536 | 0.094442 |
| 262144 | 0.343732 |
| 1048576 | 1.207408 |

I think those numbers are acceptable. If someone manages to create 1 million symbols, they should cope with a single audio dropout :-). (And by that time, the old symbol table would be bog slow anyway). With the current settings (initial size = 4096, grow factor = 2), I expect most Pd projects to only grow the symbol table 1-2 times, which takes 0.01 ms resp. 0.02 ms.

NOTE: `DEFSYMTABSIZE`, `SYMTABMAXLOAD` and `SYMTABGROW` can all be overriden at compile time, e.g. for memory constrained environments.

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -1266,6 +1266,8 @@ static void canvas_start_dsp(void)
         pd_bang(gensym("pd-dsp-started")->s_thing);
 }
 
+void symtab_rehash(void);
+
 static void canvas_stop_dsp(void)
 {
     if (THISGUI->i_dspstate)
@@ -1276,6 +1278,8 @@ static void canvas_stop_dsp(void)
         if (gensym("pd-dsp-stopped")->s_thing)
             pd_bang(gensym("pd-dsp-stopped")->s_thing);
     }
+        /* use the opportunity to clean up the symbol table */
+    symtab_rehash();
 }
 
     /* DSP can be suspended before, and resumed after, operations which
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -868,6 +868,11 @@ static void symhash_init(t_symhash *x, int size)
 #endif
 }
 
+static void symhash_free(t_symhash *x)
+{
+    freebytes(x->sh_tab, x->sh_size * sizeof(t_symbol *));
+}
+
 static t_symtab * symtab_new(void)
 {
     t_symtab *x = (t_symtab *)getbytes(sizeof(t_symtab));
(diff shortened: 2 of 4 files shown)


millerpuckette:
I don't think DSP on/off is a meaningful test of real-time-criticalness... I'd suggest doing the real-time-friendly thing in all cases.  And why not expand the symtab by a factor of 4 or even more instead of 2?  Seems like the memory hit would be minor compared to the performance gain.

I confess I'm partial to Iohannes's idea of making it a manual decision - that saves a lot of complexity.  The only reason I haven't done this 20 years ago is I wanted the code to stay simple and easy to understand.

Spacechild1:
> why not expand the symtab by a factor of 4 or even more instead of 2

I've now reverted back to a grow factor of 4

Spacechild1:
> I'd suggest doing the real-time-friendly thing in all cases.

That's fine. I'd just like to point out that with a grow factor of 4, rehashing happens *very* rarely. But if you don't like the rehashing, you can simply merge up to e604374 :-)

> I confess I'm partial to Iohannes's idea of making it a manual decision - that saves a lot of complexity

I don't see how that should work. 99% of Pd users don't even know what a symbol table is. Also, it's hard to estimate how many symbols a patch actually needs. I think that this is an implementation detail that should be hidden from users (except maybe for power users). My point is that Pd's symbol table can slow down drastically once you create a huge amount of symbols and there should be a way to automatically mitigate this situation. With a grow factor of 4 and an initial table size of 4096 (or probably larger), the symbol table expansion happens so rarely that it shouldn't be noticable. Most patches won't need to expand at all. This is really about edge cases.

> I wanted the code to stay simple and easy to understand.

Hmmm... I think the code is quite easy to understand. And it has comments ;-)

Anyway, my benchmarks show that there is a serious problem with the symbol table and we have to find a solution. This PR is just one of many ways. Actually, there is a better ways to seemlessly grow the symbol table (https://en.wikipedia.org/wiki/Hash_table#Incremental_resizing), but it's a bit more complex, so I'm sure you would like it even less :-p

Issue #1101, opened 2020-07-28
symbol box backspace doesn't work well for non ASCII characters


porres:
Some characters take two bytes or more and Pd handles them nicely in comments, messages and object boxes. But in atom boxes if you try to delete one of these "special" characters you remove one of the bytes and not the character itself. Hence, trying to delete "ç" turns it into "Ã".

The same mechanism used in the other boxes should be used.

porres:
> The same mechanism used in the other boxes should be used.

And what's that mechanism? 

It's the "u8_dec" function in https://github.com/pure-data/pure-data/blob/a92ad87194a3a82998acc7af7f0d4cb3b4460d57/src/g_rtext.c#L530 I believe

danomatika:
Actually, sending characters when editing a float or symbol atom *does not* go through that code section, which is probably why there is an issue.

danomatika:
This should be an easy fix. Looking into it now.

danomatika:
@porres Here's a build with this fix and the key handling updates: http://docs.danomatika.com/pdbuilds/0.51/Pd-0.51-3-keyfix2.zip

porres:
Works perfectly, it fixes this and https://github.com/pure-data/pure-data/issues/1138 !

thanks!

porres:
But I'm testing 0.51-3-test-3 and this is not working there yet

danomatika:
This fix is not in test3. Bug @millerpuckette.

porres:
Oh, I see the fix is in  #1195 not merged yet. Please @millerpuckette merge this one and I think 0.51-3 s ready to go ;)

umlaeute:
the fix is included in #1178 ("stable development branch") which I would prefer to get merged ;-)

PR #1192, opened 2020-10-25
[WIP] macOS Tk 8.6.10 key handling fixes

danomatika:
This PR brings fixes for the following issues with TK 8.6.10:

1. ports the macOS key release handling patch from 8.5.9 to 8.6.10 which fixes key release and key repeat\*
2. adds event handling for the character composition events, ie. creating a ü by pressing opt+u then u

With composition, an interim character is usually shown, then replaced by the final character after the composition succeeds. For ü, ¨ is shown after the opt+u until the second u is pressed. Pressing escape during composition will leave the interim character. For the Pd GUI, the easiest solution might be to send a virtual/fake BackSpace to the Pd core to remove the interim character.

Tk 8.6 has added events for when composition starts, ends, or is cancelled which I found y looking in `library.entry.tcl` to see how the entry widget handles this. I added bindings for these to the Pd patch canvas and send BackSpaces when composition ends. This seems to work Ok, however the events refer to start and end to "marked text" which makes me wonder if we need to handle some sort of selection range and/or multiple-width characters. Maybe not? I'm not 100% sure on this yet.

One current issue is that symbol atoms are not processing the backspace and the interim characters are left and/or the last character is corrupted. Object and message boxes work fine, so I'm not sure why the symbol atom buffer is handled differently here.
 
TODOs:
* [ ] fix symbol atom issue
* [ ] test on Windows & Linux, or at least those systems which implement composition handling

\* _You will need to (re)build a Tcl/Tk 8.6.10 Wish app to include the patch and use this to build the .app bundle in order to see the key release fixes._

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -217,6 +217,7 @@ compiler:
 
 * macOS 10.6: i386, x86_64, ppc
 * macOS 10.7+: i386, x86_64
+* macOS 10.15: x86_64
 
 Note: a "fat" Pd may not work on all systems and/or be able to load both 32 or
 64 bit externals. Additonally, you can specify multiple architectures directly:
--- a/mac/README.txt
+++ b/mac/README.txt
@@ -61,9 +61,9 @@ argument, a "Pd.app" is built. The version argument is only used as a suffix to
 the file name and contextual version info is pulled from configure script
 output.
 
-A pre-built universal (32/64 bit) Tk 8.5.19 Wish with patches applied is
+A pre-built universal (32/64 bit) Tk 8.6.10 Wish with patches applied is
 included with the Pd source distribution and works across the majority of macOS
-versions up to 10.14. This is the default Wish.app when using osx-app.sh. If you
+versions up to 10.15. This is the default Wish.app when using osx-app.sh. If you
 want to use a different Wish.app (a newer version, a custom build, a system
 version), you can specify the donor via commandline options, for example:
 
--- a/mac/osx-app.sh
+++ b/mac/osx-app.sh
@@ -40,7 +40,7 @@ Usage: osx-app.sh [OPTIONS] [VERSION]
 
   Creates a Pd .app bundle for macOS using a Tk Wish.app wrapper
 
-  Uses the included Tk 8.4 Wish.app at mac/stuff/wish-shell.tgz by default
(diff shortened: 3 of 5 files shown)


porres:
do you have a mac build so I can test?

danomatika:
It's the same one you've already tested.

danomatika:
@Spacechild1 I know this is marked "macOS" but can you try the pd_bindings.tcl update on Windows? The IMEText events should fire there is the compose mechanism is used, as far as I can tell from the Tk sources.

Spacechild1:
For everything looks alright (except for the backspace bug in the symbol atom). But I'm using a German keyboard layout. I tried some Umläute and French accents, but maybe @Lucarda can also test?

Lucarda:
For me on a Spanish (Spain) keyboard layout all is good except the ¨backspace bug in the symbol atom¨.

Win10 Pd **0.51.2** with `pd_bindings.tcl` from `bugfix/tk8610-keyfix`

danomatika:
@millerpuckette You will need to remove the tempo posts and stub events in pd_bindings.tcl.

danomatika:
Nevermind. I pushed a commit for this. Do a pull, then you can cherry-pick b5aacba9db549de6f3a0c67e91f11ceeb1cdb2e1 into `master`.

millerpuckette:
oops... just 'committed" test2.... will grab this and re-build...

M

danomatika:
@millerpuckette And I hav a fix for #1101 which would be good to go with this.

millerpuckette:
Hmm... I can't find b5aacba9db549de6f3a0c67e91f11ceeb1cdb2e1 - perhaps because
the PR closed after I merged it...

It's not GIT, but can you just e-mail me your commits?

thanks
M

danomatika:
I see it on Github: https://github.com/pure-data/pure-data/commit/b5aacba9db549de6f3a0c67e91f11ceeb1cdb2e1

 Maybe you need to pull the branch itself: `bugfix/tk8610-keyfix`

millerpuckette:
OK, that worked (I must have been on the wrong PR or something.

Now to mess with test3... :)

M

danomatika:
If you';re quick, you can always delete the tag on the server helpfully before anyone else pulled it... :)

PR #1195, opened 2020-10-27
fixed symbol atoms not handling multi-byte chars on backspace

danomatika:
This PR is a fix for #1101 by walking back to the previous unicode code point which could multiple-characters.


umlaeute:
merged into `develop` branch (#1178)

PR #1196, opened 2020-10-28
html updates

Lucarda:
updates to the html documentation

### DIFF
--- a/doc/1.manual/1.introduction.txt
+++ b/doc/1.manual/1.introduction.txt
@@ -12,9 +12,8 @@ Reference documentation for Pd lives in:
 or online:
     http://msp.ucsd.edu/Pd_documentation/
 
-More extensive FLOSS documentation is available on:
-    https://en.flossmanuals.net/pure-data/_full/ (English)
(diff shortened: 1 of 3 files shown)


doc/1.manual/1.introduction.txt
```diff
@@ -13,7 +13,7 @@ or online:
     http://msp.ucsd.edu/Pd_documentation/
 
 More extensive FLOSS documentation is available on:
-    http://en.flossmanuals.net/PureData (English)
+    https://en.flossmanuals.net/pure-data/_full/ (English)
     http://fr.flossmanuals.net/PureData (Francais)
```
umlaeute, line 17: that should be https:// as well.

(and really: the floss-manuals are horribly outdated)

Lucarda: > that should be https:// as well.

You mean the French one?

> (and really: the floss-manuals are horribly outdated)

I agree. We pull them off here?

Lucarda: How about:

```
More extensive documentation is available on:
    http://puredata.info/docs/BooksAboutPd/
```

PR #1198, opened 2020-10-28
stable development branch

umlaeute:

supersedes #1178 

---

- updated translations (fr,de,es)
- minor documentation updates (typos,...)
- minor code cleanups (to shut up pernickety compilers)
- macOS permissions ("entitlements", #1145)


Lucarda:
I wonder why:


```
#, tcl-format
msgid ""
"Do you want Pd to use the existing documents directory for patches and "
"external libraries?\n"
"\n"
"Location: %s\n"
"\n"
"You can change or disable this later in the Path preferences."
msgstr ""
```


```
msgid "removed GUI settings"
msgstr ""

```
```
msgid "no Pd-GUI settings to clear"
msgstr ""
```

didn't get included in `es.po`

how can we fix this?

umlaeute:
@Lucarda i didn't include it, because gettext tried to be clever and suggested similar - but wrong - translations (e.g. "removed GUI settings" would be translated as "Archivo .pdsettings eliminado").

rather than committing the wrong translations, i decided it would be better to not have any translations (and i just didn't feel like going through es, eu, hu, it, pt_br, pt_pt translations and removing the wrong translations; so i just left them out).

the proper way to proceed is: someone who speaks one of those languages should do the following:

1. update the PO-files for their language:

~~~
touch po/template.pot
make -C po
~~~

2. start `poedit` (or whatever they use) and fix the translations for their language.
3. submit a PR for the updated translation.

Lucarda:
Ok. I'll try.

PR #1204, opened 2020-11-04
stable development branch

umlaeute:

supersedes #1198 

---

- macOS permissions again ("entitlements", #1145)
- fixing spelling mistakes in documentation and code (comments)


danomatika:
Latest commit brings in a couple small portaudio updates from upstream:

* fixes [deprecated API usage going back to 10.8](https://github.com/PortAudio/portaudio/pull/338), this is a warning that is now an error when building on macOS 11.0
* adds support to [disable loading ALSA plugins via an env var](https://github.com/PortAudio/portaudio/pull/312)

PR #1215, opened 2020-11-07
fix sys_audiodevnumbertoname() using dynamic memory allocation

pierreguillot:
This PR is related to issue reported on libpd: https://github.com/libpd/libpd/issues/312.
For reason that I don't fully understand the size of `outdevlist[]` seems too big (`MAXNDEV*DEVDESCSIZE` with `MAXNDEV`is 64 and `DEVDESCSIZE`is 1024 and I set `MAXNDEV` to 1 and the problem was fixed but obviously was not a good solution). This seems to be related to the use of pd in a mutli-instance and multi-thread context. The best solution seemed to use dynamic allocation.

### DIFF
--- a/src/s_audio.c
+++ b/src/s_audio.c
@@ -27,7 +27,7 @@ typedef long t_pa_sample;
 #define SYS_XFERSAMPS (SYS_DEFAULTCH*DEFDACBLKSIZE)
 #define SYS_XFERSIZE (SYS_SAMPLEWIDTH * SYS_XFERSAMPS)
 #define MAXNDEV 128
-#define DEVDESCSIZE 1024
+#define DEVDESCSIZE 100
 
 static void audio_getdevs(char *indevlist, int *nindevs,
     char *outdevlist, int *noutdevs, int *canmulti, int *cancallback,


danomatika:
@millerpuckette This fixes a crash when using multi instances with threads, ie. the libpd C `pdtest_thread` example.

millerpuckette:
What if we just set DEVDESCSIZE to 100 instead of 1024?

umlaeute:
@millerpuckette the value used to be "*80*" until 2017. until somebody changed it to "1024" in order to work around a bug with garbage being displayed on windows.

the relevant commit is 5a87929fff8003154c00ae5bbdcce85e3a7f13fc

so it seems that changing it back to a low value like 100 is just going to introduce a regression.

Spacechild1:
Reading the commit message, this had to do with (incorrectly) using `sprintf_s`. But this function is not used (anymore) in Pd (except in `pd~.c`, but I will fix that). In fact, `pa_getdevs` and others all use `snprintf` where necessary.

So I think, going back to a lower value should be fine. If there is still a memory bug, we should fix and not throw enough memory at it until it magically disappears ;-)

Anyway, `glob_audio_properties` currently has 3 x 100kB buffers sitting on the stack. With changing `DEVDESCSIZE` back to 100, that would be 3 x 10kB, so heap allocation might still be more appropriate. After all, this function is not supposed to be realtime safe anyway.

Spacechild1:
Not really related, but I noticed that `buf` in `glob_audio_properties` is way too big:

`char buf[1024 + 2 * MAXNDEV*(DEVDESCSIZE+4)];`

This is how `buf` is actually populated:

```
sprintf(buf,
"pdtk_audio_dialog %%s \
%d %d %d %d %d %d %d %d \
%d %d %d %d %d %d %d %d \
%d %d %d %d %d %d\n",
        audioindev1, audioindev2, audioindev3, audioindev4,
        audioinchan1, audioinchan2, audioinchan3, audioinchan4,
        audiooutdev1, audiooutdev2, audiooutdev3, audiooutdev4,
        audiooutchan1, audiooutchan2, audiooutchan3, audiooutchan4,
        rate, advance, canmulti, (cancallback ? callback : -1),
        (flongform != 0), blocksize);
```

pierreguillot:
If you think it's better to fix the use of `sprintf_s`, I can do a test this weekend with DEVDESCSIZE to 100.

Spacechild1:
As I've written, there are no remaining uses of `sprintf_s`, so you can already try with DEVDESCSIZE set to 100. If you still get crashes, let's try to fix the actual bug :-)

pierreguillot:
It works, I updated the branch (force push).

danomatika:
OK sweet. If this looks good, we can merge it into the `develop` branch.

umlaeute:
hmm, well.
as @Spacechild1 said (well he didn't; i'm just reading that into what he wrote), just changing some magic number until issues go away is not a real "fix". shouldn't we find the underlying problem (or at least have an idea what the underlying problem is)?

pierreguillot:
If I understand well, there's a stack overflow. I don't really understand why it happens in a multithreaded context and not in a "normal" context. Anyway, if this is a stack overflow, using dynamic allocation or reducing the size (as it's possible) seem to be a good solution but I can be wrong. In this case I don't think I can help more :( but feel free to edit and fix the PR.

danomatika:
I can also say that running libpd `samples/c/pdtest_thread` no longer crashes with this change and #1216.

Spacechild1:
@umlaeute 

> as @Spacechild1 said (well he didn't; i'm just reading that into what he wrote), just changing some magic number until issues go away is not a real "fix". 

This was meant towards 5a87929 which *increased* `DEVDESCSIZE` to avoid a memory error (most likely buffer overrun).

> shouldn't we find the underlying problem (or at least have an idea what the underlying problem is)?

I'm pretty sure the crash in libpd/libpd#312 is caused by a stack overflow:
1) there are huge buffers on the stack
2) the crash goes away be drastically reducing the size

---

Of course, the best fix would be to use a `std::vector<std::string>` :-p

Spacechild1:
@pierreguillot 

>  I don't really understand why it happens in a multithreaded context and not in a "normal" context.

that's a good question... I've read that certain runtimes use a different default stack size for the main thread and secondary threads, e.g. https://developer.apple.com/library/archive/qa/qa1419/_index.html

danomatika:
Fixed via 0022b548b56ad1d907d83531dae4c0e64f0652b8

PR #1216, opened 2020-11-07
fix thread-safety of pdinstance_free()

pierreguillot:
I explain the fix on https://github.com/libpd/libpd/issues/271.


danomatika:
@millerpuckette this fixes a possible crash when libpd sets the current instance.

PR #1222, opened 2020-11-15
macOS default font Menlo (for now)

danomatika:
This PR changes the default font for macOS to Menlo. This is a hack until the render issue (#988) with DejaVu Sans Mono can be fixed on macOS 10.15+.

This only sets a new default font family name. DejaVu Sans Mono is still included in the app bundle for now which should help with testing once we have a fix for the rendering issue. Documentation on this change is included in `font/README.txt` and system-specific `mac/README.txt`.

### DIFF
--- a/font/README.txt
+++ b/font/README.txt
@@ -18,6 +18,13 @@ The default font weight on Linux is bold.
 
 ## macOS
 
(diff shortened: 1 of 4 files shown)


millerpuckette:
Works!  (on 10.7 and I presume on 10.15 too).  Time to release this...

Issue #1183, opened 2020-10-15
Terminal's font smaller in 0.51-3 test 1 (macOS Mojave)


porres:
Here's a screenshot, left is 0.51-2, right is 0.51-3 test 1

<img width="1028" alt="Screen Shot 2020-10-15 at 12 54 06" src="https://user-images.githubusercontent.com/5360142/96154873-a85e8b00-0ee5-11eb-9806-2c425799b431.png">

Anyway, feels pretty small to me, makes me reach for my glasses :)

umlaeute:
why don't you just change the font to whatever you want it to be?

porres:
It seems like a bug because new objects are still be created in the font size = 12, the new default. But I see now the font size somehow is defined as "10", even though we create new objects as "12", so yeah, I dunno, something looks werd.

porres:
I have to close this as I can't reproduce this anymore. I tried a fresh install for 0.51-2 and then opened a newly downloaded 0.51-3 test 1 and nothing really happened!

I guess the issue may have happened somehow when I was compiling Pd before the test release was available. I noticed the font was smaller/funny but thought that the release would be distributed correctly and was probably some compiling issue (like not carrying the correct tcl/tk version or something, I dunno).

umlaeute:
so what exactly is the bug?
- that the default font-size of new objects is not the same as the default font-size of the menu?
but hey, even the *font* (family) is different, so what is the point?
(DejaVuSans looks nice in the console, but a bit weird in the menus;
otoh, i would *love* to have a fixed-width font in both the patches and the console; but i guess the DejaVu lobby and the pixel geeks would see to prevent that)

- or is it, that the font-size of the menu is different compared to the menu font-size in the last iteration of Pd (as your title suggests)?
because for me, it is the other way round (and i'm actually glad about that).
![pdconsole](https://user-images.githubusercontent.com/214034/96183465-32830f80-0f37-11eb-82be-6967bca39f89.png)

umlaeute:
also, the underlying reason for the change is, to make the font *much* bigger (and actually readable) on modern displays (retina, hidpi,...)

porres:
> or is it, that the font-size of the menu is different compared to the menu font-size in the last iteration of Pd (as your title suggests)?

yeah, it was smaller

> because for me, it is the other way round

like, bigger, right? Weird.

Well, like I said, now for me is the same... so... I'm really lost here... 

And I wouldn't mind that it'd look bigger I guess

[...]

PR #1223, opened 2020-11-20
html updates

Lucarda:
- Include notes from issue #1183


danomatika:
@umlaeute We could merge this into `develop`.

Issue #1211, opened 2020-11-05
[writesf~] + invalid file path locks up Pd


Spacechild1:
If you pass an invalid file path to `[writesf~]`, e.g. `[open /tmp/foo.wav(` on Windows, and start recording, Pd will become unresponsive.

More specifically, it the console gets flooded with messages like:
```
writesf waiting for disk write..
(head 262016, tail 0, room 128, want 128)
... done waiting.
```

umlaeute:
i can confirm that this also happens on Linux (where `/tmp/foo.wav` is a legal path): `[open /foo.wav(` (as non-root, so no write-permissions on `/`) locks Pd-0.51-2 while flooding the console.

just for the record: there was a report of a somewhat similar issue on [StackOverflow](https://stackoverflow.com/questions/64510543).

mxa:
Maybe it would be wise to 

1. make Pd not lock up when the directory is write protected but give explicit user feedback about the problem
2. include that information in the help files of objects which are writing to disk, and - if possible - use a path in the examples which is writable on all systems (if such exists).

Spacechild1:
> make Pd not lock up when the directory is write protected

It's not about write protection, it's about `write` failing (for whatever reason).

> include that information in the help files

Which information? That writing fails if the path doesn't exist or you don't have sufficient permissions? I thought people would already know that :-p

umlaeute:
> Maybe it would be wise to
> 
>     1. make Pd not lock up when the directory is write protected but give explicit user feedback about the problem

apart from what @Spacechild1 already mentioned, i would also like to add that:
- we don't do explicit checks whether the target directory is writable. we don't even split the target into a "directory" and "basename" component. we just pass "everything" to the OS.
- even if we did perform a is-directory-writable check (and this made sense - which it doesn't necessarily do as @Spacechild1 pointed out), we would still be susceptible to race conditions.

> 
>     2. include that information in the help files of objects which are writing to disk, and - if possible - use a path in the examples which is writable on all systems (if such exists).

unfortunately such path does not exist.
(afaict, there's a common concept of user-writable directories (their home-directory; a tempdir) on all platforms; but they all look very different)

Spacechild1:
> use a path in the examples which is writable on all systems (if such exists).

That is definitely something to consider, but as @umlaeute said, such path does not exist. Two possible solutions:
1) use bang + `[openpanel]` instead of a hardcoded path
[...]

PR #1224, opened 2020-11-20
writesf~: fix unfinite loop on I/O error

Spacechild1:
Make sure that Pd doesn't get stuck in an infinite loop if `open` or `write` fails in `writesf_child_main`. Also print error message.

The code basically follows the one in `[readsf~]`, with slight adaptions where necessary.

Fixes https://github.com/pure-data/pure-data/issues/1211

NOTE: this has been incorporated into #855

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -2413,8 +2413,7 @@ static void *writesf_child_main(void *zz)
                 pute("open failed\n");
                 pute(filename);
 #endif
-                x->x_requestcode = REQUEST_NOTHING;
-                continue;
+                goto bail;
             }
             /* check if another request has been made; if so, field it */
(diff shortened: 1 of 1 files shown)


danomatika:
I can also incorporate this into the `update/soundfile` branch.

danomatika:
I've added this to #855, thanks! One note: shouldn't you use `sys_close()` for consistency?

Spacechild1:
> shouldn't you use sys_close() for consistency?

Probably :-). I just copy-pasted from `readsf~`. Feel free to replace it.

On the other hand, there are quite a few `close()` calls scattered across the whole codebase, so maybe that could be a seperate PR to clean those up?

danomatika:
>so maybe that could be a seperate PR to clean those up?

#855 :)

PR #1226, opened 2020-11-21
Fix disabled Help menu items on macOS

danomatika:
This PR adds a patch for Tk 8.6.10 on macOS which fixes items in the Help menu being disabled after windows are cycled and not re-enabled. This is basically a fresh bugfix back port from upstream Tk development and is an alternate to #991.

Also, the `mac/patches/README.txt` has been updated with info on what each patch does.

### DIFF
--- a/mac/patches/README.txt
+++ b/mac/patches/README.txt
@@ -40,6 +40,7 @@ Fixes key release events and adds repeat handling, see
 Fixes help menu items disabling after switching windows, see
 
 	https://core.tcl-lang.org/tk/info/ad7edbfafdda63e6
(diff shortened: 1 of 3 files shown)


danomatika:
@sebshader Commit 9eb4c9a keeps the Tcl reenable code but reframes it. I think it's fine to leave it for now in case people use a different Tk, ie. installed via homebrew, but maybe I'm wrong about that and its only 1 person.

Actually, I found that I had put the original hack in place in f1ddb580e97907b189c0864a9833e23fd24efbd1. It must have been a quick thing in the middle of some UI work as I don't remember doing it.

PR #1228, opened 2020-11-21
Update/soundfile

Spacechild1:
* fix bound checking + redundant/wrong `strncat` usage in `*_addextension` methods (found by a GCC warning)
* small improvements + fix some race conditions in readsf~/writesf~ child threads (passes my tests, but please double check!)


danomatika:
Couldn't the filename buffer be overrun when using `strcpy` if the filename is absurdly long?

Spacechild1:
We already check the size with `if (len + 4 >= size)`, so we know that there's enough space.

danomatika:
Ok, looks good to me then.

danomatika (approved):

Issue #1229, opened 2020-11-24
[net objects] problem with Android


Ant1r:
Latest version of network objects `[netsend]` and `[netreceive]` don't work in Android (at least for me).
They signal the error `Invalid value for ai_flags`.

I discovered that removing the setting of `AI_ALL` and `AI_V4MAPPED` flags in:
https://github.com/pure-data/pure-data/blob/c8af32521e9c723bfb2c83fbdd798b30461876f7/src/s_net.c#L55-L65
was doing the job: the error is cleared and my patches work properly.

These flags seem not to be allowed for this platform, see for example https://stackoverflow.com/questions/39674147/getaddrinfo-on-android-returning-error-eai-badflags .

Not sure how it should be cleanly fixed however.

Spacechild1:
Thanks for the report!

`The getaddrinfo function as implemented in	FreeBSD	currently does not
     support AI_ALL and	AI_V4MAPPED flags and returns EAI_BADFLAGS if one of
     them is specified.`

https://www.freebsd.org/cgi/man.cgi?query=getaddrinfo&sektion=3&manpath=FreeBSD+7.2-RELEASE

*Ugh*

So how can you send to an IPv4 address from an IPv6 socket if you can't obtain IPv4 mapped addresses!?

One thing we could do is catch `EAI_BADFLAGS` and simply default to IPv4...

@danomatika Any other ideas?

danomatika:
>One thing we could do is catch EAI_BADFLAGS and simply default to IPv4...

Do this in a FREEBSD define as it's clearly a platform-specific limitation for now.

danomatika:
Same for Android, if that's how the platform works as well. If they both do that in the same way, then maybe just do without without the defines.

danomatika:
This is also *bad news* for libpd, great.

Spacechild1:
yeah, it is certainly easier to simply disable IPv6 networking on BSDs and Android.

However, by catching the error, IPv6 support will magically work once the libc bug gets fixed.

> I wanted to release libpd 0.12 based on the 0.51-3 tag. :P

Yeah, bummer.

Anyway, I can make a PR and ask @Ant1r to test it. Maybe Miller is willing to make a 0.51-4 release. He wouldn't even need to create new packages, as the bug doesn't affect Windows and macOS. Also, there have been pettier reasons for bugfix releases in the past. :-)

danomatika:
[...]

PR #1232, opened 2020-11-26
addrinfo_get_list: hack for Android and BSDs

Spacechild1:
The BSD libc currently has a bug where `getaddrinfo()` returns `EAI_BADFLAGS` for the `AI_ALL` and `AI_V4MAPPED` flags.
This essential means that we can't use dual stack sockets, so we simply fall back to IPv4.

NOTE: this apparantly also affects Android!

Fixes https://github.com/pure-data/pure-data/issues/1229

### DIFF
--- a/src/s_net.c
+++ b/src/s_net.c
@@ -71,6 +71,14 @@ int addrinfo_get_list(struct addrinfo **ailist, const char *hostname,
          * so we fall back to IPv4 networking... */
     if (result == EAI_BADFLAGS)
     {
+        static int warned = 0;
+        if (!warned)
+        {
+            fprintf(stderr, "Warning: can't create IPv6 dual-stack socket - falling "
+                "back to IPv4. (This is a known bug in the BSD libc, which doesn't "
+                "implement the AI_ALL and AI_V4MAPPED flags for getaddrinfo().)\n");
+            warned = 1;
+        }
         hints.ai_family = AF_INET;
         hints.ai_flags = AI_PASSIVE;
         result = getaddrinfo(hostname, portstr, &hints, ailist);
--- a/src/s_net.c
+++ b/src/s_net.c
@@ -69,7 +69,6 @@ int addrinfo_get_list(struct addrinfo **ailist, const char *hostname,
          * NOTE: this also seems to affect Android!
          * In practice, this means we can't use dual stack sockets,
          * so we fall back to IPv4 networking... */
-#if defined(__ANDROID__) || defined(__FreeBSD__ ) || defined(__NetBSD__) || defined(__OpenBSD__)
     if (result == EAI_BADFLAGS)
     {
         hints.ai_family = AF_INET;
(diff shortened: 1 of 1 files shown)


Spacechild1:
This PR was successfully tested by @Ant1r on Android, see https://github.com/pure-data/pure-data/issues/1229#issuecomment-734219532.

Spacechild1:
@millerpuckette Hi Miller, I think Dan would like to make a new libpd release based on the most recent Pd tag, but the current netobject issue is a showstopper, as libpd is used a lot on Android...

Could you imagine doing a quick Pd 0.51-4 release only containing this fix? It could be source only, there's no need to produce any binaries, because it doesn't affect Windows or macOS.

src/s_net.c
```diff
@@ -62,7 +63,21 @@ int addrinfo_get_list(struct addrinfo **ailist, const char *hostname,
...
     portstr[0] = '\0';
     sprintf(portstr, "%d", port);
-    return getaddrinfo(hostname, portstr, &hints, ailist);
+    result = getaddrinfo(hostname, portstr, &hints, ailist);
+        /* There's currently a bug in the BSD libc where getaddrinfo()
+         * will return EAI_BADFLAGS for the AI_ALL and AI_V4MAPPED flags.
+         * NOTE: this also seems to affect Android!
+         * In practice, this means we can't use dual stack sockets,
+         * so we fall back to IPv4 networking... */
+#if defined(__ANDROID__) || defined(__FreeBSD__ ) || defined(__NetBSD__) || defined(__OpenBSD__)
```
umlaeute, line 72: why is this protected by ifdefs?
or put otherwise: what's the drawback on other OSs, to try again IPv4 only, if the dual-stack attempt failed (for whatever reasons, bug or not)

Spacechild1: I first had this without the #ifdef. My reasoning for the #ifdef was that we should only use the fallback on platforms which have the known BSD libc bug and on other platforms report the error as is. On the other hand, those other platforms probably won't ever show this error in the first place... So yes, I might get rid of the #ifdef again and only keep the comment.

BTW, what about printing a warning to stderr the first time 'EAI_BADFLAGS` occurs?

Ant1r:
Cool, will try on Android ASAP.

Ant1r:
OK, this last version seems working fine on Android!

danomatika:
This looks good to me and can be modified in the future if/when this changes.

danomatika:
@millerpuckette Would it be possible to accept this for a small, source-only 0.51-4 release? I'm ready to put out libpd 0.12 and it would be great to base it off of a Pd version tag. Originally, this was going to be Pd 0.51-3 but without this fix, networking is broken for BSD and Android.

millerpuckette:
I've been considering putting out a 0.51-4 anyway to add the windows
save-to-/tmp/foo.wav bug fix.  I think I can get it togehter this weekend
to put an official release out.
cheers
M

danomatika:
@millerpuckette Ok, what we can do is merge this PR into the develop branch and collect anything else obvious. I assume you want to keep it quick and limited. I will write a mail to pd-dev with a list.

PR #1236, opened 2020-12-08
change width and height to be correct after update idletasks in ::pd_bindings::patch_configure

sebshader:
due to the different way tk handles timing (on osx at least) in the current 8.6 branch, the incorrect width and height might be the last values sent to pd (as of tk commit: https://core.tcl-lang.org/tk/info/491f5a56ddf503da). Therefore they should be set to the correct values after update idletasks

### DIFF
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -347,6 +347,8 @@ proc ::pd_bindings::patch_configure {mytoplevel width height x y} {
     if {$width == 1 || $height == 1} {
         # make sure the window is fully created
         update idletasks
+        set width [winfo width $mytoplevel]
+        set height [winfo height $mytoplevel]
     }
     pdtk_canvas_getscroll [tkcanvas_name $mytoplevel]
     # send the size/location of the window and canvas to 'pd' in the form of:


danomatika:
This looks good to me. I can't imagine it breaking anything. @sebshader is there an example patch that causes issues this solves?

sebshader:
@danomatika if you use the current 8.6 branch from tk (anytime after the above commit) you get weird behavior with window geometry bc the "Configure" reports the width and height as 1 to canvas setbounds before the window is fully created. Here's the simple patch I was using to test: 
[smallwin.txt](https://github.com/pure-data/pure-data/files/5668492/smallwin.txt)
If you open the patch and click the subpatch it opens the first time with correct geometry. But then if you close and re-open the window it has a width and height of 1.

danomatika:
What macOS version? I'm not seeing an issue with macOS 10.14 and Pd 0.51-3. It's not that I don't believe you, I;'m just assuming it's another 10.15+ problem.

sebshader:
@danomatika are you using pd with a tk/wish compiled from their current core-8.6 git branch? that's where the problem is. I'm on mojave 10.14.6

danomatika:
No. I'm using the "canonical 8.6.10" which we ship with the Pd sources and which is pre-built by Miller. I don't routinely keep track with ongoing Tk development, unless needed.

danomatika:
>they fixed an issue with scrollbars updating

So we need another patch? Let me know the details now and it could possibly get into 0.51-4.

sebshader:
@danomatika I couldn't track down where exactly they fixed it because there were a lot of scrollbar-related commits and it was hard to sort out the relevant stuff that long after the 8.6.10 release (after that many changes to the relevant files). However, it was broken shortly before the commit mentioned above, but working in the commits immediately prior to it I think so the final fix for that seems to have happened shortly before it I think.

PR #1238, opened 2020-12-10
Fix undefined references when linking a static build of libpd in Windows

myQwil:
In order for Windows to successfully link applications with a
static build of libpd, the __declspec portion of EXTERN needs
to be left out, which can be done by defining EXTERN manually.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -29,6 +29,7 @@ extern int pd_compatibilitylevel;   /* e.g., 43 for pd 0.43 compatibility */
 #endif /* _MSC_VER */
 
     /* the external storage class is "extern" in UNIX; in MSW it's ugly. */
+#ifndef EXTERN
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


Spacechild1:
I think the easiest and most flexible solution would be to wrap the whole `EXTERN` definition in a `#ifndef EXTERN`, so other projects can simply override `EXTERN` with whatever they think is appropriate.

@umlaeute Any thoughts on this?

umlaeute:
i don't really have an opinion on whether `PD_STATIC`, overridable `EXTERN` or both are preferrable.

Spacechild1:
@myQwil I just saw you changed it to overridable `EXTERN`, which is cool, but your PR description is now out of date ;-)

@danomatika libpd can then simply define `EXTERN` as `extern` when building a static library on Windows. What do you think? If you agree, I'd go ahead and push this to develop.

danomatika:
Sure.

PR #1240, opened 2020-12-13
allow helpbrowser to display any amount of columns

### LINKED ISSUE #1239: Help Browser directory level doesn't go beyond the 5th level
Hi all.
The Help Browser seems to have a limit in directory level so it won't go beyond the 5th level and after that the only choice is to double click the entry to open the system browser/explorer.
Help Browser Dir Levels

sebshader:
comment out lines creating the limits for # of columns in helpbrowser.tcl. Closes #1239


umlaeute:
i tend to agree with https://github.com/pure-data/pure-data/issues/1239#issuecomment-744442674 that the issue should be properly fixed, rather than just removing the limits.

however, if the "proper fix" *is* to just remove the limits, then please do remove them rather than just commenting them out.

sebshader:
@Ant1r thanks for testing. I tried adding a final "update" at the end of the creation and that seemed to fix it on my linux vm (for some reason). hopefully fixed..

Ant1r:
fixed here, thanks!
it flickers a little when opening though.

danomatika:
It's working for me with Miller's 0.52-0test2 build, but suffers the same issue as #1481 on macOS. I'd suggest adding the update lines for my PR to yours. This may also solve @Ant1r's issue.

Another nice point would be if the horizontal scrollbar could hide when the number of columns is below a certain amount etc, but that's not a big deal since the vertical scrollbars are always present.

@umlaeute @millerpuckette I'd suggest reviewing this for 0.52 a test3 release.

sebshader:
@danomatika thanks, should be fixed now. I had ```update idletasks```d on the doclistbox but not the liblistbox before. included blocking ctrl-z too.

danomatika:
I am still seeing some display timing artifacts on my system. I'm not sure if others see the same thing, so I made a couple screen recordings. Basically, I am seeing a visual jump between sizes when, for example, the window is first created and is empty then suddenly filled with data and resized. The second jump is still the listbox contents clearing, then filling again, although not as bad the behavior before my fix in #1481.

This is Pd-0.50-0test2 with #1481 applied

https://user-images.githubusercontent.com/480637/143409981-952d5c95-0860-45c3-9fee-83903ed2b72e.mov

And this Pd-0.50-0test2 with #1240 (this PR) applied (current as of 25 Nov 10am CET):

https://user-images.githubusercontent.com/480637/143410144-a5aaf150-161c-404b-acb5-101e4fd83906.mov

danomatika:
Also, it may be worth comparing to a macOS Pd build (+ TK wish) compiled on a newer system to make sure this isn't an issue with Tk compiled on an older system, ie. Miller's builds.

sebshader:
@danomatika tbh I don't see much difference between the two videos

danomatika:
> @danomatika tbh I don't see much difference between the two videos

Second video: Initial small size of window is visible on creation then resizes for filled list content. The refresh when changing listbox entries on the rightmost panel also "flashes" as the emptying of the listbox is visible then it is filled with the new content. Ideally, you would not see these interim states and their would be "update flash."

umlaeute:
i would just go for this PR:
- the flickering seems to be minor enough on my system resp. eyes (but obviously i cannot speak for other windows)

i would drop the traiiling whitespace, though

tcl/helpbrowser.tcl
```diff
@@ -29,18 +29,24 @@ proc ::helpbrowser::open_helpbrowser {} {
         if {$::windowingsystem eq "aqua"} {
             .helpbrowser configure -menu $::dialog_menubar
         }
+        
+       
+
```
umlaeute, line 34: it seems that there are trailing whitespaces in these lines.

tcl/helpbrowser.tcl
```diff
@@ -64,18 +70,58 @@ proc ::helpbrowser::check_destroy {level} {
...
+        }
+    }
+    bind $mytoplevel.c.f <Configure> {
+        if {"%W" eq ".helpbrowser.c.f"} {
+            .helpbrowser.c configure -scrollregion [list 0 0 \
+                [winfo reqwidth .helpbrowser.c.f] \
+                [winfo height .helpbrowser.c]] \
+                -height [winfo reqheight .helpbrowser.c.f]
+        }
+
```
umlaeute, line 102: and here

tcl/helpbrowser.tcl
```diff
@@ -85,12 +131,17 @@ proc ::helpbrowser::make_rootlistbox {{select true}} {
...
-                             -highlightcolor white -selectborderwidth 0 \
-                             -height 20 -width 24 -exportselection 0 -bd 0]
-    pack $current_listbox [scrollbar "$b-scroll" -command [list $current_listbox yview]] \
-        -side left -fill both -expand 1
+    set current_listbox [listbox "[set b .helpbrowser.c.f.root0]" \
+        -yscrollcommand "$b-scroll set" \
+        -highlightbackground white -highlightthickness 5 \
+        -highlightcolor white -selectborderwidth 0 \
+        -height 20 -width 24 -exportselection 0 -bd 0]
+
```
umlaeute, line 139: and here

tcl/helpbrowser.tcl
```diff
@@ -85,12 +131,17 @@ proc ::helpbrowser::make_rootlistbox {{select true}} {
...
+        -yscrollcommand "$b-scroll set" \
+        -highlightbackground white -highlightthickness 5 \
+        -highlightcolor white -selectborderwidth 0 \
+        -height 20 -width 24 -exportselection 0 -bd 0]
+    
+    grid $current_listbox -column 0 -row 0 -sticky ns
+    grid [scrollbar "$b-scroll" -command [list $current_listbox yview]] \
+        -sticky ns -row 0 -column 1
+    grid rowconfigure .helpbrowser.c.f 0 -weight 1
+
```
umlaeute, line 144: and here

tcl/helpbrowser.tcl
```diff
@@ -307,23 +369,26 @@ proc ::helpbrowser::make_doclistbox {dir count {select true}} {
...
     # select first entry & update next col
     if {$select && [$current_listbox size] != "0"} {
         $current_listbox selection set 0
         dir_navigate_key "$dir" $count $current_listbox false
     }
-
     # force display update
     update idletasks
-
+
```
umlaeute, line 379: and here

umlaeute (commented):
apart from the trailing whitespace nitpick, i'm fine with this PR

sebshader:
I believe I got all of those whitespace areas

tcl/helpbrowser.tcl
```diff
@@ -1,5 +1,5 @@
 
-package provide helpbrowser 0.2
+package provide helpbrowser 0.3
```
danomatika, line 2: The version needs to be bumped in pkgIndex.tcl as well.

tcl/helpbrowser.tcl
```diff
@@ -30,17 +30,21 @@ proc ::helpbrowser::open_helpbrowser {} {
...
         wm resizable .helpbrowser 0 1
         ::helpbrowser::make_frame .helpbrowser
 
         # hit up, down, or Tab after browser opens to focus on first listbox
-        bind .helpbrowser <KeyRelease-Up> "focus .helpbrowser.frame.root0"
-        bind .helpbrowser <KeyRelease-Down> "focus .helpbrowser.frame.root0"
+        bind .helpbrowser <KeyRelease-Up> "focus .helpbrowser.c.f.root0"
+        bind .helpbrowser <KeyRelease-Down> "focus .helpbrowser.c.f.root0"
+
+        # ignore undo bindings?
```
danomatika, line 44: This seems to be an extra copy/paste

tcl/helpbrowser.tcl
sebshader: oh yeah I should've checked

tcl/helpbrowser.tcl
sebshader: I just kept what was already there. what do you mean?

danomatika: I see this section twice in my local checkout and the separate file view, just not in the diff:

Screen Shot 2022-08-15 at 1 09 26 AM

sebshader: ohh must've happened bc the changes from #1481 got merged twice

sebshader: @danomatika think I got it

danomatika:
Ouch, a force-push? I will see if I can pull it... In any case, I believe I am close to a solution to the startup flash, then I can look into the col creation flash.

sebshader:
I'm still getting the '.helpbrowser: no such object' error here on linux..

danomatika:
The creation flash fix is in a new `update/helpbrowser-nolimit` branch: https://github.com/pure-data/pure-data/tree/update/helpbrowser-nolimit

I also want to fix:
* [ ] Flash on creation of new columns. This is likely the canvas being resized after the window resizes.
* [ ] First selection on keyboard using up/down does not create column.

danomatika:
I don't know the best workflow for this, but you can fetch the branch, then do a cherry-pick into yours.

danomatika:
I think the focus can be removed from the scrollbar, canvas, and frame so tab works again, ie. hit tab after creation to select first item. I'm unsure if this breaks any sort of scroll handling with the keyboard though.

danomatika:
I went ahead and removed the scrollbar focus in the next commit. We have the keyboard already bound for movement, so let's try this and remove later if needed.

danomatika:
First entry selection now working with latest commit.

danomatika:
I can't find the col update flash yet, will look tomorrow.

danomatika:
> I can't find the col update flash yet, will look tomorrow.

Not having any luck, so far. I can get rid of the flash if I comment out the `update idletask` in the make listbox process, *but* things only work well if I do a `::pdwindow::post` which makes me think there is a timing issue, at least on macOS, since posting takes some time and updates a different part of the GUI.

Ant1r:
What about merging this PR for 0.55? It works fine here on Linux, and the current 5-columns limitation is sometimes really annoying.
Personally I'd call this a bug fix, rather than a feature.

porres:
I agree it is a bug and this is very much needed for my work, ping @millerpuckette

danomatika:
I can look at this again but I believe there is still the issue of a visible update aka flash as the window resizes. It would be nice to fix that before simply merging. This is, in fact, the only reason why I haven't pushed it as I don't believe it's fully done yet. Perhaps I am the only one seeing it? I don't know.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Mar 18, 2024, at 11:10 AM, Antoine Rousseau ***@***.***> wrote:﻿
What about merging this PR for 0.55? It works fine here on Linux, and the current 5-columns limitation is sometimes really annoying.
Personally I'd call this a bug fix, rather than a feature.

Ant1r:
@danomatika no flash here; there's an issue that after switching fullscreen on and off, the help browser window cannot auto-resize anymore, but this is already the case on current master, so it's not a bug coming from this PR.

umlaeute:
like @Ant1r , i don't see any problems (that are not already in the existing implementation):

| component | version |
|---|---|
| Debian/GNU | sid|
| Linux| 6.7.9|
| Xorg| 7.7|
| XFce4 | 4.18 |
| Tcl/Tk| 8.6.13|

danomatika:
Thanks for checking. The problem I was referring to before seems to have been resolved with the newer Tcl/Tk in recent Pd 0.54 builds. It works fine for me on macOS 14.2.1 with Miller's Pd 0.54-1 release (after modification).

danomatika:
Moving to 0.56 milestone. This should be good to go, I believe.

PR #1242, opened 2020-12-15
macOS fixes for 0.51-4

### LINKED ISSUE #1237: No scrollbars visible on macOS using tk 8.6.10
Using Pd-51.0-3 from miller's site on osx mojave 10.14.6 the scrollbars aren't visible. They can be dragged once found, and the page can still be scrolled with the scroll wheel. This doesn't occur with the current git core-8-6-branch from tk. (seems to have been fixed sometime before https://core.tcl-lang.org/tk/info/3fae4554efd427a2)
I thought I saw a report/discussion about this but couldn't find it..

danomatika:
## Update

I'm lumping in a few other fixes to this PR in lieu of opening a new one (lazy).

* [x] added Tk patch which fixes font dialog crash on systems with the Touchbar
* [x] audio and midi dialogs should no longer be set with `wm attribute -topmost 1` on macOS as they appear over *all* application windows
* [x] removed a couple other uses of `wm attribute -topmost 1` which were never actually applied as they use the wrong window variable name (copy/paste leftovers?), this can be reverted if I'm wrong
* [x] match font dialog ok button sizing to that of other dialogs 
* [x] fixed Pd showing Dark Mode on first launch, before disabling key is set in defaults preferences file

## Original Description

This PR fixes #1237 by adding a Tk 8.6.10 patch which backports commit for scrollbars on macOS 10.14 (and probably newer versions as well).

Apparently, they were not being updated on some events, so the TK scrollbars are not (re)drawn resulting in this for a patch with 1 object with a far xy location:

Screen Shot 2020-12-15 at 11 19 09 PM

The same type of patch looks like this with this PR applied:

Screen Shot 2020-12-15 at 11 20 08 PM

### DIFF
--- a/tcl/dialog_startup.tcl
+++ b/tcl/dialog_startup.tcl
@@ -23,8 +23,6 @@ proc ::dialog_startup::chooseCommand { prompt initialValue } {
     wm minsize .inputbox 450 30
     wm resizable .inputbox 0 0
     wm geom .inputbox "450x30"
-    # not all Tcl/Tk versions or platforms support -topmost, so catch the error
-    catch {wm attributes $mytoplevel -topmost 1}
 
     button .inputbox.button -text [_ "OK"] -command { destroy .inputbox } \
         -width [::msgcat::mcmax [_ "OK"]]
--- a/mac/README.txt
+++ b/mac/README.txt
@@ -227,4 +227,12 @@ longer their inner text and text selection positioning is off.
 
 To remedy this for now, Pd 0.51-3 changed Pd's default font for macOS to Menlo
 which is included with the system since 10.6. Menlo is based on Bitstream Vera
-Mono and DejaVu Sans Mono, so there should be no issues with patch sizing or positioning.
+Mono and DejaVu Sans Mono, so there should be no issues with patch sizing or
+positioning.
+
+## Dark Mode
+
+Pd currently disables Dark Mode support by setting the
+NSRequiresAquaSystemAppearance key to true in both the app bundle's Info.plist
+and the GUI defaults preference file. This restruction may be removed in the
+future once Dark Mode is handled in the GUI.
--- a/mac/patches/README.txt
+++ b/mac/patches/README.txt
@@ -29,6 +29,12 @@ To skip applying patches, use the tcltk-wish.sh --no-patches commandline option.
 
 ## Current Patches
 
+### tk8.6.10_scrollbars.patch
+
+Backport commit which fixes scrollbars not (re)drawing, see
+
+    https://core.tcl-lang.org/tk/info/71433282feea6ae9
+
 ### tk8.5.19_keyfix.patch
 
 Fixes key release events and adds repeat handling, see
--- a/mac/patches/README.txt
+++ b/mac/patches/README.txt
@@ -29,17 +29,16 @@ To skip applying patches, use the tcltk-wish.sh --no-patches commandline option.
 
 ## Current Patches
 
+### tk8.6.10_zombiewindows.patch
+
+Backport fix for zombie windows on systems with the Touchbar which could cause
+Pd to crash when closing the Font dialog due to a nil window pointer.
+
 ### tk8.6.10_scrollbars.patch
 
 Backport commit which fixes scrollbars not (re)drawing, see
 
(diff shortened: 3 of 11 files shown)


danomatika:
Here is a test build: http://docs.danomatika.com/pdbuilds/0.52/Pd-0.51-4-test1-scrollbars.zip

sebshader:
looking good here

sebshader:
@danomatika just looking into the console "stretch" bug that crashes pd: it seems like this patch causes a segmentation fault due to Tk_SetInternalBorderEx being called after the window it references has already been destroyed. The tcl/tk event system seems to have undergone some changes prior to the commit (since the tk tree compiled at that commit doesn't crash.)

sebshader:
@danomatika grabbing the TkWmDeadWindow function in macosx/tkMacOSKWm.c from the same tk commit as the patch fixes the crash bug

sebshader:
@danomatika here's an updated scrollbar.patch that seems to fix the crash with the font dialog. I had to take the TkWmDeadWindow function from tkMacOSXWm.c and add an objective-c declaration it uses to tkMacOSXPrivate.h.
[tk8.6.10_scrollbars.txt](https://github.com/pure-data/pure-data/files/5701166/tk8.6.10_scrollbars.txt)

sebshader:
@danomatika just fyi I didn't add in all of their fixes to for the touchbar issues, I just didn't separate them from the fixed function (and I don't think it will hurt anything)

danomatika:
@sebshader Thanks. I put it in a separate patch, otherwise things start to get messy.

danomatika:
Here is a test build with the current changes applied: http://docs.danomatika.com/pdbuilds/0.52/Pd-0.51-4-macupdates1.zip

danomatika:
Here is another test build which disables Dark Mode in the Info.plist: http://docs.danomatika.com/pdbuilds/0.52/Pd-0.51-4-macupdates2.zip

tcl/dialog_startup.tcl
```diff
@@ -23,8 +23,6 @@ proc ::dialog_startup::chooseCommand { prompt initialValue } {
     wm minsize .inputbox 450 30
     wm resizable .inputbox 0 0
     wm geom .inputbox "450x30"
-    # not all Tcl/Tk versions or platforms support -topmost, so catch the error
-    catch {wm attributes $mytoplevel -topmost 1}
```
umlaeute, line 27: (hmpf forgot to send my original post)

i don't really understand, why the `Preferences->Startup...` dialog does not have the `topmost` flag set, while Audio/Midi dialogs do.
what's the difference?

shouldn't we just fix the `$mytoplevel` instead?

~~~tcl
catch {wm attributes .inputbox -topmost 1}
~~~

danomatika:
I thought about, but figured if it has already been this way for years and no one is bothered by it, why not leave the behavior and remove the line? Up to you, you can revert that separate commit if you want and fix it.

I actually am inclined to not use topmost in most situations as the Tk docs say the intention is for the window to be above all open windows including other applications. I’m not if that is what we want. I think it was originally used to make sure the dialog appeared above other Pd windows on open but this is already achieved by calling raise and focus. I think I remember adding explicit raise and focus to some dialogs when I overhauled them.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
fair enough.
but in which sense are the Audio & MIDI dialogs different?

danomatika:
> fair enough.
> but in which sense are the Audio & MIDI dialogs different?

I don't know. They "were that way before" and I'm hesitant to change them in other platforms. OTOH it's probably safe to do so.

danomatika:
This PR is more or less ready to go for a 0.51-4-test2 build.

millerpuckette:
I think "topmost" was put in my Hans when he did the total GUI rewrite, and he was Mac only, so if "topmost" is mostly a bother on Mac I thin kit can get chucked altogether.  I'll merge these and go in and excise them altogether afterward.  It's pretty obnoxious, I agree....

millerpuckette:
OK, cranking it out now.  (I think it's fine to remove "topmost" everywhere;
I'll make that a separate commit.)

cheers
Miller

PR #1244, opened 2020-12-16
bugfix when undoing object move while zooming x 2

jyg:
with zoom = 2, the undo behaviour was inconsistent after moving an object with mouse / keyboard arrows


umlaeute:
doh, that's actually a duplicate of #858

Issue #1133, opened 2020-08-21
Edit-Font Stretch param gives (nonfatal) tcl error


chainjazz:
(original report redundant, see below)

umlaeute:
1. could you please post detailed instructions on how to reproduce this error? ("(original report redundant, see below)" is **not** very helpful; i don't know how it is redundant, nor what you are referring to with "below")

> Does Pd explicitly runs its own wish/tclsh?

yes (unless i'm mistaken)

chainjazz:
1., set logging to level 4 all
2. Open a patch window.
3. Select Edit -> Font menu item
4.. Change the "Stretch" field to something else then 100 (in my case 125)
5. Click apply or OK
6. Error

The "below" refers to the next comment in order as it appears on GitHub website by default.

The redundancy part, I had copied the complete log which featured the same error message repeated over and over, before updating to the latest version, so in order to make it easier to read the comment feed, I put I note (also, couldn't find how to remove the initial comment)

umlaeute:
> 5. Error

not here.
i'm assuming that with "Open a patch window" you really mean <kbd>Control</kbd>-<kbd>n</kbd> (to get a new, empty patch window).
if not, please provide a minimal patch that triggers the problem.

> 6.1 If you don't see the error message, try changing the font via command line parameters -font-face, etc.

no, seriously.
please provide a full example that triggers the problem.
if you use custom startup flags, you must provide **all** of them, so we can reproduce the problem.
if you use a special font, you must say so.

make sure you run Pd with `-noprefs`.
uninstall/disable all GUI-plugins that you happen to have.
don't load any externals.


> The "below" refers to the next comment in order as it appears on GitHub website by default.
[...]
> (also, couldn't find how to remove the initial comment)

so why don't you just put the updated information in place of your original comment?

sebshader:
it happens when there's a GOP object in a (non-abstraction) subpatch

sebshader:
it happens when there's a gop object connected to another object in a non-abstraction subpatch of the patch whose "font" dialog it is

umlaeute:
it would be fantastic if you could post such a patch here, with in-patch instructions (with numbers, if the order is important) on where you need to do what.

umlaeute:
like so:

~~~
#N canvas 749 176 450 300 12;
#N canvas 915 331 450 300 fuu 0;
#N canvas 5 53 450 300 bar 0;
#X obj 289 130 inlet;
[...]

Issue #345, opened 2018-04-04
Pd-GUI crashes when printing too much text to the console


umlaeute:
as reported on the [Pd-list](https://lists.puredata.info/pipermail/pd-list/2018-04/122237.html), Pd can crash when printing in an infinite loop.

i guess the reason is, that the Pd-console will keep all messages - and in an infinite loop this will eventually exhaust the memory.

a solution would be to limit the number of lines/messages stored (and make that number configurable by the user)

umlaeute:
actually, the Pd-console seems to **not** be the culprit:

something as stupid as:

~~~tcl
proc ::pdwindow::resay {msg num} {
    for {set i 0} {$i < $num} {incr i} {
        ::pdwindow::logpost {} 3 "$i: $msg\n"
    }
}
::pdwindow::resay 4294967295 "hello world"
~~~

...hangs forever but doesn't crash Pd.

otoh, this patch:

~~~
#N canvas 1052 273 450 300 12;
#X msg 175 224 bang;
#X obj 170 90 until;
#X obj 170 115 i;
#X obj 170 140 + 1;
#X obj 170 165 t f f;
#X msg 211 91 0;
#X msg 168 57 bang;
#X obj 175 249 print ====================;
#X msg 170 190 \$1 hello world;
#X connect 0 0 7 0;
#X connect 1 0 2 0;
#X connect 2 0 3 0;
#X connect 3 0 4 0;
#X connect 4 0 8 0;
#X connect 4 1 2 1;
#X connect 5 0 2 1;
#X connect 6 0 1 0;
#X connect 8 0 7 0;
~~~

crashes with:
> max size for a Tcl value (2147483647 bytes) exceeded
> Aborted (core dumped)
> pd output pipe: Connection reset by peer

here's a backtrace:

~~~sh
(gdb) run
Starting program: /usr/bin/wish8.6 ../tcl/pd-gui.tcl
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff67e9700 (LWP 1445236)]
[Detaching after fork from child process 1445237]
input channels = 0, output channels = 0
Pd-0.51.2 ("") compiled 23:56:40 Sep 28 2020
float precision = 32 bits
pd output pipe: Connection reset by peer
max size for a Tcl value (2147483647 bytes) exceeded

Thread 1 "wish8.6" received signal SIGABRT, Aborted.
[...]

PR #1246, opened 2020-12-16
 stable development branch

umlaeute:

supersedes #1235 

----

- GUI
  - fix undo movement when zoomed in
  - fix displacement of objects inserted via triggerize after objects with unusual height
  - fix Tcl-error when font-stretching hidden GOPs (Closes: #1133)
  - fix pasting of unicode text, e.g. CJK (Closes: #1284)
  - fix font-metric calculation for extreme fonts (to avoid division-by-zero) (Closes: #1285)
  - turn cursor back to normal if reading a file fails (so it doesn't hang spinning)
- core
  - use `const` specifiers in utf-8 function prototypes
  - fix compilation of libportmidi on Linux
  - prefix ingress/outgress specifiers when debugging the GUI-core connection
  - AppleWatch support
- help patches
  - normalize the "updated to Pd version <version>" string
 - deken
   - import deken-0.8.0
- crashes
  - initialise structs in expr, so it doesn't crash so easily (Closes: #1255)
  - fix crashes when sending manymany printouts from core to GUI (Closes: #345)

### DIFF
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -1278,26 +1278,6 @@ static void garray_edit(t_garray *x, t_floatarg f)
     x->x_edit = (int)f;
 }
 
-static void garray_vis_msg(t_garray *x, t_floatarg f)
-{
-    int viswas, vis = f != 0;
-    t_template *scalartemplate;
-    if (!(scalartemplate = template_findbyname(x->x_scalar->sc_template)))
-    {
-        error("array: no template of type %s",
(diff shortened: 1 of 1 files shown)


danomatika:
@millerpuckette These might be good for 0.51-4, except for maybe 44b622f which changes the JACK port numbering. The overridable EXTERN is needed for static libpd builds on Windows.

PR #1245, opened 2020-12-16
make sure object is visible for drawing cords and gatom labels when displacing

sebshader:
Fixes #1133

PR #1247, opened 2020-12-18
clip objects in GOP subpatch correctly in font dialog "stretch"

sebshader:
I noticed some more gop clipping issues when working on the font stretch tcl error bug #1245 
glist_redraw was not cutting it, canvas_redraw seems to be necessary. also set the canvas "dirty" flag after font change.

the problem was that, when using the font "stretch" entry: if a GOP non-abstraction subpatch was on the canvas, the graphcal items in the gop object would be "gobj_displace"d and redrawn on the parent canvas instead of clipped out of the child GOP, since the graphical objects contained had no way of knowing to clip themselves. (by the time glist_redraw was called the objects had already been drawn to the wrong canvas)
the window dimensions were also not getting updated correctly

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -4768,24 +4768,22 @@ static void canvas_dofont(t_canvas *x, t_floatarg font, t_floatarg xresize,
 {
     t_gobj *y;
     x->gl_font = font;
-    if (xresize != 1 || yresize != 1)
-    {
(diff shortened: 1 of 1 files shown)


sebshader:
dang this broke the regular "font size" selection

sebshader:
ok should be working now

umlaeute:

PR #1248, opened 2020-12-18
Don't apply "stretch" in font dialog when selecting different font sizes

sebshader:
In the 2nd commit, an "Apply" button was added to the stretch labelframe to apply the stretch. Then the "OK" button just closes the dialog.


umlaeute:
which problem is this trying to fix?

sebshader:
@umlaeute currently when using the font dialog, the "stretch" entry field value is applied to the canvas whenever a radiobutton is selected, and when the "OK" button is clicked. This is very unintuitive. This PR would change it so that selecting the font size radiobutton only changes the font size, and then there is a separate "Apply" button for applying the "stretch" ratio to the canvas.

umlaeute:
thanks, makes sense.

probably the `OK` button could be renamed to `Close`, as it has nothing to do with consent to the changes (as the name might imply).

umlaeute:

PR #1259, opened 2021-01-13
documentation updates

### LINKED ISSUE #1258: better document program change messages (as indexed from 1)
For reference, [this](http://midi.teragonaudio.com/tech/midispec/pgm.htm) says that although program change numbers are from 0-127, there's no standard and some devices can index from 1, and I see Pd does index from 1, see:
- https://github.com/pure-data/pure-data/blob/master/src/x_midi.c#L307
- https://github.com/pure-data/pure-data/blob/master/src/x_midi.c#L710

So I just think we need to add a note in the documentation saying that it is indexed from 1.

### LINKED ISSUE #1315: clone help and output~ abstraction
[clone] is the only one that uses the output~ abstraction found in the audio examples. 

The way it is called, though, generates a very large GOP area.

see:

Screen Shot 2021-05-19 at 02 30 08

while the abstraction was meant to look like this 

Screen Shot 2021-05-19 at 02 30 22

One way would be to hide the object name in the settings, but that's bad I think.

Another would be to use [declare -path ../3.audio.examples/], which works but I think it's not an elegant solution.

The only other way would be to include the output~ abstraction in the same folder. This would create a third copy of this abstraction, which is now also part of the data structures examples and used only once in 07.sequencer.pd

Again, only clone uses this, so it'd be yet another copy just to be used once. Nonetheless I like this option and would like to use it. Moreover, I think I can find more excuses to use it in other examples of the help file.

porres:
commit 1: midi-help.pd

I only added this information and reogranized the patch to fit this.

"** Program change values in [pgmin] and [pgmout] are indexed from 1, which means that the possible values are from 1 to 128 (not 0 to 127)!"

closes: https://github.com/pure-data/pure-data/issues/1258

----------

commit 2: makefilename.pd

minor changes to fix typos and similar issues with the text

----------

commit 3: update tabsend~/tabreceive~

-----------

commit 4: remove unused pd-help.pd

-----------

commit 5: added abbreviation: [s~] / [r~] to send~/receive~ help file

-----------

commit 6: remove 0_all_guis-INTRO

-----------

commit 7: tabwrite info on clipped indexes (closes https://github.com/pure-data/pure-data/issues/1281)

-----------

commit 9: pipe (closes https://github.com/pure-data/pure-data/issues/1314)

-----------

commit 8 and 10: update router, closes https://github.com/pure-data/pure-data/issues/1302

-----------

commit 11: added the abbreviations of objects, some suggested in #1300

-----------

commit 12: clarified list behaviour in [pack], [pipe]

-----------

commit 13: clarified 'send' message in [symbol]

-----------

commit 14: adds output~ to reference, closes #1315

-----------

commit 15: adds output~ tabosc4~ partially addresses https://github.com/pure-data/pure-data/issues/1316

-----------

commits 16-18: improve [declare] help, closes https://github.com/pure-data/pure-data/issues/1318

-----------

commit 19: slop~ update, closes https://github.com/pure-data/pure-data/pull/1259/commits/6669426e69c6da59531e4a67d45263e2abfa3d44

-----------

commit 20-21: expr update closes https://github.com/pure-data/pure-data/issues/1327


millerpuckette:
Git's complained that doc/8.topics/expr.html has trailing spaces... I tried to fix this but am not sure it worked.

Meanwhile I reverted to old versions of these files:

 doc/5.reference/czero_rev~-help.pd
 doc/5.reference/czero~-help.pd
 doc/5.reference/lop~-help.pd
 doc/5.reference/rzero_rev~-help.pd
 doc/5.reference/rzero~-help.pd
doc/5.reference/operators-help.pd
doc/5.reference/otherbinops-help.pd

PR #1260, opened 2021-01-17
increase ATOMBUFSIZE to 200, max symbol atom width to 120

reduzent:
I know about symbol table pollution etc., still, that artificial limit is annoying. If there is no technical reason to keep it that low, I'd like to have it increased. Also, I don't understand why one is allowed to create symbol atoms with a width larger than the largest symbol one can type. I'd rather have it the other way around.

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -507,7 +507,7 @@ void canvas_msg(t_glist *gl, t_symbol *s, int argc, t_atom *argv)
 
 /* ---------------------- the "atom" text item ------------------------ */
 
-#define ATOMBUFSIZE 40
+#define ATOMBUFSIZE 200
 #define ATOM_LABELLEFT 0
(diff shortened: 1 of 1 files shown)


porres:
is there a relation to https://github.com/pure-data/pure-data/pull/1172 ?

reduzent:
No, I just want to be able to type larger symbols into a symbol atom.

reduzent:
It seems cf44c8bd4e addresses part of this PR, since there seems no atom size limit anymore. However, the symbolatom  is still restricted to the maximum width of 80.

Can we relax this restriction a bit?

millerpuckette:
increased to 1000 (still arbitrary but that would be a wide display.)

umlaeute:
i guess this fixes the issue at hand.
commit is 5b65acd55b6417b51b0023f8789793a757449e4d

PR #1265, opened 2021-02-01
list-help: Rename the "serializer" example to "list iterator"

jamshark70:
The so-called "serializer" example in list-help.pd is extremely important for a number of reasons (some not even directly related to lists) -- but I had overlooked it for a long time, because it's mislabeled.

"Serializing" (AFAIK) refers to rendering a complex, multi-layered data structure into a flat data stream (such as to save into a file) to be unpacked later.

[pd example3] in this help document doesn't demonstrate that. It's just a simple list iterator.

This PR simply changes "serializer" to "list iterator" in the comment box.


umlaeute:
side note: i think such trivial re-wording changes are usually best dealt with using a text-editor (rather than Pd itself).
this way, the line-numbers (and thus object-ids) stay the same, and the changeset becomes much smaller and easier to compare and manage (from a git-perspective).

porres:
I will take care of this along other changes in my documentation PR

porres:
hi, we can close this as my documentation branch got merged and included this

PR #1267, opened 2021-02-03
Fix various typos

luzpaz:
Found via codespell

### DIFF
--- a/doc/5.reference/inlet-outlet-help.pd
+++ b/doc/5.reference/inlet-outlet-help.pd
@@ -135,7 +135,7 @@ object to pass arguments to initialize signal inlets in abstractions.
 #X restore 80 259 pd subpatch-with-inlets-and-outlets;
 #N canvas 0 0 600 400 (subpatch) 0;
 #X restore 172 524 pd;
-#X text 70 97 Inlets/outlets are used to receive and get informatin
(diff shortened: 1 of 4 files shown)


luzpaz:
Any chance I could submit a fixes for these source typos? 

https://github.com/pure-data/pure-data/search?q=timout   
https://github.com/pure-data/pure-data/search?q=comitted

umlaeute:
> Any chance I could submit a fixes for these source typos?

hmm. most of the occurences are commented out anyhow.
and some are just otherwise unused variables (which we should probably get rid of instead)

doc/5.reference/inlet-outlet-help.pd
```diff
@@ -135,7 +135,7 @@ object to pass arguments to initialize signal inlets in abstractions.
 #X restore 80 259 pd subpatch-with-inlets-and-outlets;
 #N canvas 0 0 600 400 (subpatch) 0;
 #X restore 172 524 pd;
-#X text 70 97 Inlets/outlets are used to receive and get informatin
+#X text 70 97 Inlets/outlets are used to receive and get information
```
luketeaford, line 138: I was going to open a PR for this one-- would be very nice to fix this since it is discoverable.

PR #1272, opened 2021-02-17
allow large amounts of data to be sent to the GUI (and limit the Pd-console's length)

umlaeute:
currently the GUI crashes when printing *large* amounts of data, as can happen when accidentally `[print]`ing a counter that 
 (#345).

a very simple example of such a crashing patch is:
~~~pd
#N canvas 851 401 450 300 12;
#X msg 86 46 1e+07;
#X obj 86 71 until;
#X obj 86 121 print;
#X msg 86 96 1 2 3 4;
#X connect 0 0 1 0;
#X connect 1 0 3 0;
#X connect 3 0 2 0;
~~~

there were several underlying issues:
- the Pd-console keeps all messages ever printed (unless explicitely cleared). this obviously fills the memory over time.
- an integer-overflow when allocating more memory for gui messages
- untested and broken code leading to data loss (and thus un-executable tcl-code) in the core->gui communication was triggered whenever such an integer-overflow occured
- the GUI - hoping to recover from the data-loss - would allocate more and more memory until it eventually crashed.
- the GUI would also crash if the cmdbuffer received from the core would not end with `LF` (`\n`) for more than 2GB.

this PR fixes this in multiple ways:
- prevent integer-overflow when allocating more memory for the GUI-communication
- fix the code that synchronously flushes the data to the GUI when the queue gets stuck (this is an emergency handling to keep the system alive, and as such it is not realtime friendly at all)
- improve the code on the (receiving) GUI side that checks whether the cmdbuffer can actually be executed.
- limit the maximum number of lines displayed in the Pd-console (whenever the `::pdwindow::maxlogbuffer` limit is exceeded, the Pd-console is cleared except for the last `::pdwindow::keeplogbuffer` lines. these values are currently set to `21000` (max. number of lines) resp. `1000` (lines to keep when autoclearing); the idea is to LATER make this settable via GUI-preferences)

### DIFF
--- a/m4/universal.m4
+++ b/m4/universal.m4
@@ -65,7 +65,7 @@ if test "$universal_binary" != no; then
     
     dnl add to arch list if it passes the linker
     AC_MSG_CHECKING([if linker accepts arch: $arch])
-    AC_TRY_LINK([], [return 0;], [
+    AC_LINK_IFELSE([AC_LANG_PROGRAM([],[])], [
       _pd_universal="$_pd_universal -arch $arch"
       AC_MSG_RESULT([yes])
     ], [
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -669,7 +669,17 @@ void sys_closesocket(int sockfd)
 
 static void sys_trytogetmoreguibuf(int newsize)
 {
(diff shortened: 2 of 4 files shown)


danomatika:
Without looking into the diffs yet in detail, it all sounds good.

Issue #1273, opened 2021-02-28
request: add "freeze" option for delwrite~


porres:
A 'freeze' option would keep the buffer in the delay line, it seems max's tapin~/tapout~ has this functionality. This can be a <freeze float> message where non zero freezes and zero unfreezes.

PR #1276, opened 2021-03-02
adding "freeze" functionality do [delwrite~]

porres:
closes https://github.com/pure-data/pure-data/issues/1273

### DIFF
--- a/src/d_delay.c
+++ b/src/d_delay.c
@@ -30,6 +30,7 @@ typedef struct _sigdelwrite
     int x_sortno;   /* DSP sort number at which this was last put on chain */
     int x_rsortno;  /* DSP sort # for first delread or write in chain */
     int x_vecsize;  /* vector size for delread~ to use */
+    int x_freeze;   /* freeze flag */
     t_float x_f;
 } t_sigdelwrite;
 
@@ -58,6 +59,10 @@ static void sigdelwrite_clear (t_sigdelwrite *x) /* added by Orm Finnendahl */
     memset(x->x_cspace.c_vec, 0, sizeof(t_sample)*(x->x_cspace.c_n + XTRASAMPS));
 }
 
+static void sigdelwrite_freeze (t_sigdelwrite *x, t_floatarg f) /* added by Porres */
+{
+    x->x_freeze = (int)(f != 0);
+}
 
     /* routine to check that all delwrites/delreads/vds have same vecsize */
 static void sigdelwrite_checkvecsize(t_sigdelwrite *x, int vecsize)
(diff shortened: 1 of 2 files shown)


src/d_delay.c
```diff
@@ -97,16 +103,24 @@ static t_int *sigdelwrite_perform(t_int *w)
     t_sample *in = (t_sample *)(w[1]);
     t_delwritectl *c = (t_delwritectl *)(w[2]);
     int n = (int)(w[3]);
+    t_sigdelwrite *x = (t_sigdelwrite *)(w[4]);
     int phase = c->c_phase, nsamps = c->c_n;
     t_sample *vp = c->c_vec, *bp = vp + phase, *ep = vp + (c->c_n + XTRASAMPS);
     phase += n;
 
     while (n--)
```
sebshader, line 111: would it be better to put the freeze check outside of the while loop?
like:

(outside of sigdelwrite_perform):
```
static void sigdelwrite_wrapphase(t_sample **loc, t_sample *vector, t_sample *end, int *phase, int nsamps)
{
    if(*loc == end)
    {
        vector[0] = end[-4];
        vector[1] = end[-3];
        vector[2] = end[-2];
        vector[3] = end[-1];
        *loc = vector + XTRASAMPS;
        *phase -= nsamps;
    }
}
```
then in perform:
```
if(x->x_freeze)
{
    while(n--)
    {
        bp++;
        sigdelwrite_wrapphase(&bp, vp, ep, &phase, nsamps);
    }
}
else
{
    while(n--)
    {
        t_sample f = *in++;
        if (PD_BIGORSMALL(f))
            f = 0;
        *bp++ = f;
        sigdelwrite_wrapphase(&bp, vp, ep, &phase, nsamps);
    }
}
```
just a thought, not sure if it's worth it or if the compiler would do that anyways

porres: maybe you can have an alternative PR and if it's considered best I'll remove this

sebshader: @porres maybe I'll just try to add commits to this.. that seems cleaner

porres: let's wait to hear the usual wizard like @Spacechild1 and @umlaeute

Spacechild1: Usually, the compiler can hoist a check out of the loop if it can prove that the value won't change. This is very likely the case here (for an explanation, you can google "strict aliasing rule").

If you want to make sure, you can cache `x->x_freeze` in a local variable and no sane compiler would do the check on every iteration.

porres:
What about just doing this?

Screen Shot 2021-03-03 at 09 52 16

patch: [test.pd.zip](https://github.com/pure-data/pure-data/files/6076238/test.pd.zip)

Issue #717, opened 2019-08-09
Rearrage Audio Settings to choose backend first


mxa:
It's confusing to be able to set samplerate and blocksize for the Jack backend, when they are clearly set by the jack server and are ignored anyways.
A solution would be to move the audio backend selection up and then show the settings which make sense in the respective backend underneath.

I've seen some proposals for a tabbed settings dialog, this may coincide.
Possibly something for the 0.51 milestone?

danomatika:
Maybe it's simple enough to disable or hide the sample rate and blocksize dropdown boxes for JACK? I agree they don't affect the actual backend.

I did look into reading the values from JACK and displaying them in the dialog but it ended top being a bit more complicated and I didn't get it working.

Spacechild1:
IIRC, we already hide the "block size" (= hardware buffer size) setting on macOS because it has no effect, so IMO it makes perfectly sense to hide the samplerate setting for Jack.

danomatika:
Actually, I think it makes sense to show the sample rate from JACK without being able to change it, ie. what the JACK server is using.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

mxa:
I'd like to contribute (in a couple of weeks) please suggest me either how to quickly mock-up a gui or - if not too hard - where to edit the tcl? code.

reduzent:
>  when they are clearly set by the jack server and are ignored anyways.

I think it's a well known fact by now, but the bad part is that is _not_ ignored.

danomatika:
Actually  I started doing this (showing JaCK sample rate) maybe a year ago but didn’t finish at the time. I could clean up that and push it to a branch.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
Ok, now implemented in #1287.

PR #1287, opened 2021-03-08
Don't set samplerate if it's fixed by the audio API

### LINKED ISSUE #1283: Samplerate mess when using JACK
The "Audio Settings"-dialog from menu "Media" -> "jack" shows buttons for changing samplerate. However, Pd cannot effectively change samplerate when using JACK as audio backend since jack clients "inherit" the samplerate defined by the JACK server. Changing the samplerate in the dialog seems to affect Pd's notion of its own samplerate. For instance, when JACK is running at 44.1k and 48k is set and applied in the dialog, the test tone is played at a slightly lower pitch. When switching to 96k, it drops by one octave. To make matters worse, the object [samplerate~] still reports the samplerate JACK server is running with. See attached screenshot.

I propose to simply disable the samplerate buttons in the 'jack' dialog. This setting makes only sense for other back-ends. By no way should Pd be allowed to change its (perceived) samplerate when using JACK. 

pd-jack-samplerate-bug

danomatika:
This PR fixes #1283 & addresses #717 by ensuring that the sample rate cannot be set when JACK is the current audio api as the sr is set by the JACK server. The audio dialog now also checks for JACK and disables the sample rate & blocksize UIs while simply showing the sr being used by JACK. The sr value is basically taken out of the internal audio dialog state when a fixed sr audio api is in use.

Screen Shot 2021-03-08 at 11 21 26 PM

Additionally, this also removes the focus selection on the sample rate entry when the audio dialog is opened. I added this some time ago to facilitate quicker keyboard edits and navigation but I think it make it far to easy to accidentally change the sample rate value without really meaning to.

A few questions:

1. Should the blocksize by hidden when using JACK or perhaps show JACK's current bs?
2. Should JACK set Pd's blocksize somehow? It doesn't appear to in `s_audio_jack.c`.
3. Should the delay entry also be disabled/hidden or show JACK's nominal latency?


mxa:
That's great, I think we should even go a step further and put the backend selection on top of this dialog in the form of a drop down menu. Beneath only show the applicable options for that particular backend. Get rid of the back end selection in the Media menu even if that is one klick less it is in reality not a time saver as you normally would want to check/change the further settings in the `Audio Settings... `dialog anyways.

danomatika:
I'd rather this be a focused bug fix for now. Anything more would require a lot more time and that's better spent combing all dialogs into a tabbed preferences window.

Spacechild1:
> Should the delay entry also be disabled/hidden or show JACK's nominal latency?

Does Pd always use the callback scheduler for the Jack API? If yes, then it makes sense, because `delay` only affects the ringbuffer scheduler.

src/s_audio.c
```diff
@@ -82,6 +82,16 @@ int audio_isopen(void)
             || (audio_naudiooutdev > 0 && audio_audiochoutdev[0] > 0)));
 }
 
+int audio_isfixedsr(void)
+{
+#ifdef USEAPI_JACK
+    /* JACK server sets it's own samplerate */
+    return (sys_audioapi == API_JACK);
+#else
+    return 0;
```
umlaeute, line 91: remove the `#else/#endif` clause and have an unconditional `return 0` at the end.

i think this makes code more readable.

umlaeute:
> Does Pd always use the callback scheduler for the Jack API?

i don't think (but haven't checked and i'm not sure).

in any case, i think the GUI shouldn't assume anything about the implementation of the audio backends (similar in spirit as https://github.com/pure-data/pure-data/pull/1287#pullrequestreview-607185657)

tcl/dialog_audio.tcl
```diff
@@ -202,6 +203,17 @@ proc ::dialog_audio::pdtk_audio_dialog {mytoplevel \
     pack $mytoplevel.inputs.in1f.x1 -side left -fill x -expand 1
     pack $mytoplevel.inputs.in1f.x2 $mytoplevel.inputs.in1f.l2 -side right
 
+    # FIXME: simple hack to disable samplerate when using JACK
```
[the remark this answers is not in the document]
danomatika: I find it useful to display the current sr used by JACK, partially as a confirmation that it matches that displayed in something like qjackctl. However, another approach would be to simply hide all of the sr guis if the sr is fixed. IN this case, as you suggest, a value of `0` would indicate sr is fixed and no GUI is required.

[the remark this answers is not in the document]
danomatika: I like the idea of a negative number to indicate something, but I'm unsure if they could lead to blowups if either side (GUI or core) misinterpret the number...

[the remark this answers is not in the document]
umlaeute: i kind of like the `!` prefix as it is *obviously* an invalid number - so if something goes wrong it will blow-up sooner rather than later, making it easy to spot.

it would also allow to use just `!` as an indicator for *no-display* while still allowing the `0` value.

src/s_audio.c
```diff
@@ -134,7 +151,8 @@ void sys_save_audio_params(
         sys_audiodevnumbertoname(1, audiooutdev[i],
             &audio_outdevnames[i * DEVDESCSIZE], DEVDESCSIZE);
     }
-    audio_rate = rate;
+    if (!audio_isfixedsr() || audio_rate == 0)
+        audio_rate = rate;
     audio_advance = advance;
     audio_callback = callback;
     audio_blocksize = blocksize;
```
danomatika, line 158: `audio_blocksize` should probably also be set conditionally.

src/s_audio.c
```diff
@@ -82,6 +82,23 @@ int audio_isopen(void)
...
+#ifdef USEAPI_JACK
+    /* JACK server sets it's own samplerate */
+    return (sys_audioapi == API_JACK);
+#endif
+    return 0;
+}
+static int audio_isfixedblocksize(void)
+{
+#ifdef USEAPI_JACK
+    /* JACK server sets it's own samplerate */
```
danomatika, line 96: update comment: `/* JACK server sets it's own blocksize */`

danomatika:
I think the value handling with the `!` character should be documented clearly, both on the GUI side and in a comment to `glob_audio_properties`.

src/s_audio.c
```diff
@@ -65,7 +65,7 @@ static int audio_naudiooutdev = -1;
 static int audio_audiooutdev[MAXAUDIOOUTDEV];
 static int audio_audiochoutdev[MAXAUDIOOUTDEV];
 static char audio_outdevnames[MAXMIDIINDEV * DEVDESCSIZE];
-static int audio_rate;
+static int audio_rate = 0;
 static int audio_advance = -1;
 static int audio_callback;
 static int audio_blocksize;
```
danomatika, line 71: Probably init to 0 so it will be set in `sys_get_audio_params`.

umlaeute: actually, `static int` values without an explicit value will always be initialized to **0** in C.

danomatika: Ok, although I always like to be explicit when it's something expected/important. :)

tcl/dialog_audio.tcl
```diff
@@ -169,19 +185,30 @@ proc ::dialog_audio::pdtk_audio_dialog {mytoplevel \
...
+    }
+
     label $mytoplevel.settings.bsc.bs_label -text [_ "Block size:"]
     set blocksizes {64 128 256 512 1024 2048}
     set bsmenu \
         [eval tk_optionMenu $mytoplevel.settings.bsc.bs_popup audio_blocksize $blocksizes]
-
     pack $mytoplevel.settings.bsc.bs_popup -side right
     pack $mytoplevel.settings.bsc.bs_label -side right -padx {0 10}
-    if {$audio_callback >= 0} {
```
danomatika, line 179: I think this still needs to be checked for as I now see the "Use callbacks" checkbox on macOS when it didn't show in the last 0.51-4 release.

umlaeute: i see this as a feature ;-) : have a consistent layout of the dialog, regardless of the OS and backend.

compare this:
> A: please check the checkbox FOO
> B: hmm, the FOO checkbox is disabled

against this:
> A: please check the checkbox FOO
> B: there's no FOO checkbox
> A: sure there is; just above the BAR field
> B: no there isn't. check my screenshot
> A: ah, you are on macOS and you are using PortAudio

anyhow: i'm not particularily opinionated about this.
if it bothers you much, feel free to re-introduce the checkbox hiding (using `$audio_isfixedcallback` instead of `$audio_callback`)

danomatika: I didn't introduce the change, I believe Miller did, but I appreciate that it's no longer there and doesn't take space in the layout when it's not relevant. Now that we can disable things, *maybe* it can stay but this would a flip/flop from the last version.

danomatika: I'll re-remove it for now.

danomatika:
Related:

Looking at this dialog again, also makes me wonder if the naming of "Delay (msec)" could be improved.

For the longest time, I *still* don't know what it's referring to. Looking in `s_audio.c`, it's called `sys_schedadvance` and used to compute `sys_advance_samples` which, I assume, is used to set the idle time between scheduler updates. Is this the same value as the `-sleepgrain` command line flag?

```
-sleepgrain <n>  -- specify number of milliseconds to sleep when idle
```

If so, that naming makes more sense. I think changing "Delay (msec)" to "Scheduler Sleep" or something similar would help.

umlaeute:
> JACK buffer sizes can be < 64 (16 & 32), so I wonder if we should
> round to 64 when displaying in the popup?

nah.
i think the real blocksize should be displayed, even if it is not in the list.

danomatika:
> JACK buffer sizes can be < 64 (16 & 32), so I wonder if we should round to 64 when displaying in the popup?
> nah. i think the real blocksize should be displayed, even if it is not in the list.

I agree and, in testing, it displays ok.

danomatika:
>IMO, "Latency (msec)" would be more descriptive.

Perhaps, but this leads to questions like "I'm using JACK, does Latency refer to the JACK server latency or something inside Pd." I think there needs to be an additional explanation somehow like "Pd Latency": or "Ringbuffer Latency" etc.

Spacechild1:
> Perhaps, but this leads to questions like "I'm using JACK, does Latency refer to the JACK server latency or something inside Pd?

That's true. In fact, some ASIO drivers label the hardware buffer size as "latency" as well.

> "Pd Latency": or "Ringbuffer Latency" etc.

"Pd Latency" could be fine. Or maybe "Extra latency (msec)"? "Ringbuffer Latency" is certainly too technical.

danomatika:
>"Pd Latency" could be fine. Or maybe "Extra latency (msec)"?

"Pd Latency" at least helps differentiate. I think "Extra latency (msec)" is cary and everyone will set it to 0... why would you want "extra latency?"

danomatika:
As a stopgap, maybe can bin down your earlier description into a single sentence which can be displayed via a tooltip?

The sentence can be reused as a label if/when the dialogs are reworked into a larger preferences panel with more space.

Spacechild1:
Or what about "Pd buffer size"? BTW, the relevant startup flag is `-audiobuf <n>    -- specify size of audio buffer in msec`, so "Audio buffer size" could work as well, but it could still be confused with the hardware buffer size.

> As a stopgap, maybe you can boil down your earlier description into a single sentence which can be displayed via a tooltip?

Ok, I'll try:

'...' determines the size of Pd's internal audio buffer. It allows to "absorb" CPU spikes and interruptions in DSP computation, at the expense of additional latency. Generally, `...` shouldn't be set lower than the hardware buffer size (in msec).

danomatika:
@Spacechild1 I think that's too long, but it may not matter as Tk doesn't;t have built-in tooltip support anyway :-S.

I'm thinking more along the lines of what you initially wrote: "delay only affects the ringbuffer scheduler" could be reworded to something like "Pd's audio latency affects how often samples are passed & processed between Pd and the audio hardware."

danomatika:
I suppose it's probably explained in the manual, but it might help if there is an overview that's quicker to find, as you suggest.

danomatika:
At this point, I think we should keep further changes for other PRs and test this as it is.

reduzent:
Looks good to me. Hopefully it can be merged into next release

>     3. Should the delay entry also be disabled/hidden or show JACK's nominal latency?

From what I experience, changing this values doesn't do anything while using JACK backend. I'd say hide it completely when using JACK. I can't imagine how the number shown there could have  a clear meaning. The amount would reflect only a theoretical latency (not taking into account hardware ADC/DAC latency). Showing it in Pd would suggest that audio passing through Pd would impose a latency of the value shown. However, that would not be true, because the number would only represent the overall latency of the signal passing from system:capture to sytem:playback.

All three graphs have the same latency:

system:capture -> system:playback
system:capture -> Pd -> system:playback
system:capture -> Pd -> Pd -> system:playback

Spacechild1:
> I can't imagine how the number shown there could have a clear meaning. The amount would reflect only a theoretical latency (not taking into account hardware ADC/DAC latency).

As I've explained earlier (https://github.com/pure-data/pure-data/pull/1287#issuecomment-793990156), the "Delay" parameter has nothing to do with ADC/DAC latency. It is the size of Pd's internal ringbuffer used by the polling scheduler. AFAICT, Jack uses the polling scheduler by default (like all other audio APIs), so "Delay" certainly *has* a meaning.

reduzent:
> > I can't imagine how the number shown there could have a clear meaning. The amount would reflect only a theoretical latency (not taking into account hardware ADC/DAC latency).
> 
>  AFAICT, Jack uses the polling scheduler by default (like all other audio APIs), so "Delay" certainly _has_ a meaning. 

Ok, I clearly don't understand it, then. What number should it show then? Frames/Period * Periods/Buffer converted to ms? The default value of 25ms shown in my Pd seems not to relate to anything I can think of.

What do you propose to do with that number?

danomatika:
> What do you propose to do with that number?

We've already been discussing this: https://github.com/pure-data/pure-data/pull/1287#issuecomment-793990156

Spacechild1:
@reduzent TL;DR: The "Delay" parameter is the *additional* latency added by Pd to deal with CPU spikes and other irregularities. But I agree with @danomatika that this is out of scope for this PR, as it is not affected by the audio API.

reduzent:
> @reduzent TL;DR: The "Delay" parameter is the _additional_ latency added by Pd to deal with CPU spikes and other irregularities. But I agree with @danomatika that this is out of scope for this PR, as it is not affected by the audio API.

Ok, thanks for the explanation. Indeed - after checking more carefully - I was able to notice that Pd adds latency, which justifies a number to be displayed and labelled 'latency ms' (or whatever). However, changing the value doesn't affect the perceived latency (unlike with ALSA, where changing the number clearly affects latency). So, the number could be a fixed value instead of a editable widget, when using JACK (or am I wrong here again?)

Anyway, I agree with not trying to pack too much into that PR. Not allowing the user to screw with timing/pitch by sticking to JACK's advertized samplerate is already a much appreciated change. Thanks for that!

millerpuckette:
Aha... I've noticed that setting 'delay' when on jack seems not to actually
add latency.  I think the FIFO code in s_audio_jack.c isn't working.  This is
a different issue, and it looks like there's at least one bad thing here:

    if (jack_filled >= jack_out_max)
        pthread_cond_wait(&jack_sem,&jack_mutex);

... that 'if' shouild be 'while'.  I'll try to get some time to look at this
later this week.  But not related to this PR, no.  I think this PR is ready
to merge.

cheers
M

reduzent:
Thanks for clearing this up, Miller!

umlaeute:
@millerpuckette et al: i've merged this into the `develop` branch.

reduzent:
I know this already has been merged into `develop` and is closed now. But let me nevertheless mention a recent finding of mine that only affects this specific branch.

When I run pd with JACK backend and save all preferences, i cannot run Pd afterwards at all when there is no jack server running. What happens is that jack is configured in .pdsettings, but no JACK server is available mandating a samplerate. The reported samplerate of Pd is 0 and also cannot be changed by the `-r` flag. When sr is set to 0, basically only the gui works to some degree, but Pd is frozen otherwise. I guess this is a side effect of not allowing to override what JACK is reporting.

In current master, you always get Pd up and running, regardless of preferences, even if JACK is configured and no JACK server is running.

danomatika:
That's a bug. At a minimum, it should set a valid sample rate if Jack is the API and is not running.

umlaeute:
found and fixed the culprit. Pd now runs nicely again, even if JACK is selected but not running.

umlaeute:
as a sidenote: only now i noticed that the underlying assumption behind the question

> Should the blocksize by hidden when using JACK or perhaps show JACK's current bs?


is, that the blocksize in JACK is actually fixed (well, i understood this assumption; only know i noticed that the assumption might actually be wrong).

at least on my system (JACK with dbus interface), it *is* possible to change the JACK blocksize at runtime.

in practice I can change the blocksize in things like e.g. `qjackctl`, and I frequently do that (e.g. when getting dropouts)

just telling. (and i'm actually quite happy that Pd shows the blocksize as fixed when using JACK. i wouldn't want any of my JACK-clients to fuddle with the blocksize, as this parameter affects all the other clients as well)

danomatika:
I believe I added the jack buffer side callback so Pd will know, it’s just that you’d have to close and reopen the audio dialog to see a change.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
sure, it works nicely.

i was merely rtying to add a justification to the future self<kbd>^M</kbd>reader why the `blocksize` is disabled in the audio-dialog even though it need not be fixed *technically*.

reduzent:
> found and fixed the culprit. Pd now runs nicely again, even if JACK is selected but not running.

Confirmed. Thanks for the fix!

PR #1296, opened 2021-04-15
Don't force the pdwindow into the upper-left corner of the screen

umlaeute:
instead let the WM choose

this most likely will put the PdWindow in the center.
on my system (Linux, multi-monitor) it puts the PdWindow in the center of
the currently active monitor; which is nice if the leftmost monitor is
occupied by something else and you actually work on a righthand monitor)

## update

this PR lets the Window Manager (WM) choose the position of the PdWindow ("Pd console") and the *initial* position of a newly created patch window.
given that window managers have invested significant brainpower into how to best place windows for the user (and some of them even let the user configure how they want this window positioning to happen), i think we should rely on their powers rather than inventing our own ad-hoc solution for the problem (which some people - including myself - find irritating).

once a patch window gets saved, the window position is stored within the patch and respected once the patch gets re-opened.

the PR also improves accurate storing/restoring of window positions, regardless of the size of the window borders.
(the only exception is on Windows, where a slim window with a wrapped menu-bar gets wrongly positioned; this has been a known issue for some time and is not a regression; it seems there really is [no general solution for *this* problem](https://stackoverflow.com/questions/67177411))

for those who really want a fixed position of the main PdWindow, a trivial GUI-plugin can do that job.

e.g. putting the following `pdwindow-position-plugin.tcl` into the `extra/` folder (or some other path search by Pd on startup) will restore the original behaviour:

```
if {[winfo exists .pdwindow]} {
    wm geometry .pdwindow  =500x400+20+50
}
```

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -38,16 +38,6 @@ typedef struct _canvas_private
 #define GLIST_DEFCANVASWIDTH 450
 #define GLIST_DEFCANVASHEIGHT 300
 
-/* since the window decorations aren't included, open new windows a few
-pixels down so you can possibly move the window later.  Apple needs less
-because its menus are at top of screen; we're more generous for other
-desktops because the borders have both window title area and menus. */
-#ifdef __APPLE__
-#define GLIST_DEFCANVASYLOC 22
-#else
-#define GLIST_DEFCANVASYLOC 50
-#endif
-
 /* ---------------------- variables --------------------------- */
 
 t_class *canvas_class;
@@ -331,7 +321,7 @@ t_canvas *canvas_new(void *dummy, t_symbol *sel, int argc, t_atom *argv)
     t_canvas *owner = canvas_getcurrent();
     t_symbol *s = &s_;
     int vis = 0, width = GLIST_DEFCANVASWIDTH, height = GLIST_DEFCANVASHEIGHT;
-    int xloc = 0, yloc = GLIST_DEFCANVASYLOC;
+    int xloc = GLIST_DEFCANVASXLOC, yloc = GLIST_DEFCANVASYLOC;
     int font = (owner ? owner->gl_font : sys_defaultfont);
     glist_init(x);
     x->gl_obj.te_type = T_OBJECT;
@@ -488,10 +478,10 @@ t_glist *glist_addglist(t_glist *g, t_symbol *sym,
     x->gl_font =  (canvas_getcurrent() ?
         canvas_getcurrent()->gl_font : sys_defaultfont);
     x->gl_zoom = g->gl_zoom;
-    x->gl_screenx1 = 0;
+    x->gl_screenx1 = GLIST_DEFCANVASXLOC;
     x->gl_screeny1 = GLIST_DEFCANVASYLOC;
-    x->gl_screenx2 = 450;
-    x->gl_screeny2 = 300;
+    x->gl_screenx2 = GLIST_DEFCANVASWIDTH;
+    x->gl_screeny2 = GLIST_DEFCANVASHEIGHT;
     x->gl_owner = g;
     canvas_bind(x);
     x->gl_isgraph = 1;
@@ -1061,8 +1051,9 @@ static void *subcanvas_new(t_symbol *s)
 {
     t_atom a[6];
     t_canvas *x, *z = canvas_getcurrent();
+
     if (!*s->s_name) s = gensym("/SUBPATCH/");
-    SETFLOAT(a, 0);
+    SETFLOAT(a+0, GLIST_DEFCANVASXLOC);
     SETFLOAT(a+1, GLIST_DEFCANVASYLOC);
     SETFLOAT(a+2, GLIST_DEFCANVASWIDTH);
     SETFLOAT(a+3, GLIST_DEFCANVASHEIGHT);
--- a/src/g_canvas.h
+++ b/src/g_canvas.h
@@ -51,6 +51,13 @@ extern "C" {
 #define GLIST_DEFGRAPHWIDTH 200
 #define GLIST_DEFGRAPHHEIGHT 140
 
+#define GLIST_DEFCANVASXLOC 0
+#ifdef __APPLE__
+#define GLIST_DEFCANVASYLOC 22
+#else
+#define GLIST_DEFCANVASYLOC 50
+#endif
+
 /* ----------------------- data ------------------------------- */
 
 typedef struct _updateheader
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -1981,11 +1981,19 @@ void canvas_vis(t_canvas *x, t_floatarg f)
             t_undo *undo = canvas_undo_get(x);
             t_undo_action *udo = undo ? undo->u_last : 0;
             canvas_create_editor(x);
-            sys_vgui("pdtk_canvas_new .x%lx %d %d +%d+%d %d\n", x,
-                (int)(x->gl_screenx2 - x->gl_screenx1),
-                (int)(x->gl_screeny2 - x->gl_screeny1),
-                (int)(x->gl_screenx1), (int)(x->gl_screeny1),
-                x->gl_edit);
+            if ((GLIST_DEFCANVASXLOC == x->gl_screenx1) && (GLIST_DEFCANVASYLOC == x->gl_screeny1)) /* initial values for new windows */
+            {
+                sys_vgui("pdtk_canvas_new .x%lx %d %d {} %d\n", x,
+                    (int)(x->gl_screenx2 - x->gl_screenx1),
+                    (int)(x->gl_screeny2 - x->gl_screeny1),
+                    x->gl_edit);
+            } else {
+                sys_vgui("pdtk_canvas_new .x%lx %d %d +%d+%d %d\n", x,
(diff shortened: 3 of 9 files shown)


umlaeute:
i've asked for feedback about this [on the Pd-list](https://lists.puredata.info/pipermail/pd-list/2021-04/129471.html)

sebshader:
I like where it is and then having custom window-positions be done thru tcl plugins. Imo it would be more inconvenient for new users to constantly have to move it in order to see it at the same time they are patching when starting pd, bc it would be more difficult for them to create tcl-plugins to customize it right when they're getting started.

umlaeute:
> I like where it is and then having custom window-positions be done thru tcl plugins.

the problem being that I think it is practically impossible, to have both: as soon as you tell the WM to override the automatic positioning, you cannot go back.

> Imo it would be more inconvenient for new users to constantly have to move it in order to see it at the same time they are patching when starting pd,

however, i don't understand this at all: why would new users "constantly have to move it around in order to see it"?
what is in effect currently that saves them from this chore?

sebshader:
@umlaeute

> > I like where it is and then having custom window-positions be done thru tcl plugins.
> the problem being that I think it is practically impossible, to have both: as soon as you tell the WM to override the automatic positioning, you cannot go back.
I'm not clear on why not having automatic positioning would be an issue. Ideally there would be an easy way for tcl/tk to interact with multi-monitor setups (I realize this might not be the case right now) so if someone were to change the positioning of the window thru a tcl plugin (like they would if they wanted automatic positioning back) they would be able to specify where they wanted it anyways.
Imo the default behavior should be to put the window at a specified location but on whatever monitor is active.

> > Imo it would be more inconvenient for new users to constantly have to move it in order to see it at the same time they are patching when starting pd,
> however, i don't understand this at all: why would new users "constantly have to move it around in order to see it"?
> what is in effect currently that saves them from this chore?

just when starting pd. Because I assume people would want to patch in the center of their screens and have the console on the side or not in the center so they have more contiguous space for windows. so every time they open pd they would have to move the console to the side.

umlaeute:
>  just when starting pd. Because I assume people would want to patch in the center of their screens and have the console on the side or not in the center so they have more contiguous space for windows. so every time they open pd they would have to move the console to the side.

i don't know about your patching, but for me the current situation is:
- main Pd-window (console) is almost at the top left (`+20+50`)
- any newly created patch window is created at very the top left (Pd-core asks for `+0+50`, but because of window-decoration this effectively sticks the window in the top left corner)

the net result being, that everytime i want to see both the pd-console and my patch window, i have to move either of them.

this sounds very much like the thing you want to save people from - but it's happening to me *right now*, whereas with my patch applied, the two windows are distinct and i see both.

>  Imo the default behavior should be to put the window at a specified location but on whatever monitor is active.

kind of what i'm suggesting: the window ends up at a specified location on whatever monitor is active (the "specified location" being the middle of the screen :-P)

umlaeute:
just for the record, here's two screenshots on how windows are placed.
the screenshots cover my entire desktop to give context.

what i'm doing is:
1. start Pd
2. without touching any window (nor the mouse), open a new file with <kbd>Ctrl</kbd>-<kbd>n</kbd>
3. make a screenshot

## this is what I see with Pd-0.51-4
pdwindow org

## this is what I see with this patch applied:
pdwindow new

clearly the new behaviour is better to work with.

sebshader:
@umlaeute if new patches open in the top left then why don't we have the console open in the top left and have the patches open with the automatic positioning?

umlaeute:
because patch windows are much more problematic, as they get shared (with other people that have different screen layouts) and thus must have a decent default position. (check the mailinglist archives).
i'm pretty sure we do want to save the window position of a patch, as people actually build there interfaces around this.
(and i can't possibly imagine somebody relying on the main pdwindow showing up at `+20+50`)

sebshader:
@umlaeute I think if a patch opens in the top left a user is going to move it around regardless. (so that they don't stack on top of each other when multiple windows are open). People probably mostly will want a set up where they are able to open patch windows while having the console window visible and having a contiguous area for windows that won't obscure the console is the more convenient way to achieve that. This patch would make it so people would have to move 1 more window when they start pd every time they want that, when they are beginning with pd.

To the point about patches having consistent saved position when they're saved: having a default console position in the top left would make it more likely for these saved positions to not obscure the console when they are shared as the middle of the screen can be an arbitrary number of pixels away from the side. Knowing that another user has the same console starting-position would make patches easier to work with when shared since the layout would be less likely to be user-independent or have as much variation between users than if every user has a different default console-pixel position.

Perhaps patch/canvas positions shouldn't be saved.. then at runtime they could open in a way that best suits whatever system is running and it would make patches more compatible with different geometries when they're shared.

porres:
just wanna say I get really annoyed how every new window comes up on the top left corner and. have to drag it

thanks

umlaeute:
it seems you are preventing a fix for a specific problem in order to allow for another solution which - to my knowledge - is not implementable right now (and not in a cross-platform way).
that seems to be a bad deal.

### should we write our own window manager?

removing the explicit window position places the window in the center of the active monitor in *my setup* (xfwm4,...).
this is consistent across many applications (e.g. if i create an xterm, it will be placed in the center of the active monitor).
i'm pretty sure that this can be configured (for my window manager) and is probably different for other window managers.
and this is a good thing. if you write a window manager you should give a lot of thought to this problem (and i supposed authors of window managers do) and their solutions will evolve over time to a (local) optimum. if you don't like the window position strategy choose a different window manager). it's not Pd's business to second guess how people will move their windows to them and how much of a problem this is for *new* users (actually i don't know what's special about "new users". i'm an old user and it annoys me plentily).

### should we leave window positioning entirely to the WM?
this might seem somewhat contradictory to what i just wrote above, but: i think this won't fly.
i'm convinced that (some) people use careful window placement to build their multiwindow  Pd applications  - at least I did so in the past.
removing the possibility to place a window at a given position, would break these patches.

it might be an idea to make absolute positioning configurable through the patch itself though.

reduzent:
I think the order of precedence should be:

1. What the user configures
2. What the WM defaults to
3. What the application defaults to

Since the Pd console is not configured by the user (unless you use a TCL plugin), I think the WM should position the window. 

BTW, on GNOME (Ubuntu 20.04) the Pd window appears on the top left or with some (100px?) margin at the top and the left, depending on the size of the window relative to screen pixel dimensions. So, with current Pd console size on a 1920x1080 screen, the result is almost the same as without the patch.

sebshader:
> it's not Pd's business to second guess how people will move their windows to them and how much of a problem this is for new users

I think it is somewhat an application's responsibility to make application-specific window management decisions if workflow is a factor, and especially for new users that won't have anything configured and are just looking into it. Experienced users can just use a tcl-plugin to put it wherever they want it.
The issue isn't so much second guessing how people will move their windows as it is how the windows pop up in the first place.. (& I feel like many applications make the same decision to have a window in a corner).
Especially in an app like pd where the workflow often involves many differently-sized and differently-placed windows, it's good/utilitarian to have the console window off to the side by default.

it's kinda off-topic but now I'm liking the idea of a preference to have root canvases open up in the "default wm location" and subpatches open over their text-boxes.. maybe I'll open a pr..

umlaeute:
i just checked my settings, and with Xfce4, I can switch the automatic positioning from "center of the active monitor" to "under (or: near) the mouse pointer".

sebshader:
or better yet, what if there were just preferences for:
console window position (could be left blank to default to wm positioning)
if new patches should be placed in the corner or not (if not default to wm positioning)
if root canvases should open with their saved positions (if not use default wm positioning)
if subpatches should open with their saved positions (if not open where clicked/their text boxes are, possibly with settable offset)

reduzent:
What I really like about this patch is that it let's the WM do sensible things like placing the Pd console to an empty space on the screen (I only just found out now). I'm all for letting the WM play its strengths.

umlaeute:
afaict, this patch might introduce a slight regression with regards to window-offset calculation (#108).

umlaeute:
some test reports:

- Debian
- Mojave (VM)
- Wine
  - if the window has a wrapped menu-bar, the position is constant, but the window height shrinks with every `[vis 0, vis 1(`. need to check on real windows and without the patches applied. (this does *not* happen on Debian)

umlaeute:
so for the record, here's screenshots of the original and new automatic screen positioning (now with patch-windows also defaulting to automatic positioning).

all of the screenshots show the following:
1. clear the desktop
2. start Pd
3. create a about 6 new  patches with <kbd>Ctrl</kbd>+<kbd>n</kbd>

## Linux (Debian/Sid, Xfce4)

the actual screen is (much) bigger, so i just grabbed a portion from the upper-left part...

### unpatched
Xfce4/unpatched
### patched

Xfce4/patched

again: actual screen is much bigger, and there's actually 1½ columns of patchwindows missing to the right.

also, I tweaked the windowmanager slightly, lowering the "Minimum size of windows to trigger smart placement" so that Patch-Windows fall under this category (normally they are too small and will just be positioned either "At the center of the screen" or "Under the mouse pointer", depending on the user's choice)


## macOS (mojave)

### unpatched
macOS/unpatched
### patched
macOS/patched

## Windows10
### unpatched
Win10/unpatched
### patched
Win10/patched

umlaeute:

PR #1298, opened 2021-04-16
Release cursor if binbuf_read fails

myQwil:
The cursor is grabbed when Pd expects that a canvas is going to be created, but if `binbuf_read` fails, `proc pdtk_canvas_new` won't get called and the cursor won't be released.

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -2086,5 +2086,6 @@ void glob_open(t_pd *ignore, t_symbol *name, t_symbol *dir, t_floatarg f)
         canvas_vis(gl, 1);
         return;
     }
-    glob_evalfile(ignore, name, dir);
+    if (!glob_evalfile(ignore, name, dir))
+        sys_vgui("::pdwindow::busyrelease\n");
 }


umlaeute:
which problem is this solving (and before i dig into the code: where is the cursor grabbed?)

myQwil:
line 18 of wheredoesthisgo.tcl in `proc open_file`
it calls `::pdtk_canvas::started_loading_file` which in turn calls `::pdwindow::busygrab`

In the event that a seemingly valid patch file fails to load properly, the mouse in its grabbed state won't be able to do anything. It effectively freezes Pd, or at least the main window.

umlaeute:
so far the `m_binbuf.c` file is free of anything related to GUI. let's keep it like that.

wouldn't the proper place for the release be `glob_open` (possibly checking the return value of `glob_evalfile()`)

could you provide a minimal "seemingly valid patch file fails to load properly"?

myQwil:
A simple test would be to change line 866 of m_binbuf.c to say `return (1)` and then open up a patch

umlaeute:
an even simpler test is just to revoke read-permissions from an arbitrary patch file and try to open it.

umlaeute:
i just checked, and I think this is a better fix for the problem:

~~~diff
diff --git a/src/g_canvas.c b/src/g_canvas.c
index 32dc7d8f..b9d34f82 100644
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -2078,6 +2078,7 @@ t_pd *glob_evalfile(t_pd *ignore, t_symbol *name, t_symbol *dir)
 void glob_open(t_pd *ignore, t_symbol *name, t_symbol *dir, t_floatarg f)
 {
     t_glist *gl;
+    t_pd*x;
     if (f != 0)
         for (gl = pd_getcanvaslist(); gl; gl = gl->gl_next)
             if (name == gl->gl_name && gl->gl_env && gl->gl_env->ce_dir == dir)
@@ -2086,5 +2087,7 @@ void glob_open(t_pd *ignore, t_symbol *name, t_symbol *dir, t_floatarg f)
         canvas_vis(gl, 1);
         return;
     }
-    glob_evalfile(ignore, name, dir);
+    if(!glob_evalfile(ignore, name, dir)) {
+        sys_vgui("::pdwindow::busyrelease\n");
+    }
 }
~~~

if you agree, feel free to force-push the change to this PR.

myQwil:
That looks like a good solution. Works on my end.

PR #1306, opened 2021-04-30
respect key repeat system preference on osx in tk 8.6.11

sebshader:
currently in tk 8.6.11 on osx key repeating will occur regardless.

This patch (hopefully) respects the system preference, as the previous keyfix patches have
fixes sebshader/pdnext#4

### DIFF
--- a/mac/patches/tk8.6.11_keyrepeat.patch
+++ b/mac/patches/tk8.6.11_keyrepeat.patch
@@ -1,0 +1,13 @@
+diff --git a/macosx/tkMacOSXKeyEvent.c b/macosx/tkMacOSXKeyEvent.c
+index 1dc39827e..08301b3ef 100644
+--- a/macosx/tkMacOSXKeyEvent.c
++++ b/macosx/tkMacOSXKeyEvent.c
+@@ -255,6 +255,8 @@ static NSUInteger textInputModifiers;
+      */
+ 
+     if (type == NSKeyDown && [theEvent isARepeat]) {
++    // ignore repeats if they are turned off in the system settings
++    if ([NSEvent keyRepeatDelay] < 0) return theEvent;
+ 
+ 	xEvent.xany.type = KeyRelease;
+ 	Tk_QueueWindowEvent(&xEvent, TCL_QUEUE_TAIL);


danomatika:
I will check this later. It would be great if that's the only patch we need for 8.6.11~

sebshader:
@danomatika I did find 1 other small bug but so far it only affects the ceammc library, and shouldn't be very relevant with the rest of pd. (mainly relevant in rendering elements of the "alt" ttk theme)
https://core.tcl-lang.org/tk/tktview/4c7d935e9736a1046a1da3ddd0d9039eb232c943

umlaeute:
hmm. i'm using this patch to test:

~~~
[key]
|
[t b b]
|     |
[timer]
|
[print keydelta]
~~~

on macOS-10.14 (in a VM) with Tcl/Tk-8.6.12 i observe the following numbers when i keep a key pressed:

| "repeat rate" in macOS *System Settings* | keydelta |
|----------------------|--------------|
| set to `fast` | ~33ms |
| set to `slow` | ~2000ms |
| set to `off` | ~2000ms |
| set to `fast` (again) | ~33ms |
| *now* set to `off` | ~33ms |

as you can see, turning the repeat to `off` doesn't really work - instead it uses the last setting.
*however* i observe the same behavior in other apps (e.g. the Terminal)

sebshader:
@umlaeute do you mean this patch adapted to 8.6.12? and when you say the terminal displays the same behavior, how is that measured? some key-logging command-line thing, or through tk?

sebshader:
@umlaeute when you say "in terminal" do you mean some keylogging command-line program, or wish/tk?

sebshader:
@umlaeute I am not getting key repeats using compiled Wish 8.6.12 on MacOs 10.14.6 with the key repeat set to off, using the Wish.app or in Pd using your test patch

umlaeute:
when i say "in the terminal" i mean: `/System/Applications/Utilities/Terminal.app` where i just press and hold a key "on the cmdline".

maybe the behaviour i'm experiencing is somehow related to running in a VM.

and good to see that it works for you with Tcl/Tk-8.6.12: Does this mean that this issue should be closed once we switch to that version?

sebshader:
@umlaeute yes, it seems like we can close with 8.6.12 afaict.
actually, terminal doesn't respect the key repeat set to "off" here either. (but pd with tk 8.6.12 does)

danomatika:
>as you can see, turning the repeat to off doesn't really work - instead it uses the last setting.
>however i observe the same behavior in other apps (e.g. the Terminal)

I remember going over all of this when first making the TK patches and I believe that is the normal behavior. I think I did some tests and also some research online but I honestly have flushed that from my main memory once the release was out. :)

From informal testing here on macOS 11.6.1 and a fresh universal (arm64, x86_64) build of Pd-0.52-0 with Tk 8.6.12, key repeat seems to be working as intended. Thankfully without patches.

umlaeute:
should we close this?
or wait until Tk-8.6.12 is used by default?

danomatika:
I was waiting for @sebshader to close it, but otherwise we could wait until the release is finalized...

sebshader:
I was gonna wait until we actually switch. like #1438 #1441

Issue #1308, opened 2021-05-02
actively prevent duplicate connections


umlaeute:
while the user-interface prevents it, it is actually possible to create multiple connections (e.g. with dynamic patching, but probably with some other glitches as well) between a single pair of iolets.

such as:

~~~pd
#N canvas 6 61 450 300 12;
#X msg 109 102 bang;
#X obj 109 127 print;
#X connect 0 0 1 0;
#X connect 0 0 1 0;
~~~

This is usually a source of great sorrow and hard to find bugs (e.g. #1135, #1307).

I suggest that Pd should either (or both):
- prevent any duplicate connection between a single pair of iolets
- actively notify the user (with a trackable error) whenever such a duplicate connection is (attempted to) being made.

PR #1309, opened 2021-05-03
prevent duplicate connections

umlaeute:
this PR simply checks whether a pair of iolets is already connected, and if so refuses to continue.

Closes: https://github.com/pure-data/pure-data/issues/1308

this also prints some info why connecting failed (at debug-loglevel), complete with clickable printout to easily find the problem.

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -4341,7 +4341,7 @@ void canvas_connect(t_canvas *x, t_floatarg fwhoout, t_floatarg foutno,
 
         /* check if objects are already connected */
     if (canvas_isconnected(x, objsrc, outno, objsink, inno)) {
-        logpost(src, 3, "this pair of iolets is already connected");
+        logpost(src, 3, "io pair already connected");
         goto bad;
     }
 
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -4341,7 +4341,7 @@ void canvas_connect(t_canvas *x, t_floatarg fwhoout, t_floatarg foutno,
 
         /* check if objects are already connected */
     if (canvas_isconnected(x, objsrc, outno, objsink, inno)) {
-        logpost(src, 3, "refusing duplicate connection");
+        logpost(src, 3, "this pair of iolets is already connected");
         goto bad;
     }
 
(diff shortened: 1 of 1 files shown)


src/g_editor.c
```diff
@@ -4320,14 +4320,30 @@ void canvas_connect(t_canvas *x, t_floatarg fwhoout, t_floatarg foutno,
...
-        !(objsink = pd_checkobject(&sink->g_pd)))
-            goto bad;
+        !(objsink = pd_checkobject(&sink->g_pd))) {
+        logpost(src?src:sink, 3, "cannot connect unpatchable object");
+        goto bad;
+    }
+
+        /* check if objects are already connected */
+    if (canvas_isconnected(x, objsrc, outno, objsink, inno)) {
+        logpost(src, 3, "refusing duplicate connection");
```
danomatika, line 4344: The log line works here but I'd find something like "already connected" nice, although perhaps it's too ambiguous?

umlaeute: definitely nicer ;-)

src/g_editor.c
```diff
@@ -4341,7 +4341,7 @@ void canvas_connect(t_canvas *x, t_floatarg fwhoout, t_floatarg foutno,
 
         /* check if objects are already connected */
     if (canvas_isconnected(x, objsrc, outno, objsink, inno)) {
-        logpost(src, 3, "refusing duplicate connection");
+        logpost(src, 3, "this pair of iolets is already connected");
```
danomatika, line 4344: Hah, now I feel like that's too much. Which pair are we referring to? It's not like it will select the two objects or the IOs directly to show the user. Also, would "iolet" be understandable to most people?

Maybe "inlet-outlet pair already connected" or "inlet/outlet already connect" or "inlet/outlet pair already connect"

umlaeute: sure.
i'm offline now. please feel free to force-push.

danomatika: I made a small teak comment to the branch.

src/g_editor.c
```diff
@@ -4341,7 +4341,7 @@ void canvas_connect(t_canvas *x, t_floatarg fwhoout, t_floatarg foutno,
 
         /* check if objects are already connected */
     if (canvas_isconnected(x, objsrc, outno, objsink, inno)) {
-        logpost(src, 3, "this pair of iolets is already connected");
+        logpost(src, 3, "io pair already connected");
```
umlaeute, line 4344: this i actually don't like. the term "io" is nowhere used in Pd, and i think it just adds more confusion.

## more suggestions

this one uses the same "cannot connect ..." pattern as the other error-messages in the same code-block:

> cannot connect multiple times

or simply:

> already connected

umlaeute:

PR #1324, opened 2021-05-25
generate tcl event before window closes

### LINKED ISSUE #1323: generate virtual tk event before patch window is deleted
This would allow objects using tcl to clean up/save before they are actually deleted (add a virtual event to ```wm protocol``` in ::pd_bindings::patch_bindings in pd_bindings.tcl)

sebshader:
This allows tcl/tk gui objects to store their state before they're deleted
closes #1323

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2019,6 +2019,7 @@ void canvas_vis(t_canvas *x, t_floatarg f)
                 canvas_destroy_editor(x);
             return;
         }
+        sys_vgui("event generate .x%lx <<WindowClosing>>\n", x);
         glist_noselect(x);
         if (glist_isvisible(x))
             canvas_map(x, 0);


umlaeute:
i'm not entirely sure why this would be needed.
in your Tcl-code, can't you just do `bind ${winid} <Destroy> {+::myguiplugin::cleanup %W}`?

this will call the `::myguiplugin::cleanup` proc whenever your `${winid}` (or one of it's children) is destroyed.

true, the callback is only called after the Tk widget has been destroyed (so you cannot do something like `${winid} cget -data`), but if your actual state is stored in a regular Tcl variable (e.g. `::myguiplugin::state(${winid})`) then you can still access (and persist) it.

sebshader:
Now this was 5 years ago, but iirc the situation was that it was a 'text' canvas item and I wanted to send the state to the pd process from the text that the tk object had
If I understand the suggestion here, the text state would be duplicated 3 times rather than 2: 1 in pd, 1 in the tk object, and 1 in some kind of class-specific global container keyed on the object ID

umlaeute:
> Now this was 5 years ago

yes i know...

> If I understand the suggestion here, the text state would be duplicated 3 times rather than 2: 1 in pd, 1 in the tk object, and 1 in some kind of class-specific global container keyed on the object ID

there are of course a number of alternatives:
- since we are talking about programmatically closing the window, you could of course just issue an explicit "retrieve data" command in your program (patch).
- just sync your states between Pd-GUI and Pd-core continuously: whenever the user enters something in the text widget, immediately send the new state to the core. if the widget disappears, nothing is lost.


on writing this , I realize that my assumption about "programmatically closing the window" was not true: clicking the "close window" button will trigger the code path that you modified (so I guess this is what you had in mind...)


in any case: I strongly believe that such events are an implementation detail of the GUI. and therefore should be handled soleley on the GUI side. generating the event in Pd-core means that we have to provide this event generation code (with its semantics) for any toolkit we want to use (e.g. the brewing GTK one), which i think is a no-go.

so the proper place to inject such an event generator would be the `pdtk_canvas_delete` proc (which did not exist when you created your PR).

---

apart from my ramblings about where to implement such events, i realize now that I should have commented on #1323 rather than here.

umlaeute:
yet another possibility would be to just override the `destroy` proc in your GUI plugin:

~~~tcl
namespace eval ::myguiplugin:: { }
if { [info commands ::myguiplugin::destroy] eq {} } {
  rename ::destroy ::myguiplugin::destroy
  proc destroy args {
    foreach a $args {
      event generate ${a} <<WindowClosing>>
      ::myguiplugin::destroy ${a}
    }
  }
}
~~~

PR #1340, opened 2021-06-11
[WIP] i18n help

umlaeute:
this is a work-in progress branch to enable i18n help files (that can be provided by the user).

there has been a bit of discussion about this on https://github.com/pure-data/pddp/issues/7

- supports help-patches in multiple languages
- help paths can now be saved in the preferences
- help patches can be dynamically added via a new `add-to-helppath` global message (cf. the old `add-to-path` message)

## language selection
- language selection is based on the `LANG` environment variable (just like the GUI already does).
- for a given `LANG=so_DJ`, Pd will first check whether there is a help-patch `bang-help.so_dj.pd` (`<language>_<country>`), i that fails for `bang-help.so.pd` and finally fall back to `bang-help.pd`
  - locale specifiers are normalized to lowercase; hyphens replaced by underscore; everything after a `.` (typically: the encoding) is ignored. thus `om-KE.UTF-8` becomes `om_ke`
  - if the local specifier does not include the `_<country>` suffix, it won't be searched for

### search order
- similar to the search paths,  there are now three different help paths
  - temporary help paths (added via the `-helppath` cmdline flag)
  - preference help paths
  - static help paths (`@PD@/doc/5.reference`)
 
this is the order a help-path gets searched for:
  - absolute paths
  - relative to current `dir` (as specified when calling `open_via_helppath()`)
  - temporary (cmdline) help paths
  - temporary (cmdline) search paths
  - preference help paths
  - preference search paths
  - static (built-in) help paths
  - static (built-in) search paths

just for reference, this is how it used to be:
  - absolute paths
  - relative to current `dir` (as specified when calling `open_via_helppath()`)
  - temporary (cmdline) search paths
  - temporary (cmdline) help paths
  - static (built-in) help paths
  - static (built-in) search paths

so the changes are:
- `-helppath` cmdline-flag takes precedence over `-path`
- ordinary search-paths (for finding libraries,...) that are added via the preferences are now searched too
- obviously there weren't any "preference" help paths in the original version

### user provided localisation

one of the key ideas was to allow users to add i18n support for a given language, e.g. by downloading a language pack via deken.

this allows the Pd core to not have to worry about maintaining help patches in multiple languages.

such packages should be able to override the default documentation without exposing the gory details to the end user.

for this to work, we need a way to dynamically register new help paths with Pd, hence the new `add-to-helppath` global message.

### binary compatibility
for the static/temporary helppaths the t_pdinstance struct had to be extended with a new member.
this new member `st_private` is a simple `(void*)` pointer that completely hides the implementation from any external, using the PIMPL idiom.
nevertheless, the size of the `t_pdinstance` changes.

there's a seemingly unused member `double st_time_per_dsp_tick` which i would like to reuse (as it has the same or a larger size) for the new `st_private`.
Pd's code is full of comments that it should get rid of the `st_time_per_dsp_tick` member (but assumes that GEM might use it. i've checked the Gem (including its history) and haven't found anything like that).

if we could reuse that member, binary compatibility should be OK.

## Tasks

- [x] load help patches based on their language code
       using `<objectname>-help.<locale>.pd`
- [x] dynamically addable help paths
       so that language packs can install their help patches anywhere, and only need to tell Pd where to look for
- [x] store help paths in preferences
       so that dynamically added help paths persist
       - temporary help paths (added via `-helppath`) and built-in help paths are **not** stored
- [x] help path preferences GUI
      help paths can be saved to (and restored from) the preferences. the "Path" preferences have been enhanced to also allow manipulation of the (stored) help paths.
- [ ] automatic handling of newer versions
      if you have a language pack installed for the help-files and install a new version of Pd (with updated help-patches), you would get the translated & outdated help patches, but you really should get the fresh & non-translated ones.

# Testing binaries

If you want to test this branch but cannot compile yourself, you can get binaries with the `i18n-features` from our nightly snapshots:
- [macOS/64bit](https://git.iem.at/pd/pure-data/-/jobs/artifacts/feature%2Fi18n-help/download?job=macOS)
- [Windows/64bit](https://git.iem.at/pd/pure-data/-/jobs/artifacts/feature%2Fi18n-help/download?job=w64)
- [Windows/32bit](https://git.iem.at/pd/pure-data/-/jobs/artifacts/feature%2Fi18n-help/download?job=w32)

---

PS: please use the  https://github.com/pure-data/pddp/issues/7 ticket (or create  a new ticket at https://github.com/pure-data/pddp/issues/) for brainstorming ideas about i18n help patches.

comments about the specifics of the current implementation are of course welcome here...

### DIFF
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -48,6 +48,8 @@ void x_midi_newpdinstance( void);
 void x_midi_freepdinstance( void);
 void s_inter_newpdinstance( void);
 void s_inter_free(t_instanceinter *inter);
+void s_path_newpdinstance( void);
+void s_path_freepdinstance( void);
 void g_canvas_newpdinstance( void);
 void g_canvas_freepdinstance( void);
 void d_ugen_newpdinstance( void);
@@ -60,11 +62,13 @@ void s_stuff_newpdinstance(void)
     STUFF->st_externlist = STUFF->st_searchpath =
         STUFF->st_staticpath = STUFF->st_helppath = STUFF->st_temppath = 0;
     STUFF->st_schedblocksize = STUFF->st_blocksize = DEFDACBLKSIZE;
+    s_path_newpdinstance();
 }
 
 void s_stuff_freepdinstance(void)
 {
     freebytes(STUFF, sizeof(*STUFF));
+    s_path_freepdinstance();
 }
 
 static t_pdinstance *pdinstance_init(t_pdinstance *x)
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -204,7 +204,7 @@ EXTERN void pdinstance_free(t_pdinstance *x)
     pd_setinstance(x);
     sys_lock();
     pd_globallock();
-    
+
     instanceno = x->pd_instanceno;
     inter = x->pd_inter;
     canvas_suspend_dsp();
--- a/src/m_glob.c
+++ b/src/m_glob.c
@@ -32,6 +32,7 @@ void glob_midi_setapi(t_pd *dummy, t_floatarg f);
 void glob_start_path_dialog(t_pd *dummy, t_floatarg flongform);
 void glob_path_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_addtopath(t_pd *dummy, t_symbol *path, t_float saveit);
+void glob_addtohelppath(t_pd *dummy, t_symbol *path, t_float saveit);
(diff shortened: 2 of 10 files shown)


Lucarda:
Tested on Windows @ 2cfa0a1 and is working fine.

Lucarda:
I also tried to find "white space" troubles in path but couldn't find them.
Tested on Windows10.

umlaeute:
closing this, as i don't think that it will ever be accepted

PR #1346, opened 2021-06-29
Compatibility for watchOS

hbarfu:
I successfully compiled and ran Pd on a watchOS device recently by using a modified libpd wrapper. Had to make some minor changes to the pure-data source code itself too, see "Files changed" tab. These changes theoretically break the GUI on watchOS, but you don't ever use the GUI with libpd anyway, so this shouldn't be a problem IMO. I don't see a way to achieve this without modifying s_inter.c.

### DIFF
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -39,6 +39,7 @@ that didn't really belong anywhere. */
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <glob.h>
+#include <TargetConditionals.h>
 #else
 #include <stdlib.h>
 #endif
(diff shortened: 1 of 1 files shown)


umlaeute:
this got rejected by @millerpuckette  in 2b5c219656771a311f36d8ac826281f153e0ea3e.

i'm re-opening this PR (so it doesn't get lost) until the "different idea" emerges in the repository.

millerpuckette:
People do use libpd with the Pd GUI enabled (at least I do :).  But I'm hitting the same trouble as with this watch-OS, in trying to get Pd running under freeRTOS.  My solution was to rearrange things so that it's not too painful to run Pd with no GUI at all.  (I
also jettisoned all networking, audio, MIDI, FFTs, and some other stuff, in favor of running a bluetooth/serial connection.)

I'd like to start working toward getting all the OS-dependent conditional compilations out of the C code (even if there remain a few in the ".h" files - I don't think it's possible ever to clean them out entirely.)

PR #1348, opened 2021-07-07
html updates.

Lucarda:
This PR includes what has been discussed on https://github.com/pure-data/pddp/issues/9 :

- reworks "Chapter 1"

  - more explicit explanation of the "doc" sub-folders.

  - warns of outdated FLOSS manuals.

- allow the reader to change the font (serif - sans-serif) and the screen mode (dark mode - light mode)

- updated images.

- some other minor fixes

-------------------------

[click here](https://lucarda.github.io/pd-manual-preview-1/) to preview.


millerpuckette:
can this be adopted without the javascript code?  I'm keeping the website passive so that it stays future-proof.

Lucarda:
https://github.com/pure-data/pure-data/pull/1349 gives an alternate without javascript that does not allow to switch "font" or "dark-mode".

> I'm keeping the website passive so that it stays future-proof.

As far as I understand the javascript file is just another file that is given to the browser. The website will stay passive with the javascript file.

umlaeute:
> As far as I understand the javascript file is just another file that is given to the browser. The website will stay passive with the javascript file.

you could also include the js-code directly in the HTML-file (though of course this would mean you have to duplicate it for each and every .html-file which is probably not what we want).

but anyhow: the JS is executed within the client (aka: browser), so the server-side of things will stay *passive*.
whether this means that the "website will be kept passive (or not)" is probably subjective.

Lucarda:
the `.js` file acts similarly as the `.css` file. Both can be included in each HTML file but that is harder to maintain.

I can't get how any implementation of the JS code exceeds the **passive website** threshold.

Anyway Miller can choose this PR or https://github.com/pure-data/pure-data/pull/1349 as he sees fit his needs.


@millerpuckette if you are wondering about the `localStorage` in the javascript code: this is stored in the client's browser, not in the server. (https://developer.mozilla.org/en-US/docs/Web/API/Web_Storage_API)

umlaeute:
> I can't get how any implementation of the JS code exceeds the passive website threshold.

the HTML-code is no longer purely descriptive ("passive") but ("actively") runs code.

the current, dumb HTML-code (supposedly) runs on *Netscape Navigator*, *Chrome* and *SuperMax 3000*.
the same is probably not true if you add JS (you'll have to wait for *SuperMax* to be released, but you could try "Netscape" right now).

Lucarda:
in the particular case of this PR the dumb-HTML is the main structure and the JS just adds/removes a class to each HTML tag. If the browser does not support JS (very unlikely) the dumb-HTML is correctly rendered anyway but without the option to change font or dark-mode.

Lucarda:
Closing this as #1349 has been merged.

Issue #1148, opened 2020-09-22
fast-forward behaves weird when audio is off


umlaeute:
if audio is `off` and we use the `fast-forward` message, then the system will not advance time after the fast-forward until the real-time has caught up.

e.g. consider the following patch:

~~~
[loadbang]               [; pd fast-forward 10000(
|
[metro 1000]
|
[print]
~~~

- initially, we get "bang" every second.
- when clicking the `fast-forward` message, `[metro]` will be fast-forwarded and output 10 bangs immediately.
- after that there will be a 10 second pause (in which the real-time can catch up with the logical time)
- after 10 seconds, the `[metro]` starts to output bangs again.

similarly, banging the 2nd inlet of `[timer]` after the `fast-forward` will output the same value until the real-time has caught up.

this does not happen when audio is on (using a polling scheduler, as observed in #1126)

Spacechild1:
I guess this would also be fixed by moving the loop inside `sched_tick`, as mentioned here: https://github.com/pure-data/pure-data/issues/1126#issuecomment-696722093

Spacechild1:
I'm working on a fix for this and #1126

umlaeute:
afaict this has been fixed in the meantime.

PR #1350, opened 2021-07-13
fast-forward with audio turned off

umlaeute:
as described in #1148, the `fast-forward` message to Pd does not work if the DSP is turned off.

this PR fixes this by special handling the AUDIO_NONE scheduler.


Spacechild1:
This one really looks like a trivial fix to me. Maybe it can go to develop?

Spacechild1:
already fixed in master

PR #1360, opened 2021-07-22
bytestruct: parse lists of bytes into messages according to a struct definition (and vice versa)

umlaeute:
Pd has a couple of objects to convert between binary data and Pd messages (`list {to,form}symbol`, `osc{parse,format}`, `fudi{parse,format}`,...)

this PR adds another - more generic - one: `bytestruct`.
it allows one to specify a memory layout of a [POD structure](https://en.wikipedia.org/wiki/Passive_data_structure) via a format-string, and it will turn lists of bytes into an atom-list according to this definition.

it is closely modelled after the [Python `struct` standard library](https://docs.python.org/3/library/struct.html)

as with `text`, `array`, `list` & `file`, the `bytestruct` uses verbs to distinguish between different modes of operation:

- `[bytestruct pack]` turns a general atom-list into a bytelist
- `[bytestruct unpack]` turns a bytelist back into a general atom-list
- `[bytestruct size]` returns the number of bytes required to represent the structure

the format specification is rather flexible and can express, integers (8bit, 16bit, 32bit, 64bit; all signed and unsigned), floats (16bit(!), 32bit, 64bit), strings (wchar[], char[], pascal strings), bools, bytes and padding bytes (that are present in the binary representation but are ignored otherwise).

### limitations
the struct format string, describes a fixed-sized struct. while this is quite flexible, it does not lend itself to variable size-data (zero-terminated strings, run-length encoded data).

### examples
the help-patch comes with some examples:
- creating/parsing OSC-bundles
- reading a WAV file info (using the `fmt` chunk)
- reading loops from a WAV-file (using the `smpl` chunk)


umlaeute:
note: this PR also includes #1361 (in order for the examples of parsing WAV-files to become readable)

umlaeute:
@millerpuckette should this PR be closed (or merged)?

`[bytestruct]` kind of a wishlist object that would be nice to have, but is of course a bit beyond the scope of Pd itself.
i would like to get rid of open PRs (and just drop those that have little chance to get merged ever)

Ant1r:
Personally, I would prefer it to be merged :-) 
When dealing with communication with hardware (which I don't consider as being beyond Pd's scope), it could be really handy.

umlaeute:
closed in favour of #2288

PR #1361, opened 2021-07-22
bugfix/file: improvements and bugfixes for the new [file] object

### LINKED ISSUE #1358: file handle: read with a list fails
when in read-mode, `[file handle]` expects a single float containing the number of bytes to read.

however, when sending a `list` message with a single-float atom (e.g. `[list 12(`) i get an error on the console but not an error-bang on right-hand output:
> error: no way to handle 'list' messages while reading file

since a single `[trigger anything]` will silently turn a `float`-message into a `list` message, this is rather awkward:

automatically reading a file might turn into an endless loop (as the read does not proceed  and no EOF is ever emitted)

~~~
[open foo.txt(
|
[t b     b               a]
|        |      [r stop] |
|        [until]         |
[close(  |               |
|        [float 1(       |
|        /               /
| +-----+---------------+
|/
[t a]
|
[file handle]
|           [s stop]
~~~

umlaeute:
of course there are always bugs, even if you test a lot.

this PR fixes #1358 

also, when working on the `bytestruct` object (https://github.com/pure-data/pure-data/pull/1360), I noticed that the current interface made working on datasets that need to be parsed incrementally a bit awkward, as it requires a lot of going forward and backward in the patch.

### the problem

e.g. here's a patch that parses (parts of) a WAV-file (extracting information from the `fmt ` and the `smpl` chunks):

loop-bad

note that there's some code hidden in subpatches to make the patch "cleaner"

### a solution
so this PR adds the possibility to share a file-handle between multiple `[file handle]` objects, making the result much easier to read (i think):

loop-nice

note the (new) `[file define]` object that *owns* the actual file-handle.

the patch is a bit larger than the ugly other one, as it shows the *entire* code.
obviously it is easier to follow the data flow.

### non-advanced use-cases
it's still possible to use a `[file handle]` (without the shared label) as a standalone object to do all the operations on a single file.

### DIFF
--- a/src/x_file.c
+++ b/src/x_file.c
@@ -215,11 +215,17 @@ static const char*do_errmsg(char*buffer, size_t bufsize) {
 }
 #endif /* !_WIN32 */
 
-
+typedef struct _file_handler {
+    int fh_fd;
+    int fh_mode; /* 0..read, 1..write */
+} t_file_handler;
+#define x_fd x_fhptr->fh_fd
+#define x_mode x_fhptr->fh_mode
 typedef struct _file_handle {
     t_object x_obj;
-    int x_fd;
-    int x_mode; /* 0..read, 1..write */
+    t_file_handler x_fh;
+    t_file_handler*x_fhptr;
+    t_symbol*x_fcname; /* multiple [file handle] object can refer to the same [file define] */
 
     mode_t x_creationmode; /* default: 0666, 0777 */
     int x_verbose; /* default: 0 */
@@ -264,6 +270,7 @@ static void do_parse_args(t_file_handle*x, int argc, t_atom*argv) {
     t_symbol*flag_m = gensym("-m");
     t_symbol*flag_q = gensym("-q");
     t_symbol*flag_v = gensym("-v");
+    x->x_fcname = 0;
     while(argc--) {
         const t_symbol*flag = atom_getsymbol(argv);
         if (0);
@@ -289,7 +296,12 @@ static void do_parse_args(t_file_handle*x, int argc, t_atom*argv) {
                 x->x_creationmode = mode;
             }
         } else {
-            pd_error(x, "unknown flag %s", flag->s_name);
+            int filearg = (!argc);
+            if(filearg) {
+                x->x_fcname = (t_symbol*)flag;
+            } else {
+                pd_error(x, "unknown flag %s", flag->s_name);
+            }
             break;
         }
         argv++;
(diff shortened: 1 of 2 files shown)


src/x_file.c
```diff
@@ -1411,7 +1502,8 @@ void x_file_setup(void)
         gensym("close"), 0);
     class_addmethod(file_handle_class, (t_method)file_handle_seek,
         gensym("seek"), A_GIMME, 0);
-    class_addfloat(file_handle_class, file_handle_float);
+    class_addmethod(file_handle_class, (t_method)file_handle_set,
+        gensym("set"), A_DEFSYMBOL, 0);
```
Spacechild1, line 1506: Maybe it would be more idiomatic to set the symbol via a second inlet to match the syntax of the `[text]` and `[array]` objects.

umlaeute: actually there *is* a 2nd inlet to set it.

Spacechild1: Ah, sorry, now I see it.

src/x_file.c
```diff
@@ -1308,7 +1358,38 @@ t_class *file_split_class,*file_join_class,*file_splitext_class, *file_splitname
...
+   x->x_fhptr = &x->x_fh;
+   x->x_fh.fh_fd = -1;
+   x->x_canvas = canvas_getcurrent();
+   x->x_creationmode = 0666;
+   x->x_verbose = 0;
+   if(1 == argc && A_SYMBOL == argv->a_type) {
+       x->x_fcname = atom_getsymbol(argv);
+       pd_bind(&x->x_obj.ob_pd, x->x_fcname);
+   } else {
+       pd_error(x, "%s requires an argument: handle name", s->s_name);
```
Spacechild1, line 1378: This will print an error message whenever you open the help patch :-) `[text define]` and `[array define]` wouldn't complain and just set to an empty symbol.

umlaeute: true.
but with `[text define]` resp `[array define]` you can actually *do something* with the object, even if there's no name attached to it.
whereas `[file define]` is useless without a name.

but yes, i probably could change this...

umlaeute: or change the help-patch, to use `[file define $0.file]` instead...

Spacechild1: > but with [text define] resp [array define] you can actually do something with the object,

Yes, you can do *something*, but you can't do anything meaningful. All three objects expect a symbol to be actually useful, but only `[file define]` would complain if the symbol is missing. It's not too important, just an inconsistency.

umlaeute: i've added an argument to the help-file so it doesn't through any warnings no more

PR #1364, opened 2021-08-04
added x-puredata.xml file

mxa:
add this file to the system by

`xdg-mime install linux/x-puredata.xml`

this command should be added to the makefile if it is run on linux.


umlaeute:
merged into `develop` (unfortunately i merged before i set the target-branch, so I have to close manually)

PR #1366, opened 2021-08-16
add float method to [random] to allow setting the range and generating a value

sebshader:
This adds a float method to random in order to be able to simply send a float to the left inlet to set the range and immediately generate a value.
motivated by https://forum.puredata.info/topic/13592/random-hot-inlet-doesn-t-allow-float

### DIFF
--- a/doc/5.reference/random-help.pd
+++ b/doc/5.reference/random-help.pd
@@ -1,14 +1,13 @@
 #N canvas 490 69 497 442 12;
-#X msg 54 217 bang;
-#X obj 54 306 random 5;
-#X floatatom 107 279 0 0 0 0 - - -;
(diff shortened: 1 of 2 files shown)


jamshark70:
This issue came up in a lesson yesterday, whereupon I came up with this demonstration:

pd-random

That is, there is precedent within Pd itself for *not* requiring the bang to be sent separately. (Or, if there's a concrete reason why [delay] allows it and [random] doesn't, I find it difficult to discern.)

millerpuckette:
It's added now (with new commit to avoid conflicts in this one)

porres:
cool, let's close this then

porres:
cool, let's close this then

PR #1371, opened 2021-08-31
don't prefix [print] ouput with "verbose(2)" when sending to stderr or via printhook

Ant1r:
In https://lists.puredata.info/pipermail/pd-dev/2021-08/022830.html , Roman Haefeli reports that the [print] object printout is garbled with the "verbose(2): " prefix when printing to stderr.

This is due to the "trackable print" addition (see #464).

Here is a possible fix, that just restores the original [print] behavior when sending to stderr or using a printhook.

### DIFF
--- a/src/x_interface.c
+++ b/src/x_interface.c
@@ -5,10 +5,19 @@
 /* interface objects */
 
 #include "m_pd.h"
+#include "s_stuff.h"
 #include <string.h>
 
 /* -------------------------- print ------------------------------ */
 static t_class *print_class;
+
+  /* avoid prefixing with "verbose(2): "
+  when printing to stderr or via printhook. */
+#define print_startlogpost(object, level, fmt, args...) do{ \
+    if (sys_printhook || sys_printtostderr) \
+        startpost(fmt, args); \
+    else startlogpost(object, level, fmt, args); \
+} while(0)
 
 typedef struct _print
 {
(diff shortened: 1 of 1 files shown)


Spacechild1:
I think the long term goal we also want to filter the console output by its verbosity level, so I don't think it's good to fallback to `startpost`.

As I've already suggested: why not add another API method `logatom` which mirrors the behavior of `postatom`? After all, printing atoms is exactly what `[print]` does.

Spacechild1:
> So when the ouput is redirected to stderr, every "atom" is prefixed with "verbose() :".

`logatom` would simply have no prefix. That's perfectly fine because you would start with `startlogpost` anyway.

> BTW: what should be the right loglevel for [print]?

I'm pretty sure "Normal" (as it currently is).

> Does the [print] output really need to be filtered?

That's what we already do when setting the log level in the GUI. "Error" won't show any output of `[print]`.

Spacechild1:
I just realized we probably also need `endlogpost`.

Here's an example:
```
t_atom msg[3];
for (int i = 0; i < 3; ++i) {
    SETFLOAT(&msg[i], i+1);
}

startlogpost(x, 0, "foo:");
logatom(x, 0, 3, msg);
endlogpost(x, 0);
```

This would print `verbose(0): 1 2 3` to the console.

Spacechild1:
BTW, we could probably go ahead and also add `logfloat` and `logstring` for the sake of completeness. Also, we could rearrange the declarations in `m_pd.h` to make the structure more clear:

```
EXTERN void post(const char *fmt, ...);
EXTERN void startpost(const char *fmt, ...);
EXTERN void poststring(const char *s);
EXTERN void postfloat(t_floatarg f);
EXTERN void postatom(int argc, const t_atom *argv);
EXTERN void endpost(void);

EXTERN void logpost(const void *object, const int level, const char *fmt, ...)
    ATTRIBUTE_FORMAT_PRINTF(3, 4);
EXTERN void startlogpost(const void *object, const int level, const char *fmt, ...)
    ATTRIBUTE_FORMAT_PRINTF(3, 4);
EXTERN void logstring(const void *object, int level, const char *s);
EXTERN void logfloat(const void *object, int level, t_floatarg f);
EXTERN void logatom(const void *object, int level, int argc, const t_atom *argv);
EXTERN void endlogpost(const void *object, int level);

EXTERN void verbose(int level, const char *fmt, ...) ATTRIBUTE_FORMAT_PRINTF(2, 3);
EXTERN void bug(const char *fmt, ...) ATTRIBUTE_FORMAT_PRINTF(1, 2);
EXTERN void pd_error(const void *object, const char *fmt, ...) ATTRIBUTE_FORMAT_PRINTF(2, 3);

Ant1r:
I think `endlogpost()` isn't needed (`endpost()` is enough).
Can you show your `logatom()` (and other ones) implementation? They don't call `dologpost()`, right?

Anyway do you think the initial `verbose(x):` prefixing is tolerable? I mean, it will force people that are parsing stderr to adapt their script or code.

Spacechild1:
> I think endlogpost() isn't needed (endpost() is enough).

We need `endlogpost()` to match the verbosity of `startlogpost`. Otherwise we might get missing newlines with "Error" or "Critical".

> Can you show your logatom() (and other ones) implementation? They don't call dologpost(), right?

They are still in my head :-) But I can also write them myself, I'm just busy till next week.

> Anyway do you think the initial verbose(x): prefixing is tolerable? I mean, it will force people that are parsing stderr to adapt their script or code.

Good question. I see what you mean. `[print]` used to post without prefix, but now it would start with `verbose(2)`. I'm not sure...

Ant1r:
> `[print]` used to post without prefix, but now it would start with `verbose(2)`

yes... 
That's why I think the current PR is OK for now, it just fixes the original issue in quite a simple way without adding new public functions. IMO it should be merged before the next release (alternatively I could revert the whole "trackable print" thing, but it would be sadly...).

And we could revisit all this while solving the verbose/logpost conflict.
It could be worth it to add a Tcl `::pdwindow::continue_logpost` method, to reduce the memory cost of logging long list of atoms.

I'm thinking that I could remove the `startlogpost()` declaration from `m_pd.h`, so that the function can be removed later when a better solution is found. If it stays public, an external developer could start using it...

Ant1r:
so I hid `startlogpost()`.

@millerpuckette I think this can be merged.

Ant1r:
@umlaeute maybe this could be merged into `develop`, to try to avoid the [print]->stderr regression in the release to come? (I'm feeling guilty because I was the one who caused this regression ;-)

PR #1381, opened 2021-09-16
fix MSC compilation for the recent addition of [trace] (fixes CI checks)

Ant1r:
Now that `m_obj.c` uses `snprintf`, it must define it as a shortcut for `_snprintf` in the case of MSC, just like e.g m_binbuf.c:
```
#ifdef _MSC_VER
#define snprintf _snprintf
#endif

```


umlaeute:
thanks.

i've cherry picked this into the `develop` branch https://github.com/pure-data/pure-data/pull/1354, so i'm closing this PR in favour of the other one ;-)

PR #1383, opened 2021-09-26
h/vslider-help: Correct description of logarithmic range

jamshark70:
Updates a minor detail in h- and v-slider help. The subpatch about linear or logarithmic ranges claimed that a log range must be positive. This isn't true.

The real condition is that `hi / lo` must be positive, which is the case if both numbers are positive *or* both numbers are negative (and of course non-zero). A slider with a range e.g. -100 to -1 works perfectly well.


porres:
nice, you should also update the help file of [nbx] for that matter

porres:
I will include this in my documentation updates branch, along with 'nbx', so you may close this

porres:
hi, we can close this as my documentation branch got merged and included this!

PR #1389, opened 2021-10-04
fix t_loglevel usage

Spacechild1:
Changes:
* `t_loglevel` is meant to be used with `logpost()` instead of `verbose()`
* deprecate `verbose()` in favor of `logpost()`
* replace all `verbose(...)` calls in the code base with `logpost(NULL, ...)`

We have already discovered with horror that `verbose()` and `logpost()` use different values for the log level. Unfortunately, there is no good overall solution to make `t_loglevel` work with both functions, so I would just go ahead and only allow it for `logpost()`.

Also, I went ahead and marked `verbose()` as deprecated, since `logpost()` is the newer (and better) function. I'm wondering if we should even remove it from `m_pd.h`, just like Miller has done with `error()`...

### DIFF
--- a/src/s_print.c
+++ b/src/s_print.c
@@ -146,7 +146,7 @@ static void dologpost(const void *object, const int level, const char *s)
     }
 }
 
-void logpost(const void *object, const int level, const char *fmt, ...)
+void logpost(const void *object, int level, const char *fmt, ...)
 {
     char buf[MAXPDSTRING];
     va_list ap;
@@ -256,6 +256,7 @@ void error(const char *fmt, ...)
 }
 #endif
 
+/* deprecated in favor of logpost() */
 void verbose(int level, const char *fmt, ...)
 {
     char buf[MAXPDSTRING];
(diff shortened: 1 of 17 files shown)


src/s_print.c
```diff
@@ -146,7 +146,7 @@ static void dologpost(const void *object, const int level, const char *s)
     }
 }
 
-void logpost(const void *object, const int level, const char *fmt, ...)
+void logpost(const void *object, int level, const char *fmt, ...)
```
umlaeute, line 149: i know that the `const` qualifier for the level is of little actual use, but is there a specific reason to remove it?

umlaeute (commented):
looks good
(there's a wee comment on the `const` qualifier for the level-argument of `logpost()`, but that's more a question than a real issue)

umlaeute (approved):
oops, pressed the <kbd>Comment</kbd> button too fast. it should have been <kbd>Approve</kbd>...

src/s_print.c
Spacechild1: because there is no reason for it to be there in the first place :-)

PR #1390, opened 2021-10-04
stable development branch

umlaeute:

it supersedes #1354 

---

- Update French translation
- Update German translation
- Use shared Travis-CI configuration (so we no longer need to molester miller whenever Travis-CI decides to break)
- Fix doc/8.topics/expr.html


Spacechild1:
@umlaeute I have a couple of no-brainer fixes for Pd 0.52-test1, some of them critical. Should I push them to develop or make seperate PRs to master? I guess the former would be more practical.

umlaeute:
Yes: just push 'develop' and create a new PR.

thx

PR #1397, opened 2021-10-13
gui preferences

umlaeute:
this merges the various preferences (Startup, Path, Audio, MIDI,...) into a single multi-tabbed preferences window available via "Preferences -> Edit Preferences..." where you can add all the various settings.

the preferences dialog also includes new settings to
- set whether new windows are started zoomed or not (this used to be a *non-persistent* tick-box, in which place the new "GUI..."-preferences show up )
- select language (for menus)
  - languages can be one of the existing localisations, or the system default or none at all (defaults to the language Pd is written in, thus English)
- change the preference layout to non-tabbed

### Issue Tracker
- Closes: https://github.com/pure-data/pure-data/issues/1423
- Closes: https://github.com/pure-data/pure-data/issues/1446
- Closes: https://github.com/pure-data/pure-data/issues/1955


Lucarda:
> I had an error in the Pd console complaining about the "zoom new window" flag on the first "Apply" push (and I think the toggle was grayed out, like if it wasn't neither 'true' nor 'false') but I cannot make it appear again, even after deleting my .pdsettings file.

Same here on Windows (only the first push).

Ant1r:
a small remark about `pd_i10n.tcl`: isn't it "**i18n**" (for internationalization=i+18letters+n) versus "**l10n**" (for localization=l+10letters+n)?..

umlaeute:
doh.

umlaeute:
DoH.

i always confuse the two.
so which one are we actually talking about here?

danomatika:
O.o Popup? Prepare for the pitchforks...

umlaeute:
call it a 2nd-level-dialog then ;-)

umlaeute:
regardless of the actual naming of "default language" vs "default", "system language" or "baggage claim", i think the current implementation is almost good to go.

### what needs to be done yet?
#### issues
- `ttk::notebook` currently crashes on macOS. wtf?
- ...
#### improvements
- remember last tab across restarts
- select audio/midi backend from within preferences (see #1423 )
- ...

Ant1r:
very cool! All seems to work fine here, and does look great. I hope it will on other OSes as well!
congrats and thanks

Lucarda:
On Windows (amd64)

  - It took me a while to understand settings when not using "use tabbed preferences" (i now understood it and can live with it (or not because I prefer "tabbed" preferences", :)
  
  - I'm getting troubles selecting ASIO devices: when I select the input I get in the output devices: "(same as input device)...". Since I can't change the output device I get in the console `error opening audio: Illegal combination of I/O devices`.
  
    - can this be reverted. I mean the "(same as input device)..." ?

umlaeute:
> It took me a while to understand settings when not using "use tabbed preferences" (i now understood it and can live with it (or not because I prefer "tabbed" preferences", :)

i wonder what the trouble was?
- the default is to use tabbed preferences
- there's an option that allows you to use scrolled preferences instead of the tabbed ones
- you don't like scrolled preferences (fine - just use the default; i don't like MIDI, so i use OSC and don't bother about the ugliness of MIDI)

so what's the issue? (in the end, the checkbox will probably go away; being able to disable the tabbed preferences was initially an attempt to mediate some crashes on macOS, but that failed; in the meantime i'd like having a quick way to test the fallback scenario is your tcl/tk version does not provide a tab-container)

> when I select the input I get in the output devices: "(same as input device)..."
> -  can this be reverted. I mean the "(same as input device)..." ?

i noticed this too (but since i don't use windows i just assumed that this is the way things are in that land, and happily headed back to my linux world). your comment indicates that this is instead a regression.
just to be sure: with the normal Pd-0.52-0test0 (current `master`) you can select different input and output devices when using ASIO, right?

Lucarda:
> i wonder what the trouble was?

> - there's an option that allows you to use scrolled preferences instead of the tabbed ones.

yes. it was just me. I didn't understood the scroll and thinking "how do I: go back to "tabbed", how do I: set "startup prefs", ... " 

:)

>  with the normal Pd-0.52-0test0 (current master) you can select different input and output devices when using ASIO, right?

Yes.

danomatika:
>yes. it was just me. I didn't understood the scroll

Maybe the option should be a radio button set with both options spelled out ie. Preferences presentation: 1. tabbed 2. scroll.

Also, @umlaeute do you have details on the macOS crashes? I remember seeing some commits in the Tk 8.6.12-rc branch related to ttk::notebook issues. I believe this is something that is likely fixed when 8.6.12 is released, hopefully soon. In the meantime, I can do some testing on my system.

umlaeute:
> Hopefully this fixes the jp device naming issues in the audio & midi dialogs.

yes, it should.
i haven't managed to get a japanese Win machine to run, but I tested on a chinese Win7 and it seems to work fine now

tcl/pd_i18n.tcl
```diff
@@ -56,7 +56,11 @@ proc ::pd_i18n::get_system_language {} {
     } else {
         set lang ""
     }
-    return [file rootname [string tolower $lang]]
+    set lang [file rootname [string tolower $lang]]
+    if { $lang == "c" } {
```
danomatika, line 60: good catch, I can see "c" causing problems...

umlaeute: ?
i committed this mostly for cosmetic reasons...

danomatika: I was imagining a platform that might not match the locale due to case-sensivity (or lack thereof). ;)

porres:
this got merged and I'm testing my build, but when I try to go to Preferences --> Edit Preferences, Pd crashes. 

I'm on a 2013 macbook air mojave 10.4.6

oh, and I get one of these...

Screen Shot 2023-06-04 at 17 22 31

umlaeute:
argh. 
- which TclTk is this?
- is this self-compiled, or are you using one of the pre-built binaries?

some TclTk versions are known to crash with the ttk-notebook (which is used for tabbed preferences).

you can disable it via the preferences (catch 22).
you can also disable it by setting the `use_ttknotebook` GUI-preference to a value of `0`.

This is most easily done with the tclprompt-plugin, where you simply have to type the following in the tclprompt:
```
::pd_guiprefs::write "use_ttknotebook" 0
```

you could also write a GUI-plugin that calls this (once).

this is obviously only to get your Pd back up and running for now.
we need a better solution for Pd-0.54 like:
- find out which versions of TclTk do crash (and which don't), and blacklist these versions
- allow the user to disable (and probably enable) tabbed preferences without having to use the preferences (or some GUI-plugin), e.g. by using a checkbox-menu.

porres:
self-compiled in the worst way i guess

porres:
> this is obviously only to get your Pd back up and running for now.

it runs fine if I reopen, I only can't go to preferences.

now, wouldn't it be nice if we could provide 'nightly builds' from this repo with the correct pre built binaries and everything?

sebshader:
@porres can you give the specific steps you take to compile and package/run pd? I feel like some similar issues have come up in the past (e.g. our [keycode] external failing).
And what tcl/tk version does it say?

porres:
log 4 gives me

------------------ done with main ----------------------
connection from 'pd' to 'pd-gui' on 127.0.0.1:56219
Tk 8.6.10
detected font: DejaVu Sans Mono
using font: Menlo normal
the Pd window filtered 5 lines
Loading plugin: /Applications/Pd-0.54.0.app/Contents/Resources/Scripts/pd_deken.tcl

danomatika:
Should be Tk 8.6.12. Do you build with “make app”, if so perhaps the prebuilt wish.app is out of date.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

Lucarda:
Just some information for Windows:

I had tested this with `Tk 8.5.10` (the one that gets extracted from `/msw/pdprototype.tgz`) and it works fine.

umlaeute:
> it runs fine if I reopen, I only can't go to preferences.

sure.
but that is bad enough.

> 
> now, wouldn't it be nice if we could provide 'nightly builds' from this repo with the correct pre built binaries and everything?

as in https://git.iem.at/pd/pure-data/-/jobs/artifacts/master/download?job=macOS ? 
- it's not exactly "nightly" but gets updated whenever there's some changes on the `master` branch
- this uses the same infrastructure as the signed releases for macOS
- it's been there since miller offers signed releases for macOS (the signed releases are taken from this very CI)

danomatika:
That doesn’t surprise me. The Tk focus on Windows was higher and there were issues with earlier version of Tk on macOS. Thankfully development has picked up and Tk 8.6.10+ seems to be much improved. On newer systems, though we want 8.6.12 as a minimum and should check out 8.6.13 which I believe is released now.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
i knew i fixed this once: https://github.com/pure-data/pure-data/blob/4126532807d17f8f60e367446882d2f149950054/tcl/pd_deken.tcl#L1645-L1652

i'll submit a patch for the `develop` branch in a second.

porres:
yeah, I built with “make app”

porres:
> as in https://git.iem.at/pd/pure-data/-/jobs/artifacts/master/download?job=macOS ?

cool, where is the general "link" for all OSes?

umlaeute:
https://git.iem.at/pd/pure-data/-/pipelines

millerpuckette:
I think this is successfully merged.

PR #1403, opened 2021-10-21
[cputime] and Windows

Lucarda:
This PR makes [cputime] work again on Windows. It was showing a 'couldn't create...' when opening 'list of objects'. Not sure which commit broke it.

Note: it still have the old behavior that gives values only when DSP is on (or only the very first time a bang gets in the right inlet if this was banged prior to any left bang).


Lucarda:
ping @Spacechild1 

This can go to 'develop' ?

it's a no brainer.

Spacechild1:
sure enough :-)

Issue #1384, opened 2021-09-29
No audio on macos with jack


fdch:
Well, these are two cases where there is no audio with `jackd -d coreaudio -r 48000` (jack version 1.9.19 installed from homebrew on Mojave 10.14.6 ). 

- Using Miller's Pd-0.51-4, I get `Can't open Jack (it seems not to be installed)` on the Pd Console.
- Using a compiled [1] pd from latest git. Pd connects to Jack (after I turn on DSP I see the pure_data jack ports) [2], but there is no dsp processing on the Test Audio and Midi patch. 

Both cases are either with connecting to Jack from the `Media > Jack` menu or running pd like `./path/to/bin/pd -jack -channels 2 -r 48000`. In both cases, portaudio works.

[1] configured like this:
```
./configure --enable-fftw --disable-jack-framework --enable-jack
```
[2] to my surprise, I noticed that pd jack ports now start at 1, not 0 as before...

fdch:
>  does Pd still react otherwise? 
Yes, it does.
> are `callbacks` turned on or off?
When `callbacks` are turned off, it does not work. Otherwise, there is audio!

fdch:
One more thing. As [this comment](https://github.com/pure-data/pure-data/issues/1190#issuecomment-790190424) says on the issue you mention, when starting with `brew services start jack`, pd works.

However, callbacks act funny here:
- callbacks off: normal sine tone on the Test Audio and Midi patch
- callbacks on: sine tone shifts pitch down, almost a half-step

umlaeute:
i'm not sure i can make sense of the last two comments.
it seems that:
- *everything*¹ works if you start jack with `brew services start jack`
- if you don't start jack, it works with `callbacks: ON` but not otherwise (i have no idea why you would get sound if JACK is not running)

> callbacks on: sine tone shifts pitch down, almost a half-step

that looks a lot like a 44.1kHz vs 48kHz issue

fdch:
> - if you don't start jack, it works with `callbacks: ON` but not otherwise (i have no idea why you would get sound if JACK is not running)
> 

No, that is not the case. When I start jack with `jackd -d coreaudio -r 48000` and then pd with `pd -jack -channels 2` I get audio when `callbacks` are ON, but no audio when `callbacks` are OFF. 


> that looks a lot like a 44.1kHz vs 48kHz issue

yes, but I did not change the sample rate :)

umlaeute:
you say:

[...]

Issue #1374, opened 2021-09-05
Opening Pd document on Mac


mmturner2468:
After updating to version 0.51-4, I find that when I double click on a pd document (e.g., my_file.pd), the program opens the console but the patcher window itself does not. I need to double click the _my_file.pd_ document a second time for it to launch. 

Pd acts as expected once I've gotten the document open. 

I'm using: MacBook Pro (13 inch, M1, 2020), MacOS 11.5.2, Pd 0.51-4

Thanks for any help,
Mitch

umlaeute:
> The resolution seems to be that this is an issue on macOS 10.15 but is fixed with newer versions of Tk and macOS 11:

hmm, i don't see this on macOS-10.15/Catalina with Pd-0.51-4: the patch opens just fine.

danomatika:
I have a universal test build of (prerelease) 0.52-0 with TK 8.6.12:

http://docs.danomatika.com/pdbuilds/0.52/Pd-0.52-0-universal.zip

On my M1 MBP running macOS 11.6.1, I don't see this issue any more. It thankfully seems to have been fixed with 8.6.12.

danomatika:
This should be fixed in Pd 0.52. Can you double-check and close this issue if everything is working?

mmturner2468:
Hi, Dan. Yes, the Pd 0.52 fixed my issue. Will close the ticket today. 

Thanks for getting this solved.

— Mitch

Issue #1303, opened 2021-04-22
Right-click menu disappears/strange mouse behavior occurs when moving canvas windows to different monitor/desktop (MacOS)


thouldcroft:
I've come across a strange behavior with canvas windows when moved to a different desktop/monitor on MacOS. I discovered this when moving Pd windows to a different monitor for teaching. When you move the window to a desktop/monitor that the program didn't originally open on, you lose the ability to right-click on an object and choose from the pop up menu. When the window gets moved back to the original screen, the mouse appears as if you are click-holding and creating  a selection rectangle when you move the mouse into the window, however, right-click functionality still works on the original desktop/monitor. 

Interestingly, if you tell Pd to open on the other desktop/monitor, the right-click functionality works; so only when moving the canvas window to a desktop/monitor that Pd didn't originally open on, does this problem happen.

emviveros:
I can confirm this misbehavior in MacOS 10.13.6 using Pd 0.51-4

umlaeute:
could you please perform some testing, with the following patch, e.g. `test.pd`?

~~~
#N canvas 889 157 450 300 12;
#N canvas 5 51 450 300 test 0;
#X restore 180 162 pd test;
~~~

- start Pd from the cmdline using something like `/Applications/Pd-0.51-4.app/Contents/Resources/bin/pd -d 3 test.pd 2>pd.log`
- right click the `[pd test]` subpatch
- close Pd

this will write a file `pd.log` that contains a lot of debugging information. please follow the steps above exactly an ddon't do anything more (the file is full of stuff, and it's already hard to find the relevant information).

repeat the procedure on the "other screen" (make sure to not overwrite the original `pd.log` file, by either renaming the first or using another name of the second run).

in the end you should have two logfiles, one where right-clicking did work, and one where it did not.
(clean the two files of uninteresting information,) zip them and post them here.

thouldcroft:
I've never created a Pd log before, so I wasn't quite sure what data to clear, the two logs seem to be identical up to `pdtk_canvas_getscroll .x10072fcc0.c` so I cleared everything before that.
[Pd-Logs.zip](https://github.com/pure-data/pure-data/files/6367404/Pd-Logs.zip)

emviveros:
[pd.log.zip](https://github.com/pure-data/pure-data/files/6370079/pd.log.zip)

umlaeute:
thanks for the logs.

it seems like Pd is opening the right-click menu, but somewhere off screen.
[...]

Issue #1190, opened 2020-10-23
PD doesn't work with the "new" jack2 on macOS


dyfer:
PureData is not able to use the "new" jack2 (as of Q4 2020) on macOS, AFAICT. Tested with a [recent jack2](https://github.com/jackaudio/jack2-releases/releases) on macOS 10.15 (Catalina). After selecting `jack` as the backend, PD reports that jack is not installed.

IIUC, the issue is that PD is linked with the jack "framework" that was provided with JackOSX, but is not provided with the current jack2.

I'm mostly logging this to raise awareness about the issue with jack(2) development on macOS and backward compatibility. I also reported compatibility issue in the jack2 repo: https://github.com/jackaudio/jack2/issues/640

danomatika:
>PureData is not able to use the "new" jack2 (as of Q4 2020) on macOS, AFAICT.

I assume you are referring to the precompiled versions you can download, which is true as the older framework is still weak linked by default.

 if you build Pd yourself, you can disable this and explicitly link to jack installed on the system, ie. via homebrew:

    ./configure   --disable-jack-framework --enable-jack

The problem with this, is that it's not distributable as the same version of jack has to be present on the system running Pd. Currently, Pd is built for such a wide range of systems, going back to 10.6, that this is a bit more difficult than what the older JackOSX lib provided.

The basic info on this is in the [INSTALL.txt](https://github.com/pure-data/pure-data/blob/master/INSTALL.txt#L233), however most general users don't need to look into it.

danomatika:
I did some checking and can confirm building Pd without the framework allows me to connect to the Homebrew jack server.

The issue is more that we need some sort of replacement for the framework that allows us to distribute Pd with jack enabled on macOS *without* requiring it on user systems. Maybe this is actually easy, [ala `-weak_library jack`](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPFrameworks/Concepts/WeakLinking.html#//apple_ref/doc/uid/20002378-107026) but I wasn't able to get libtool to link with it set.

dyfer:
I have limited experience with this, but I came across information implying that weak linking to a framework is easier than to a library. Again, my experience is very limited. AFAIU current jack2 provides only the shared library, and not the framework.
In the linked thread @falkTX mentioned https://github.com/x42/weakjack/ as a possible library to easily weak link jack, I have not tried it though.

danomatika:
[...]

PR #1414, opened 2021-10-27
[WIP] macOS updates

danomatika:
This PR is meant to collect some commits which update the build on macOS for the upcoming Pd 0.52 release. A couple of things I can think of are:

* [x] Update Tcl/Tk build script to support arm64
* [x] Use Tcl/Tk [8.6.12](https://core.tcl-lang.org/tk/timeline?r=core-8-6-12-rc) once it's released (now released)
* [x] ~~Add signing and notarization helper script, ie. you only need to provide developer signature & .app~~ (do this in a separate PR later)
* [ ] Update precompiled Wish in `mac/stuff` and/or...
* [ ] ...update workflow regarding compiling Wish.app based on system (10.6 or newer 10.13+)
* [ ] Link to newer jack distribution on builds for new systems (10.13+) while keep weak linking on old systems
* [x] Hope/pray that macOS 12 doesn't break anything (using Tk 8.6.12, so should be ok)

Current macOS issues:

* [ ] #1384
* [ ] #1190
* [x] #1374
* [x] #1306
* [ ] #1303
* [x] #1251
* [x] #1094
* [x] #576
* [x] #514
* [x] #353
* [x] #239

Anything else I can remember...

* [x] Hope/pray that multi monitor window positioning is no longer broken when reopening a patch (can't find issue right now)... **this seems fine**


umlaeute:
Tcl/Tk 8.6.12 has been released a few days ago :tada: 

on a related note:

for whatever reasons my Tcl/Tk-8.6.10 builds (compiled on macOS-10.15/Catalina) fail to run on macOS-10.12.
Pd-GUI just crashes with an error (`__objc_empty_cache: symbol not found`). This symbol is only referenced in *.../Tk.framework/Versions/8.6/Tk*.

Wish is built with `mac/tcltk-wish.sh 8.6.10` (and then incorporated into the Pd-app with `mac/osx-app.sh --wish Wish-8.6.10.app`.

I've tested the build on the following systems:
- macOS-10.12 (Sierra): fails :boom:
- ~~macOS-10.12 (High Sierra)~~ (i currently don't have access to that system)
- macOS-10.14 (Mojave) :tada:
- macOS-10.15 (Catalina) :tada:


any ideas what might go wrong?
i haven't tweaked the Tcl/Tk build system at all, but it seems it is invoked with `-mmacosx-version-min=10.6`, so that should be good.

You can find the build-logs at https://git.iem.at/pd/pure-data/-/jobs/32332 (and you can also download the created .app from this page (artifacts -> <kbd>Download</kbd>)
There's also a signed/notarized version available at https://git.iem.at/pd/pure-data/-/jobs/32346 (again: artifacts -> <kbd>Download</kbd>).

iirc the artifact links will only be valid for a week or so.

umlaeute:
i also tried with Tcl/Tk-8.6.12 and the `udpate/macos-build` branch, but the problem persists (as was expected)

millerpuckette:
If I'm reading this correctly I can happily merge this PR for the 0.52 release, correct?

umlaeute:
@miller: yes i think so. the PR so far only contains:
- documentation updates
- script hygiene (removal of bashisms)
- universal-arch detection (based on available XCode)
- fallback-logic if you forgot to add the `.app` suffix for Wish

so merging should be save (although it won't fix many of the issues mentioned by @danomatika in the initial post).

possibly a PR (even WIP) is not the very best place to collect these open issues.

danomatika:
No please don’t merge yet. I was only waiting first for Tk 8.6.12 then I can test the build and make suggestions on handling a few of those points. I am under pressure right now on other projects, so I haven’t been able to be very proactive this release, but I at least want to revisit those issues as many may probably have been fixed in Tk by now.

For instance, I consider the double key binding events on non-english locales a major problem.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

sebshader:
@umlaeute I feel like the ```ld: warning: URGENT: building for OSX, but linking against dylib (/usr/lib/libpthread.dylib) built for (unknown). Note: This will be an error in the future.```
seems odd.. does reinstalling command line tools make those warnings go away? It seems like the machine is mojave? on my mojave machine I get no such warnings when building tk 8.6.10 (in a fairly current ```master``` branch)

umlaeute:
@danomatika wrote:
> For instance, I consider the double key binding events on non-english locales a major problem.

I agree that there are still (important) things todo.
however, i would also like to have any fixes get into Pd asap, however minor they may seem.
(as in: I see zero (0!) reasons why we shouldn't include a fix for the JACK linking problem in `master` while we are waiting for Tcl/Tk-8.6.12)

so i think that a PR is just the wrong tool for collecting problems that *should be* solved (it's the right tool for collecting problems that **are** solved).
so i went ahead and created a ["Project"](https://github.com/pure-data/pure-data/projects/3) that collects all the issues (and PRs) that hopefully can replace the "collecting issues" aspect of this PR.

danomatika:
I updated the configure `--enable-universal` option to support `arm64`. I can now make a universal app on my macOS 11.6.1 system via:

~~~
./autogen.sh
./configure --universal
make
mac/tcltk-wish.sh --universal 8.6.1
mac/osx-app.sh -w Wish-8.6.12 0.52-0
~~~

The resulting Pd-0.52-0:
~~~
% file Pd-0.52-0.app/Contents/MacOS/Pd
Pd-0.52-0.app/Contents/MacOS/Pd: Mach-O universal binary with 2 architectures: [x86_64:Mach-O 64-bit executable x86_64] [arm64:Mach-O 64-bit executable arm64]
Pd-0.52-0.app/Contents/MacOS/Pd (for architecture x86_64):	Mach-O 64-bit executable x86_64
Pd-0.52-0.app/Contents/MacOS/Pd (for architecture arm64):	Mach-O 64-bit executable arm64
~~~

PR #1424, opened 2021-10-31
fix resizing gatoms with mouse by requesting the correct object font width

### LINKED ISSUE #1413: new gatoms don't resize well when their font size isn't the same as the canvas
example here, the canvas has (default) font size 12, and one number-gatom is 24 while the other one is 8.

resize_gatom

this also happens with the symbol-gatom and the list-gatom.

Ant1r:
fixes #1413


Ant1r:
422a6a2d9a6198439a31ede07ea54461ea7fad1f implements this.

PR #1426, opened 2021-11-01
fix t_methodentry bug introduced by using unsigned char instead of t_atomtype

pierreguillot:
Fix `t_methodentry` bug introduced by using `unsigned char` instead of `t_atomtype` in this commit https://github.com/pure-data/pure-data/commit/0a727e594ffa4a85f2c3c8c18d51730ffcce160b. 

The methods' arguments are not well forwarded in `pd_typemess ()` because of the type of `me_arg` mismatches with the structure.

The bug is really annoying because when opening a patch the connection cannot be created, etc.  

I don't know why the bug doesn't seem to happen in Pd 0.52.0 (this might be related to my compiler - XCode 13.0 - AppleClang 13.0.0.13000029). 

I tested the fix using PDINSTANCE=1 (for Camomile) but  it should also work with PDINSTANCE=0.

### DIFF
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -138,9 +138,7 @@ static void class_addmethodtolist(t_class *c, t_methodentry **methodlist,
     m = (*methodlist) + nmethod;
     m->me_name = sel;
     m->me_fun = (t_gotfn)fn;
-    i = 0;
-    while ((m->me_arg[i] = args[i]))
-        i++;
+    memcpy(m->me_arg, args, MAXPDARG+1);
 }
 
 #ifdef PDINSTANCE
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -114,7 +114,7 @@ static t_pdinstance *pdinstance_init(t_pdinstance *x)
 }
 
 static void class_addmethodtolist(t_class *c, t_methodentry **methodlist,
-    int nmethod, t_gotfn fn, t_symbol *sel, t_atomtype *args,
+    int nmethod, t_gotfn fn, t_symbol *sel, unsigned char *args,
         t_pdinstance *pdinstance)
 {
     int i;
@@ -639,7 +639,7 @@ void class_addmethod(t_class *c, t_method fn, t_symbol *sel,
     }
     else
     {
-        t_atomtype argvec[MAXPDARG+1];
+        unsigned char argvec[MAXPDARG+1];
         nargs = 0;
         while (argtype != A_NULL && nargs < MAXPDARG)
         {


Spacechild1:
To be honest, I can't really spot the problem with the original code. Can you show me which lines actually caused the issue?

pierreguillot:
Of course, the problem should have been more explicit sorry.

As you can see in the function `class_addmethodtolist()` at the [l.142 of m_class.c](https://github.com/pure-data/pure-data/blob/master/src/m_class.c#L142), `(m->me_arg[i] = args[i])` but `m->me_arg` is an array of `unsigned char` and `args` an array of `t_atomtype`. 
The function could work if the argument `args` is really an array of `t_atomtype` (even if this is really insecure) like this is the case in the function `class_addmethod()`  [l.661 of m_class.c](https://github.com/pure-data/pure-data/blob/master/src/m_class.c#L661). But this not the case in the function `pdinstance_new()` [l.183 of m_class.c](https://github.com/pure-data/pure-data/blob/master/src/m_class.c#L183) where an array of unsigned char` is used for `args`.

I hope it clarifies :) Let me know if you need further explanation.

Spacechild1:
Thanks for the reply!

> But this not the case in the function pdinstance_new() l.183 of m_class.c where an array of unsigned charis used forargs`.

That is the critical part! I didn't realize that `pdinstance_new()` would pass an `unsigned char *` to `class_addmethodtolist`. That's why the bug only manifests with `-DPDINSTANCE`. Actually, this shouldn't even compile if the enum has an incompatible underlying type:

> 6.7.2 4 says each enumerated type (an enum type) is compatible with an implementation-defined choice of char or a signed or unsigned integer type. So a pointer to some enum can be assigned to a pointer to one char or integer type (and vice-versa), but you cannot know which one without knowing something about the particular C implementation. https://stackoverflow.com/a/52408718

(As always, C++ is more sane and doesn't allow implicit pointer casts between `enum` and `int` at all!)

As a matter of fact, I can't compile the current Pd master branch on Windows at all with `-DPDINSTANCE`:

```
m_class.c: In function 'pdinstance_new':
m_class.c:186:39: warning: passing argument 6 of 'class_addmethodtolist' from incompatible pointer type [-Wincompatible-pointer-types]
  186 |                     c->c_methods[0][i].me_arg, x);
      |                     ~~~~~~~~~~~~~~~~~~^~~~~~~
      |                                       |
      |                                       unsigned char *
m_class.c:117:57: note: expected 't_atomtype *' but argument is of type 'unsigned char *'
  117 |     int nmethod, t_gotfn fn, t_symbol *sel, t_atomtype *args,
      |                                             ~~~~~~~~~~~~^~~~
```


> I don't know why the bug doesn't seem to happen in Pd 0.52.0

Because it's compiled without `-DPDINSTANCE`, so there is no incompatible pointer cast.

The question is rather: why can you successfully compile it with `-DPDINSTANCE`? Maybe a bug in AppleClang?

---

Anyway, your PR fixes the compilation error, so it's fine!

Spacechild1:
BTW, with GCC 10.2 I get the following warning:

```
m_class.c: In function 'class_addmethodtolist':
m_class.c:142:26: warning: writing 1 byte into a region of size 0 [-Wstringop-overflow=]
  142 |     while ((m->me_arg[i] = args[i]))
      |            ~~~~~~~~~~~~~~^~~~~~~~~~
In file included from m_class.c:7:
m_imp.h:19:19: note: at offset 6 to object 'me_arg' with size 6 declared here
   19 |     unsigned char me_arg[MAXPDARG+1];
      |                   ^~~~~~
g++ -shared 
```

It's really a false positive. However, the loop can be replaced with a simple `memcpy(m->me_arg, args, MAXPDARG+1);`. It would be less confusing and fix the warning. What do you think? Could you add this to the PR?

pierreguillot:
Yes, of course!

pierreguillot:
Done

Spacechild1:
Since this is really a trivial fix, I went ahead and merged it into develop.

Issue #1429, opened 2021-11-02
[samplerate~] in 0.52 outputs 48000 instead of 44100 if dsp is off when you load Pd


porres:
When you load pd and try to get the sample rate with samplerate~ when the dsp is still off, it considers it to be 48000, even though it is actually set as 44100

<img width="457" alt="Screen Shot 2021-11-02 at 19 28 51" src="https://user-images.githubusercontent.com/5360142/139960752-85b36b69-5dbb-462a-b8ba-247bd2da74b2.png">

after you turn dsp on and try again, it correctly outputs 44100

Spacechild1:
> after you turn dsp on and try again, it correctly outputs 44100

Well, Pd can't know the sample rate before you turn on the audio. 48000 is just a default value. Note that in Pd 0.52 the default sample rate has changed from 44100 to 48000. Either way, without turning on DSP at least once, `[samplerate~]` doesn't really know what to output, so it could be literally anything.

porres:
> Note that in Pd 0.52 the default sample rate has changed from 44100 to 48000.

that explains something, but how come I get 44100 when I open the Pd?

Spacechild1:
> but how come I get 44100 when I open the Pd?

I guess you meant "start DSP"? It's probably because you have saved your audio settings with 44.1 kHz. When you start DSP it will recall your audio settings and `[samplerate~]` will output the "real" sample rate.

porres:
I completely removed Pd and reinstalled so I can see now that, in this case, it loads 48000 by default.

 don't really get why change it, cause it is more CPU intensive, but anyway, can't pd use this value from what has been saved under preferences and use it as the default value for samplerate~?

Lucarda:
how about:
![sr](https://user-images.githubusercontent.com/25239522/139965799-8b4dd87c-bb0e-43ff-9b14-b1c6527a639b.png)

Spacechild1:
> can't pd use this value from what has been saved under preferences and use it as the default value for samplerate~?

But why? The value simply is not meaningful unless DSP is turned on. In your patch you should call `[samplerate~]` whenever you need the value. It is always a mistake to call it only once in the beginning. What if the user changes the sample rate while the patch is running?

Spacechild1:
If you *really* need to cache the value, you can do `[r pd-dsp-started] -> [samplerate~]`, so it will automatically update whenever the sample rate has changed.

porres:
[...]

PR #1447, opened 2021-11-13
revert audio settings behavior

Spacechild1:
This PR does the following:

1) Apply audio settings after argument parsing, so that objects will immediately see the specified sample rate or channel count, even if DSP hasn't been started yet.
A prominent example is `[samplerate~]`. A patch might precalcalculate things that are sample rate dependent. If the default sample rate is different from the saved sample rate, it would have to recalculate everything again after DSP has been started.
Other patches (wrongly) assume a fixed sample rate and only bang `[samplerate~]` once during load time. Starting with a different sample rate than the one saved by the user can break such patches.

2) Revert the default sample rate to 44100.
44100 Hz has been the default sample rate for ages and I don't think it is a good idea to change it without a very compelling reason.
Many patches rely on the default sample rate. Yes, they are probably buggy, but IMO we should still try to avoid breaking them if possible.
Also, certain patches can't be written in a sample rate independent manner. For example, `[readsf~]` can't resample, so it only works with the sample rate of the accompanying sound files. If you ship such a patch, you would now have to tell the user to change the sample rate or supply a script with starts Pd with "-r 44100". Older existing projects would simply break.

Fixes https://github.com/pure-data/pure-data/issues/1429

NOTE: I am fully aware that a well behaved Pd patch should dynamically react to sample rate changes, e.g. with `[r pd-dsp-started] -> [samplerate~]`, but in reality many people haven't been aware of the problem. I assume that there are *many* Pd patches that (wrongly) assume a fixed samplerate.


Lucarda:
This VOTE took me some time and I don't want to be impartial.

"revert default samplerate to 44100"

+1

reduzent:
I think that many patches (wrongly) assume a fixed samplerate and check it only once at launch time is a strong point. It would be friendly not to break them.

millerpuckette:
ok - friendliness here trumps correctness :)

Spacechild1:
will fix the merge conflict

PR #1428, opened 2021-11-02
stable development branch

Spacechild1:
this PR only contains curated, small, no-brainer changes for the upcoming Pd 0.52 release.

Content (should be updated after each push):

* [regression] fix extra verbose `-stderr` print out
* [regression] fix crash in `[pd~]` subprocess (Closes: #1418)
* [regression] fix compilation error with `-DPDINSTANCE`
* [regression] fix Windows audio backend/settings
* [regression] fix broken `[cputime]` on Windows
* fix occasional wrong usage of `PDINSTANCE` define
* get rid of redundant `x_ctl` member in filter objects
* add missing files to build system
* update HTML docs
* fix some typos
* allow devices with non-ASCII filenames on Windows (hello :cn: , :jp:,...; Closes #980)
* [regression] fix crash when using `[framp~]` without instantiating an `[sqrt~]` object first (Closes: #1432)
* fix connection-lines when resizing new list gatom (Closes: #1407)
* Fix copy-and-paste error for "--enable-libpd-*" flags
* on macOS always link against the CoreFoundation framework (even if no audio backend is selected)
* `[soundfiler]`: fix `-ascii` not overriding the `.txt` extension
* fix `-r` command line option
* fix sized of zoomed labels of gatoms

PR #1450, opened 2021-11-14
fix the gatom's label font size when zoomed in

Ant1r:
this PR fixes bad gatoms' label font size when the canvas is zoomed-in, see:
https://lists.puredata.info/pipermail/pd-list/2021-11/130208.html


umlaeute:
merged into 'develop' (PR #1428 )

Issue #1453, opened 2021-11-17
[0.52-test2] list atom crashes Windows.


Lucarda:
Pd @ 90ad800 Windows amd64.

Steps to reproduce:

1. create a list atom
2. type `hello world` and press enter.
3. re-enter to edit it and add `10`

   * `hello world 10`
   * press enter : &#128165;

#### GDB back trace:

```
Thread 1 received signal SIGSEGV, Segmentation fault.
rtext_key (x=0x202c9ae4ee0, keynum=0, keysym=0x3) at g_rtext.c:660
660         else if (!strcmp(keysym->s_name, "Home"))
(gdb) bt
#0  rtext_key (x=0x202c9ae4ee0, keynum=0, keysym=0x3) at g_rtext.c:660
#1  0x00007ffa1d1064d7 in gatom_key (z=<optimized out>,
    keysym=<optimized out>, f=<optimized out>) at g_text.c:803
#2  0x00007ffa1d0ed53b in rtext_key (x=0x202c9ae4ee0, keynum=keynum@entry=10,
    keysym=keysym@entry=0x202c9aea660) at g_rtext.c:594
#3  0x00007ffa1d0d1421 in canvas_key (x=0x202c9ae4510, av=<optimized out>,
    ac=<optimized out>, s=<optimized out>) at g_editor.c:3057
#4  0x00007ffa1d11d963 in binbuf_eval (x=<optimized out>,
    target=0x202c9ae4a80, target@entry=0x0, argc=-1073741637, argc@entry=0,
    argv=0x7f7e1ff3c0, argv@entry=0x0) at m_binbuf.c:790
#5  0x00007ffa1d12a22d in socketreceiver_read (fd=736, x=0x202c9ae7f50)
    at s_inter.c:658
#6  socketreceiver_read (x=0x202c9ae7f50, fd=736) at s_inter.c:590
#7  0x00007ffa1d1295d0 in sys_domicrosleep (microsec=microsec@entry=0)
    at s_inter.c:226
#8  0x00007ffa1d12b08b in sys_pollgui () at s_inter.c:976
#9  0x00007ffa1d1260df in sched_idletask () at m_sched.c:299
#10 0x00007ffa1d126b6c in m_pollingscheduler () at m_sched.c:375
#11 m_mainloop () at m_sched.c:432
#12 0x00007ff62b8a13b1 in ?? ()
#13 0x00007ff62b8a14c6 in ?? ()
[...]

Issue #1422, opened 2021-10-30
weird characters appear when typing in new symbol and list atoms


Spacechild1:
when typing things in the new symbol and list atoms, a random characters starts to appear right before the `...`.

Examples:

![symbol1](https://user-images.githubusercontent.com/16126632/139531927-3f818406-c021-450b-a51f-a5232a8c33c5.JPG)

![symbol2](https://user-images.githubusercontent.com/16126632/139531932-bdc2dfa1-5973-4603-ad2b-8c541925c960.JPG)

Pd 8897538f883c4bec907647989c1198a9ef9495d7 on Windows 7

Lucarda:
Confirmed

Pd(amd64) 8897538 on Windows 10

Ant1r:
this doesn't seem to happen on Linux (neither Metacity nor gnome-shell).

umlaeute:
an i can't reproduce it in a VM either.

however, running Pd (`develop` 6892cc3f519f6aefe979303dd910868d6e63ee63  on Linux) through Valgrind indicates that there's a memory problem:

~~~
==1068044== Invalid read of size 1
==1068044==    at 0x1D2E51: u8_nextchar (s_utf8.c:246)
==1068044==    by 0x18AB06: rtext_formatatom (g_rtext.c:323)
==1068044==    by 0x18AB06: rtext_senditup (g_rtext.c:393)
==1068044==    by 0x18C099: rtext_key (g_rtext.c:716)
==1068044==    by 0x1A49F9: gatom_key (g_text.c:792)
==1068044==    by 0x1BCAA7: binbuf_eval (m_binbuf.c:790)
==1068044==    by 0x1C8B6D: socketreceiver_read (s_inter.c:658)
==1068044==    by 0x1C8B6D: socketreceiver_read (s_inter.c:590)
==1068044==    by 0x1C7B33: sys_domicrosleep (s_inter.c:226)
==1068044==    by 0x1C9AF0: sys_pollgui (s_inter.c:976)
==1068044==    by 0x1C495B: sched_idletask (m_sched.c:299)
==1068044==    by 0x1C54A6: m_pollingscheduler (m_sched.c:375)
==1068044==    by 0x1C54A6: m_mainloop (m_sched.c:432)
==1068044==    by 0x4BD3E49: (below main) (libc-start.c:314)
==1068044==  Address 0x53bc1d1 is 0 bytes after a block of size 1 alloc'd
[...]

PR #1456, opened 2021-11-17
Fix memory corruption in t_rtext

Spacechild1:
`t_rtext` works on unterminated strings, but it calls utf8 functions which expects null-terminated strings. This can lead to all kinds of weirdness and memory corruption.

Solution: whenever we allocate/resize the content of `x_buf`, we make space for an extra 0 byte. `x_bufsize`, however, still behaves like strlen(), i.e. doesn't include the 0 byte. The advantage is that we only need to do this in a few places (when the buffer is allocated/resized) and don't have to touch the actual (complicated) logic in the rest of the code.

Fixes https://github.com/pure-data/pure-data/issues/1453 and https://github.com/pure-data/pure-data/issues/1422

This PR also includes 2 small drive-by commits:

* remove a few unnecessary casts when passing a pointer to ".x%lx".
* fix GUI debug printing on Windows

### DIFF
--- a/src/s_print.c
+++ b/src/s_print.c
@@ -48,14 +48,6 @@ char* pdgui_strnescape(char *dst, size_t dstlen, const char *src, size_t srclen)
     return dst;
 }
 
-static char* strnpointerid(char *dest, const void *pointer, size_t len)
-{
-    *dest=0;
-    if (pointer)
-        snprintf(dest, len, ".x%lx", (unsigned long)pointer);
-    return dest;
-}
-
 static void dopost(const char *s)
 {
     if (sys_printhook)
@@ -105,12 +97,8 @@ static void doerror(const void *object, const char *s)
 #endif
     }
     else
-    {
-        char obuf[MAXPDSTRING];
-        sys_vgui("::pdwindow::logpost {%s} 1 {%s}\n",
-                 strnpointerid(obuf, object, MAXPDSTRING),
-                 pdgui_strnescape(upbuf, MAXPDSTRING, s, 0));
-    }
+        sys_vgui("::pdwindow::logpost .x%lx 1 {%s}\n",
+            object, pdgui_strnescape(upbuf, MAXPDSTRING, s, 0));
 }
 
 static void dologpost(const void *object, const int level, const char *s)
@@ -141,12 +129,8 @@ static void dologpost(const void *object, const int level, const char *s)
 #endif
     }
     else
-    {
-        char obuf[MAXPDSTRING];
-        sys_vgui("::pdwindow::logpost {%s} %d {%s}\n",
-                 strnpointerid(obuf, object, MAXPDSTRING),
-                 level, pdgui_strnescape(upbuf, MAXPDSTRING, s, 0));
-    }
+        sys_vgui("::pdwindow::logpost .x%lx %d {%s}\n",
+            object, level, pdgui_strnescape(upbuf, MAXPDSTRING, s, 0));
 }
 
 void logpost(const void *object, int level, const char *fmt, ...)
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -747,7 +747,7 @@ static void graph_vis(t_gobj *gr, t_glist *parent_glist, int vis)
             i -= glist_fontheight(x);
             sys_vgui(".x%lx.c create text %d %d -text {%s} -anchor nw "
                 "-font {{%s} -%d %s} -tags [list %s label graph]\n",
-                (long)glist_getcanvas(x),  x1, i,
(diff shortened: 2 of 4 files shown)


Spacechild1:
*sigh*. I just noticed that after this the right margin of all text objects is too small:

grafik

LEFT: `rtext-fix`, RIGHT: `master`

I have no idea why...

Lucarda:
@Spacechild1 

 the weird characters are gone with this PR but I'm getting the same crash as in #1453

this is the **GDB**:

```
Thread 1 received signal SIGSEGV, Segmentation fault.
rtext_key (x=0x120273d9a00, keynum=0, keysym=0x3) at g_rtext.c:673
673         else if (!strcmp(keysym->s_name, "Home"))
(gdb) bt
#0  rtext_key (x=0x120273d9a00, keynum=0, keysym=0x3) at g_rtext.c:673
#1  0x00007fff89fd6577 in gatom_key (z=<optimized out>,
    keysym=<optimized out>, f=<optimized out>) at g_text.c:803
#2  0x00007fff89fbd5db in rtext_key (x=0x120273d9a00, keynum=keynum@entry=10,
    keysym=keysym@entry=0x120273da590) at g_rtext.c:600
#3  0x00007fff89fa1421 in canvas_key (x=0x120273d9230, av=<optimized out>,
    ac=<optimized out>, s=<optimized out>) at g_editor.c:3057
#4  0x00007fff89feda03 in binbuf_eval (x=<optimized out>,
    target=0x120273d97a0, target@entry=0x0, argc=-1073741637, argc@entry=0,
    argv=0xba5bff2b0, argv@entry=0x0) at m_binbuf.c:790
#5  0x00007fff89ffa30d in socketreceiver_read (fd=676, x=0x120273d5eb0)
    at s_inter.c:667
#6  socketreceiver_read (x=0x120273d5eb0, fd=676) at s_inter.c:599
#7  0x00007fff89ff96b0 in sys_domicrosleep (microsec=microsec@entry=0)
    at s_inter.c:226
#8  0x00007fff89ffb17b in sys_pollgui () at s_inter.c:997
#9  0x00007fff89ff617f in sched_idletask () at m_sched.c:299
#10 0x00007fff89ff6c0c in m_pollingscheduler () at m_sched.c:375
#11 m_mainloop () at m_sched.c:432
#12 0x00007ff6e9ea13b1 in ?? ()
#13 0x00007ff6e9ea14c6 in ?? ()
#14 0x00007fffd6157c24 in KERNEL32!BaseThreadInitThunk ()
   from C:\WINDOWS\System32\kernel32.dll
#15 0x00007fffd706d4d1 in ntdll!RtlUserThreadStart ()
   from C:\WINDOWS\SYSTEM32\ntdll.dll
#16 0x0000000000000000 in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
(gdb)
```

umlaeute:
> `rtext_key (x=0x120273d9a00, keynum=0, keysym=0x3) at g_rtext.c:673`

urgh. that `keysym` does not look goog.

Spacechild1:
@millerpuckette we definitely need this before the final release.

src/s_print.c
```diff
@@ -48,14 +48,6 @@ char* pdgui_strnescape(char *dst, size_t dstlen, const char *src, size_t srclen)
...
 
-static char* strnpointerid(char *dest, const void *pointer, size_t len)
-{
-    *dest=0;
-    if (pointer)
-        snprintf(dest, len, ".x%lx", (unsigned long)pointer);
-    return dest;
-}
-
 static void dopost(const char *s)
```
umlaeute, line 51: i'm not super-excited by this change :frowning_face: 

instead i think it would be better to use a canvas-to-guiID wrapper everywhere, so there's a single documented way to calculate the ID (rather than having assuming it in multiple places) - even if it is a bit slower.

Spacechild1: Well, it's a private function and only used in two places. It has been introduced quietly without discussion.

In the future we might refactor the whole codebase and replace all those `.x%lx` strings with something else. But even then it isn't clear to me that a function would be the best approach. This could be also solved with a macro:

```
#define CANVAS_ID ".x%lx.c"
#define OBJECT_ID ".x%lx"

sys_vgui(CANVAS_ID " delete GOP\n",  glist_getcanvas(x));
```

That's why I deleted the function for now. We can later reintroduce it *if* we agree it's really the best approach.

Spacechild1: Ha, true! The culprit was HCS :-)

I've now pushed the commit to develop and added an issue about fixing the pointer ID problem: https://github.com/pure-data/pure-data/issues/1474

Spacechild1:
merged to develop

PR #1460, opened 2021-11-18
[0.52-test2] html updates

Lucarda:
This PR:

- few corrections to chapter3


doc/1.manual/x1.htm
```diff
@@ -108,14 +108,12 @@ <H3> <A id=s2> 1.2. other resources </A> </H3>
...
-<A href=https://en.flossmanuals.net/pure-data/_full/>
-en.flossmanuals.net/pure-data/_full/</A> (English) and
-<A href=http://fr.flossmanuals.net/PureData/>
-fr.flossmanuals.net/PureData/</A> (French).
+	
+<P>You can check <A href=http://puredata.info/docs/BooksAboutPd/>
+puredata.info/docs/BooksAboutPd/</A> for more documentation. But please 
+be aware that some references might be very old and outdated. For instance, 
+the FLOSS Manuals deal with installing Pd Extended, which is now an abandoned
+fork of Pd since 2013! Please refer to this manual for more up to date information.
```
umlaeute, line 116: i think it is generally a bad idea to encode information about the outdatedness of webpages in the offline documentation.
at least where it is possible: just put that info on the https://puredata.info/docs/BooksAboutPd/ *now* and everyone will be receive the info, regardless of which version of Pd they are using (actually the page already says that FLOSS-manuals are outdated)

porres: see https://github.com/Lucarda/pure-data/pull/8

PR #1481, opened 2021-11-25
helpbrowser fixes for 0.52-0test2

danomatika:
This PR addresses a couple of issues found on macOS for the help browser in Pd-0.52-0test2:

* browser list boxes (columns) not fully refreshing their contents on navigation
* browser complaint about undo/redo events

System tested: macOS 11.6.1

It looks like changes in Tk 8.6.12 may have affected the Tk idle timing, so forcing via `update idletasks` ensures the listbox display is refreshed.

For the binding, I'm just catching CMD+z and CMD+Z (aka SHIFT-CMD-z) but perhaps there is a better place or way to do this?


sebshader:
@danomatika I've finished the x-scrolling/expandable/non-limited helpbrowser, I wonder if we should just use that instead #1240

PR #1484, opened 2021-11-27
Fix single object election drag and resize behavior 

danomatika:
This is an edited version of #922 which brings it up to date with the current master. (_Apologies to @
jamshark70, but this seems the quicker way to get it ready for testing._)

As outlined in the referenced PR, this bug fix deals with the following issues in on the canvas in edit mode when a single object is selected:

1. After dragging, the object's text is selected when the mouse is released. If the user wants to drag the same object again, the object must be manually deselected first. Expected behavior is more likely to be that the object can simple be clicked on and dragged again with the deselect and that the object's text should only be selected for editing when the user manual clicks inside the object when its clicked on.

2. Dragging the right edge of the object allows for it to be manually resized. If the object is selected, however, the resize handle is not available until the object is manually deselected. Expected behavior is more likely to be that the object can be resized irregardless if it's selected or not.


Spacechild1:
Works fine for me!

PS: when you create a new PR anyway, you can safely rebase on top on master instead of merge.

Spacechild1 (approved):

danomatika:
@danomatika >PS: when you create a new PR anyway, you can safely rebase on top on master instead of merge.

["Explain how."](https://www.youtube.com/watch?v=dgct3Jn8pFA) :)

Spacechild1:
`git rebase master` :-) or if `master` is not up to date:  `git pull --rebase master`

if you happen to have uncommited changes, you have to save them with `git stash` before the rebase and afterwards get them back with `git stash pop`.

In case there are conflicts, git pauses the rebasing process so that you can fix them. Once resolved, you can continue with `git rebase --continue`.

Generally, rebasing should be avoided for public branches because it rewrites history. But for private branches it's much better than merging because you don't get those redundant merge commits. Finally, interactive rebasing (`git rebase -i`) is extremely powerful because you can change individual commits after the fact, remove them, join them, etc.

danomatika:
Ok, makes sense. For this PR, I checked out the last commit to master in the original branch when making the new branch, then I pulled in the original branches commits, then merged with master to bring it up to date. If rebase is the better approach, I will try it in the future...

PR #1490, opened 2021-12-02
Fixed bug where LibPd was ignoring init parameters

myQwil:
While working with LibPd, I noticed that its `PdBase::init` function was completely ignoring all of its parameters and falling back to default values. This is the simplest solution I could come up with to fix that. We just need to tell `sys_get_audio_settings` that audio settings were already initialized in `sys_set_audio_settings`

### DIFF
--- a/src/s_audio.c
+++ b/src/s_audio.c
@@ -197,9 +197,10 @@ static void audio_compact_and_count_channels(int *ndev, int *devvec,
 
 /* ----------------------- public routines ----------------------- */
 
+static int initted = 0;
+
 void sys_get_audio_settings(t_audiosettings *a)
 {
-    static int initted;
     if (!initted)
     {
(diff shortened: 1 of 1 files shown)


src/s_audio.c
```diff
@@ -197,9 +197,10 @@ static void audio_compact_and_count_channels(int *ndev, int *devvec,
 
 /* ----------------------- public routines ----------------------- */
 
+static int initted = 0;
```
umlaeute, line 200: any reason to move this out of the `sys_get_audio_settings()` body?

myQwil: that was done so that it can be seen by `sys_set_audio_settings()`

umlaeute: i see (i didn't spot this due to the way GitHub displays diffs...)

PR #1494, opened 2021-12-03
Documentation updates

porres:
1st commit (there may be more): 

added output~ to help-intro.pd and edited release notes

since we now have output~ in 'extra' I added it to the object list under the 'extra' category. I also mentioned this in the release notes. Ressi's collaboration were listed twice so I removed the redudancy. I also added how I did lots of work on the help files :) hope it's cool to brag about it :/


umlaeute:
> Ressi's collaboration were listed twice so I removed the redudancy. 

which redundancy?

there are multiple paragraphs, each mentioning a single change, and christof contributed to more than one: thus he is listed in each paragraph where he contributed.

porres:
originally, the file had a line that said

"New methods for "list store" and added features to "array" (by Christof Ressi)."

and then, much later down the text Miller would say again that we had new methods for list store and garray methods, but more in detail. This seemed like giving the news twice, so I removed the first line and keep the more descriptive ones at the end... but but Christof's name at the bottom in both cases.

Spacechild1:
Two things that aren't explicitly mentioned are the new `[vis(` and `[edit(` messages for garrays. They are easy to miss otherwise...

umlaeute:
while you are at it: could you please remove the part from your new Pd-messages.pd (which is nice btw) that says:
> Pd gets also receives messages when you set things via Menus or shortcuts and we can also send such messages from within the patch. 

it think there is no benefit in telling the user that this is what happens when they press a keyboard shortcut.
it might also just be wrong (that is: the `[; pd dsp 1(` message is independent of how the GUI turns on the DSP; it just happens to use this right now).

just tell the people that there are various aspects that you can control and list them.


of: and while working on #1397, i'm pretty sure i want to get rid of the `[; pd start-startup-dialog(` message (and friends).
i don't think it is of much use in general, and if you document it now it is much harder to get rid of it later (and then we have to keep compatibility code and more cruft which just adds up to more real BUGS)

umlaeute:
just a minor nitpick: would it be possible to put a short and meaningful description into the *title* (first line) of each commit message. `git log --name-only` already gives you which files have been touched by a commit.

i think a good example of how to have both the context (which help-file was worked on) and the intent (why was the change required) on the first glance can be found in 2ac3408d9714e9c36913b60359c9254e6dbc15fd

if having *both* the file-name and the short-description take too much space for a proper "title", prefer to drop the information that can be easily autogenerated (filename, author, date)

PR #1495, opened 2021-12-03
[0.52-$] compilation fix.

Lucarda:
This PR includes 'draw-shapes-help.pd' on autotools so that compilation finishes.

The file was renamed.

PR #1510, opened 2021-12-08
[ci] add GitHub workflow

umlaeute:
yet another CI provider...

the reason to use this one is mainly because we get instant feedback in PRs of whether the build actually succeeds (e.g. things like #1495 should be really easy to catch with this)

this (again) uses an external workflow definition so we can adapt to changes in the infrastructure without bothering with Pd's repository.

### DIFF
--- a/.github/workflows/puredata.yml
+++ b/.github/workflows/puredata.yml
@@ -1,0 +1,16 @@
+name: pure-data
+
+on:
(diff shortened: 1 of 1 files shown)


Spacechild1:
SuperCollider has been doing this for quite a while now and it is indeed very convenient!

PR #1514, opened 2021-12-15
[print]: fix avoiding 'verbose(2)' when using a printhook

Ant1r:
Since the merge of #836, the [print] output is still prefixed with `verbose(2):` when using libpd mutli-instance printhooks.
This is a fix.


Ant1r:
@umlaeute I re-target this to develop, I think it's small enough and would be good to have  in 0.52.

umlaeute:
I guess so, but I'm currently offline.
Probably @danomatika or @Spacechild1 can merge

Issue #1476, opened 2021-11-23
gatom issues in pd 0.52-0 test2


porres:
Bringing here what I raised on the list

- [x] regression: floats typed in symbol box get emptied out.

- [ ] fix shortcut for list box (maybe cmd+6 and keep cmd+4 for symbol box).
- [x] shorter default value for symbol box (20 is too large). - 7c8956e25b037ea87638c6d46eb9ad582c88a86f

- [x] regression:  '$0-x' as a send or receive name become '#0-x'  (while we're at it, we can allow "#" in labels). **- Fixed by Miller** in 7b949aa903f2531d2a9c1af9f5aaae2ff6f11f82
- [x] internally force escape for colons, semicolons and dollars in 'list box'. **- Fixed by Miller** in 09074299673036955ff8410438a454f47a01f6f2

porres:
I now realize number boxes allow you to type any character in. This wasn't the case before. Not an actual 'bug' but I don't see any advantage in letting this happen now, maybe we should filter them back as before...

PR #1515, opened 2021-12-15
fix symbol box not outputting floats as symbols & only allow meaningful characters in number boxes

porres:
this is one of the tasks in https://github.com/pure-data/pure-data/issues/1476

tested here and it seems to work just fine, but of course I'd expect some revision

I also only allowed meaningful characters in number boxes as before, see my discussion also in https://github.com/pure-data/pure-data/issues/1476#issuecomment-994235584

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -817,7 +817,11 @@ void gatom_key(void *z, t_symbol *keysym, t_floatarg f)
             rtext_key(t, '.', &s_);
             rtext_key(t, 0, gensym("Home"));
         }
-        rtext_key(t, c, keysym);
+        if (x->a_flavor == A_SYMBOL || x->a_flavor == A_LIST)
+            rtext_key(t, c, keysym);  /* insert the character */
+        else if (x->a_flavor == A_FLOAT && ((c >= '0' && c <= '9') || c == '.' ||
+            c == '-' || c == '+' || c == 'e' || c == 'E' || c == '\b'))
+                rtext_key(t, c, keysym);  /* insert the accepted characters */
     }
 }
 
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -817,11 +817,6 @@ void gatom_key(void *z, t_symbol *keysym, t_floatarg f)
             rtext_key(t, '.', &s_);
             rtext_key(t, 0, gensym("Home"));
         }
-            /* automatically escape special characters in symbols */
-        if (x->a_flavor == A_SYMBOL && (c == ' ' || c == ',' || c == ';' ||
-            c == '$' || c == '\\'))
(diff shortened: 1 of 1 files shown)


src/g_text.c
```diff
@@ -795,7 +795,13 @@ void gatom_key(void *z, t_symbol *keysym, t_floatarg f)
             while (i--)
                 rtext_key(t, '\b', &s_);
             rtext_gettext(t, &buf, &bufsize);
-            text_setto(&x->a_text, x->a_glist, buf, bufsize);
+            t_atom *ap = gatom_getatom(x);
```
umlaeute, line 798: this is not valid C89 code: please only declare variables at the beginning of a code block

src/g_text.c
```diff
@@ -811,12 +817,11 @@ void gatom_key(void *z, t_symbol *keysym, t_floatarg f)
...
-        if (x->a_flavor == A_SYMBOL && (c == ' ' || c == ',' || c == ';' ||
-            c == '$' || c == '\\'))
-                rtext_key(t, '\\', &s_);
-            /* and at last, insert the character */
-        rtext_key(t, c, keysym);
+        if (x->a_flavor == A_SYMBOL || x->a_flavor == A_LIST)
+            rtext_key(t, c, keysym);  /* insert the character */
+        else if (x->a_flavor == A_FLOAT && ((c >= '0' && c <= '9') || c == '.' ||
+            c == '-' || c == '+' || c == 'e' || c == 'E' || c == '\b'))
+                rtext_key(t, c, keysym);  /* insert the accepted characters */
```
umlaeute, line 824: while this is not a regression, i wonder whether we shouldn't use a validator function (`strtod()`) instead of filtering by character.
e.g. the above (and *almost* Pd<0.52) will happily allow the user to type `1.2-3.4+5` which then ends up as *1.2*

umlaeute (changes requested):
please use C89

src/g_text.c
porres: like this? https://github.com/pure-data/pure-data/pull/1515/commits/d6f80c6c93b4b73f2db7b8b2c51bef57b061bfa0

umlaeute: yes, like this.
though typically you could want to keep variable scope as tight as possible, so the `t_atom*ap` declaration could really go here: https://github.com/pure-data/pure-data/blob/d6f80c6c93b4b73f2db7b8b2c51bef57b061bfa0/src/g_text.c#L790-L791

but anyhow. the C89 problem is fixed by d6f80c6

src/g_text.c
umlaeute: it was more like a general suggestion. for *this* PR i think it is ok to restore the original behaviour.

PR #1518, opened 2021-12-15
Fix compilation of Tk-8.6.12 on OSX<10.12

umlaeute:
This is a backport of a build-fix for Tk-8.6.12: https://github.com/tcltk/tk/commit/e3db95d5882e8cf0750989b7311f69c3fdabf065, that allows running `mac/tcltk-wish.sh 8.6.12` on older OSX-versions.

so far i've tried:
- OSX-10.7: the build now fails elsewhere
- OSX-10.8: the build succeeds (but it's not possible to use `-universal` to create an i386/amd64 binary; I don't think there is any necessity for i386 builds)

i only have access to an OSX-10.7  and an OSX-10.8 machine remotely, so I cannot test whether the resulting binary actually works.

However, running ` ./Wish-8.6.12.app/Contents/MacOS/Wish` does not crash (but it just waits forever), which i think indicates that nothing is fundamentally wrong. (if i run the same on command with the universal binary compiled on macOS-11 provided by @millerpuckette a few days ago, I get an `[NSProcessInfo operatingSystemVersion]: unrecognized selector sent to instance` error...

in any case: here's a [binary built on OSX-10.8](https://users.iem.at/zmoelnig/pd/download/Wish-8.6.12.osx107.app.tgz)

please test and if it doesn't work, this PR should probably be just closed.

### DIFF
--- a/mac/patches/tk8.6.12_NSBitmapFormatAlphaFirst.patch
+++ b/mac/patches/tk8.6.12_NSBitmapFormatAlphaFirst.patch
@@ -1,0 +1,65 @@
+From e3db95d5882e8cf0750989b7311f69c3fdabf065 Mon Sep 17 00:00:00 2001
+From: culler <culler>
+Date: Tue, 9 Nov 2021 20:34:22 +0000
+Subject: [PATCH] Fix [e699a7b606]: Fix build on macOS<10.12 - patch from
+ Christopher Chavez.
+
+---
+ macosx/tkMacOSXConstants.h | 1 +
+ macosx/tkMacOSXImage.c     | 7 ++++---
+ 2 files changed, 5 insertions(+), 3 deletions(-)
+
+diff --git a/macosx/tkMacOSXConstants.h b/macosx/tkMacOSXConstants.h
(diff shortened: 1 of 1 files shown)


umlaeute:
sidenote: even if this patch allows us to bulid Tcl/Tk for old OSX versions (and even if this build might actually appear to work on newer macOS versions), i think we should use a universal amd64/arm64 binary as the Wish for the default macOS Pd.

so this patch is really only meant to *also* allow building Pd for older OSXs.

(for newer macOS builds the patch is a no-op)

danomatika:
Sounds good if it indeed works. Maintaining differing Tk versions can only be unwieldy.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

millerpuckette:
Let me know if I'm missing something... but I think the existing Tcl/Tk versions
are working, adn the thing this introduces is the ability to grind out
compiled versions more easily in the future - if this is so, it should be OK to
merge the TclTk changes after 0.52-0 goes out.

cheers
Miller

PR #1517, opened 2021-12-15
documentation updates

porres:
documentation updates as usual

Closes: https://github.com/pure-data/pure-data/issues/1525

Closes: https://github.com/pure-data/pure-data/pull/1568

closes https://github.com/pure-data/pddp/issues/6

closes https://github.com/pure-data/pure-data/issues/1578

closes https://github.com/pure-data/pure-data/pull/1517 

closes https://github.com/pure-data/pddp/issues/56

closes https://github.com/pure-data/pure-data/issues/1527

Issue #1459, opened 2021-11-18
performance degradation with gatoms in Pd 0.52-test2


Spacechild1:
The following patch contains a large amount of float gatoms:

[many-gatoms.zip](https://github.com/pure-data/pure-data/files/7559225/many-gatoms.zip)

On my system (Windows 7), the idle CPU load of Pd without DSP is ca. 0.25%

When I open this patch with Pd 0.51-4 and move the mouse, the CPU load rises up to a max. of 0.5%.

When I do the same with Pd 0.52-test1 or Pd 0.52-test2, the CPU load rises up to a max. of 5%!! On my 4-core laptop, this is 1/5 of a core just for moving the mouse!

Can someone confirm this?

Lucarda:
`many-gatoms.pd`

Intel Core i5-520M
Windows 10
amd64
DSP On

|Pd 0.51-4 | Pd 0.52-test2|
|:----|:----|
|~2%|~20%|

Ant1r:
Same here: 
Ubuntu 16.04 / AMD Phenom II forced to 2.2GHz

| Pd 0.51-4 | Pd 0.52-test2 |
|         ----    |     -------          |
| ~2%         | ~22%            |

umlaeute:
just to clarify: this is about moving the mouse across the patch without any buttons pressed (that is: it's not about updating the gatoms, but their checking whether the mouse is over them)?

and consequently, the CPU usage is measured in the `pd`(-core) process rather than the `pd-gui`?

Ant1r:
Yes, for my part I was just moving the mouse without any button pressed, and I measured the CPU load with the `Media/Load Meter`.

Lucarda:
Same as @Ant1r , I measured it with Windows "task manager".

Spacechild1:
@umlaeute yes and yes!

Lucarda:
This has to be looked deeply but i found that commenting

https://github.com/pure-data/pure-data/blob/master/src/g_canvas.c#L499

improves back performance and gives clues also for #1410

Lucarda:
This has come down substantially to normal figures


|Pd 0.51-4|	Pd 0.52-test4|
|:--|:--|
|~2%	|~3% |

porres:
> This has come down substantially to normal figures

with what? latest merges?

Lucarda:
Not sure exactly where. Now testing 621fb52 "**Pd 0.52-test4**"

[...]

PR #1519, opened 2021-12-15
stable development branch

umlaeute:

(as a general rule-of-thumb we shouldn't include changes to Pd-files in this PR, as it is just too easy to end up with an unresolvable file conflict)

it supersedes #1498

---

fixes so far (this list should be manually updated whenever new fixes are pushed):

- use C89-style declarations (preventing variable declaration in mid-code) everywhere (we somehow missed [src/s_midi_alsa.c](https://github.com/pure-data/pure-data/blob/621fb52400df4d5a964e9c2606ec06f9d9a2c59d/src/s_midi_alsa.c#L135))
- drop executable bit from `.pd` and `.txt` files
- use consistent line-endings


millerpuckette:
Anything else to do before calling it 0.52-0? Once this is merged it looks ready to me.

Spacechild1:
To me it looks good. AFAICT, the only remaining regression is https://github.com/pure-data/pure-data/issues/1459, but this one isn't really a show stopper and can wait for the next bugfix release.

I completely lost track of the macOS wish thing. Is this solved now?

millerpuckette:
I _think_ the MacOS Wish thing is OK now.  The remaining problem is that I should probably include compiled wish 8.6.10 with
Iohannes's patches, but I think he has a workaround for the CI system and I'm afraid of introducing bugs, so would rather put that off for after 0.52 if that's workable for everyone.

porres:
here are a couple of regression bugs, no show stoppers either but I have provided a fix https://github.com/pure-data/pure-data/pull/1515

I'm also done with https://github.com/pure-data/pure-data/pull/1517 did work today on revising 7.stuff/soundfile-tools

umlaeute:
> I _think_ the MacOS Wish thing is OK now.

This is basically my understanding as well.

>  The remaining problem is that I should probably include compiled wish 8.6.10 with
>Iohannes's patches,

Somewhat.
My patches are for 8.6.12, and my wish was to have a pre-compiled (and patched) Wish-8.6.12 included...


> but I think he has a workaround for the CI system

but this is correct from my POV: the iem-ci builds use a pre-compiled Wish-8.6.12(universal amd64+arm64) that is hosted on the iem-ci infrastructure (so it is a different binary from the one included with Pd)

PR #1531, opened 2021-12-25
Changing pkgconfig to pkg-config on windows mingw building instructions

abreubacelar:
hello, just learning how to compile pd on windows 10, and in both instructions about mingw it says to install pkgconfig, but trying that gives me:

~~~
$ pacman -S pkgconfig
error: target not found: pkgconfig
~~~

changing it to pkg-config it worked: 

~~~
$ pacman -S pkg-config
resolving dependencies...
looking for conflicting packages...

Packages (1) pkg-config-0.29.2-4
~~~

So I changed both INSTALL.txt and x6.htm, hope this is ok

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -319,12 +319,12 @@ Open an Msys2 shell and install the compiler chain, autotools, & gettext via:
 
     # 64 bit
     pacman -S mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang \
-              make pkg-config autoconf automake libtool \
+              make autoconf automake libtool \
               mingw-w64-x86_64-gettext
 
     # 32 bit
     pacman -S mingw-w64-i686-toolchain mingw-w64-i686-clang \
-              make pkg-config autoconf automake libtool \
+              make autoconf automake libtool \
               mingw-w64-i686-gettext
 
 Install git if you want to clone the Pd sources from Github, etc:
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -319,12 +319,12 @@ Open an Msys2 shell and install the compiler chain, autotools, & gettext via:
 
     # 64 bit
     pacman -S mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang \
-              make pkgconfig autoconf automake libtool \
+              make pkg-config autoconf automake libtool \
               mingw-w64-x86_64-gettext
 
     # 32 bit
     pacman -S mingw-w64-i686-toolchain mingw-w64-i686-clang \
-              make pkgconfig autoconf automake libtool \
+              make pkg-config autoconf automake libtool \
               mingw-w64-i686-gettext
 
 Install git if you want to clone the Pd sources from Github, etc:
--- a/doc/1.manual/x6.htm
+++ b/doc/1.manual/x6.htm
@@ -292,12 +292,12 @@ brew link --force gettext</pre>
 
 <pre># 64 bit
 pacman -S mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang \
-          make pkgconfig autoconf automake libtool \
+          make pkg-config autoconf automake libtool \
           mingw-w64-x86_64-gettext
 
 # 32 bit
 pacman -S mingw-w64-i686-toolchain mingw-w64-i686-clang \
-          make pkgconfig autoconf automake libtool \
+          make pkg-config autoconf automake libtool \
           mingw-w64-i686-gettext</pre>
 		  
 <p>Install git if you want to clone the Pd sources from Github, etc:</p>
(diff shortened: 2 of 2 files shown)


umlaeute:
if one of our windows gurus (@Spacechild1 , @Lucarda ,...) could confirm that this is indeed a typo (it sounds very reasonable to me, but i cannot check msys/mingw packages right now), i'd happily merge this into `develop` (or @Spacechild1 could do this :-))

Lucarda:
@umlaeute 

> could confirm that this is indeed a typo

confirmed. I forgot to update the pd docs but did it a year ago on http://puredata.info/docs/developer/WindowsMinGW 

it was not a typo (it worked a year ago). they suddenly change the pkg name.

Spacechild1:
Actually, there is `pkg-config` and there is `pkgconf`, the latter being a replacement of the former. Personally, I have `pkgconf` installed, but `pkg-config` seems to work as well. Not sure which one we should recommend?

Spacechild1:
```
Name            : mingw-w64-x86_64-pkgconf
Version         : 1.7.3-5
Description     : pkg-config compatible utility which does not depend on glib
Architecture    : any
URL             : https://github.com/pkgconf/pkgconf
Licenses        : ISC
Groups          : mingw-w64-x86_64-toolchain
Provides        : mingw-w64-x86_64-pkg-config
Depends On      : None
Optional Deps   : None
Required By     : None
Optional For    : None
Conflicts With  : mingw-w64-x86_64-pkg-config
Replaces        : mingw-w64-x86_64-pkg-config
Installed Size  : 476.38 KiB
Packager        : CI (msys2-autobuild/599a5db8/378501773)
Build Date      : Mon Nov 23 09:31:45 2020
Install Date    : Wed Mar 17 22:57:15 2021
Install Reason  : Explicitly installed
Install Script  : No
Validated By    : Signature
```

Spacechild1:
Maybe recommend the newer `pkgconf`, but add a note that `pkg-config` also works?

@Lucarda Do you have `pkgconf` or `pkg-config` installed?

Lucarda:
@Spacechild1 I have `pkg-config`.

Lucarda:
@Spacechild1 

Is this pkg used when `./autogen.sh` ?

A couple of month ago I updated msys2 (fresh install) and when doing `./autogen.sh` i got lots of warnings.

I can post them here if needed.

Spacechild1:
I don't see a problem in `autogen.sh`.

I *think* I have been using `pkg-config` before, but had to switch to `pkgconf` for openFrameworks.

I think this PR is fine, but we might add a note that you can also use `pkgconf` if you have already have it installed (because switching back to `pkg-config` might break things).

umlaeute:
> because switching back to pkg-config might break things

but not for Pd's build system.

probably something generic along the lines of 
> `pkg-config` or a compatible replacement

otoh: the documentation touched by this PR is really code-snippets that you are supposed to copy-paste to get yourself a build environment quickly.
if you are a developer that handles multiple different projects and one of them (e.g. ofx) is requiring a fork of a common tool like `pkg-config` i would expect you to be able to handle this yourself.

> Maybe recommend the newer `pkgconf`, but add a note that `pkg-config` also works?

having said the above i would prefer `pkg-config` as the default as this is a well-established tool and this is also what we are using.

> Is this pkg used when `./autogen.sh` ?

`pkg-config` is used by the `./configure` script (`autogen.sh` only "compiles" the `configure` script from some aramaic language to modern POSIX shell) - or so i thought. doing a quick grep it seems it's not actually used anywhere and could potentially be dropped altogether.

e.g. when building the Debian package of puredata, no `pkg-config` variant is installed at all.
of course there might be subtle differences between building Debian packages and windows binaries, but most likely `pkg-config` is needed nowhere...

Spacechild1:
> having said the above i would prefer pkg-config as the default as this is a well-established tool and this is also what we are using.

Agreed. That's what I wrote in my last post ;-)

> but most likely pkg-config is needed nowhere...

I just uninstalled `pkgconf` and I can still successfully do `./autogen.sh`, `configure` and `make`, so it's really not needed at all. I would suggest to remove it from the list.

Lucarda:
> Maybe @Lucarda can double check? You can do pacman -Qi > packages.txt to get a list of all installed packages.

on my win10 machine if I `pacman -Qi > packages.txt` i get `pkg-config` in the package itself and also "required by" `libiconv` (this last package is not required to build Pd I guess, see below)


> I would suggest to remove pkg-config from the list, so we don't potentially confuse people:

I agree. I just did a fresh install of msys2 on a win8.1 machine without installing `pkg-config` and it all went good. I also tried:

````
    touch template.pot
    make po
````

which went good, but 

````
    touch ../tcl/pd-gui.tcl
    make template
````	

didn't ( I got: `Make: nothing to be done for 'template'`). but this is unrelated as I get same results on the above win10 machine.

Lucarda:
@abreubacelar can you make one more commit that wipes out the 4 occurrences of `pkg-config` ?

I think we have quadruple checked the thing.

PR #1533, opened 2022-01-01
portmidi: fix rate limiting (again)

Spacechild1:
I just noticed that https://github.com/pure-data/pure-data/pull/641 missed a couple of things of my original PR.

Most notably, `LIMIT_RATE` is still hardcoded in `pmmacosxcm.c`, so `-DLIMIT_RATE=0` in `portmidi/Makefile.am` has no effect, since. I noticed this because of a compiler warning.

@danomatika Can you have a second look? Can this go into develop? Should we also send this upstream?

### DIFF
--- a/portmidi/portmidi/pm_mac/pmmacosxcm.c
+++ b/portmidi/portmidi/pm_mac/pmmacosxcm.c
@@ -63,7 +63,9 @@
 
      Note that this distorts accurate timestamps somewhat.
  */
+#ifndef LIMIT_RATE
 #define LIMIT_RATE 1
+#endif
 
 #define SYSEX_BUFFER_SIZE 128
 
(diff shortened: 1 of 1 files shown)


danomatika:
Is this being applied by hand? I use the `portmidi/update.sh` script to automate pulling in the portmidi sources and apply any needed patches. It was my understanding these changes were supposed to be included upstream. If not, then perhaps applying the patch failed the last time we updated portmidi.

Spacechild1:
> It was my understanding these changes were supposed to be included upstream. If not, then perhaps applying the patch failed the last time we updated portmidi.

I see! I think that's what happened.

> Is this being applied by hand?

I think, I just took the changes from #641. I wasn't aware of the patch file and `update.sh`. I've now tried to apply the patch by hand and run `update.sh`, but both failed...

danomatika:
The patch needs to be updated I guess as they’ve made changes upstream. Maybe we should restart the discussion with the portmidi devs about integrating this.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

millerpuckette:
Is this ready to merge into 0.52-3?  It looks OK but I don't have a Mac+MIDI interface I can test it on.

danomatika:
@millerpuckette This is safe since we had this before and it was accidentally removed, I think when I updated portmidi.

PR #1540, opened 2022-01-09
update file-help.pd wrong send in pd base+ext

abreubacelar:
The symbol atom in pd base+ext seems to be wrong, and was triggering the receive in another subpatch (pd split+join)

Edit: i also changed a comment that seems to be wrong too... it was talking about splittext, but was saying splitname


umlaeute:
note: i think this was already handled by @Ant1r ...

abreubacelar:
ah sorry then, closing...

abreubacelar:
the changes are other then the ones from Ant1r, so I open another pull request on porres documentation branch

Issue #1542, opened 2022-01-10
editing values in array "list view" only works on first page


umlaeute:
opening the "List View" of an array, I can double click on an item and enter a new value.

however, this only works on the first page.
editing a value on a non-first page, will change the value on the corresponding index on the first page instead.

e.g. 
- change the 3rd value on the 2nd page from "0.123" to "3.14" and pressing <kbd>Return</kbd>
- the displayed values (of the 2nd page) apparently did not change
- going back to the 1st page, it becomes apparent that the 3rd value on the 1st page became "3.14"

umlaeute:
the problem is obviously that `::pd_array_listview_pagesize` defaults to **0** (but it should be the same value as [`ARRAYPAGESIZE`](https://github.com/pure-data/pure-data/blob/bad155d26cceead85e845da4f7c3d41498ea0574/src/g_array.c#L13)

Issue #1541, opened 2022-01-09
[array] (Tcl) UNHANDLED ERROR: window name starts with an upper-case letter


hwkcgn:
when creating an array with a name starting with an uppercase letter (e.g. E05-tap) and then trying to see list view in properties, pd throws an error (observed on mac 11.6.2. with with Pd-0.52-1-really and Pd-0.51-4-macosx and as per @Spacechild1 also with Pd 0.49): 

~~~
(Tcl) UNHANDLED ERROR: window name starts with an upper-case letter: "E05-tabArrayWindow"
   while executing
"toplevel $windowName -class DialogWindow"
   (procedure "pdtk_array_listview_new" line 6)
   invoked from within
"pdtk_array_listview_new .gfxstub7fe688fc4330 E05-tab 0"
   ("uplevel" body line 2)
   invoked from within
"uplevel #0 $docmds"(Tcl) UNHANDLED ERROR: window name starts with an upper-case letter: "E05-tabArrayWindow"
   while executing
"toplevel $windowName -class DialogWindow"
   (procedure “pdtk_array_l
~~~

PR #1544, opened 2022-01-11
array-editor fixes & improvements

umlaeute:
this PR fixes a couple of issues with the "array editor" and improves it otherwise (<kbd>List View</kbd> in the array preferences)

### bugfixes
- allow the user to open the <kbd>List View</kbd> for arrays with names that start with a capital letter (Closes: #1541)
- allow the user to open the <kbd>List View</kbd> for arrays with names that contain spaces
- fix editing of values that are not on the first page (Closes: #1542 )
- remove code absurdities (like `[expr $x + [expr $y + [expr $a * $b]]]`
- remove assignment absurdities (like `set $winid.x [entry $winid.x]`)
- refactored and simplified code

### improvements
- abstract interface for sending paged array-data from Pd-core to Pd-GUI
  - Pd-core no longer needs to worry how the data is stored (or represented) on the GUI side
  - we can send multiple values at once, rather than element by element
  - the GUI figures out how to display the values by itself
- (optionally) uses a nicer table-widget for showing the data
- add an entry box to directly go to a data page
- copy/paste now works on all platforms
  - either via the context-menu (that used to be windows only)
  - or via the default bindings (<kbd>Ctrl</kbd>+<kbd>C</kbd> / <kbd>Ctrl</kbd>+<kbd>V</kbd>)

### DIFF
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -475,6 +475,20 @@ void garray_arraydialog(t_garray *x, t_symbol *name, t_floatarg fsize,
 }
 
 /* jsarlo { */
+static int garray_arrayviewlist_setpage(t_symbol*arrayname, t_array*a, int page, int pagesize) {
+
+        /* make szre the requested page is within range */
+    int maxpage = (a->a_n - 1) / pagesize;
+    if(page < 0)
+        page = 0;
+    if(page > maxpage)
+        page = maxpage;
+
+    sys_vgui("::dialog_array::listview_setpage {%s} %d %d %d\n",
+        arrayname->s_name,
+        page, maxpage+1, pagesize);
+    return page;
+}
 void garray_arrayviewlist_new(t_garray *x)
 {
     int i, yonset=0, elemsize=0, page=0;
@@ -495,9 +509,7 @@ void garray_arrayviewlist_new(t_garray *x)
             0);
     gfxstub_new(&x->x_gobj.g_pd, x, cmdbuf);
 
-    sys_vgui("::dialog_array::listview_setpage {%s} %d\n",
-        x->x_realname->s_name,
-        (int)page);
+    page = garray_arrayviewlist_setpage(x->x_realname, a, page, ARRAYPAGESIZE);
 
     sys_vgui("::dialog_array::listview_setdata {%s} %ld",
         x->x_realname->s_name,
@@ -529,16 +541,7 @@ void garray_arrayviewlist_fillpage(t_garray *x,
         return;
     }
 
-        /* make sure that page is within range */
-    if (page < 0) {
-        page = 0;
-    }
-    else if ((page * ARRAYPAGESIZE) >= a->a_n) {
-        page = (int)(((int)a->a_n - 1)/ (int)ARRAYPAGESIZE);
-    }
-    sys_vgui("::dialog_array::listview_setpage {%s} %d\n",
-        x->x_realname->s_name,
-        (int)page);
+    page = garray_arrayviewlist_setpage(x->x_realname, a, page, ARRAYPAGESIZE);
 
     sys_vgui("::dialog_array::listview_setdata {%s} %ld",
         x->x_realname->s_name,
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -475,32 +475,54 @@ void garray_arraydialog(t_garray *x, t_symbol *name, t_floatarg fsize,
 }
 
 /* jsarlo { */
-static int garray_arrayviewlist_setpage(t_symbol*arrayname, t_array*a, int page, int pagesize)
+static void garray_arrayviewlist_fillpage(t_garray *x,
+                                   t_float fPage,
+                                   t_float fTopItem)
 {
+    int i, size=0, topItem=(int)fTopItem;
+    int pagesize=ARRAYPAGESIZE, page=(int)fPage, maxpage;
+    t_word *data=0;
 
-        /* make szre the requested page is within range */
-    int maxpage = (a->a_n - 1) / pagesize;
-    if(page < 0)
-        page = 0;
+    if(!garray_getfloatwords(x, &size, &data)) {
+        pd_error(x, "error in %s()", __FUNCTION__);
+        return;
+    }
+
+        /* make sure the requested page is within range */
+    maxpage = (size - 1) / pagesize;
     if(page > maxpage)
         page = maxpage;
+    if(page < 0)
+        page = 0;
 
     sys_vgui("::dialog_array::listview_setpage {%s} %d %d %d\n",
-        arrayname->s_name,
+        x->x_realname->s_name,
         page, maxpage+1, pagesize);
-    return page;
+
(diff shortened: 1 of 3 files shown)


danomatika:
Screenshot(s)?

umlaeute:
here's a (not super-interesting) screenshot:

dialog_array

the merit of the PR is mostly in the cleanup of the codebase.

danomatika:
>the merit of the PR is mostly in the cleanup of the codebase.

Yup. Great. Little by little...

danomatika:
Should we cherry-pick 1edd501f6820f1d15410829bb3280fa108c83bfb into the develop branch?

umlaeute:
> Should we cherry-pick 1edd501 into the develop branch?

no. it depends on a couple of commits that precede it.

src/g_array.c
```diff
@@ -475,34 +475,54 @@ void garray_arraydialog(t_garray *x, t_symbol *name, t_floatarg fsize,
...
+
+    sys_vgui("::dialog_array::listview_setdata {%s} %ld",
+        x->x_realname->s_name,
+        (long long)(page * ARRAYPAGESIZE));
+    for (i = page * ARRAYPAGESIZE;
+         (i < (page + 1) * ARRAYPAGESIZE && i < a->a_n);
+         i++)
     {
-        yval = *(t_float *)(a->a_vec +
+        yval = *(t_float *)(a->a_vec + \
```
Spacechild1, line 521: I guess the `(t_float *)` cast is a relict of the 32-bit era. It happens to work, but it would be more correct to cast to the actual type (`t_word`).

Also, what's up with that backslash? :-)

Spacechild1: Actually, why not simply use `garray_getfloatwords`? For garrays, `elemsize` and `yonset` are not relevant.

src/g_array.c
```diff
@@ -516,46 +536,34 @@ void garray_arrayviewlist_fillpage(t_garray *x,
...
+
+    sys_vgui("::dialog_array::listview_setdata {%s} %ld",
+        x->x_realname->s_name,
+        (long long)(page * ARRAYPAGESIZE));
     for (i = page * ARRAYPAGESIZE;
          (i < (page + 1) * ARRAYPAGESIZE && i < a->a_n);
          i++)
     {
         yval = *(t_float *)(a->a_vec + \
                elemsize * i + yonset);
```
Spacechild1, line 554: see `garray_arrayviewlist_new`

umlaeute: also fixed (by dropping the duplicate implementation )

src/g_array.c
```diff
@@ -475,90 +475,64 @@ void garray_arraydialog(t_garray *x, t_symbol *name, t_floatarg fsize,
...
+    int i, size=0, topItem=(int)fTopItem;
+    int pagesize=ARRAYPAGESIZE, page=(int)fPage, maxpage;
+    t_word *data=0;
 
-        /* make szre the requested page is within range */
-    int maxpage = (a->a_n - 1) / pagesize;
-    if(page < 0)
-        page = 0;
+    if(!garray_getfloatwords(x, &size, &data)) {
+        pd_error(x, "error in %s()", __FUNCTION__);
```
Spacechild1, line 487: Small nitpick: I think that `bug()` would be more appropriate here, as the function should never fail. Same in `garray_arrayviewlist_new`.

umlaeute: probaby.
but then, `pd_error()` allows us to trace the error to the problematic array, whereas `bug()` does not (but gives us an extra 'consistency check failed').

which one should be preferred?

(`error()` has been used in this place since the original implementation in 2004 :stuck_out_tongue: )

danomatika:
In looking into this PR, I notice a *slightly* related issue:

1. click on an `[array define]` object to open the table view
2. right-click to open the context menu
3. choose any option and the console prints a `array: no method for 'done-popup'` error

danomatika:
Also... how do I show the list view again? It's been a while since I looked at it and I don't remember how to trigger the dialog. :P

Spacechild1:
> Also... how do I show the list view again? It's been a while since I looked at it and I don't remember how to trigger the dialog. :P

It's in the array properties ("Show list view")

millerpuckette:
While you're looking at this... what if we just replaced that whole 'list view'
with a simple text window?  I'm thinking we could just have "copy" copy all
the numbers from the array in ascii and vice versa for 'paste', and add a
plain old text window to Pd that would also be optionally bound to a 'text
define'.  That would be far simpler and I think more powerful.

cheers
Miller

danomatika:
Ok, on my macOS 11.6.1 system, the list view headers are slightly broken using the Tk 8.6.10 Wish in `mac/stuff` while a Pd.app made with Tk 8.6.12 Wish works ok.

Tk 8.6.10:
Screen Shot 2022-01-23 at 5 52 58 PM

Tk 8.6.12:
Screen Shot 2022-01-23 at 5 54 09 PM

I'd agree with Miller's suggesting about (re)using the text view editor, however there would need to be numeric validation and some sort of line numbering.

umlaeute:
> Ok, on my macOS 11.6.1 system, the list view headers are slightly broken using the Tk 8.6.10

Thanks for checking.
I think the brokenness can be ignored (that is: the columns are pretty self-explanatory and don't require any headings, even if they are nice)


>I'd agree with Miller's suggesting about (re)using the text view editor, however there would need to be numeric validation and some sort of line numbering.

I'm not fully sure about Miller's suggestion, basically because I think your "howevers" are crucial:
- for such large uniform datasets line numbering is crucial, and I haven't seen a nice tk widget yet
- I don't think it's much fun to edit a list of numbers in a text window. There must be a reason why excel uses cells (apart from evil plans for world dominion)

additionally:
- we probably do want paging, as the size of tables can be pretty large
- I *really* as much code of the representation specifics to live on the GUI side rather than the "great unified everything-looks-the-same view and let the core figure out what to do with the data" scheme (Obviously we could still have a simple text window represent the data)

I think the most compelling reason to have a simple text window is that we don't have to implement another view (but then: the other view is already implemented...)

danomatika:
> I think the brokenness can be ignored (that is: the columns are pretty self-explanatory and don't require any headings, even if they are nice)

I disagree on this.

umlaeute:
> > I think the brokenness can be ignored (that is: the columns are pretty self-explanatory and don't require any headings, even if they are nice)
> 
> I disagree on this.

you are welcome :-)

i any case, i think it is not a regression: the original code lacked headings as well.

danomatika:
> > > I think the brokenness can be ignored (that is: the columns are pretty self-explanatory and don't require any headings, even if they are nice)
> > 
> > 
> > I disagree on this.
> 
> you are welcome :-)
> 
> i any case, i think it is not a regression: the original code lacked headings as well.

In any case, the release build looks fine with Tk 8.6.12 on a macOS 12 system (thankfully):

Screen Shot 2022-03-30 at 11 58 15 AM

PR #1560, opened 2022-01-23
fix PERTHREAD build error by setting macos version min 10.9

### LINKED ISSUE #1443: PDINSTANCE support on macOS?
it seems that there is some problem with the `PERTHREAD` declaration on macOS:

~~~sh
$ ./configure --enable-libpd --enable-libpd-instance
[...]
$ make
[...]
/bin/sh ../libtool  --tag=CC   --mode=compile gcc -DPACKAGE_NAME=\"pd\" -DPACKAGE_TARNAME=\"pd\" -DPACKAGE_VERSION=\"0.52.0\" -DPACKAGE_STRING=\"pd\ 0.52.0\" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"pd\" -DVERSION=\"0.52.0\" -DHAVE_STDIO_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_STRINGS_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_UNISTD_H=1 -DSTDC_HEADERS=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DHAVE_FCNTL_H=1 -DHAVE_LIMITS_H=1 -DHAVE_NETDB_H=1 -DHAVE_NETINET_IN_H=1 -DHAVE_STDDEF_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_SYS_IOCTL_H=1 -DHAVE_SYS_PARAM_H=1 -DHAVE_SYS_SOCKET_H=1 -DHAVE_SYS_TIME_H=1 -DHAVE_SYS_TIMEB_H=1 -DHAVE_UNISTD_H=1 -DHAVE_ALLOCA_H=1 -DHAVE_ALLOCA=1 -DHAVE_FORK=1 -DHAVE_VFORK=1 -DHAVE_WORKING_VFORK=1 -DHAVE_WORKING_FORK=1 -DHAVE_MALLOC=1 -DHAVE_REALLOC=1 -DRETSIGTYPE=void -DHAVE_DUP2=1 -DHAVE_FLOOR=1 -DHAVE_GETCWD=1 -DHAVE_GETHOSTBYNAME=1 -DHAVE_GETTIMEOFDAY=1 -DHAVE_MEMMOVE=1 -DHAVE_MEMSET=1 -DHAVE_POW=1 -DHAVE_REGCOMP=1 -DHAVE_SELECT=1 -DHAVE_SOCKET=1 -DHAVE_SQRT=1 -DHAVE_STRCHR=1 -DHAVE_STRERROR=1 -DHAVE_STRRCHR=1 -DHAVE_STRSTR=1 -DHAVE_STRTOL=1 -DHAVE_LIBDL=1 -I.  -DPD -DPD_INTERNAL -DUSEAPI_DUMMY  -DPDINSTANCE=1 -DPDTHREADS=1 -DLIBPD_EXTRA=1 -DPD_FLOATSIZE=32  -mmacosx-version-min=10.6  -ffast-math -funroll-loops -fomit-frame-pointer -O3 -MT libpd_la-z_libpd.lo -MD -MP -MF .deps/libpd_la-z_libpd.Tpo -c -o libpd_la-z_libpd.lo `test -f 'z_libpd.c' || echo './'`z_libpd.c
libtool: compile:  gcc -DPACKAGE_NAME=\"pd\" -DPACKAGE_TARNAME=\"pd\" -DPACKAGE_VERSION=\"0.52.0\" "-DPACKAGE_STRING=\"pd 0.52.0\"" -DPACKAGE_BUGREPORT=\"\" -DPACKAGE_URL=\"\" -DPACKAGE=\"pd\" -DVERSION=\"0.52.0\" -DHAVE_STDIO_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_INTTYPES_H=1 -DHAVE_STDINT_H=1 -DHAVE_STRINGS_H=1 -DHAVE_SYS_STAT_H=1 -DHAVE_SYS_TYPES_H=1 -DHAVE_UNISTD_H=1 -DSTDC_HEADERS=1 -DHAVE_DLFCN_H=1 -DLT_OBJDIR=\".libs/\" -DHAVE_FCNTL_H=1 -DHAVE_LIMITS_H=1 -DHAVE_NETDB_H=1 -DHAVE_NETINET_IN_H=1 -DHAVE_STDDEF_H=1 -DHAVE_STDLIB_H=1 -DHAVE_STRING_H=1 -DHAVE_SYS_IOCTL_H=1 -DHAVE_SYS_PARAM_H=1 -DHAVE_SYS_SOCKET_H=1 -DHAVE_SYS_TIME_H=1 -DHAVE_SYS_TIMEB_H=1 -DHAVE_UNISTD_H=1 -DHAVE_ALLOCA_H=1 -DHAVE_ALLOCA=1 -DHAVE_FORK=1 -DHAVE_VFORK=1 -DHAVE_WORKING_VFORK=1 -DHAVE_WORKING_FORK=1 -DHAVE_MALLOC=1 -DHAVE_REALLOC=1 -DRETSIGTYPE=void -DHAVE_DUP2=1 -DHAVE_FLOOR=1 -DHAVE_GETCWD=1 -DHAVE_GETHOSTBYNAME=1 -DHAVE_GETTIMEOFDAY=1 -DHAVE_MEMMOVE=1 -DHAVE_MEMSET=1 -DHAVE_POW=1 -DHAVE_REGCOMP=1 -DHAVE_SELECT=1 -DHAVE_SOCKET=1 -DHAVE_SQRT=1 -DHAVE_STRCHR=1 -DHAVE_STRERROR=1 -DHAVE_STRRCHR=1 -DHAVE_STRSTR=1 -DHAVE_STRTOL=1 -DHAVE_LIBDL=1 -I. -DPD -DPD_INTERNAL -DUSEAPI_DUMMY -DPDINSTANCE=1 -DPDTHREADS=1 -DLIBPD_EXTRA=1 -DPD_FLOATSIZE=32 -mmacosx-version-min=10.6 -ffast-math -funroll-loops -fomit-frame-pointer -O3 -MT libpd_la-z_libpd.lo -MD -MP -MF .deps/libpd_la-z_libpd.Tpo -c z_libpd.c  -fno-common -DPIC -o .libs/libpd_la-z_libpd.o
In file included from z_libpd.c:20:
In file included from ./z_libpd.h:20:
./m_pd.h:890:8: error: thread-local storage is not supported for the current target
extern PERTHREAD t_pdinstance *pd_this;
       ^
./m_pd.h:883:19: note: expanded from macro 'PERTHREAD'
#define PERTHREAD __thread
                  ^
z_libpd.c:57:8: error: thread-local storage is not supported for the current target
static PERTHREAD t_atom *s_argv = NULL;
       ^
./m_pd.h:883:19: note: expanded from macro 'PERTHREAD'
#define PERTHREAD __thread
                  ^
z_libpd.c:58:8: error: thread-local storage is not supported for the current target
static PERTHREAD t_atom *s_curr = NULL;
       ^
./m_pd.h:883:19: note: expanded from macro 'PERTHREAD'
#define PERTHREAD __thread
                  ^
z_libpd.c:59:8: error: thread-local storage is not supported for the current target
static PERTHREAD int s_argm = 0;
       ^
./m_pd.h:883:19: note: expanded from macro 'PERTHREAD'
#define PERTHREAD __thread
                  ^
z_libpd.c:60:8: error: thread-local storage is not supported for the current target
static PERTHREAD int s_argc = 0;
       ^
./m_pd.h:883:19: note: expanded from macro 'PERTHREAD'
#define PERTHREAD __thread
                  ^
5 errors generated.
~~~

full build logs at https://git.iem.at/pd/pure-data/-/jobs/32431

i don't know whether this is a limitation of clang/macOS/... or a bug in the use of `PERTHREAD`

danomatika:
This PR fixes #1443 by adding logic to configure to set the macOS version min to 10.9 when building with `--enable-libpd-instance`. Building with instances on < 10.9 will most likely fail anyway, so there aren't any macOS version checks.

There is also an indentation fix to the added libpd stuff.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -11,7 +11,7 @@ AC_CANONICAL_HOST
 
 
 ########################################
-#### ohoh. automake<<1.16 do not like $(top_srcdir) in _SOURCES.
+#### ohoh. automake<<1.16 does not like $(top_srcdir) in _SOURCES.
 AS_IF([test "x${am__api_version}" = "x1.15" ], [
   enable_dependency_tracking=no
   AC_MSG_WARN([legacy automake detected: disabled dependency-tracking!
@@ -69,8 +69,8 @@ AS_CASE([$host],
         # a set of search paths are used on macOS in s_inter.c
         wish="default search paths"
 
-        # required for dlopen & weak linking for older OSX version support
-        CFLAGS="-mmacosx-version-min=10.6 $CFLAGS"
+        # required for dlopen & weak linking for older macOS version support
+        macos_version_min=10.6
     ],[
         platform=iOS
         locales=no
@@ -284,6 +284,8 @@ AS_IF([test x$debug = xyes],[
 
 #########################################
 ##### Configure Options #####
+
+##### libpd #####
 AC_ARG_ENABLE([libpd],
     [AS_HELP_STRING([--enable-libpd], [additionally build libpd])])
 AC_ARG_ENABLE([libpd-utils],
@@ -539,15 +541,26 @@ AS_IF([test ! "x${WISH}" = "x"],[
 
 ##### libpd #####
 AS_IF([test "x${enable_libpd}" = "xyes"],[
-        libpd="yes"
-        libpd_=""
-        AS_IF([test "x${enable_libpd_instance}" = "xyes"],[ libpd_="${libpd_}+multi "])
-        AS_IF([test "x${enable_libpd_utils}" = "xyes"],[ libpd_="${libpd_}+utils "])
-        AS_IF([test "x${enable_libpd_extra}" = "xyes"],[ libpd_="${libpd_}+extra "])
-        AS_IF([test "x${enable_libpd_setlocale}" = "xno"],[ libpd_="${libpd_}-setlocale "])
-        AS_IF([test "x${libpd_}" != "x"],[ libpd="${libpd} ( ${libpd_})"])
+    libpd="yes"
+    libpd_=""
+    AS_IF([test "x${enable_libpd_instance}" = "xyes"],[
+        libpd_="${libpd_}+multi "
+        AS_IF([test "x$MACOSX" = "xyes"],[
+            macos_version_min=10.9
+            AC_MSG_NOTICE([Setting macOS version min 10.9 for libpd instance])
+        ])
+    ])
+    AS_IF([test "x${enable_libpd_utils}" = "xyes"],[ libpd_="${libpd_}+utils "])
+    AS_IF([test "x${enable_libpd_extra}" = "xyes"],[ libpd_="${libpd_}+extra "])
+    AS_IF([test "x${enable_libpd_setlocale}" = "xno"],[ libpd_="${libpd_}-setlocale "])
+    AS_IF([test "x${libpd_}" != "x"],[ libpd="${libpd} ( ${libpd_})"])
 ],[
-        libpd="no"
+    libpd="no"
+])
+
+##### macOS version min #####
+AS_IF([test "x$macos_version_min" != "x"],[
+    CFLAGS="-mmacosx-version-min=$macos_version_min $CFLAGS"
 ])
 
 #########################################
--- a/configure.ac
+++ b/configure.ac
@@ -47,6 +47,11 @@ midi_backends=""
 DEBUG_CFLAGS="-O0"
 RELEASE_CFLAGS="-ffast-math -funroll-loops -fomit-frame-pointer -O3"
 
+# common flags applied to all pd binaries
+PD_CPPFLAGS=""
+PD_CFLAGS=""
+PD_LDFLAGS=""
+
 #########################################
 ##### OS Detection #####
 
@@ -80,19 +85,19 @@ AS_CASE([$host],
     # homebrew paths
     AS_IF([test -e /usr/local],[
         AM_CPPFLAGS="-I/usr/local/include $INCLUDES"
-        LDFLAGS="-L/usr/local/lib $LDFLAGS"
+        PD_LDFLAGS="-L/usr/local/lib $PD_LDFLAGS"
     ])
 
     # fink paths
     AS_IF([test -e /sw],[
         AM_CPPFLAGS="-I/sw/include $INCLUDES"
-        LDFLAGS="-L/sw/lib $LDFLAGS"
+        PD_LDFLAGS="-L/sw/lib $PD_LDFLAGS"
     ])
 
     # macports paths
     AS_IF([test -e /opt/local],[
         AM_CPPFLAGS="-I/opt/local/include $INCLUDES"
-        LDFLAGS="-L/opt/local/lib $LDFLAGS"
+        PD_LDFLAGS="-L/opt/local/lib $PD_LDFLAGS"
     ])
 
     EXTERNAL_LDFLAGS="-bundle -undefined dynamic_lookup"
@@ -276,10 +281,15 @@ AC_ARG_ENABLE([debug],
     [AS_HELP_STRING([--enable-debug], [use debugging support])],
     [debug=$enableval], [debug=no])
 AS_IF([test x$debug = xyes],[
-    CFLAGS="$CFLAGS $DEBUG_CFLAGS"
+    PD_CFLAGS="$DEBUG_CFLAGS $PD_CFLAGS"
 ],[
-    CFLAGS="$CFLAGS $RELEASE_CFLAGS"
-    CPPFLAGS="$CPPFLAGS -DNDEBUG"
+    PD_CFLAGS="$RELEASE_CFLAGS $PD_CFLAGS"
+    PD_CPPFLAGS="-DNDEBUG $PD_CPPFLAGS"
+])
+
+##### macOS version min #####
+AS_IF([test "x$macos_version_min" != "x"],[
+    PD_CFLAGS="-mmacosx-version-min=$macos_version_min $PD_CFLAGS"
 ])
 
 #########################################
@@ -307,8 +317,8 @@ AM_CONDITIONAL(LIBPD_NO_SETLOCALE, test x$enable_libpd_setlocale = xno)
 PD_CHECK_UNIVERSAL(ARCH, [universal=yes], [universal=no])
 AM_CONDITIONAL(UNIVERSAL, test x$universal = xyes)
 AS_IF([test x$universal = xyes],[
-    CFLAGS="$ARCH_CFLAGS $CFLAGS"
-    LDFLAGS="$ARCH_LDFLAGS $LDFLAGS"
+    PD_CFLAGS="$ARCH_CFLAGS $PD_CFLAGS"
+    PD_LDFLAGS="$ARCH_LDFLAGS $PD_LDFLAGS"
 ])
 
 ##### Gettext #####
@@ -543,13 +553,7 @@ AS_IF([test ! "x${WISH}" = "x"],[
 AS_IF([test "x${enable_libpd}" = "xyes"],[
     libpd="yes"
     libpd_=""
-    AS_IF([test "x${enable_libpd_instance}" = "xyes"],[
-        libpd_="${libpd_}+multi "
-        AS_IF([test "x$MACOSX" = "xyes"],[
-            macos_version_min=10.9
-            AC_MSG_NOTICE([Setting macOS version min 10.9 for libpd instance])
-        ])
-    ])
+    AS_IF([test "x${enable_libpd_instance}" = "xyes"],[ libpd_="${libpd_}+multi "])
     AS_IF([test "x${enable_libpd_utils}" = "xyes"],[ libpd_="${libpd_}+utils "])
     AS_IF([test "x${enable_libpd_extra}" = "xyes"],[ libpd_="${libpd_}+extra "])
     AS_IF([test "x${enable_libpd_setlocale}" = "xno"],[ libpd_="${libpd_}-setlocale "])
@@ -558,10 +562,10 @@ AS_IF([test "x${enable_libpd}" = "xyes"],[
     libpd="no"
 ])
 
-##### macOS version min #####
-AS_IF([test "x$macos_version_min" != "x"],[
-    CFLAGS="-mmacosx-version-min=$macos_version_min $CFLAGS"
-])
+# pass common flags via @PD_*@ AM variables for use in Makefiles
+AC_SUBST(PD_CPPFLAGS)
+AC_SUBST(PD_CFLAGS)
+AC_SUBST(PD_LDFLAGS)
 
 #########################################
 ##### Output files #####
@@ -610,9 +614,9 @@ AC_MSG_NOTICE([
     Installation prefix:  $prefix
 
     Compiler:             $CC
-    CPPFLAGS:             $CPPFLAGS
-    CFLAGS:               $CFLAGS
-    LDFLAGS:              $LDFLAGS
+    CPPFLAGS:             $PD_CPPFLAGS $CPPFLAGS
+    CFLAGS:               $PD_CFLAGS $CFLAGS
+    LDFLAGS:              $PD_LDFLAGS $LDFLAGS
     INCLUDES:             $AM_CPPFLAGS
     LIBS:                 $LIBS
 
--- a/extra/bob~/GNUmakefile.am
+++ b/extra/bob~/GNUmakefile.am
@@ -17,11 +17,11 @@ EXTRA_DIST = makefile
 dist_external_DATA = $(PATCHES) $(OTHERDATA)
 
 AUTOMAKE_OPTIONS = foreign
-AM_CFLAGS = @EXTERNAL_CFLAGS@
-AM_CPPFLAGS	+= -I$(top_srcdir)/src -DPD
+AM_CFLAGS = @EXTERNAL_CFLAGS@ @PD_CFLAGS@
+AM_CPPFLAGS	+= -I$(top_srcdir)/src -DPD @PD_CPPFLAGS@
 bob__la_LIBADD = $(LIBM)
 AM_LDFLAGS = -module -avoid-version -shared @EXTERNAL_LDFLAGS@ \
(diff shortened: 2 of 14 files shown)


danomatika:
The potential issue I see with this is applying `-mmacosx-version-min=10.9` when building Pd itself, but that shouldn't result in anything major as older externs should still load ok.

danomatika:
Another approach would be to set the flag in the Makefile.am when building libpd only. My first approach was the simplest.

configure.ac
```diff
@@ -539,15 +541,26 @@ AS_IF([test ! "x${WISH}" = "x"],[
...
+    libpd="yes"
+    libpd_=""
+    AS_IF([test "x${enable_libpd_instance}" = "xyes"],[
+        libpd_="${libpd_}+multi "
+        AS_IF([test "x$MACOSX" = "xyes"],[
+            macos_version_min=10.9
+            AC_MSG_NOTICE([Setting macOS version min 10.9 for libpd instance])
+        ])
+    ])
+    AS_IF([test "x${enable_libpd_utils}" = "xyes"],[ libpd_="${libpd_}+utils "])
```
umlaeute, line 553: oh, this breaks the nice layout and is much harder to parse.
i think it would be nicer to separate the `+multi` suffixing from the `macos_version_min` assignment:

~~~m4
AS_IF([test "x${enable_libpd_instance}" = "xyes"],[ libpd_="${libpd_}+multi "])
# ...
AS_IF([test "x${enable_libpd_instance}" = "xyes" && test "x$MACOSX" = "xyes"],[
    macos_version_min=10.9
    AC_MSG_NOTICE([Setting macOS version min 10.9 for libpd instance])
])
~~~

see also my other comment

configure.ac
```diff
@@ -539,15 +541,26 @@ AS_IF([test ! "x${WISH}" = "x"],[
...
+    AS_IF([test "x${enable_libpd_setlocale}" = "xno"],[ libpd_="${libpd_}-setlocale "])
+    AS_IF([test "x${libpd_}" != "x"],[ libpd="${libpd} ( ${libpd_})"])
 ],[
-        libpd="no"
+    libpd="no"
+])
+
+##### macOS version min #####
+AS_IF([test "x$macos_version_min" != "x"],[
+    CFLAGS="-mmacosx-version-min=$macos_version_min $CFLAGS"
```
umlaeute, line 563: :+1:

umlaeute (commented):
since now the `-mmacosx-version-min=` is only added if `macos_version_min` is set, wouldn't it make more sense for the libpd-instance case to just unset the the variable?

i'm mostly afraid of breaking builds with older XCode versions (that do not support `10.9` yet) - no idea what miller is using to produce his OSX10.6 builds.

not defining the macos-version-min ~~might~~<kbd>^W</kbd>will produce a library that is not usable on older versions of XCode.

umlaeute:
> not defining the macos-version-min ~~might~~^W*will* produce a library that is not usable on older versions of XCode.

ah, this is of course bad as it affects not only the libpd binary but also the main Pd binary.

in that case i would actually prefer:
~~~automake
if LIBPD_INSTANCE
libpd_la_CFLAGS += -mmacosx-version-min=10.9
endif
~~~

probably even overriding only if the macos-version-min is not already set (moving the `AS_IF([test "x$macos_version_min" != "x"]` check from configure.ac to  Makefile.am)

danomatika:
> wouldn't it make more sense for the libpd-instance case to just unset the the variable?

If the variable is *not* set, the build can only load dynamic libs similarly built on the same system version. It should *always* be set.

>i'm mostly afraid of breaking builds with older XCode versions (that do not support 10.9 yet) - no idea what miller is using to produce his OSX10.6 builds.

10.9 is only used for PERTHREAD and this won't work on earlier versions of Xcode anyway. Personally, I see this build option as an endcase.

danomatika:
From Apple [doc](https://developer.apple.com/library/archive/documentation/Porting/Conceptual/PortingUnix/compiling/compiling.html):

>-mmacosx-version-min version
>Specifies the version of OS X you are targeting. You must target your compile for the oldest version of OS X on which you want to run the executable. In addition, you should install and use the cross-development SDK for that version of OS X.

There is a better page somewhere, but I couldn't find it.

Spacechild1:
> i'm mostly afraid of breaking builds with older XCode versions (that do not support 10.9 yet) - no idea what miller is using to produce his OSX10.6 builds.

@umlaeute Such XCode versions won't be able to build multi-instance Pd anyway because of the missing support for thread local storage. Miller only ships "regular" single-instance Pd builds.

I think @danomatika's approach is fine. Actually, I did almost the same in an experimental Pd branch of mine that requires thread local storage.

Spacechild1:
On the other hand, since libpd rebuilds all the sources, it would make sense to only apply `PDINSTANCE` to the libpd sources (and raise the macosxc min version), but not to the sources for the main executable. After all, a multi-instance Pd *app* is not very useful.

umlaeute:
> After all, a multi-instance Pd app is not very useful.

i think it would have been super-useful to have a multi-instance Pd.app from the beginning (so externals are always build for multi-instance Pd, regardless of whether they are then used by a Pd.app or some libpd)

Spacechild1:
> i think it would have been super-useful to have a multi-instance Pd.app from the beginning

I would have also been super-useful to have proper multi-instance support from the beginning (without the thread-local-storage hack).

What I meant was that a multi-instance Pd app is not useful *now* because of the current ecosystem. This doesn't mean that we can't switch everything to `PDINSTANCE` in the future.

Ant1r:
> This doesn't mean that we can't switch everything to `PDINSTANCE` in the future.

Moreover this would need to recompile every external...

Spacechild1:
> Moreover this would need to recompile every external...

Yes. This is what I meant when I referred to the "current ecosystem".

danomatika:
So I've tried the following:

~~~
# PERTHREAD requires macOS 10.9+ SDK
if LIBPD_MULTIINSTANCE
libpd_la_CFLAGS = -mmacosx-version-min=10.9
endi
~~~

However, this *prepends* to the CFLAGS so 10.9 is set then 10.6 is set afterward coming from configure and overrides 10.9. I could imagine adding an additional `MACOSX_VERSION_MIN` variable but then it would need to be manually added to all targets ala pd, `extra` externals, etc.

umlaeute:
> However, this prepends to the CFLAGS  [...]

i see.
the problem is that we are currently abusing the `CFLAGS` variable for setting obligatory CFLAGS (like `-mmacos-min-version`).

`CFLAGS` (or rather any non-qualified `*CFLAGS` - including `CPPFLAGS`, `CXXFLAGS`, `LDFLAGS`) are supposed to provide default values that are overridable during the `make` invocation (e.g. `make CFLAGS=-O0`) and which in turn can be used to override any system-provided default flags. therefore it comes last.

we should switch to using distinct `*FLAGS` variables for obligatory flags.

danomatika:
What do you propose? Use `CPPFLAGS` instead?

danomatika:
Actually, note that providing CFLAGS during configure or make still appends to allow overriding, I'm just pointing out that it doesn't work similarly when using the libtool variable.

umlaeute:
> What do you propose? Use `CPPFLAGS` instead?

no, not at all.
that way we would just abuse a variable from another *domain* (that just happens to sort in the right order).

i'd propose using something like `OBLIGATORY_CFLAGS` instead of `CFLAGS` in `configure.ac` and then use that in the various `Makefile.am` to set the specific `CFLAGS` targets.

e.g.
~~~make
OBLIGATORY_CFLAGS = @OBLIGATORY_CFLAGS@
pd_CFLAGS = $(OBLIGATORY_CFLAGS)
libpd_CFLAGS = $(OBLIGATORY_CFLAGS)

# [...]

pd_CFLAGS += ...
~~~

obviously the name `OBLIGATORY_CFLAGS` is to be (very much) improved.
(and hey, it's probably not even fully *obligatory*; that's the reason i picked the `OBLIGATORY_CFLAGS = @OBLIGATORY_CFLAGS@` scheme above, which would actually allow overriding (as in "replacing it's value") in the `make` invocation)

and yes. this is of course work as it is touching a number of files. but i think it ought to be done regardless (and while we are there: we *should* move things like `-DPD` from `CFLAGS` to `CPPFLAGS`)


> Actually, note that providing CFLAGS during configure or make still appends to allow overriding,

yes, but that's just a hack to get the behaviour we want approximately.
also *appending* is not exactly the same as *overriding* (and i'm guilty as charged in having just used "overriding" in my previous comment, even though in reality it was "appending")

> it doesn't work similarly when using the libtool variable.

to be precise, this is an automake variable, not a libtool variable.

danomatika:
Ok, I was also thinking along these lines but wasn't sure if there was some other approach. It would be good to fix up this behavior, regardless. Also, I'd suggest the following naming `PD_CFLAGS`, `SHARED_CFLAGS`, `STANDARD_CFLAGS`, etc.

umlaeute:
i would avoid `PD_CFLAGS` as it is very close to `pd_CFLAGS` which is reserved.

danomatika:
Ah yes. Maybe something combined ala `PD_GENERAL_CFLAGS` or `PD_SHARED_CFLAGS` or some variation to denote that they are applied to Pd sources specifically and to all of the various build targets? That leaves open adding further derivations if/when needed.

danomatika:
Latest commit refactors `CFLAGS`, `CPPFLAGS`, and `LDFLAGS` handling into `PD_CFLAGS`, PD_CPPFLAGS`, and `PD_LDFLAGS`, respectively. These are now passed down as Automake variables and applied explicitly, ie.  `pd_CFLAGS = @PD_CLFAGS@` or `AM_CPPFLAGS = @PD_CPPFLAGS`. I tried to keep ordering basically the same, so there shouldn't be any flag ordering issues. Setting user flags is now guaranteed to be applied last.

I know, I know, the suggestion was to avoid `PD_*` since there are targeted variables ala `pd_CFLAGS` & `PD_CFLAGS`, however I felt that:

1. they are accessed as Automake variables in a Makefile.am with ampersands, `@PD_CFLAGS@`, which I feel is sufficiently distinct from `pd_CFLAGS`
2. the longer names seemed like overkill when actually working with them
3. existing Automake variables don't follow a `PD_` prefix naming pattern, ex. `EXTERNAL_CFLAGS`

---

One problem I see is that now CFLAGS is empty by default, autotools is setting it again to `-g -02`, at least on my system. Since CFLAGS is always applied last in the ordering, our optimizations are overridden.

---

Another small point: I think the `--enable-libpd-instance` option should be named `--enable-libpd-multi` as 1. the libpd makefile uses `MULTI` ala `make MULTI=true` and 2. the libpd options at the end of configure uses `+multi` in its listing.

umlaeute:
> Since CFLAGS is always applied last in the ordering, our optimizations are overridden.

personally I don't see this as too troublesome. in the end it boils down to using `-O2` (the autotools default) vs `-O3` (the value overridden in `configure.ac`)
if i want a properly optimized builds, i'll have to pass more flags any how, and `CFLAGS` is the proper way to do that.

if you have strong feelings about being able to do `--enable-debug`, you could set `CFLAGS` directly (unless empty) *before* the [first call to `AC_PROG_CC`](https://www.gnu.org/software/autoconf/manual/autoconf-2.71/html_node/Preset-Output-Variables.html#index-CFLAGS-4)  (which is also called by `AC_PROG_LIBTOOL`).
This boils down to moving the ["Build with debug flags?"-check](https://github.com/pure-data/pure-data/blob/bad155d26cceead85e845da4f7c3d41498ea0574/configure.ac#L274-L282) up before the initial [Check for programs, libs, & headers](https://github.com/pure-data/pure-data/blob/bad155d26cceead85e845da4f7c3d41498ea0574/configure.ac#L177-L183).



> These are now passed down as Automake variables and applied explicitly, ie. `pd_CFLAGS = @PD_CFLAGS@`

actually I prefer using a make-variable for this:

~~~mak
PD_CFLAGS=@PD_CFLAGS@
pd_CFLAGS = $(PD_CFLAGS)
~~~

this allows  you to specifically override (as in "replace") these flags
(though it kind of re-introduces the naming-problem, which you've hidden away)



> Another small point: I think the --enable-libpd-instance option should be named --enable-libpd-multi as 1. the libpd makefile uses MULTI ala make MULTI=true and 2. the libpd options at the end of configure uses +multi in its listing.

fine with me.
it might be nice to keep the original option as fallback (or not: as it just adds cruft)

umlaeute:
> This boils down to moving the "Build with debug flags?"-check up before the initial "Check for programs, libs, & headers".

however, keep in mind, that other things like adding `-DNDEBUG` should probably be added regardless of the default flags, so this should be done *later*. (which probably means that there would be two `AS_IF([test x$debug = xyes], ...` invocations)

Spacechild1:
>> Since CFLAGS is always applied last in the ordering, our optimizations are overridden.

> personally I don't see this as too troublesome. in the end it boils down to using -O2 (the autotools default) vs -O3 (the value overridden in configure.ac)

Be careful! GCC doesn't do autovectorization at `-O2`: https://godbolt.org/z/nfqafE97W.

(Clang, on the other hand, does: https://godbolt.org/z/aa1MTzP86)

I think we should specify the release optimization level and not leave it to autotools.

danomatika:
I agree. We should keep our current default as it works fine so far and is explicit.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

millerpuckette:
I'll be amazed if this all works... but it's not my tango so just merging and waiting for CLI to pronounce on it.

millerpuckette:
CLI->CI, oops, showing off my lack of hipness again

danomatika:
> CLI->CI, oops, showing off my lack of hipness again

I use "CLI" or "console" although I believe I have never actually used what would be considered a "real" mainframe terminal/console. :)

PR #1566, opened 2022-01-28
fix soundfiler read reading samples without tablename

danomatika:
This fixes an incorrect cast introduced by 56da2e187d7ad9f72017941c7163a24e9066860a and an bad check for raw status which forced [soundfiler] to *always* read samples even when only header info should be read, ie. without a given tablename.

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -1101,7 +1101,7 @@ done:
 
     /* soundfiler_read ...
 
-       usage: read [flags] filename table ...
+       usage: read [flags] filename [tablename] ...
        flags:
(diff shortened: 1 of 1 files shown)


danomatika:
It might be good to double check the `size_t` and `ssize_t` comparisons and the casting introduced by 56da2e187d7ad9f72017941c7163a24e9066860a for additional issues.

danomatika:
I converted the `ssize_t` casts. I think the rest are ok since those ints casted to `size_t` should never be less than 0.

Lucarda:
Tested on Windows and it works fantastic.

umlaeute (approved):

Issue #1562, opened 2022-01-24
iemgui: troubles with curlies in labels


umlaeute:
while investigating #1559 , i noticed some more problems with the iemgui preferences and curly braces in labels.

it's possible to create a label for an iemgui that contains curly braces.

---

e.g. the following will create a label `foo{bar}`:

~~~
[102 111 111 123 98 97 114 125(
|
[list tosymbol]
|
[label $1(
|
~~~

this seems to work, but when i open the object's preferences (where I still see `foo{bar}` as the label), and click on <kbd>Apply</kbd>, the label is reduced to `foobar` (so the curlies are removed).

---

there's more fun if you only have an opening curly.

e.g. the following will create a label `foo{bar`:

~~~
[102 111 111 123 98 97 114(
|
[list tosymbol]
|
[label $1(
|
~~~

this still seems to work on first glance, only:
- hiding+restoring the window (or switching desktops) freezes the GUI
- opening the preferences dialog, freezes the GUI as well

"freezing the GUI" means that nothing is drawn any more.
clicking on buttons (if you know where they are supposed to be, as they might not get drawn) or other interactive elements still sends data to the core which reacts on them (but printing to the Pd-console doesn't work anymore either).

afaict, Pd-GUI is just waiting for the curlies to close.
this can be confirmed by 
- set `foo{bar` (dangling open curly) as the label of an iemgui
- hide+unhide the window (notice that almost everything is gone)
- set  `foo}bar` (dangling closing curly) as the label of an iemgui
- hide+unhide the window again
- notice that the label is now a full page of tcl-tk commands (all those commands that were lost inbetween)
- --> the label of the

Issue #1559, opened 2022-01-23
iemgui objects trigger Tcl error when sending label msg with double quotes


reduzent:
Sending a `label` message containing  double quotes to a iemgui object like [cnv], [hsl], etc. triggers a Tcl error on the Pd window:

```
(Tcl) UNHANDLED ERROR: extra characters after close-quote
    while executing
".x55b1ffb49040.c itemconfigure 55b1ffb544e0LABEL -text [::pdtk_text::unescape ""s"
    ("uplevel" body line 23)
    invoked from within
"uplevel #0 $docmds"
``` 
After triggering that messages, (many) iemgui objects are not visually updated anymore, though still trigger messages when clicked. [tgl]s are still updated(?)

reduzent:
Tested on on current master (bad155d26c) on Ubuntu 20.04.

reduzent:
Pd 0.51-4 doesn't exhibit the problem.

reduzent:
Also, the problem only occurs through the `label` message, but not when setting the label through properties.

umlaeute:
afaict this was introduced with fcd13008d (@HenriAugusto), where the curly-quotes where replaced with double-quotes.

PR #1570, opened 2022-02-07
bugfix: iemgui labels (display and dialog)

umlaeute:
this PR fixes various problems with iemgui labels:
- https://github.com/pure-data/pure-data/issues/622
- https://github.com/pure-data/pure-data/issues/1559
- https://github.com/pure-data/pure-data/issues/1562
- the various state buttons are now pulldown-menus, which makes it a lot clearer what the current state is and what the other possible states are.

while doing so, i couldn't refrain from cleaning up the iemgui code *a bit* (don't get overly excited):
- use understandable variable names (fix questions like: "What's `rng_sched`?" or: "Why does setting `var_iemgui_lin0_log1` to *1* show *`scale`* in a vu-meter?")
- use tcl-arrays rather for resolving per-window variables, rather than `eval` indirection
- use shared (but private) functions in the core for doing the same-same thing (like: opening up a dialog)

the iemgui dialog code is still ugly, but a bit less cringy (at least the tcl/tk part of it)

things unresolved so far:
- backslash in labels somehow gets lost when saving
- ~~the various state-buttons should be replaced with radiobuttons/pulldown-menus (i can *never* remember whether <kbd>linear</kbd> means  that the fader is currently in *linear* mode, or whether it's going to be linear if I click it. using buttons that change their labels also makes problems with the layout (as the label width changes)~~

---

this PR includes all the changes from #1633 (which closes about a dozen iemgui related issues)

---

this PR has merged the current master which includes [#1696 Refactor TCL-specific string handling into pdgui_vmess()](https://github.com/pure-data/pure-data/pull/1696), and has been updated to use the new `pdgui_vmess()` function for the communication with Pd-GUI.


danomatika:
Yeah, I can never remember what state the damn push buttons are in. Radio, drop down, of checkbox would be better.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
ok, i've implemented the pulldown menu for states.
it still resizes dynamically here, depending on the length of the label of the selection, but that's not really a regression (aka: it did that before as well...)

umlaeute:
i've retracted the PR to "draft" state as currently there's some regressions


- a sr-name with spaces (not that I would advise anybody to use *that*) like `foo bar` will be **displayed** as `foo\ bar` in the dialogs
  - a label with spaces (again `foo bar`) is displayed correctly
- a label with $args (e.g. `$0-foo` will be **saved** in the expanded form (e.g. `1003-foo`)
  - a sr-name with $args (again `$0-foo`) is saved correctly

umlaeute:
the issues with $args in labels and spaces in send/receive names are now resolved, i'm thus marking this PR as ready to merge.

it has also been rebased on #1633 (as both PRs deal with iemguis, and they collided a bit)

umlaeute:
no idea why github shows this as unmergeable: i've just merged the `improvement/merge-iemgui` branch into this branch and resolved all the conflicts manually with 6ade4c783d289ab6fd842aa9cff423a2be24d809 - this also includes a merge with today's `master` in the merge-commit 8812c1b090b0d41ed6b9e7ec9d862acbb2213e81, so there shouldn't be any conflict.


@millerpuckette i would prefer this PR to be merged in favour of #1633 (which is only a subset).

obviously, the merging would need to be done manually.

Issue #1579, opened 2022-02-27
[text sequence]: can't omit array name when providing flags


Spacechild1:
Unlike other objects like `[text define]` or `[array define]`, the flags for `[text sequence]` must come *after* the positional array name argument. If you omit the array name, a subsequent flag (`-g`, `-w` or `-t`) is wrongly interpreted as an array name. As a consequence, you always have to provide a dummy array name if you want to use any of these flags.

PR #1581, opened 2022-02-28
closes https://github.com/pure-data/pure-data/issues/1579

porres:
allows flags to come first in [text sequence] as they should but also allow the wrong order from before for backwards compatibility.

### DIFF
--- a/src/x_text.c
+++ b/src/x_text.c
@@ -1516,13 +1516,49 @@ static void *text_sequence_new(t_symbol *s, int argc, t_atom *argv)
 {
     t_text_sequence *x = (t_text_sequence *)pd_new(text_sequence_class);
     int global = 0;
-    text_client_argparse(&x->x_tc, &argc, &argv, "text sequence");
     x->x_waitsym = 0;
     x->x_waitargc = 0;
     x->x_eaten = 0;
     x->x_loop = 0;
     x->x_lastto = 0;
     x->x_clock = clock_new(x, (t_method)text_sequence_tick);
+    
+    while (argc && argv->a_type == A_SYMBOL &&
+        *argv->a_w.w_symbol->s_name == '-')
+    {
+        if (!strcmp(argv->a_w.w_symbol->s_name, "-w") && argc >= 2)
+        {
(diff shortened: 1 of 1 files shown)


umlaeute:
would it be possible to use a meaningful subject for the commit (and the PR) that doesn't require anybody to go to a website to find out what this is about?

typically, you would just use the "Closes:" stanza in the *body* of the commit message.
as you can see, either putting it in the subject or (more likely) the missing colon failed to link the PR with the issue.

umlaeute:
sidenote: the build failure on MinGW is a caching problem with the github runners for the CI job (so not a problem with the PR).

it's not your fault, but i was hoping for no PRs (or other triggers of the CI) for a full 7 days, as this would have cleaned the cache...
(that was a secret hope and i didn't tell anybody; but once you've force-pushed a commit with a nicer commit-message i would like to ask whether it's possible to not push anythung that triggers a CI build in this repository for 8 days or so)

Spacechild1:
> but i was hoping for no PRs (or other triggers of the CI) for a full 7 days, as this would have cleaned the cache...

Does this also apply to forks? Should I also wait for 7 full days after the last CI job?

umlaeute:
@Spacechild1 hmm, probably yes.

Pd uses a [common pipeline definition](https://github.com/pure-data/pure-data-ci/blob/main/.github/workflows/puredata.yml), so i guess the caching is done in the context of the https://github.com/pure-data/pure-data-ci repository.

i didn't think of that before. bummer... that makes it probably unfeasible to expect an expiration of the broken cache...

umlaeute:
maybe i should just disable the mingw job for now...

umlaeute:
@porres btw, the other [failed job](https://github.com/pure-data/pure-data/runs/5364024496?check_suite_focus=true) is indeed an indication that something should be fixed...

millerpuckette:
The flags currently come _after_ the name of teh text sequence to keep this coherent with the other "text" objects like "text get"... of course, in the "get" case the extra arguments aren't flags as they are with "sequence" but I still think it's the more logical way to do it (and anyway it keeps the source code tidier).

PR #1585, opened 2022-03-20
Free all canvases before closing Pd

myQwil:


src/s_inter.c
```diff
@@ -1551,8 +1551,22 @@ void sys_bail(int n)
...
+            sys_closebang((t_canvas *)y);
+    canvas_closebang(x);
+}
+
 void glob_exit(void *dummy, t_float status)
 {
+    t_glist *g;
+    for (g = pd_getcanvaslist(); g; g = g->gl_next)
+        sys_closebang(g);
+
```
Spacechild1, line 1569: Actually, we also want to *free* the root canvases. `canvas_free` calls `glist_delete` on all children, which in turn calls `canvas_loadbang` if it's a canvas - and so on recursively... This means we only miss a call to `canvas_closebang` for the root canvas itself. `sys_closebang` is not necessary.

Note that `canvas_free` already removes a root canvas from the canvas list. This means the code would be something like:
```
t_glist *c;
while (c = pd_getcanvaslist())
{
    canvas_closebang(c);
    pd_free(&c->g_pd);
}
// ...
```

NB: the order of close bangs is not "correct", but this is a general problem. I've already proposed a fix here: https://github.com/pure-data/pure-data/pull/596. With this PR, the code above would do the right thing.

myQwil: This is occurring inside of `glob_exit`, the function which shuts down Pd. For that reason, it doesn't seem like it's really necessary to free anything since everything will be freed when Pd closes.

Spacechild1: Just becomes the OS will eventually collect the memory, doesn't mean that you shouldn't properly release your resources. It's not only about memory, either. Objects might want to perform arbitrary tasks on shutdown and we should give them a chance to do so.

As a bonuse, the canvas destructor already sends closebangs to its children, so I do not see why we should *not* use it.

Spacechild1:
This is also implemented in https://github.com/pure-data/pure-data/pull/1756/commits/fadec1967599d6043020cdec5c37743dff0d9a19 as part of https://github.com/pure-data/pure-data/pull/1756

Spacechild1:
This is now handled in `pd_term`:

https://github.com/pure-data/pure-data/blob/b00725132b13a29566b9188f01a6bf277c5c7a07/src/m_pd.c#L331-L346

As you can see, we do send closebangs to the canvases, but we do not actually free them (yet).

Issue #1539, opened 2022-01-08
add [file] verb (object) to resolve relative paths to current patch directory


Spacechild1:
Discussions on the mailing list and in https://github.com/pure-data/pure-data/issues/1536#issuecomment-1007849130 showed that we need an easier way to resolve file paths to the current patch.

I'd like to propose `[file resolve]` which takes a file path, and if it's relative, resolves it to the current patch directory. Absolute paths would pass unchanged.

Passing a `[bang(` could simply output the patch directy.

Finally, we might provide an argument to specify the parent level, like `[pdcontrol]`'s `[dir(` method.

`[file resolve]` would complement `[file which]`, so you can implement reading and writing files like in most other Pd objects (array, text, etc.) `[file resolve]` would be used for writing, `[file which]` would be used for reading.

umlaeute:
i don't really know about `[file resolve]`: for me this somehow clashes with `[file which]` (e.g. i expect having to look up the help path to resolve which one is which every time i use them).

`[file patchpath]` doesn't really have this problem (and is pretty clear about what it does, i think)

what do you think?

umlaeute:
on the [mailinglist](https://lists.puredata.info/pipermail/pd-list/2022-01/130663.html) you wrote:

> `[file patchdir]` would be the other way round: it's main purpose is to 
> get the patch directory, but as an extra feature it should probably also 
> allow to resolve a file path (because such a common task shouldn't 
> require several objects).
> 
> Both would essentially provide the same things. IMO, `[file resolve]`
> would be more intuitive because after all that's what people likely want 
> to do when they ask for the current patch directory.


but on recapping the thread, i see that it ended up pretty univocal:

- call it `[file patchpath]` ([pd-list](https://lists.puredata.info/pipermail/pd-list/2022-01/130667.html))
- it takes a `[bang(` to get the current patch directory, or a `[symbol ...(` that is resolved via the current patchdirectory (unless it's an absolute path) ([pd-list](https://lists.puredata.info/pipermail/pd-list/2022-01/130666.html))
- it can be parametrized to set the parent level ([pd-list](https://lists.puredata.info/pipermail/pd-list/2022-01/130668.html))

so sorry for the noise...

umlaeute:
things have been implemented in https://git.iem.at/pd/file

i only need to create a PR (and review the sources :-)

[...]

Issue #1537, opened 2022-01-08
add [file] verb (object) that returns current working directory


Spacechild1:
Sometimes you want to read/write a file relative to the current working directory, but all "high-level" Pd objects - unlike the "low-level" [file] API - resolve relative paths to the canvas environment. Therefore it would be nice to have an object that returns the current working directory, e.g. `[file cwd]`.

Knowing the cwd is handy when writing Pd based command line utilities. If I pass a relative path to such a utility, I want to resolve it to the cwd and not to the Pd patch itself. Currently I have to wrap the whole thing in a shell script/batch script (which I would anyway) and turn all paths into absolute paths before passing them on to Pd. Not a big deal, but still.

Another case is where I want to generate files and put them in the cwd if the user did not provide an output directory.

umlaeute:
as mentioned on the mailinglist, `[file cwd]` could also be used to *set* the current working directory.

Issue #1535, opened 2022-01-07
add [file isabsolute] and/or [file isrelative]


Spacechild1:
It would be very handy to have objects for determining whether a file path is relative or absolute, e.g. `[file isabsolute]` resp. `[file isrelative]`.

Strictly speaking, we would only really need one of the two, but I think it doesn't hurt to have both. After all, we also have `[file isfile]` *and* `[file isdirectory]`.

umlaeute:
hmm, i'd rather only have one, and i'd go for `[file isabsolute]` (simply because it maps to `sys_isabsolutepath`).

alternatives:
- Tcl provides a [`file pathtype`](https://tcl.tk/man/tcl/TclCmd/file.htm#M27) which would output a symbol `absolute` or `relative` (or `volumerelative`, whatever that is)
- an alternative way in Tcl to check whether a path is absolute or not is to use [`file normalize`](https://tcl.tk/man/tcl/TclCmd/file.htm#M25) to normalize the pathname and then check if it changed (not a very good test; but `[file normalize]` might be interesting nevertheless)

and btw: `[file isfile]` is *not* simply the negation of `[file isdirectory]` (at least on systems that have more types of "files" than directories and ordinary files, e.g. devices, named pipes,...)

Spacechild1:
> hmm, i'd rather only have one, and i'd go for [file isabsolute] (simply because it maps to sys_isabsolutepath).

Yes, `[file isabsolute]` is probably enough.

---

> [file normalize] might be interesting nevertheless)

I also thought about such an object.

Actually, "normalize" does not necessarily imply turning a path into an absolute path. In some libraries/platforms, "normalize" is just a string manipulation, removing redundant slashes or resolving double dots between path components. They have a different function for resolving relative paths. In C++, for example, `std::filesystem::path` normalizes and `std::filesystem::absolute` resolves relative paths. Another example: https://stackoverflow.com/a/10823859/6063908.

umlaeute:
things have been implemented in https://git.iem.at/pd/file

i only need to create a PR (and review the sources :-)

iirc, the main reason for stalling this is that @porres was updating the help-patches (including that of `[file]`) and i dodn't want to get into some conflict, so I'd rather wait until the #1517 got merged. 

see also #1539

PR #1590, opened 2022-03-23
[file] new verbs 'isabsolute', 'normalize', 'cwd', 'patchpath'

umlaeute:
this adds 4 new subcommands to `[file]`:

- `[file normalize]` - normalize a path by removing redundant components
- `[file isabsolute]` - check whether a path is absolute 
   - Closes: #1535
- `[file cwd]` - get/set the current working directory
  - Closes: #1537
- `[file patchpath]` - get the patch's path (kind of obsoletes the `[dir(` message to `[pdcontrol]`) 
   - Closes: #1539

the `file-help.pd` reference has also been updated accordingly, which will most likely lead to conflicts with #1517.
i have first synched the reference patch with #1517 (as of 675591aacd684ad3f07dece2f7385a19d8624646), so i guess the resolution would be to use the `file-help.pd` of **this** PR.


umlaeute:
out of mere habit this PR is based on `develop`...

Spacechild1:
> out of mere habit this PR is based on develop...

why not rebase onto master? ;-)

src/x_file.c
```diff
@@ -87,6 +93,30 @@ static int sys_remove(const char *pathname) {
     u8_utf8toucs2(ucs2buf, MAXPDSTRING, pathname, MAXPDSTRING);
     return do_delete_ucs2(ucs2buf);
 }
+static char* sys_getcwd(char *buf, size_t size) {
+    int err=0;
+    size_t ucs2size = size;
+    uint16_t*res = 0, *ucs2buf = getbytes(ucs2size*sizeof(*ucs2buf));
```
Spacechild1, line 99: why not simply use `uint16_t src[MAX_PATH]`, like in the functions above?

umlaeute: good point.
the reason was file's `sys_*` closely wrap the POSIX functions (which has a `size`) parameter, and it appeared to me that this dynamic size ought to be passed thru as far as possible.

anyhow, i've now switched to a stack-based allocation.

umlaeute:

note that this PR is actually driven by some requests by others (notably @Spacechild1; but also a few others chimed in)

Spacechild1:
@millerpuckette It would be really great if this could make it into Pd 0.54!

PR #1591, opened 2022-03-23
set dirty flag from font dialog only

### LINKED ISSUE #1511: 'font' message to canvas makes patch dirty
The following patch sets the dirty flat of the patch when the message is clicked:

```
[namecanvas $0.font-test]

[font 16 100 100(
|
[s $0.font-test]
```

Up to 0.51-4, this message didn't render the patch dirty. Since 0.52, GOP abstractions using this message for dynamically changing the font size of datastructure labels (for instance) trigger a pop-up when the main patch is closed: "Do you want to save changes in ..." . 

Since other dynamic modifications of the patch (creating objects, clearing a subpatch) do not set the dirty flag, I consider this a regression.

sebshader:
closes #1511

send the "dirty" message to the canvas from tcl in "apply" procs, instead of checking in canvas_font


sebshader:
sorry for all of the force-pushes again..

PR #1593, opened 2022-03-25
lrshift-set

czhenry:
I have made changes to extra/lrshift~.c to add a set method and a float inlet.

To make it work, I removed the leftshift_perform and rightshift_perform functions and used just one lrshift_perform function in it.

### DIFF
--- a/extra/lrshift~/lrshift~.c
+++ b/extra/lrshift~/lrshift~.c
@@ -7,53 +7,53 @@ static t_class *lrshift_tilde_class;
 typedef struct _lrshift_tilde
 {
     t_object x_obj;
-    int x_n;
-    t_float x_f;
+    t_float  x_n;
+    t_float  x_f;
 } t_lrshift_tilde;
 
-static t_int *leftshift_perform(t_int *w)
+static t_int *lrshift_perform(t_int *w)
 {
-    t_sample *in = (t_sample *)(w[1]);
-    t_sample *out= (t_sample *)(w[2]);
-    int n = (int)(w[3]);
-    int shift = (int)(w[4]);
-    in += shift;
-    n -= shift;
-    while (n--)
-        *out++ = *in++;
-    while (shift--)
-        *out++ = 0;
-    return (w+5);
-}
-
-static t_int *rightshift_perform(t_int *w)
-{
-    t_sample *in = (t_sample *)(w[1]);
-    t_sample *out= (t_sample *)(w[2]);
-    int n = (int)(w[3]);
-    int shift = (int)(w[4]);
-    n -= shift;
-    in -= shift;
-    while (n--)
-        *--out = *--in;
-    while (shift--)
-        *--out = 0;
-    return (w+5);
+    t_lrshift_tilde *x = (t_lrshift_tilde *) w[1];
+    t_sample *in = (t_sample *)(w[2]);
+    t_sample *out= (t_sample *)(w[3]);
+    int n = (int)(w[4]);
+    
+    //test for abs(shift) >= n and limit shift between -(n-1) and n
+    //when shift equals n, only the "*out++ = 0.0f" loop runs
+    int shift = (int)x->x_n;
+    shift = (((shift<0)?-1*shift:shift) < n)?shift:n;
+    
+    //The loops are written this way to avoid "read after write" errors
+    //when *in and *out point to the same array
+    t_sample *ptr, *in_end = in + n-1;
+    if (shift >= 0)  //shift left, shift is a postive int or zero 
+    {
+        for( ptr = in + shift; ptr<=in_end;)
+            *out++ = *ptr++;
+        for( int i = 0; i<shift; i++)
+            *out++ = 0.0f;
+    } else //shift right, shift is a negative int
+    {
+        t_sample *out_end = out + n-1;
+        for( ptr = in_end + shift; ptr>=in;)
+            *out_end-- = *ptr--;
+        for( int i = 0; i>shift; i--)
+            *out_end-- = 0.0f;
+    }
+    
+    return(w+5);
 }
 
 static void lrshift_tilde_dsp(t_lrshift_tilde *x, t_signal **sp)
 {
-    int n = sp[0]->s_n;
-    int shift = x->x_n;
-    if (shift > n)
-        shift = n;
-    if (shift < -n)
-        shift = -n;
-    if (shift < 0)
-        dsp_add(rightshift_perform, 4,
-            sp[0]->s_vec + n, sp[1]->s_vec + n, (t_int)n, (t_int)(-shift));
-    else dsp_add(leftshift_perform, 4,
-            sp[0]->s_vec, sp[1]->s_vec, (t_int)n, (t_int)shift);
+    t_int n = sp[0]->s_n;    
+    dsp_add(lrshift_perform, 4, x,
+        sp[0]->s_vec, sp[1]->s_vec, n);
+}
+
+static void lrshift_tilde_set(t_lrshift_tilde *x, t_floatarg f)
+{
+    x->x_n = f;
 }
 
 static void *lrshift_tilde_new(t_floatarg f)
@@ -61,6 +61,7 @@ static void *lrshift_tilde_new(t_floatarg f)
     t_lrshift_tilde *x = (t_lrshift_tilde *)pd_new(lrshift_tilde_class);
     x->x_n = f;
     x->x_f = 0;
+    floatinlet_new(&x->x_obj, &x->x_n); 
     outlet_new(&x->x_obj, gensym("signal"));
     return (x);
 }
@@ -73,4 +74,6 @@ void lrshift_tilde_setup(void)
     CLASS_MAINSIGNALIN(lrshift_tilde_class, t_lrshift_tilde, x_f);
     class_addmethod(lrshift_tilde_class, (t_method)lrshift_tilde_dsp,
         gensym("dsp"), 0);
+    class_addmethod(lrshift_tilde_class, (t_method)lrshift_tilde_set,
+        gensym("set"), A_DEFFLOAT, 0);
 }


extra/lrshift~/lrshift~.c
```diff
@@ -7,60 +7,61 @@ static t_class *lrshift_tilde_class;
...
-    return (w+5);
+    t_lrshift_tilde *x = (t_lrshift_tilde *) w[1];
+    t_sample *in = (t_sample *)(w[2]);
+    t_sample *out= (t_sample *)(w[3]);
+    int n = (int)(w[4]);
+    
+    //test for abs(shift) >= n and limit shift between -(n-1) and n
+    //when shift equals n, only the "*out++ = 0.0f" loop runs
+    int shift = (int)x->x_n;
+    shift = (((shift<0)?-1*shift:shift) < n)?shift:n;
```
Spacechild1, line 24: Pd uses spaces before and after operators.

extra/lrshift~/lrshift~.c
```diff
@@ -7,60 +7,61 @@ static t_class *lrshift_tilde_class;
...
+    
+    //The loops are written this way to avoid "read after write" errors
+    //when *in and *out point to the same array
+    t_sample *ptr, *in_end = in + n-1;
+    if (shift >= 0)  //shift left, shift is a postive int or zero 
+    {
+        for( ptr = in + shift; ptr<=in_end;)
```
Spacechild1, line 31: we always put a space after `for` and no space after open parenthesis

extra/lrshift~/lrshift~.c
```diff
@@ -7,60 +7,61 @@ static t_class *lrshift_tilde_class;
...
+            *out++ = *ptr++;
+        for( int i = 0; i<shift; i++)
+            *out++ = 0.0f;
+    } else //shift right, shift is a negative int
```
Spacechild1, line 35: `else` goes to its own line

extra/lrshift~/lrshift~.c
```diff
@@ -7,60 +7,61 @@ static t_class *lrshift_tilde_class;
...
+    t_sample *in = (t_sample *)(w[2]);
+    t_sample *out= (t_sample *)(w[3]);
+    int n = (int)(w[4]);
+    
+    //test for abs(shift) >= n and limit shift between -(n-1) and n
+    //when shift equals n, only the "*out++ = 0.0f" loop runs
+    int shift = (int)x->x_n;
+    shift = (((shift<0)?-1*shift:shift) < n)?shift:n;
+    
+    //The loops are written this way to avoid "read after write" errors
```
Spacechild1, line 26: we use C-style comments

extra/lrshift~/lrshift~.c
```diff
@@ -7,60 +7,61 @@ static t_class *lrshift_tilde_class;
...
+    //when *in and *out point to the same array
+    t_sample *ptr, *in_end = in + n-1;
```
Spacechild1, line 28: Pd is a C89 code base, which means that *all* local variables have to be declared at the top of the block.

extra/lrshift~/lrshift~.c
Spacechild1: This is way too complicated. Later you have to branch over the value of `shift` anyway. So why not simply do it like this:
```
if (shift > 0) /* left shift */
{
    /* validate shift value, set pointers and perform */
}
if (shift < 0) /* right shift */
{
    /* validate shift value, set pointers and perform */
}
else /* no shift */
{
    /* just copy */
}
```

Spacechild1:
That's a very welcome addition!

I pointed to some coding style issues in my comments.

---

Generally, Pd objects never use a float inlet and a method that do the same thing. You have to decide for one of the two. Personally, I would prefer a float inlet because it's easier to use.

---

Actually, we might only create the inlet if the user does *not* provide a creation argument. This would allow us to optimize for the case where the shift is constant, so that the branching and validation would already happen in the "dsp" method - as it is currently done - and not in the perform routine. We just need to add two helper functions:
```
static void leftshift(const t_sample *in, t_sample *out, int n, int shift)
{
    /* perform the left shift */
}

static void rightshift(const t_sample *in, t_sample *out, int n, int shift)
{
    /* perform the right shift */
}

static t_int *rightshift_perform(t_int *w)
{
    /* dispatch to leftshift() */
}

static t_int *rightshift_perform(t_int *w)
{
    /* dispatch to rightshift() */
}

static t_int *lrshift_perform(t_int *w)
{
    /* validate shift value and dispatch to either leftshift(), rightshift(), or just copy */
}

static void lrshift_tilde_dsp(t_lrshift_tilde *x, t_signal **sp)
{
    /* If we have an inlet, add lrshift_perform().
     * Otherwise, check and validate x_n and use either leftshift_perform(), 
     * rightshift_perform(), or dsp_add_copy().
     * /
}
```

extra/lrshift~/lrshift~.c
czhenry: Thanks very much for your feedback and advice.  There's a lot of room to simplify lrshift~.c and I think it should be done.

I wrote it the way I did because it was symmetric and short.  Some of the conditional logic is contained in the fact that a while/for loop may run 0 times.  
If shift=0, the result can be produced by either block.  Only the 1st loop in each block runs.
If abs(shift) > n-1, the input is never read.  Only the 2nd loop in each block runs.  So, these are trivial cases.

I'm slightly intimidated by the idea of optimizing lrshift~.c at the moment.  I'm reading d_arithmetic.c for reference while I'm working on it, and many routines there are written with "perform" and "perf8" perform routines.  I could emulate the style, but I am not a sophisticated programmer.

I'm in favor of having "code that speaks for itself." I have these places I left the comments in for the 1st draft, I hope it helps you read it quickly.  I can remove those comments as they become un-needed.  Maybe we can settle on a version of it that's clear enough it doesn't need them.

lrshift~ really ought to be a dirt simple and nearly optimal piece of code.  The specification is really, really simple: turn a floatarg into the integer number of samples to shift while copying one array to another.  This is a really basic array operation.  

The motivating case is in abstractions that have variable block sizes, lrshift~ should be able to change its parameters along with changes to the blocksize using messages.  I'm in favor of being able to set the number as default and also update it from float messages later.  

I've tested this code (as shift~.c external) and produced identical output as lrshift~.c does.  I'm learning the process of making a pull request for the first time--do you want to see a test patch or do any other tests with it next?

I was considering the number of conditional statements being evaluated on each block as well.  I can save one conditional per tick by moving the validation inside the if-statement blocks and it's less complicated.  I've made the change there.  I also noticed I could write the second for loop as a while loop with the same count when shift is a negative int.  

I think I addressed your style notes.  Thanks for helping me with those.
Please take another look at my perform routine again, and see if you think it should have more changes.

extra/lrshift~/lrshift~.c
czhenry: Thanks.  Updated.

extra/lrshift~/lrshift~.c
czhenry: Updated.

extra/lrshift~/lrshift~.c
czhenry: Updated.

extra/lrshift~/lrshift~.c
czhenry: Updated.

extra/lrshift~/lrshift~.c
Spacechild1: The code is now much more readable. Thanks!

extra/lrshift~/lrshift~.c
```diff
@@ -7,68 +7,67 @@ static t_class *lrshift_tilde_class;
...
+        for (ptr = in + shift; ptr <= in_end;)
+            *out++ = *ptr++;
+        while (shift--)
+            *out++ = 0.0f;
+    }
+    else /* shift right, shift is a negative int */
+    {
+        if (shift < -n)
+            shift = -n;
+        t_sample *out_end = out + n-1;
```
Spacechild1, line 40: variables must be declared at the top of a block.

extra/lrshift~/lrshift~.c
```diff
@@ -7,68 +7,67 @@ static t_class *lrshift_tilde_class;
...
-    return (w+5);
-}
+    t_lrshift_tilde *x = (t_lrshift_tilde *) w[1];
+    t_sample *in = (t_sample *)(w[2]);
+    t_sample *out= (t_sample *)(w[3]);
+    int n = (int)(w[4]);
+    int shift = (int)x->x_n;
+    t_sample *ptr,
+        *out_end = out + n - 1,
+        *in_end = in + n - 1;
```
Spacechild1, line 23: `out_end` is shadowed in the else-block below. `in_end` is only really necessary for the if-block.

In this case, there is no need to calculate any pointers outside the if-else-blocks at all:
```
    if (shift >= 0)  /* shift left, shift is a postive int or zero */
    {
        if (shift > n)
            shift = n;
        n -= shift;
        in += shift;
        while (n--)
            *out++ = *in++;
        while (shift--)
            *out++ = 0.0f;
    }
    else /* shift right, shift is a negative int */
    {
        if (shift < -n)
            shift = -n;
        out += n - 1;
        in += n + shift - 1;
        n += shift;
        while (n--)
            *out-- = *in--;
        while (shift++)
            *out-- = 0.0f;
    }
```
Note: I didn't test this. Please make sure I haven't made a typo :-)

---

As a rule of thumb: first write the most natural soluation for each branch and only then see if there actually is common code that can be hoisted outside the loop.

Also, by convention "end" pointers point to one past the last element, so `out_end` should rather be `out + n`. Consequently, you would use `ptr < out_end

extra/lrshift~/lrshift~.c
```diff
@@ -7,68 +7,67 @@ static t_class *lrshift_tilde_class;
...
-    if (shift < -n)
-        shift = -n;
-    if (shift < 0)
-        dsp_add(rightshift_perform, 4,
-            sp[0]->s_vec + n, sp[1]->s_vec + n, (t_int)n, (t_int)(-shift));
-    else dsp_add(leftshift_perform, 4,
-            sp[0]->s_vec, sp[1]->s_vec, (t_int)n, (t_int)shift);
+    t_int n = sp[0]->s_n;
+    dsp_add(lrshift_perform, 4, x,
+        sp[0]->s_vec, sp[1]->s_vec, n);
```
Spacechild1, line 54: By convention, we would rather write
```
    dsp_add(lrshift_perform, 4, x,
        sp[0]->s_vec, sp[1]->s_vec, (t_int)sp[0]->s_n);
```
The advantage is that it makes the (very important) cast to `t_int` more visible.

Spacechild1:
@millerpuckette What do you think about the optimization suggestion in: https://github.com/pure-data/pure-data/pull/1593#issuecomment-1078870734

TL;DR: only sprout the inlet if no argument is provided; otherwise do the branching and input validation in the "dsp" method.

This basically saves 2 branches per block, which I think is not really significant - except for small blocksizes...

Issue #1598, opened 2022-03-28
fails to build on OpenBSD


umlaeute:
- build of utilities fails with unknown `FD_SET` (and friends)

~~~console
[...]

Issue #1597, opened 2022-03-28
[text] sorting crashes on FreeBSD


umlaeute:
*sort*ing a list stored in 

running `[sort(` on a filled `[text define]` on FreeBSD, segfaults:

~~~gdb
#0  0x00007fffffffdca8 in ?? ()
#1  0x0000000800a61b2b in qsort_r () from /lib/libc.so.7
#2  0x00000000004cb299 in text_define_sort (x=0x8010eb7c0, s=<optimized out>, argc=<optimized out>, argv=<optimized out>) at x_text.c:557
#3  0x000000000049beaa in outlet_anything (x=<optimized out>, s=s@entry=0x8010549e0, argc=argc@entry=0, argv=argv@entry=0x7fffffffddf0) at m_obj.c:613
~~~

afaict, the problem is that `qsort_r` on FreeBSD (as on Darwin) is
`void qsort_r(void *base, size_t nmemb, size_t size, void *thunk, t_comparefun*compar);`
whereas on Linux it is:
`void qsort_r(void *base, size_t nmemb, size_t size, t_comparefun*compar,  void *thunk);`

Issue #1596, opened 2022-03-28
fails to build on FreeBSD


umlaeute:
building on FreeBSD fails with
- underlinking (does not link against `-lm`, as the FreeBSD case is not handled in configure/Makefile)
- missing `glob_watchdog`
- wrong endianness includes (`<endian.h>` instead of `<machine/endian.h>`), at least in s_audio_alsa.c

umlaeute:
I'm not sure how much Pd is used on FreeBSD (probably not very much; there *is* a [Pd-0.47-1](https://www.freebsd.org/cgi/ports.cgi?query=Pd&stype=all&sektion=all#pd-0.47.1_1) package available as part of the FreeBSD "ports", though that doesn't work with FreeBSD13.

However, I'd like to run my unit-tests on a diverse set of operating systems to increase coverage (and robustness), and thus it would be great if building Pd on the various BSD-variants would be possible.

umlaeute:
while we are there: on OpenBSD, Pd will use the `.pd_openbsd` extension for externals, but on FreeBSD it uses `.pd_linux` (most likely becaues of Debian/kFreeBSD); i think it would make sense to (also (?)) check for `.pd_freebsd`.

PR #1599, opened 2022-03-29
BSD build fixes

umlaeute:
this PR collects fixes to allow building Pd on FreeBSD, OpenBSD and NetBSD.

I don't really believe that there are many Pd users on these platforms, but the exercise lead to a bit of code-cleanup and generalization.

e.g. the new code tests whether `qsort_r` is available and if not falls back to our own `STUPID_SORT` implementation,
and we no longer need to discuss that for each platform.

fixes integrate with the autotools system (but i took care to keep it working with the makefile-based system)

Closes: #1596 
Closes: #1598 
Closes: #1597 

Also fixes build and runtime problems on NetBSD (but i hadn't opened a ticket for that)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,11 +37,11 @@ lib*.a
 *.la
 
 # externals
-*.pd_linux
-*.pd_darwin
+*.pd_*
 *.d_fat
 *.d_fat_o
 *.dll
+*.so
 
 # windows build artifacts
 src/*.com
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -141,7 +141,7 @@ double check which architectures Pd was built with, use the "file" command:
     ...
     Pd-0.47.1.app/Contents/Resources/bin/pd: Mach-O 64-bit executable x86_64
 
-### Linux & BSD
+### Linux
 
 Platform requirements:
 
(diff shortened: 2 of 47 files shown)


umlaeute:
i see your point (of getting rid of system-specific defines in the code).

however, i don't think that adding `/usr/include/machine/` to the search paths is necessarily the right way to fix this, as there appear to be a couple of files both in `/usr/include/machine/` and `/usr/include/`.

how about this:
- at configure time, check whether `machine/endian.h` is available and if so, define `HAVE_MACHINE_ENDIAN_H` (this is the standard macro emitted by autotools if it detects a header-file)
- in the relevant source-files, include `machine/endian.h` if the macro is set (and true), otherwise include `endian.h`

rinse and repeat for other system-dependent includes.


this still adds the ugliness of `#ifdef`ed includes, but at least they are no longer system-dependent...

Spacechild1:
Going a bit further: we should think about adding a small header for (internal) system specific definitions that are repeated over and over in the code base, like the endian stuff, `alloca`, `_snprintf` for MSVC, etc.

This utility header could then be simply included by `m_pd.h` - but only if `PD_INTERNAL` is defined.

umlaeute:
> This utility header could then be simply included by m_pd.h - but only if PD_INTERNAL is defined.

urgh, no.

`m_pd.h` is *the* public header of Pd. we do not want to add anything specifically to building Pd itself.
(the only exception is the conditional definition of `EXTERN`, so we can use the public header for building Pd as well).

so: i welcome the idea of a private header file `m_imp_utils.h`, which would then be included by whatever source files requires these utilities (and if need be by `m_imp.h` itself)

Spacechild1:
> m_pd.h is the public header of Pd. we do not want to add anything specifically to building Pd itself.

You're right. I guess I was just being lazy :-)

> so: i welcome the idea of a private header file m_imp_utils.h, which would then be included by whatever source files requires these utilities (and if need be by m_imp.h itself)

+1

umlaeute:
> Hooray for `m_private_utils.h`!

unfortunately it breaks the macOS builds (`_XOPEN_SOURCE` doesn't play nicely with `BYTE_ORDER`...).
for now this is only an issue with `x_file.c` (which requires `_XOPEN_SOURCE` but not `BYTE_ORDER`, so I'm going to demote the error to a warning...)

umlaeute:
closing the PR and rebooting it

Issue #902, opened 2020-03-06
[Feature request] new file extension for double-precision externals.


Lucarda:
There is some discussion on #900 about how to implement FAT externals (single binaries that can work on single and double precision). While this can be cool for advanced developers I think there should also be a special file extension for double-precision externals so that existing and unmaintained libraries can be compiled for Pd-double without any or very few code modifications.

There are some suggestions like `<name>.float64.l_amd64`

Since Pd-double is something many of us are interested and is almost ready, there should be an easy way for us to compile existing code and use it. It could become hard if we have to insert something into many .c files.

Also it can be a nice option for active developers to simply supply non-FAT binaries.

umlaeute:
> Also it can be a nice option for active developers to simply supply non-FAT binaries.

esp. since most likely there are externals that cannot be compiled as PHAT-binaries (shared internal functions that deal with `t_float`, where it would just be too burdensome to mangle all their names)

umlaeute:
here's another idea:

#### use the deken-specifier together with the native extension for dynamic-link libraries.

To keep things simple (and reduce the noise with `-verbose`), i would suggest to only allow *lower case* specifiers, and no arch variants (e.g. `i386` for all `x86_32` variants, and `amd64` or all `x86_64` variants)

e.g. `zexy.linux-amd64-32.so`, `zexy.darwin-arm64-32.so` (or `zexy.darwin-arm64-32.dylib`?), `zexy.windows-i386-32.dll`

##### pros
- using the system extension (as the last part of the full extension)
 - it gives quite predictable *final* extensions, which makes filtering (by filename) a lot easier
 - these extensions are *well known* (so we don't have to invent a new one for each and every platform)
 - system tools often use the file-extension to recognize 
- using the deken specifier (as part of the  full extension)
 - it full covers what we need right no (being able to distinguish by OS/binary-format, CPU instruction set and float-size
 - people can relate the files within a deken-package with the deken-package-filename
 - the problem space is the same for deken package files and externals. if we ever need to add a new parameter, the deken specifier and the externals are likely to be affected in a similar way, so we need to solve the problem only once.
- it gets rid of the super-cryptic `.<first-letter-of-the-os>_<cpu-arch>` specifier (`.o_ia64` anybody?)

##### cons
- it shares the same (final) extensions as any support libraries.
  e.g.:
[...]

PR #1605, opened 2022-03-29
filename extensions for double-precision externals (and removing system-specific quirks)

umlaeute:
This PR attempts to get rid of a number of system-specific quirks in the source code in favor of generic checks

### file extension

this is a draft implementation for filename extensions that includes the [deken-specifier](https://github.com/pure-data/deken/blob/main/developer/README.md#filename-format), in order to allow single-precision externals to coexist with double-precision externals.

The idea is to insert a string `.<os>-<arch>-<floatsize>` right before the (common) file-extension.
the "common file-extension" so far is the system's default extension for dynamic libraries (`.dll` on Windows; `.so` on Un*xes)

So an actual binary might look like `foobar.darwin-ppc-32.so`.
(There is still some discussion going on in #902 about the actual format.)

This scheme allows for generic and extendible names that adhere to standards, and can be predicted.

The actual code uses hardcoded values for common systems (Linux/macOS/Windows) and architectures (amd64, i386, arm, arm64,...) but has a mechanism to add new extensions as build options (e.g. for *BSD)

### header inclusion
traditionally, Pd's code is full of `#ifdef __OS__`  to cater for system-specific includes (like "include *this* header on macOS and *that* header on Windows").

Instead this PR attempts to detect the presence of such system-specifics (like headers) *before* compiling (in the `./configure` stage), and then use those headers that are required (like "include *this* header if it exists, otherwise include *that* header if it exists).

The autotools build-system (`./configure`) now does some elaborate checks on which headers are available and usable.
the simple `src/makefile.*` just adds the relevant defines to the system-specific makefiles.

### qsort_r

`qsort_r` is implemented differently on BSDs (including macOS) and GNU systems (Linux), and not available at all on other systems.
instead of enumerating the different beaviours of various OSs, we now do an active check whether `qsort_r` is available at all, and then check it's function prototype to see which variant is used.
based on this info, we can pick the correct sorting implementation reliably.

for the simple `src/makefile.*`, we just replicate the original OS-logic by adding the relevant defines to the system-specific makefiles.

### alloca

about 6 different source-code files use `alloca()` to do "stack-based allocation unless too much memory is required".
traditionally all 6 files would duplicate the same code (include the relevant headers, create some macros that handle the gory details), but many of them slightly different.
this PR unifies the code by providing a single re-usable implementation.

### audio backends

rather than assuming that certain audio-backends are only available on given platforms, we now check whether this is actually true (e.g. OSS exists on more than just Linux, but requires different headers and libraries to be linked; in the PortAudio case, the un*x specific code is no longer restricted to just Linux and macOS.

#### private header utilities
to keep code repetition at a minimum, there's a new header `m_private_utils.h` (only to be used by the Pd-buildsystem itself, and not by externals (including those in `extra/`)), that collects such gory details like
- including the correct headers to detect the endianness
- unify `ALLOCA` handling (include the correct headers, and provide a single macro to be used everywhere)
- redefines for POSIX functions for MSVC

#### BSD

all this was mainly triggered to support building on various BSD-variants.
as such, this PR also includes some simplifications in the BSD handling (no longer special casing the OpenBSD case) and adds build-instruction for this platform.

this PR also closes a few BSD-related bugs:

Closes: https://github.com/pure-data/pure-data/issues/1596
Closes: https://github.com/pure-data/pure-data/issues/1598
Closes: https://github.com/pure-data/pure-data/issues/1597

#### configure options

there's also some new configure options to override file-extensions for externals and the deken-specifier.
this allows us to drop some more OS-specific defines for exotic OSs.
e.g. there's no need to add extra code to allow `.pd_freebsd` on FreeBSD, if this is practically unused. instead, the few people that need it can configure Pd with `--with-external-extension=pd_freebsd`.
Alternatively (or additionally), they could configure Pd with `--with-deken-os=FreeBSD` which will also use `.freebsd-amd64-32.so` as a valid externals extension.

---
This is a reboot of https://github.com/pure-data/pure-data/pull/1599 and includes #1614, so now BSD-support is reached by making the code more general rather than adding additional special-casings.


Spacechild1:
> any idea what's wrong with the [MSVC build](https://ci.appveyor.com/project/umlaeute/pure-data/builds/43056367)?

`s_inter.c` is missing
```
#define snprintf _snprintf
#endif
```

The error message for the 32-bit build
`s_inter.obj : error LNK2019: unresolved external symbol _snprintf referenced in function _sys_deken_specifier`
is a bit confusing because it seems like it's looking for `_snprintf` already. However, the actual symbol name would be `__snprintf` (2 underscores!) because 32-bit PE files use leading underscores for all external symbols.

The error message for the 64-bit build is more obvious:
`s_inter.obj : error LNK2019: unresolved external symbol snprintf referenced in function sys_deken_specifier`

umlaeute:
:facepalm: i checked and double-checked `s_loader.s` instead of `s_inter.c`...

Spacechild1:
Fun fact: `_snprintf` is actually not safe because it doesn't guarantee null-termination.

Visual Studio 2015 finally has a proper `snprintf` - only 16 years after the release of the C99 standard...

But that's not the topic of this PR :-) For now let's continue pretending that `_snprintf` is safe.

umlaeute:

(this PR is really trying to simplify things; but then i'm aware that i tend to overengineer things a bit)

umlaeute:
ping @millerpuckette

PR #1607, opened 2022-04-04
Implement sig_tilde_perf8

myQwil:
Also, remove duplicate definitions of `sig_tilde_perform` and `sig_tilde_perf8`, and instead make these functions non-static so that they can be referred to with extern declarations.

### DIFF
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -18,7 +18,7 @@ typedef struct _sig
     t_float x_f;
 } t_sig;
 
-static t_int *sig_tilde_perform(t_int *w)
+t_int *sig_tilde_perform(t_int *w)
 {
     t_float f = *(t_float *)(w[1]);
     t_sample *out = (t_sample *)(w[2]);
@@ -28,7 +28,7 @@ static t_int *sig_tilde_perform(t_int *w)
     return (w+4);
 }
 
-static t_int *sig_tilde_perf8(t_int *w)
+t_int *sig_tilde_perf8(t_int *w)
 {
     t_float f = *(t_float *)(w[1]);
     t_sample *out = (t_sample *)(w[2]);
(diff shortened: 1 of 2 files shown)


umlaeute:
- `sig_tilde_perf8` mainly improves performance, so that's a good thing. otoh, there are two identical implementations of `sig_tilde_perf8` in https://github.com/pure-data/pure-data/blob/master/src/d_ctl.c#L31-L49 resp https://github.com/pure-data/pure-data/blob/18c9695ba4de543c51c2c013b45b99b8f1c35a19/src/d_ugen.c#L1283-L1301 - maybe we should get rid of one (and make the other non-static)
- `pdsymbol_list` looks like an oversight (the code has been lingering there since the beginning of the git history). however, acticating it *does* change the behaviour of patches (as `[list foo bar(` starts outputting `[symbol foo(` instead of `[symbol list(`

i'm not opposed to changing the behaviour of `[symbol]`, but @millerpuckette should be aware of this in case he wants to include the PR...

Spacechild1:
> pdsymbol_list looks like an oversight (the code has been lingering there since the beginning of the git history). however, acticating it does change the behaviour of patches (as [list foo bar( starts outputting [symbol foo( instead of [symbol list(

Another thing: currently, `[symbol]` "converts" floats to "symbol float" - which I think is not very useful. `pdsymbol_list` wouldn't really change that. Actually, there is already a PR by @porres takes care of both issues (and adds documentation): https://github.com/pure-data/pure-data/pull/987.

I would suggest to remove ab62849 from this PR and focus on the `sig_tilde_perf8` issue instead.

Spacechild1:
> maybe we should get rid of one (and make the other non-static)

+^1

millerpuckette:
can this be closed?  commit f316d2bac centralized scalar-to-vector copying.

myQwil:
This can be closed

PR #1614, opened 2022-04-06
move system-specific logic from compile-time to configure-time

umlaeute:
This PR attempts to get rid of a number of system-specific quirks in the source code in favor of generic

### header inclusion
traditionally, Pd's code is full of `#ifdef __OS__`  to cater for system-specific includes (like "include *this* header on macOS and *that* header on Windows").

Instead this PR attempts to detect the presence of such system-specifics (like headers) *before* compiling (in the `./configure` stage), and then use those headers that are required (like "include *this* header if it exists, otherwise include *that* header if it exists).

The autotools build-system (`./configure`) now does some elaborate checks on which headers are available and usable.
the simple `src/makefile.*` just adds the relevant defines to the system-specific makefiles.

### qsort_r

`qsort_r` is implemented differently on BSDs (including macOS) and GNU systems (Linux), and not available at all on other systems.
instead of enumerating the different beaviours of various OSs, we now do an active check whether `qsort_r` is available at all, and then check it's function prototype to see which variant is used.
based on this info, we can pick the correct sorting implementation reliably.

for the simple `src/makefile.*`, we just replicate the original OS-logic by adding the relevant defines to the system-specific makefiles.

### alloca

about 6 different source-code files use `alloca()` to do "stack-based allocation unless too much memory is required".
traditionally all 6 files would duplicate the same code (include the relevant headers, create some macros that handle the gory details), but many of them slightly different.
this PR unifies the code by providing a single re-usable implementation.

### audio backends

rather than assuming that certain audio-backends are only available on given platforms, we now check whether this is actually true (e.g. OSS exists on more than just Linux, but requires different headers and libraries to be linked; in the PortAudio case, the un*x specific code is no longer restricted to just Linux and macOS.

#### private header utilities
to keep code repetition at a minimum, there's a new header `m_private_utils.h` (only to be used by the Pd-buildsystem itself, and not by externals (including those in `extra/`)), that collects such gory details like
- including the correct headers to detect the endianness
- unify `ALLOCA` handling (include the correct headers, and provide a single macro to be used everywhere)
- redefines for POSIX functions for MSVC

#### BSD

all this was mainly triggered to support building on various BSD-variants.
as such, this PR also includes some simplifications in the BSD handling (no longer special casing the OpenBSD case) and adds build-instruction for this platform.

this PR also closes a few BSD-related bugs:

Closes: https://github.com/pure-data/pure-data/issues/1596
Closes: https://github.com/pure-data/pure-data/issues/1598
Closes: https://github.com/pure-data/pure-data/issues/1597

#### configure options

there's also some new configure options to override file-extensions for externals and the deken-specifier.
this allows us to drop some more OS-specific defines for exotic OSs.
e.g. there's no need to add extra code to allow `.pd_freebsd` on FreeBSD, if this is practically unused. instead, the few people that need it can configure Pd with `--with-external-extension=pd_freebsd`

---
This is a reboot of https://github.com/pure-data/pure-data/pull/1599, but now BSD-support is reached by making the code more general rather than adding additional special-casings.

Note that this PR is fully included in #1605 (so I would prefer if the other PR was included instead of this one)


umlaeute:
this has been merged into #1605 
i think it would be better to pick the #1605 PR instead of this, so I'm marking this "Draft" to prevent accidental merging.

Issue #1620, opened 2022-04-10
request, set data structures' colors in hexadecimal


porres:
This request is for drawing objects:
- drawpolygon/drawcurve/filledpolygon/filledcurve
- plot
- drawtext/drawnumber/drawsymbol (which are aliases of the same class).

The simplest way would be to just allow a symbol argument as well and expect the symbol argument to be in a hexadecimal format in the same way as accepted by 'iemguis'.

A '-h' could be given, maybe, but I don't see why we'd need it (yet).

A field name that would set the color in hexadecimal format would require the field type to be a symbol.

porres:
> A '-h' could be given, maybe, but I don't see why we'd need it (yet).

Ok, so, if you give it a symbol argument for the color, it understands that it is a field name for the color. So, for the sake of being "elegant", a flag may be welcome. Although I think it is unexpected that someone created a field name called "#44AAFF" for instance.

porres:
by the way, I was able to set a symbol argument for the color in hexadecimal using some unused code from g_template, I might be able to create a PR...

![Screen Shot 2022-04-10 at 00 08 41](https://user-images.githubusercontent.com/5360142/162599630-de66833b-97a7-44ff-9507-5ad62271fafc.png)

porres:
Ok, did implement for [drawtext] and I'm using a '-h' flag, it all seems to work pretty fine. The '-h' seems useful and elegant in the end, it is also used to mark the field values for the color argument to take a symbol value... most of the code was actually there already but was unused.

here's a patch for testing: 
[test.pd.zip](https://github.com/pure-data/pure-data/files/8458216/test.pd.zip)

tomara-x:
i have questions, does this add hex colors to things like plot and drawpolygon? the second question is.. can we please have this?!

porres:
There are conflicts to be solved and maybe a need to restart from scratch :/

ArturJD96:
This would be a really nice update allowing more sophisticated UI color manipulations when using structs. Current color resolution is sufficient, but shows weakness when saturation is used as an affordance cue.

PR #1621, opened 2022-04-10
add '-h' flag for hexadecimal color in Data Structures.

porres:
So far only for [drawtext]/[drawnumber]/[drawsymbol]

This takes care of one part of https://github.com/pure-data/pure-data/issues/1620

still missing this implementation for [plot] and the other shapes. But this is a first implementation draft for testing and appreciation.

### DIFF
--- a/src/g_template.c
+++ b/src/g_template.c
@@ -868,8 +868,15 @@ static void fielddesc_setfloatarg(t_fielddesc *fd, int argc, t_atom *argv)
 
 static void fielddesc_setsymbolarg(t_fielddesc *fd, int argc, t_atom *argv)
 {
-        if (argc <= 0) fielddesc_setsymbol_const(fd, &s_);
-        else if (argv->a_type == A_SYMBOL)
+    if (argc <= 0) fielddesc_setsymbol_const(fd, &s_);
+    else if (argv->a_type == A_SYMBOL)
+    {
+        t_symbol *sym = argv->a_w.w_symbol;
+        if ('#' == sym->s_name[0])
+        {
+            fielddesc_setsymbol_const(fd, argv->a_w.w_symbol);
+        }
+        else
         {
             fd->fd_type = A_SYMBOL;
             fd->fd_var = 1;
@@ -877,7 +884,8 @@ static void fielddesc_setsymbolarg(t_fielddesc *fd, int argc, t_atom *argv)
             fd->fd_v1 = fd->fd_v2 = fd->fd_screen1 = fd->fd_screen2 =
                 fd->fd_quantum = 0;
         }
-        else fielddesc_setsymbol_const(fd, &s_);
+    }
+    else fielddesc_setsymbol_const(fd, &s_);
 }
 
 static void fielddesc_setarrayarg(t_fielddesc *fd, int argc, t_atom *argv)
@@ -1206,6 +1214,19 @@ static void numbertocolor(int n, char *s)
     green = n % 10;
     sprintf(s, "#%2.2x%2.2x%2.2x", rangecolor(red), rangecolor(blue),
         rangecolor(green));
+}
+
+static void symboltocolor(t_symbol *sym, char *s)
+{
+    if ('#' == sym->s_name[0])
+    {
+        int col = (int)strtol(sym->s_name+1, 0, 16) & 0xFFFFFF;
+        sprintf(s, "#%6.6x", col);
+    }
+    else
+    {
+        sprintf(s, "#%6.6x", 0);
+    }
 }
 
 static void curve_vis(t_gobj *z, t_glist *glist,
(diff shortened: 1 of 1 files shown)


Spacechild1:
Cool! I have been proposing this a couple of times, but was too lazy to implement it :-) 

> The '-h' seems useful and elegant in the end, it is also used to mark the field values for the color argument to take a symbol value...

I think it would be much nicer if we manage to do it without a flag.

What about the following:
1. check if there is a field with that name
2. if not, try to interpret the symbol as a (hex) number
3. if that fails, we print an error message

AFAICT, this would only fail if the user wants to set the color to `#FFFFFF` *and* there is also a field with the same name. I think this is very unlikely and also something the user can control. If they want to use hex colors, they must not use field names that look like hex numbers :-)

porres:
how do we make sure that if the given field name (other than a hex-number) should expect a hex symbol to define the color?

Spacechild1:
> how do we make sure that if the given field name (other than a hex-number) should expect a hex symbol to define the color?

We know the type of the field. If it's a float field, we get a numeric color code; if it's a symbol field, we get a hex color string.

porres:
Ok, I can try it, but would like to hear others as well, like @umlaeute , @danomatika , @jmmmp and stuff

jmmmp:
I guess that's the best solution. Another extra could be to also interpret negative numbers, which are read by all guis.

danomatika:
> We know the type of the field. If it's a float field, we get a numeric color code; if it's a symbol field, we get a hex color string.

This would be my suggestion... the arg type tells you the format.

porres:
ok, no one oppose's @Spacechild1's suggestion then, I'll work on it!

> Another extra could be to also interpret negative numbers, which are read by all guis.

can we have an extra issue for that?

porres:
> Another extra could be to also interpret negative numbers, which are read by all guis.

I actually don't think that is a good, though, as 'hexadecimal' is the current best practice in iemguis, as documented.

Spacechild1:
> Another extra could be to also interpret negative numbers, which are read by all guis.

When this was discussed on the list, I have actually proposed this, but @umlaeute rightfully pointed out that it's probably not a good idea to port this "hack" to data structures since we already have a better solution (= hex color codes).

porres:
Well, I need help with this. The flag thing really made things a lot easier in the code as I also found an unused code structure and just made use of it. If I have to do things differently, it would require some surgery.

the thing is that all field names expect a float value, as all the field names are variables for float arguments. Hence, we need a way to specify that this particular field will expect a symbol value, not a float.

I know it looks uglier and it wasn't my first ideal alternative as well, but it just happens to be an elegant coding solution for now.

porres:
my opinion is that data structures is already quite hard to use and understand and some objects already have several flags, so this wouldn't be that really bad...

Spacechild1:
I'll have a look. If there is really no sane way to implement this, we can go for the flag, but we should at least try.

porres:
> I'll have a look. If there is really no sane way to implement this, we can go for the flag, but we should at least try.

great, please have a go at it, my take is done and works great, here's a test patch

[test.pd.zip](https://github.com/pure-data/pure-data/files/8459688/test.pd.zip)

I'm really unconfortable touching this and changing things beyond this

reduzent:
> about the negative numbers and the hack, i have a question. Say that the user may want to use data structures to create art, and then use many colors, gradients and etc (now we would have more resolution with hex codes). But since they are symbols, they would populate the "symbol pool" (i don't know the correct name). If negative numbers would be allowed for the same purpose, would they be more efficient for this case?

Me thinks that when someone is going for the fancy stuff like 2d gradients implemented with data structures, the extra memory consumed by the increased symbol table might not be the primary concern.  Inversely, if you target a memory-restricted system, there are better options for creating fancy graphics than Pd's data structures.

Spacechild1:
@abreubacelar 

> But since they are symbols, they would populate the "symbol pool" (i don't know the correct name). If negative numbers would be allowed for the same purpose, would they be more efficient for this case?

You're right that excessive use of hex colors can pollute the symbol table. Switching between a few colors wouldn't be an issue, but cycling through all 16777216 RGB colors would definitely bring Pd to its knees... In practice, it's more likely that you want to make a smooth gradient between two colors and you won't need more than a few dozens colors for that. But it's certainly a valid concern and there should probably be a warning in the docs.

umlaeute:
just to add some more numbers: after generating all 16777216 symbols from `#000000` to `#ffffff`, the time from adding a new symbol raised from *0.004ms* to *0.163ms* on my system (Pd-0.52.2, so all the new symboltable in effect).

this of course matches @Spacechild1's oberservation (but is explicit about creating all possible colors).

> My takeaway is that with the new symbol table size, it is very unlikely that you will run into actual performance issues, but still it's something to be aware of.

i concur.
and for "being aware of" i agree that there should be a *hint* in the docs - but that's about it

porres:
trying to resolve conflicts, but I might find it easier to reboot this :(

porres:
> trying to resolve conflicts, but I might find it easier to reboot this :(

Sorry, yeah, it was easier for me to just reboot this... here's the new one https://github.com/pure-data/pure-data/pull/2670

Issue #1299, opened 2021-04-17
oscparse gives numbers as symbols


porres:
This seems to annoy everyone and I don't know if it's intentional, but [oscparse] outputs lists containing numbers that are treated as symbols and that is very unusual for Pd.

It also makes it a pain to use [route] as it can't route a message that starts with "0" as a symbol.

I wonder if we can change this by making it check and output a float instead. For those afraid of compatibility breakage, we could have a flag for that purpose, what do you say?

Spacechild1:
To be clear: this only happens for numbers which are part of the address pattern. For example, `/foo/1/bar` would output three symbols "foo", "1" and "bar".

I agree that this is behavior is very unintuitive and it has come up a few times. However, I'm not sure if we should just fix this and risk breaking some patches, or find another solution, like adding a flag to `[oscparse]`.

porres:
>  I'm not sure if we should just fix this and risk breaking some patches, or find another solution, like adding a flag to [oscparse].

I'm sure we gotta do something as when this happens, it's a lot of trouble. Why would you oppose to a flag?

The options I see are:
- fix it;
- fix it and create a compatibility flag;
- create a flag to force floats ("-f"?)
- make route deal with floats as symbols

not sure what else to do, maybe some new object or function that converts this easily to floats?

Spacechild1:
> Why would you oppose to a flag?

I never said I would...

> * fix it;
> * fix it and create a compatibility flag;
> * create a flag to force floats ("-f"?)
> * make route deal with floats as symbols

Nr. 1 and 4. are out of question.

porres:
sorry, now I get you were not sure if we should fix it (and then implicitly create a compatibility flag) or add a flag. Well, I'd be happy in one way or another... but if I were to pick a dog in the fight, I'd go for the fix + compatibility flag

umlaeute:
otoh, what if the OSC-address is `/foo/1.00/bar`?
it is unclear whether the 2nd element (`1.00`) is meant to be a number or some weird symbolic identifier (with a zero-padded 2nd-level value).

not that it would be esp. easy to properly route `1.00`.

i'm just mentioning this to point out that blindly converting everything that looks like a number within the address is probably not what we want.

Spacechild1:
IMO, it might be ok to only convert integers because that's the common use case.

umlaeute:
including `01`, i guess.

Spacechild1:
> including 01, i guess.

[...]

PR #1623, opened 2022-04-10
[oscparse] improvements

porres:
[oscparse] always parses addresses as symbols, even if they look like floats, which make it really hard to use [route]. I propose a flag that turns addresses that look like floats into floats.

We also add an outlet that gives the index in the output list that splits the address from the data

closes https://github.com/pure-data/pure-data/issues/1299

also closes https://github.com/pure-data/pure-data/issues/1858

### DIFF
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -290,8 +290,8 @@ static t_class *oscparse_class;
 
 typedef struct _oscparse
 {
-    t_object    x_obj;
-    t_int       x_flag;
+    t_object x_obj;
+    int x_flag;
 } t_oscparse;
 
 #define ROUNDUPTO4(x) (((x) + 3) & (~3))
@@ -392,12 +392,12 @@ static void oscparse_list(t_oscparse *x, t_symbol *s, int argc, t_atom *argv)
         if(x->x_flag)
         {
             t_symbol *sym = grabstring(argc, argv, &i, 1);
-              t_float f = 0.0f;
-              char *str_end = NULL;
-              f = strtod(sym->s_name, &str_end);
-              if (f == 0 && sym->s_name == str_end)
-                  SETSYMBOL(outv+j, sym);
-              else SETFLOAT(outv+j, f);
+            t_float f = 0.0f;
+            char *str_end = NULL;
+            f = strtod(sym->s_name, &str_end);
+            if (f == 0 && sym->s_name == str_end)
+                SETSYMBOL(outv+j, sym);
+            else SETFLOAT(outv+j, f);
         }
         else SETSYMBOL(outv+j, grabstring(argc, argv, &i, 1));
     }
@@ -482,7 +482,7 @@ static t_oscparse *oscparse_new(t_symbol *s, int argc, t_atom *argv)
 {
     t_oscparse *x = (t_oscparse *)pd_new(oscparse_class);
     x->x_flag = 0;
-    if (argc && argv[0].a_w.w_symbol == gensym("-float"))
+    if (argc && argv[0].a_w.w_symbol == gensym("-n"))
     {
         x->x_flag = 1;
     }
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -10,6 +10,7 @@
 #include <math.h>
 #include <stdio.h>
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/x_misc.c
```diff
@@ -289,7 +290,8 @@ static t_class *oscparse_class;
 
 typedef struct _oscparse
 {
-    t_object x_obj;
+    t_object    x_obj;
+    t_int       x_flag;
```
Spacechild1, line 294: `t_int` is meant for pointers. If you need an int, use `int` instead.

src/x_misc.c
```diff
@@ -385,8 +387,20 @@ static void oscparse_list(t_oscparse *x, t_symbol *s, int argc, t_atom *argv)
...
         dataonset, nfield); */
     for (i = j = 0; i < typeonset-1 && argv[i].a_w.w_float != 0 &&
-        j < outc; j++)
-            SETSYMBOL(outv+j, grabstring(argc, argv, &i, 1));
+         j < outc; j++)
+    {
+        if(x->x_flag)
+        {
+            t_symbol *sym = grabstring(argc, argv, &i, 1);
+              t_float f = 0.0f;
```
Spacechild1, line 395: watch your indentation ;-)

Spacechild1:
could the flag be just `-f` instead of `-float`?

porres:
I thought '-f' would be confusing as it is used for [oscformat] but it has a different meaning there ('format').

Note that [oscparse] has no arguments, so maybe we could just have an argument instead.

 I also think there could be a message method to change this behaviour as well.

src/x_misc.c
porres: what should I do here?

Spacechild1: The lines starting at 395 seem to have 2 extra spaces of indentation.

Spacechild1:
> I thought '-f' would be confusing as it is used for [oscformat] but it has a different meaning there ('format').

I see. What about `-n` for "numeric address pattern"?

> I also think there could be a message method to change this behaviour as well.

I don't think that's necessary. After all, this feature is really a bug fix in disguise to maintain backwards compatibility. I don't think that users would need to switch dynamically between the default "bad" mode and the new "good" mode. Otherwise, they can always use two objects.

porres:
> What about -n for "numeric address pattern"?

sounds good, and what's your reasoning that this is better than an argument, it's more explicit?

Spacechild1:
I think flags are more user-friendly than numeric codes. Also, we might want to add other options in the future and positional arguments don't scale well.

umlaeute:
i was going to comment on the `-float` as well.
just for the record: my original suggestion would have been `-fp` (for: "**f**loat (in the)  **p**ath"), but `-n` looks ok as well.

umlaeute:
> After all, this feature is really a bug fix in disguise to maintain backwards compatibility.

i do not agree with that statement. my reasoning is explained in the help patch for `[file split]` (which faces a similar dilemma).
the short version is: what do you do with `/foo/02/bar` *and* `/foo/2/bar`?

I'm kind of happy with a flag that enables the output as a number (as suggested here).
(it *might* be cool, if the actual flag used could also be used for `[file split]`.)

while we are there: how about adding a 2nd outlet to `[oscparse]` that gives the index in the output list that splits the address from the data?

Spacechild1:
> while we are there: how about adding a 2nd outlet to [oscparse] that gives the index in the output list that splits the address from the data?

are you reading my mind? I was about to propose exactly the same thing, but then thought I should open a seperate feature request.

umlaeute:
> are you reading my mind?

actually it's writing, but that ambiguity is all part of the bigger picture.

reduzent:
> while we are there: how about adding a 2nd outlet to `[oscparse]` that gives the index in the output list that splits the address from the data?

Very nice. That missing info prevented me from switching to [oscformat]/[oscparse] in some projects. It didn't occur to me that it could be added so easily.

Spacechild1:
@porres Can you update the PR title and description? Maybe just something like "[oscparse] improvements" and then a description of both features?

porres:
done

porres:
yay, thanks for merging this!

PR #1625, opened 2022-04-11
Enforce black-colored font in "File Open" and "Choose Directory" dialogs

myQwil:
Certain Linux distros will try to apply a dark theme to these dialogs, but will only be successful in changing the font color to white, making text virtually impossible to read against the white background.
This also occurs with tcl's checkboxes and radio buttons.

### DIFF
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -295,6 +295,11 @@ proc init_for_platform {} {
         "x11" {
             set ::modifier "Control"
             option add *PatchWindow*Canvas.background "white" startupFile
+            # enforce black foreground in check/radio buttons and open dialogs
+            option add *Checkbutton.foreground "black" userDefault
+            option add *Radiobutton.foreground "black" userDefault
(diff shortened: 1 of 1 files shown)


umlaeute:
hmm, i'd like to be able to reproduce the problem first.

could you provide a step-by-step description on how to re-create the original problem? (e.g. here on Debian/GNU with xfce4 as the window-manager, i can manually change the Appearance to use the "Adawait Dark" theme, but it doesn't seem to have any impact on Pd whatsoever - except for the color of the window borders)

myQwil:
To be honest, I don't see the problem on my end anymore either. I recently upgraded my OS to 22.04 and now when I open Pd, the window borders are a light-grey color, whereas before they were a dark-grey color.

A dark theme is no longer being applied to Tcl. This could be due to a change made to Tcl's code, or a change made to the code of my desktop environment (KDE Plasma). For the time being, however, it's probably safe to close this MR, as the issue seems to be outside the scope of Pd.

Issue #1628, opened 2022-04-17
Use "move" for IEMGUIs / Text Objects / Data Structures' Scalars and graphs instead of redrawing or using 'coords'


porres:
for example, in [bng] https://github.com/pure-data/pure-data/blob/master/src/g_bang.c#L84 it doesn't use simpler and better 'move' code as gatoms use (see https://github.com/pure-data/pure-data/blob/master/src/g_text.c#L1049)

Edit: Scalars and graphs should also be optimized, see discussion below.

It is more efficient to just "move" than recreate and redraw things.

porres:
by the way, I tried checking how Data Structures work when "displacing" but couldn't actually find it...

sebshader:
I think it's for supporting tcl/tk 8.4. Maybe it's time to drop that support..

sebshader:
text_drawborder in g_text.c also uses "coords" instead of "move"

porres:
> it seems like data structures (and arrays) are just redrawn

where? that's quite bad huh?

sebshader:
@porres I think for instance scalars are in g_scalar.c scalar_displace which seems like it just updates the coordinates and then calls scalar_redraw

for garrays & graphs there is graph_displace which calls glist_redraw

porres:
> there is scalar_displace in g_scalar.c which seems like it just updates the coordinates and then calls scalar_redraw

Yeah, that's it, I see it, well, we could have it call a new "scalar_move" function instead  of "scalar_redraw", it should be much more efficient. And in fact, when moving lots of Data Structures Scalars, I can see them getting redrawn and it's very very slow!!!

porres:
> for garrays & graphs there is graph_displace which calls glist_redraw in g_graph.c I wonder if there could be a glist_displace...

Yup, I can also see that! But I need help figuring clarifying this. Does this mean that when moving abstractions aroung with "graph on parent" they all get redrawn?

sebshader:
> Yup, I can also see that! But I need help figuring clarifying this. Does this mean that when moving abstractions aroung with "graph on parent" they all get redrawn?

yes. but usually widgetbehaviors have a "displace" so I'm not sure why that's not being used in a glist_redraw.. maybe some z-ordering/clipping thing?

porres:
> maybe some z-ordering/clipping thing?

why? what do you mean?

[...]

PR #1629, opened 2022-04-18
Use "move" for IEMGUIs / Text Objects / Data Structures' Scalars and graphs instead of redrawing or using 'coords'

porres:
For now, only text objects, which partially closes https://github.com/pure-data/pure-data/issues/1628

more to come...

EDIT: iemguis and graphs taken care of already, only Data Structures left to go.

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -1251,8 +1251,10 @@ static void text_displace(t_gobj *z, t_glist *glist,
     {
         t_rtext *y = glist_findrtext(glist, x);
         rtext_displace(y, glist->gl_zoom * dx, glist->gl_zoom * dy);
-        text_drawborder(x, glist, rtext_gettag(y),
-            rtext_width(y), rtext_height(y), 0);
+        sys_vgui(".x%lx.c move %sR %d %d\n", glist_getcanvas(glist), rtext_gettag(y),
+                 dx * glist->gl_zoom, dy * glist->gl_zoom);
+        sys_vgui(".x%lx.c move %siolet %d %d\n", glist_getcanvas(glist), rtext_gettag(y),
+                 dx * glist->gl_zoom, dy * glist->gl_zoom);
         canvas_fixlinesfor(glist, x);
     }
 }
@@ -1442,11 +1444,11 @@ void glist_drawiofor(t_glist *glist, t_object *ob, int firsttime,
         int onset = x1 + (width - iow) * i / nplus;
         if (firsttime)
             sys_vgui(".x%lx.c create rectangle %d %d %d %d "
-                "-tags [list %so%d outlet] -fill black\n",
+                "-tags [list %so%d %siolet outlet] -fill black\n",
                 glist_getcanvas(glist),
                 onset, y2 - oh + glist->gl_zoom,
                 onset + iow, y2,
-                tag, i);
+                tag, i, tag);
         else
             sys_vgui(".x%lx.c coords %so%d %d %d %d %d\n",
                 glist_getcanvas(glist), tag, i,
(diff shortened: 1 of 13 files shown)


src/g_text.c
```diff
@@ -1251,8 +1251,10 @@ static void text_displace(t_gobj *z, t_glist *glist,
     {
         t_rtext *y = glist_findrtext(glist, x);
         rtext_displace(y, glist->gl_zoom * dx, glist->gl_zoom * dy);
-        text_drawborder(x, glist, rtext_gettag(y),
-            rtext_width(y), rtext_height(y), 0);
+        sys_vgui(".x%lx.c move %sR %d %d\n", glist_getcanvas(glist), rtext_gettag(y),
+                 dx * glist->gl_zoom, dy * glist->gl_zoom);
+        sys_vgui(".x%lx.c move %siolet %d %d\n", glist_getcanvas(glist), rtext_gettag(y),
+                 dx * glist->gl_zoom, dy * glist->gl_zoom);
```
sebshader, line 1257: I think there might be a way to do this w/o creating a new iolet tag.. i'll think about this

porres: I though about adding a global tag as I did with many others, but it got complicated and I gave up thinking it couldn't be done sanely. Please have a go at it though to make sure :)

sebshader:
too bad we can't use moveto without dropping tk 8.5

sebshader:
@porres the canvas moveto command was only introduced in tk 8.6 so if we want to support tk 8.5 we can't use it. It would greatly simplify a lot of this. for instance for the iemgui stuff we wouldn't have to add arguments to the draw function, we could simply call moveto the correct (current) coordinates

porres:
And what is the advantage of using "move" to "coords", and "moveto" to "move"?

I'm assuming the "move" command is more efficient but I didn't get where I think it's important yet, which would be moving objects within graphs and not redrawinf data structures, can you maybe help me with that @sebshader ?

sebshader:
@porres yeah, more efficient and simpler to write. moveto uses absolute coordinates, and move uses relative ones.

porres:
> I'm unsure if it's worth refactoring the iemguis 

I already did it! It wasn't bad

sebshader:
> > I'm unsure if it's worth refactoring the iemguis
> 
> I already did it! It wasn't bad

yes, but in doing so certain code gets more involved. Not only does the draw code need 2 extra arguments for the delta values, but in using only move everywhere there is slightly more c-side code like:
```
int xpos = (int)atom_getfloatarg(0, ac, av);
     int ypos = (int)atom_getfloatarg(1, ac, av);
     int dx = xpos - iemgui->x_obj.te_xpix;
     int dy = ypos - iemgui->x_obj.te_ypix;
     iemgui->x_obj.te_xpix = xpos;
     iemgui->x_obj.te_ypix = ypos;
     if(glist_isvisible(iemgui->x_glist))
     {
         (*iemgui->x_draw)(x, iemgui->x_glist, IEM_GUI_DRAW_MODE_MOVE);
         (*iemgui->x_draw)(x, iemgui->x_glist, IEM_GUI_DRAW_MODE_DISPLACE, dx*zoom, dy*zoom);
         canvas_fixlinesfor(iemgui->x_glist, (t_text*)x);
     }
```
I think a good compromise might be to use macros for generics and set specific widgetbehaviors for each iemgui. I'm looking into that a bit now

porres:
> I think a good compromise might be to use macros for generics and set specific widgetbehaviors for each iemgui.

so much more refactoring instead? I agree it's not ideal, but I was quite little intrusive

umlaeute:
as always i think that iemgui's should be refactored on a BIG scale, rather than doing some micro-optimisations.

with *BIG scale* i mean: move all the drawing code to the Tcl/Tk side, and use some well defined interface (ideally: mimicking the GUI->Core interface) to control that from the Core.

umlaeute:
re *BIG scale* refactoring:

this would then get rid of practically all the `class_setwidget` functionality: all objects (however fancy their GUI) would use the same displace/select/... functions (on the core-side), it's only the GUI side that would differentiate.

porres:
I can gladly revert my commits for the iemguis change in favor of https://github.com/pure-data/pure-data/pull/1633 how do I do that?

porres:
I guess I can just start a new cleaner PR instead with a single commit ;)

porres:
closing in favor of https://github.com/pure-data/pure-data/pull/1635

Issue #1634, opened 2022-04-21
Use "move" to displace objects instead of "coords"


porres:
This is in respect all text objects, graphs and iemguis.

iemguis is taken care of in https://github.com/pure-data/pure-data/pull/1633

PR #1635, opened 2022-04-21
Use "move" to displace objects instead of "coords"

porres:
closes https://github.com/pure-data/pure-data/issues/1634

The only thing is that all the graph objects ("gobj") inside it get redrawn, with 

```
                gobj_vis(g, x, 0);
                gobj_vis(g, x, 1);
```

much like in https://github.com/pure-data/pure-data/blob/master/src/g_graph.c#L656

I'm doing this cause I don't know better

there's a gobj_displace code in https://github.com/pure-data/pure-data/blob/master/src/g_canvas.c#L71

but if I use that, the objects get displaced alright when moving the graph on the parent but they also get moved inside the abstraction, which screws things up...

I still think that redrawing "gobj" objects is not ideal, but we can think of a way not to do that later...

### DIFF
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -947,8 +947,19 @@ static void graph_displace(t_gobj *z, t_glist *glist, int dx, int dy)
     {
         x->gl_obj.te_xpix += dx;
         x->gl_obj.te_ypix += dy;
-        if (glist_isvisible(glist)) {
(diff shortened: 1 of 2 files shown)


porres:
not only this has conflicts, but also 'move' is already used in the current master. Only thing still missing is using it for 'graphs'

PR #1639, opened 2022-04-21
allow floats as labels and don't allow float as send/receive names

### LINKED ISSUE #1081: allow float in iemguis labels but not their built in send/receive
Checking the code of iemguis I realized you can set a float as a send/receive address. This is weird, because you cannot have floats inside [send]/[receive]. Nonetheless, at least you can set a send symbol via the 2nd inlet of [send] as a symbolic float.
Now, what's actually funny is that you can only have integers, and not decimal point floats! 
Since this is the same code that handles labels in iemguis (see https://github.com/pure-data/pure-data/blob/master/src/g_all_guis.c#L200 )

This also means that you cannot have floating point values in labels, which seems like an unnecessary restriction. Allowing decimal point floats seems useful for label display. As far as send/receive, I don't see the usefulness even for integer floats, but well, what are we gonna do? Remove that option? Make a separate code not to allow it?

And in a related discussion, would it be useful to allow receive and send to load floats as send symbols?

porres:
closes #1081

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -219,17 +219,24 @@ void iemgui_verify_snd_ne_rcv(t_iemgui *iemgui)
     }
 }
 
-t_symbol *iemgui_new_dogetname(t_iemgui *iemgui, int indx, t_atom *argv)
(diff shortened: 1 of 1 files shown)


porres:
closing in favor of https://github.com/pure-data/pure-data/pull/1642

PR #1642, opened 2022-04-22
label updates for iemguis

porres:
closes https://github.com/pure-data/pure-data/issues/1081

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -219,17 +219,24 @@ void iemgui_verify_snd_ne_rcv(t_iemgui *iemgui)
     }
 }
 
-t_symbol *iemgui_new_dogetname(t_iemgui *iemgui, int indx, t_atom *argv)
+t_symbol *iemgui_new_doget_labelname(t_iemgui *iemgui, int indx, t_atom *argv)
 {
     if (IS_A_SYMBOL(argv, indx))
         return (atom_getsymbolarg(indx, 100000, argv));
     else if (IS_A_FLOAT(argv, indx))
     {
         char str[80];
-        sprintf(str, "%d", (int)atom_getfloatarg(indx, 100000, argv));
+        sprintf(str, "%g", atom_getfloatarg(indx, 100000, argv));
         return (gensym(str));
     }
     else return (gensym("empty"));
+}
+
+t_symbol *iemgui_new_dogetname(t_iemgui *iemgui, int indx, t_atom *argv)
+{
+    if (IS_A_SYMBOL(argv, indx))
+        return (atom_getsymbolarg(indx, 100000, argv));
+    return (gensym("empty"));
 }
 
 void iemgui_new_getnames(t_iemgui *iemgui, int indx, t_atom *argv)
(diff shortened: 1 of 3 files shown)


src/g_all_guis.h
```diff
@@ -268,6 +268,7 @@ EXTERN t_symbol *iemgui_dollararg2sym(t_symbol *s, int nth_arg, int tail_len, in
 EXTERN void iemgui_verify_snd_ne_rcv(t_iemgui *iemgui);
 EXTERN void iemgui_all_sym2dollararg(t_iemgui *iemgui, t_symbol **srlsym);
 EXTERN t_symbol *iemgui_new_dogetname(t_iemgui *iemgui, int indx, t_atom *argv);
+EXTERN t_symbol *iemgui_new_doget_labelname(t_iemgui *iemgui, int indx, t_atom *argv);
```
umlaeute, line 271: i think there is no reason the declare this new function in the header file, let alone declare the function as `EXTERN`.

src/g_all_guis.c
```diff
@@ -219,26 +219,33 @@ void iemgui_verify_snd_ne_rcv(t_iemgui *iemgui)
...
+}
+
 void iemgui_new_getnames(t_iemgui *iemgui, int indx, t_atom *argv)
 {
     if (argv)
     {
         iemgui->x_snd = iemgui_new_dogetname(iemgui, indx, argv);
         iemgui->x_rcv = iemgui_new_dogetname(iemgui, indx+1, argv);
-        iemgui->x_lab = iemgui_new_dogetname(iemgui, indx+2, argv);
+        iemgui->x_lab = iemgui_new_doget_labelname(iemgui, indx+2, argv);
```
umlaeute, line 248: hmm.
introducing a new function to be called exactly once seems to be an overkill.
how about:
- extending `iemgui_new_dogetname` to also return values for `A_FLOAT`
- use `!(IS_A_FLOAT(argv, indx))` to special-case the snd/rcv names

alternatively, just add a `IS_A_FLOAT()` clause to handle the float-case for labels.

umlaeute (changes requested):
this seems overly complicated.

when fixing, please `amend` your commit and `force-push` to this branch

src/g_vumeter.c
```diff
@@ -412,8 +412,10 @@ static void vu_label(t_vu *x, t_symbol *s)
 static void vu_label_pos(t_vu *x, t_symbol *s, int ac, t_atom *av)
 {iemgui_label_pos((void *)x, &x->x_gui, s, ac, av);}
 
-static void vu_label_font(t_vu *x, t_symbol *s, int ac, t_atom *av)
-{iemgui_label_font((void *)x, &x->x_gui, s, ac, av);}
+static void vu_label_font(t_vu *x, t_symbol *s, int ac, t_atom *av){
+    iemgui_label_font((void *)x, &x->x_gui, s, ac, av);
+    (*x->x_gui.x_draw)(x, x->x_gui.x_glist, IEM_GUI_DRAW_MODE_CONFIG);
+}
```
umlaeute, line 419: i wonder whether we shouldn't call `draw_config` directly from `iemgui_label_font`.

pros
- it would allow all iemguis to do more sophisticated adjustments when the font changes.
- we already do something very similar when the size changes

this would create a performance penalty when changing the label-font for all iemguis, but i think font-changes are done seldom enough to not be a real problem.

umlaeute: on re-thinking, your proposed solution is probably better than my suggestions.

porres:
Well, I trust your judging. If you want I can make changes to help you want, but you can also go ahead and do it your own way

umlaeute:
i cherry picked parts of the patches, and modified the rest.
thanks a lot.

PR #1644, opened 2022-04-23
Sliders fixes2

### LINKED ISSUE #1042: [slider] value isn't preserved accross zooming

### LINKED ISSUE #1043: [slider] buggy shift-mouse behaviour when zoomed-in and send=receive

Ant1r:
Make internal value independent from zoom factor.
This replaces  #933, now targeted to `merge-iemgui` branch.
This fixes #1042 and fixes #1043.


Ant1r:
BTW it's cool now to have only one slider file to fix! Thanks @umlaeute !

umlaeute:
i've cherry-picked this single commit as 1c9c396d5767b68be2605f3977e6e599d17f4062 to keep the history more linear.
thanks.

PR #1645, opened 2022-04-23
nbx fixes

porres:
some fixes to include into merge-iemgui

### DIFF
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -29,10 +29,13 @@ static void my_numbox_draw_update(t_gobj *client, t_glist *glist);
 
 void my_numbox_clip(t_my_numbox *x)
 {
-    if(x->x_val < x->x_min)
-        x->x_val = x->x_min;
-    if(x->x_val > x->x_max)
-        x->x_val = x->x_max;
+    if (x->x_min != 0 || x->x_max != 0)
+    {
+        if(x->x_val < x->x_min)
+            x->x_val = x->x_min;
+        if(x->x_val > x->x_max)
+            x->x_val = x->x_max;
+    }
 }
 
 void my_numbox_calc_fontwidth(t_my_numbox *x)
@@ -397,15 +400,18 @@ int my_numbox_check_minmax(t_my_numbox *x, double min, double max)
     }
     x->x_min = min;
     x->x_max = max;
-    if(x->x_val < x->x_min)
+    if(x->x_min != 0 || x->x_max != 0)
     {
-        x->x_val = x->x_min;
-        ret = 1;
-    }
-    if(x->x_val > x->x_max)
-    {
-        x->x_val = x->x_max;
-        ret = 1;
+        if(x->x_val < x->x_min)
+        {
+            x->x_val = x->x_min;
+            ret = 1;
+        }
+        if(x->x_val > x->x_max)
+        {
+            x->x_val = x->x_max;
+            ret = 1;
+        }
     }
     if(x->x_lin0_log1)
         x->x_k = exp(log(x->x_max/x->x_min) / (double)(x->x_log_height));
@@ -754,7 +760,7 @@ static void *my_numbox_new(t_symbol *s, int argc, t_atom *argv)
     int lilo = 0, ldx = 0, ldy = -8;
     int fs = x->x_gui.x_fontsize;
     int log_height = 256;
-    double min = -1.0e+37, max = 1.0e+37, v = 0.0;
+    double min = 0, max = 0, v = 0.0;
 
     if((argc >= 17)&&IS_A_FLOAT(argv,0)&&IS_A_FLOAT(argv,1)
        &&IS_A_FLOAT(argv,2)&&IS_A_FLOAT(argv,3)
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -575,7 +575,8 @@ static void my_numbox_set(t_my_numbox *x, t_floatarg f)
         else
         {
             x->x_val = ftocompare;
-            my_numbox_clip(x);
+            if (pd_compatibilitylevel < 53)
+                my_numbox_clip(x);
         }
         sys_queuegui(x, x->x_gui.x_glist, my_numbox_draw_update);
     }
@@ -734,7 +735,8 @@ static void my_numbox_key(void *z, t_symbol *keysym, t_floatarg fkey)
         x->x_buf[0] = 0;
         x->x_gui.x_fsf.x_change = 0;
         clock_unset(x->x_clock_reset);
-        my_numbox_clip(x);
+        if (pd_compatibilitylevel < 53)
+            my_numbox_clip(x);
         my_numbox_bang(x);
         sys_queuegui(x, x->x_gui.x_glist, my_numbox_draw_update);
     }
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -14,6 +14,11 @@
 #include "g_all_guis.h"
 #include <math.h>
 
+#ifdef _MSC_VER
+#include <float.h>
+#define isnan _isnan
+#endif
+
 #define MINDIGITS 1
 #define MINFONT   4
 
@@ -557,8 +562,15 @@ static void my_numbox_set(t_my_numbox *x, t_floatarg f)
         ftocompare must be t_float type like x_val. */
     if (memcmp(&ftocompare, &x->x_val, sizeof(ftocompare)))
     {
-        x->x_val = ftocompare;
-        my_numbox_clip(x);
+        if(isnan(f))
+        {
+            x->x_val = f;
+        }
+        else
+        {
+            x->x_val = ftocompare;
+            my_numbox_clip(x);
+        }
         sys_queuegui(x, x->x_gui.x_glist, my_numbox_draw_update);
     }
 }


porres:
forgot to mention that https://github.com/pure-data/pure-data/pull/1645/commits/4895b4b4986b4d6373c47d840a79c60ac3c6244d closes https://github.com/pure-data/pure-data/issues/633

umlaeute:
hmm, i had considered these form your original PR, but there are still some questions open:

1. nbx already prints `NaN`s just fine with Pd-0.52-2; so why do we need [a33faf6](https://github.com/pure-data/pure-data/pull/1645/commits/a33faf6ee6aeb254b23db197955900d89f5dd6c6)?
2. i do not really like the fix for printing `inf`. i know that the ordinary gatom numberbox uses `min=max=0` to detect "no range", but iirc nbx has always used this to fix the nbx to a specific value. so this might actually break things.
   - would it not possible to use `-inf` and `+inf` as ranges for unlimited ranges?
   - also, this should play nicely with `log`-mode (and: with toggling between *lin* and *log* mode
3. i originally decided to not include this after reading https://github.com/pure-data/pure-data/issues/633#issuecomment-939372529 - but i guess this was overcautios on my side, as my entire PR is heavily touching all iemguis, so there's no reason to exclude this specific one.

porres:
> nbx already prints NaNs just fine with Pd-0.52-2

can you give me a patch to test? cause I can't confirm this, see:

Screen Shot 2022-04-23 at 16 49 06

umlaeute:
nan

porres:
Well, it seems that for some reason this is still buggy on macOS? I wouldn't know why though.

umlaeute:
looking at the code i too don't see where such a bug could be introduced.

porres:
> looking at the code i too don't see where such a bug could be introduced.

yeah, all I know is that it worked before 0.46

Anyway, my fix for it was copied from how gatom used to do it https://github.com/pure-data/pure-data/blob/6a06fa736e501afa5cda62e845fbba44732cc2af/src/g_text.c#L597

This changed with https://github.com/pure-data/pure-data/pull/791 so gatoms do not currently use "isnan" to deal with this.

It seems number boxes now do a similar comparison and somehow this works for linux and stuff, but not for mac...

All I know is that my fix does fix this for good, but maybe you don't want to use 'isnan' and do the same thing as gatoms do now (I haven't checked what's the difference and why it works for gatoms).

umlaeute:
i do not care especially about `isnan`, but i would like to understand the problem before fixing it.

Spacechild1:
A comment from my side:

Pd is compiled with `-ffast-math` which enables a couple of "unsafe" math optimizations. Among others, it assumes that there are no NaNs or Infs. As a consequence, the result of `isnan` would be undefined. For more details, here's a good SO answer: https://stackoverflow.com/a/22931368/6063908

We might consider compiling Pd with `-ffast-math -fno-finite-math-only`. That's essentially what SuperCollider does (because they occasionally rely on non-signalling NaNs as sentinal values). Then we could get rid of the brittle `memcmp` hack and safely use `isnan` and friends.

Spacechild1:
> As a consequence, the result of isnan would be undefined.

Quick demonstration: https://godbolt.org/z/MKvzsMr3j

When compiling with `-ffast-math`, recent versions of GCC and clang simply replace `isnan` with a constant 0 (= false). By adding`-fno-finite-math-only`,  `isnan` works as expected.

umlaeute:
I'm all for adding `-fno-finite-math-only`, but I don't see how this will fix the problem (and a quick test shows, it doesn't).

What we are seeing is that `(NAN <  -1e37) == true` on macOS (wrongly, but possibly just undefined behaviour), and `false` on Linux (as it should be).

Now @porres is replacing this with:

```

if isnan(NAN)
 X=NAN;
else {
  if (NAN<-1e37)X=-1e37;
}
```

which apparently fixes the problem.
If `isnan` evaluated to 1, the fix would be a nop.

If Nan's where handled correctly, the `isnan`  would be useless...

porres:
while we're at it, "isnan" is not used anymore for gatoms and we can delete this --> https://github.com/pure-data/pure-data/blob/master/src/g_text.c#L589

porres:
So, I first did this change a while back, before #791 that changed the code of gatom and nbx, maybe it made more sense back then.

The thing now is that 'nan' values are passed through and x->x_val gets this value here 

https://github.com/pure-data/pure-data/blob/master/src/g_numbox.c#L610

I can now see the problem is in my_numbox_clip and nan is considered to be less than -1e37, so it clips to that value. That didn't happen before #791 - what happened is what I described here https://github.com/pure-data/pure-data/issues/629#issuecomment-1107836773 (the last value would come out instead).

I also clearly don't know why on linux (and windows, cause I just tested on a virtual machine) the clip function doesn't clip nans and what we need to do to prevent this from happening on macOS...

Now, one fix is test if it is a nan and not clip it then, which is actually what I'm doing in https://github.com/pure-data/pure-data/pull/1645/commits/a33faf6ee6aeb254b23db197955900d89f5dd6c6 so it fixes it.

> * if `isnan` evaluated to `false`, the fix would be a nop.
> * if nan's where handled correctly, the `isnan`  would be useless...

yes, I can see that, agree. 

Now, another thing is that nbx shouldn't clip values in the first place, so https://github.com/pure-data/pure-data/pull/1645/commits/4895b4b4986b4d6373c47d840a79c60ac3c6244d also fixes this! What we're discussing then is how to fix the issue if the compatibility flag is set to < 53, which is really worth noting.

It seems that being able to safely use 'isnan' is helpful one way or another. By the way, it seems that [expr] uses 'isnan' --> https://github.com/pure-data/pure-data/blob/2b5c219656771a311f36d8ac826281f153e0ea3e/src/x_vexp_fun.c#L197

so we need a safe way to use it.

> We might consider compiling Pd with -ffast-math -fno-finite-math-only. That's essentially what SuperCollider does (because it occasionally uses non-signalling NaNs as sentinal values). Then we could get rid of the brittle memcmp hack and safely use isnan (and others).

sounds good

umlaeute:
>I can now see the problem is in my_numbox_clip and nan is considered to be less than -1e37, so it clips to that value. 

Yes, but this is simply the wrong behaviour: a nan is not sortable, so it can never be "less than -1e37".




>I also clearly don't know why on linux (and windows, cause I just tested on a virtual machine) the clip function doesn't clip nans 


Because that us the correct behaviour: the nan value doesn't match any of the comparisions, so no clipping applies.

(Both my statements assume that there is such a thing as a "NaN" - which @Spacechild1  has shown is not necessary true. But then 'isnan' is just 'false', which brings us back to my last comment)

porres:
I get there's a bug in macOS and NaN shouldn't be clipped.

> But then 'isnan' is just 'false'

hmm, I'm confused. My fix works because it is true. And gatoms also worked before when it used 'isnan'. So it works in [expr]. I may not understood things correctly, but it seems 'isnan' might be undefined if Pd is not compiled correctly, so we could just make it sure it compiles correctly, right?

porres:
> 2\. i do not really like the fix for printing `inf`. i know that the ordinary gatom numberbox uses `min=max=0` to detect "no range", but iirc nbx has always used this to fix the nbx to a specific value. so this might actually break things.

nbx uses this to clip to a specific value if a range is given, like gatom numberbox, so there's no real difference.

> would it not possible to use -inf and +inf as ranges for unlimited ranges?

it's easy to set -inf and inf as default min/max ranges, the problem is that if you get into the properties window and save something, -inf/inf become zero (by the way, this is also the case for gatoms). I didn't try a fix for this. Funny enough, I just tested, and 'nans' get through the clip function in this case on macOS, they don't get clipped to -inf or inf or whatever :) so I guess that if we can have properties windows take -inf/inf in iemguis and gatoms, but I don't think there's a real advantage and also why nbx should be different than gatom by not considering 0,0 as unlimited range.

> also, this should play nicely with log-mode (and: with toggling between lin and log mode

maybe the answer is in here, but I don't get. As it is, if I toggle to log mode with defaut values, the range is from 1e+35 to 1e+37. How would it be better with -inf/inf? Oh, and I tried it, and what happens is that when I toggle to log mode with default values of -inf/inf it just sets the range to inf/inf

When going to Log mode one should give it a meaningful range not crossing zeros anyway...

Hence, I just don't see the problem with the current fix for printing -inf/inf, and it would be nice and consistent to gatoms.

umlaeute:
> but it seems 'isnan' might be undefined if Pd

no. it's *behaviour* might be undefined.

anyhow: the fun part is that https://github.com/pure-data/pure-data/commit/4895b4b4986b4d6373c47d840a79c60ac3c6244d short-circuits the entire clipping which yields the wrong results on macOS, so I'm just going to apply this single commit, and leave out the entire `isnan`-mess and everybody will be happy.

imho, https://github.com/pure-data/pure-data/commit/4895b4b4986b4d6373c47d840a79c60ac3c6244d *completely resolves #629 , as nbx now displays the incoming value regardless of whether it's a "nice" float, and it still is (and **should be**) impossible to use nbx to create NaN and ±inf, so the range doesn't apply for these cases.

it might still be interesting to allow for setting ±inf as a range delimiter in the properties (e.g. in double precision we can have values that are `>>1e37`, even though they might be of no practical use)

umlaeute:
closing this, as i think it's actually merged (under the assumption that the 3rd commit obsoletes the other two)

porres:
> closing this, as i think it's actually merged (under the assumption that the 3rd commit obsoletes the other two)

Sort of. One consideration is that when setting compatibility to < 53, then -inf/inf cannot be printed, but this might actually be better! Another one is that I still like the idea of setting the range to 0/0 to remove the range limitation. I not only like this better but also think it would be better for being consistent to gatom float. But I can live without it and, anyway, the idea is that we also now should improve gatom boxes and deprecate [nbx].

Anyway, thanks a lot :) really glad of all the cleaning and improving we're doing on iemguis

cheers

PR #1649, opened 2022-04-24
rename Intrrpt/Hold to min/max in bng properties

porres:
This closes https://github.com/pure-data/pure-data/issues/976 and over there there's a discussion why I think these names are terrible.

In fact, I have already documented how these work properly in my rewrite of the iemguis documentation, but it'd be much simpler and easier to explain/document if we just change the names.

I will get to the documentation itself once the iemguis-merge PR gets merged and will also take care of many other documentation related stuff - I'm keeping track of al the things I need to change.


porres:
ping @umlaeute : this seems like a no brainer, what do you think? Got some considerations?

umlaeute:
i originally doubted whether `Min:` would be meaningful in the GUI.

porres:
what do you mean? what would be the issue? What is the suggestion?

umlaeute:
applied with 3c6ec04d9994d462c09cbbb108d1d23a62818434

PR #1650, opened 2022-04-25
change vu default size

porres:
I had proposed this change in https://github.com/pure-data/pure-data/issues/920#issuecomment-615439700

I made it higher just because the scale looks less compressed and the size is also closer to vertical sliders and radios, wat do you think? see screenshot on how it looks now.

Screen Shot 2022-04-24 at 21 12 28


umlaeute:
hmm, for me it looks a bit different (and the current defaults seem to be much better).

vu-sliders

(this is with a patch fontsize of **12** (the default))

---

i guess you are just using a different font-size then (probably **10**).

here's how the defaults look for various fonts in the current `merge-iemgui` branch:
vu-sliders-org

and here's with your changes applied:
vu-sliders-porres

porres:
> i guess you are just using a different font-size then (probably 10).

I was actually using 12, but yeah, using larger fonts make the vu larger in comparison to the previous setting

umlaeute:
hmm, if you were indeed using a font-size `12`, i'm genuinely puzzled why the two look so different (and in fact look like my `10` example)

umlaeute:
applied with 22aaabcf60375a3e03392b0fd73006776cf22854 (after changing the title to something meaningful)

PR #1659, opened 2022-05-01
canvas zoom and scroll with mousewheel

### LINKED ISSUE #1726: Offer more than two zoom settings
I'm using Pure Data 0.52.1 on Ubuntu 22.04 with screen resolution of 3840 x 2400. The default size of a patch is very very small and almost illegible. 

Here is a patch shown with the rest of the desktop for scale.

Screenshot from 2022-08-28 21-34-04

Pure data offers one increase in ui zoom/scale, which makes the patch more legible. 

Here is a patch with the zoom applied (left) next to an original sized patch for scale/reference (right).

Screenshot from 2022-08-28 21-29-29

Though this is much better it would be even better the ui could go larger. This would be useful for many situations

Can Pure Data offer more than two zoom sizes?

Ant1r:
Zoom and scroll with the mouse wheel, like most modern software.

zoom

The code is adapted from https://wiki.tcl-lang.org/page/Canvas+zooming+using+mousewheel.
It will also allow to clean up the GUI C code from every `IEMGUI_ZOOM()` occurrences and logic ;-)

Note: the current zoom factor could be reported to the Pd-core side, in order to be saved to file.

### DIFF
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -93,6 +93,8 @@ proc ::pd_bindings::global_bindings {} {
     bind all <$::modifier-Key-minus>       {::pd_canvaszoom::stepzoom %W -1}
     bind all <$::modifier-Key-KP_Add>      {::pd_canvaszoom::stepzoom %W 1}
     bind all <$::modifier-Key-KP_Subtract> {::pd_canvaszoom::stepzoom %W -1}
+    bind all <$::modifier-Key-0>           {::pd_canvaszoom::setzoom %W 0}
+    bind all <$::modifier-Key-KP_0>        {::pd_canvaszoom::setzoom %W 0}
 
     # note: we avoid CMD-H & CMD+Shift-H as it hides Pd on macOS
 
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -88,11 +88,11 @@ proc ::pd_bindings::global_bindings {} {
     # take the '=' key as a zoom-in accelerator, because '=' is the non-shifted
     # "+" key... this only makes sense on US keyboards but some users
     # expected it... go figure.
-    bind all <$::modifier-Key-equal>       {::pd_menucommands::scheduleAction menu_send_float %W zoom 2}
-    bind all <$::modifier-Key-plus>        {::pd_menucommands::scheduleAction menu_send_float %W zoom 2}
-    bind all <$::modifier-Key-minus>       {::pd_menucommands::scheduleAction menu_send_float %W zoom 1}
-    bind all <$::modifier-Key-KP_Add>      {::pd_menucommands::scheduleAction menu_send_float %W zoom 2}
-    bind all <$::modifier-Key-KP_Subtract> {::pd_menucommands::scheduleAction menu_send_float %W zoom 1}
+    bind all <$::modifier-Key-equal>       {::pd_canvaszoom::stepzoom %W 1}
+    bind all <$::modifier-Key-plus>        {::pd_canvaszoom::stepzoom %W 1}
+    bind all <$::modifier-Key-minus>       {::pd_canvaszoom::stepzoom %W -1}
+    bind all <$::modifier-Key-KP_Add>      {::pd_canvaszoom::stepzoom %W 1}
+    bind all <$::modifier-Key-KP_Subtract> {::pd_canvaszoom::stepzoom %W -1}
 
     # note: we avoid CMD-H & CMD+Shift-H as it hides Pd on macOS
 
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -68,6 +68,7 @@ namespace eval ::pd_bindings:: {
     # expected it... go figure.
     setshortcuts "Edit|ZoomIn"            "${control} plus" "${control} KP_Add" "${control} equal"
     setshortcuts "Edit|ZoomOut"           "${control} minus" "${control} KP_Subtract"
+    setshortcuts "Edit|ZoomReset"         "${control} 0" "${control} KP_0"
     setshortcuts "Edit|TidyUp"            "${control} Shift R"
     setshortcuts "Edit|ConnectSelection"  "${control} K"
     setshortcuts "Edit|Triggerize"        "${control} T"
@@ -240,8 +241,9 @@ proc ::pd_bindings::global_bindings {} {
     bind  all  <<Edit|SelectAll>>         {::pd_menucommands::scheduleAction menu_send %W selectall}
 
     bind  all  <<Edit|Font>>              {::pd_menucommands::scheduleAction menu_font_dialog}
-    bind  all  <<Edit|ZoomIn>>            {::pd_menucommands::scheduleAction menu_send_float %W zoom 2}
-    bind  all  <<Edit|ZoomOut>>           {::pd_menucommands::scheduleAction menu_send_float %W zoom 1}
+    bind  all  <<Edit|ZoomIn>>            {::pd_menucommands::scheduleAction ::pd_canvaszoom::stepzoom %W 120}
+    bind  all  <<Edit|ZoomOut>>           {::pd_menucommands::scheduleAction ::pd_canvaszoom::stepzoom %W -120}
+    bind  all  <<Edit|ZoomReset>>         {::pd_menucommands::scheduleAction ::pd_canvaszoom::setzoom %W 0}
     bind  all  <<Edit|TidyUp>>            {::pd_menucommands::scheduleAction menu_send %W tidy}
     bind  all  <<Edit|ConnectSelection>>  {::pd_menucommands::scheduleAction menu_send %W connect_selection}
     bind  all  <<Edit|Triggerize>>        {::pd_menucommands::scheduleAction menu_send %W triggerize}
--- a/tcl/pd_canvaszoom.tcl
+++ b/tcl/pd_canvaszoom.tcl
@@ -66,8 +66,12 @@ proc ::pd_canvaszoom::zoominit {mytoplevel} {
 proc ::pd_canvaszoom::scroll_point_to {c xcanvas ycanvas xwin ywin} {
     set zdepth $::pd_canvaszoom::zdepth($c)
     set scrollregion [$c cget -scrollregion]
-    set scrollx [expr { ($xcanvas * $zdepth - $xwin) / [lindex $scrollregion 2]}]
-    set scrolly [expr { ($ycanvas * $zdepth - $ywin) / [lindex $scrollregion 3]}]
+    set x0 [lindex $scrollregion 0]
+    set y0 [lindex $scrollregion 1]
+    set W [expr [lindex $scrollregion 2] - [lindex $scrollregion 0]]
+    set H [expr [lindex $scrollregion 3] - [lindex $scrollregion 1]]
+    set scrollx [expr { ($xcanvas * $zdepth - $xwin - $x0) / $W}]
+    set scrolly [expr { ($ycanvas * $zdepth - $ywin - $y0) / $H}]
     $c xview moveto $scrollx
     $c yview moveto $scrolly
 }
@@ -90,8 +94,12 @@ proc ::pd_canvaszoom::setzoom {c steps} {
     set xwin [expr {[winfo pointerx $c] - [winfo rootx $c]}]
     set ywin [expr {[winfo pointery $c] - [winfo rooty $c]}]
     set scrollregion [$c cget -scrollregion]
-    set left_xview_pix [expr [lindex [$c xview] 0] * [lindex $scrollregion 2]]
-    set top_yview_pix [expr [lindex [$c yview] 0] * [lindex $scrollregion 3]]
+    set x0 [lindex $scrollregion 0]
+    set y0 [lindex $scrollregion 1]
+    set W [expr [lindex $scrollregion 2] - [lindex $scrollregion 0]]
+    set H [expr [lindex $scrollregion 3] - [lindex $scrollregion 1]]
+    set left_xview_pix [expr $x0 + [lindex [$c xview] 0] * $W]
+    set top_yview_pix [expr $y0 + [lindex [$c yview] 0] * $H]
     set xcanvas [expr ($xwin + $left_xview_pix) / $zdepth($c)]
     set ycanvas [expr ($ywin + $top_yview_pix) / $zdepth($c)]
 
--- a/tcl/pd_canvaszoom.tcl
+++ b/tcl/pd_canvaszoom.tcl
@@ -1,0 +1,86 @@
+package provide pd_canvaszoom 0.1
+
+namespace eval ::pd_canvaszoom:: {
+    namespace export zoominit
+}
+
+proc ::pd_canvaszoom::zoominit {mytoplevel {zfact {1.1}}} {
+    set c [tkcanvas_name $mytoplevel]
+    # save zoom state in a global variable with the same name as the canvas handle
+    upvar #0 $c data
+    set data(zdepth) 1.0
+    set data(idle) {}
+    # add mousewheel bindings to canvas
+    bind all <Control-Button-4> \
+        {event generate [focus -displayof %W] <Control-MouseWheel> -delta  1}
+    bind all <Control-Button-5> \
+        {event generate [focus -displayof %W] <Control-MouseWheel> -delta -1}
+    bind $c <Control-MouseWheel> "if {%D > 0} {zoom $c $zfact} else {zoom $c [expr {1.0/$zfact}]}"
+}
+
+proc zoom {c fact} {
+    upvar #0 $c data
+    # zoom at the current mouse position
+    set x [$c canvasx [expr {[winfo pointerx $c] - [winfo rootx $c]}]]
+    set y [$c canvasy [expr {[winfo pointery $c] - [winfo rooty $c]}]]
+    $c scale all $x $y $fact $fact
+    # save new zoom depth
+    set data(zdepth) [expr {$data(zdepth) * $fact}]
+    # update fonts only after main zoom activity has ceased
+    after cancel $data(idle)
+    set data(idle) [after idle "zoomtext $c"]
+}
+
+proc zoomtext {c} {
+    upvar #0 $c data
+    # adjust fonts
+    foreach {i} [$c find all] {
+        if { ! [string equal [$c type $i] text]} {continue}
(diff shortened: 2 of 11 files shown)


Ant1r:
Trying to update the line-width of lines and rectangles:

scrollzoom-linewidth-1

Spacechild1:
That's just awesome! Pd badly needs a proper zoom function...

Ant1r:
"Almost" good here!
At least edition is now possible in a zoomed canvas.
Still some bugs, with "undo" for example.
Scrolling to the zooming center still to be done too.

Ant1r:
I think it's about ready.
I've fixed some bugs, and added scrolling with mouse button 2 (scrollwheel button). Very pleasant...
Please test!

tcl/pd_canvaszoom.tcl
```diff
@@ -0,0 +1,333 @@
...
+    namespace export scalescript
+}
+
+proc ::pd_canvaszoom::zoominit {mytoplevel {zfact {1.1}}} {
+    set c [tkcanvas_name $mytoplevel]
+    # save zoom state in a global variable with the same name as the canvas handle
+    upvar #0 $c data
+    set data(zdepth) 1.0
+    set data(oldzdepth) 1.0
+    set data(idle) {}
```
umlaeute, line 15: using a global variable with a common name (the canvas handle) looks a bit brittle.

i think it would be better to use a namespaced variable

tcl/pd_canvaszoom.tcl
```diff
@@ -0,0 +1,333 @@
...
+            {event generate [focus -displayof %W] <Control-MouseWheel> -delta  1}
+        bind all <Control-Button-5> \
+            {event generate [focus -displayof %W] <Control-MouseWheel> -delta -1}
+    }
+    bind $c <Control-MouseWheel> "if {%D > 0} {zoom $c $zfact} else {zoom $c [expr {1.0/$zfact}]}"
+
+    # add button-2 bindings to scroll the canvas
+    bind $c <ButtonPress-2> {+ %W scan mark %x %y}
+    bind $c <B2-Motion> {+ %W scan dragto %x %y 1}
+}
```
umlaeute, line 29: to be honest i think i would prefer it, if continuous zoom wouldn't use the mousewheel, but instead just have <kbd>Ctrl</kbd>+<kbd>+</kbd> resp <kbd>Ctrl</kbd>+<kbd>-</kbd> (this is already the established shortcut for zoom-in/zoom-out; and the mousewheel is not very comfortable in a laptop (without a realy mouse))

sebshader: I prefer mousewheel to scroll the canvas
any of that can also be done in a tcl plugin probably though..

tcl/pd_canvaszoom.tcl
Ant1r: you're right, I lazily copied the code from https://wiki.tcl-lang.org/page/Canvas+zooming+using+mousewheel.
done now.

umlaeute:
>  the current zoom factor should be reported to the Pd-core side, in order to be saved to file.

i honestly don't think so.
this could be saved entirely on the GUI side via `::pd_guiprefs::write` (resp `::pd_guiprefs::read`

umlaeute:
i wonder whether the internal zoom-state should (also) be kept as a "step" value, so it's easier to reset to `zoom=0` (and when saving the zoom-amount to a preferences file, we can store simple integers and don't have to worry about precision)

this would of course totally break if the zoom-factor was changeable.

Ant1r:
> this could be saved entirely on the GUI side via ::pd_guiprefs::write (resp ::pd_guiprefs::read)

I was thinking that the zoom factor could be saved per-patch, I mean both zoom factor and scrolling position can be seen as part of the final presentation of the patch to the user.

> i wonder whether the internal zoom-state should (also) be kept as a "step" value

Maybe we could store something like a fractional log factor? e.g `zommfactor = pow(10, zoomsteps/100)` so:
```
-30 -> 1/2
 0  -> 1 
 30 -> 2
100 -> 10
```
etc... ?

Ant1r:
Actually `zoom factor = pow(2, steps/100)` is better in order to include exact powers of 2...

umlaeute:
yes, sthg like that.

> I was thinking that the zoom factor could be saved per-patch

hmm, interesting idea.

what i do love about this entire PR, is that it it's totally unrelated to Pd-core, and just does all the magic on the GUI side.
the only changes on the C side would be to get rid of the entire zoom-goo there.

Ant1r:
Yeah, thanks, that's what I like too. I'm dreaming of a world where Pd doesn't even know how do the objects look like on the patch...

If I had more time I'd love to try recoding my entire new knob in TCL, as an exercise. I wonder what change in the core would be needed, if any.

tcl/pd_canvaszoom.tcl
Ant1r: I find  <kbd>Ctrl</kbd>+<kbd>mousewheel</kbd> very convenient, and it's a very common shortcut.

I'm all for using <kbd>Ctrl</kbd>+<kbd>+</kbd>/<kbd>Ctrl</kbd>+<kbd>-</kbd> shortcuts together, but obviously that would mean getting rid of the old zoom functionality.

Ant1r:
> internal zoom-state should (also) be kept as a "step" value

done. Now the depth is expressed in steps, and the actual depth is `pow(2, steps / 100)`.
`zteps` is kept within the [-400 , 400] range, so `zdepth` remains within [1/16 , 16].

tcl/pd_canvaszoom.tcl
Ant1r: I reassigned the Ctrl+ and Ctrl- shortcuts, as well as Edit/Zoom* menu entries.
So now the legacy zoom function isn't reachable anymore...

Ant1r:
This PR has been brought up to date.

I would appreciate it being tested in a variety of circumstances, so that we can insist to have it included.
My own tests show no obvious problem.

Ant1r:
This PR has been updated to latest master, and debugged for MacOS and Windows.

Spacechild1:
I already gave many thumbs up, but I haven't come around to actually test it yet. I will put this on my list and then lobby relentlessly to have this included in Pd 0.55 :)

pingarelho:
Firstly, thank you for this PR, I have been using it extensively for the past month.

I noticed that when zooming in and out, and then going to the properties screen and clicking ok, the label and the value of a [nbx], for instance, can change a bit. Video below for reference.

https://github.com/pure-data/pure-data/assets/42239760/c4e68550-8849-4983-a795-ce3972679ad3

Ant1r:
Thanks for the feedback, I'll check that as soon as I can (which won't be for a good week).

Ant1r:
OK the line width bug is fixed too.
Does anyone see other bugs?

danomatika:
Is there a menu command / key binding to return to a normal 1:1 zoom? Ala CMD+0 in a web browser, on macOS at least.

Ant1r:
@danomatika all suggestions are welcome of course!
However I'd prefer to wait for Miller's decision before adding new features (which may need to touch more files).

Ant1r:
updated and rebased.

sebshader:
wouldn't it be more complete/consistent if the c-side zoom logic were taken out as well?

Ant1r:
Of course I'm all for doing it, but let's see if it's worth the effort...

Ant1r:
@millerpuckette have you had a chance to test this one?

Ant1r:
Hi,
I've updated this PR to current master, and changed the target branch to `develop` (hope it's a good idea?)

@millerpuckette could you please comment on this? thanks!

danomatika:
This looks pretty clean to me, good work. I would also like to add a MOD+0 binding to reset zoom to 1, similar to most browsers.

Ant1r:
@danomatika your wish is granted :-)

Ant1r:
@millerpuckette could you please give your opinion on this? thanks in advance!

Ant1r:
rebased to master.

@millerpuckette there may still be a few glitches, mainly: in certain situations, the mouse pointer can lose its position relative to the patch.
I would really need to know if there is any hope of this ever being taken into account before devoting more work to it... Thanks!

danomatika:
😉 It is being considered for 0.57, I  believe.

millerpuckette:
I'll have to look at it (particularly with an eye to what would happen 
if we move to
GTK or Qt :)

ATM I'm catching up on some other stuff but can start looking at things 
like this in a couple of weeks.

Ant1r:
Thanks Miller!

This zoom proposal is fully implemented in the GUI side (currently Tk), so:

- the bad news is that everything will be lost if/when we switch to another toolkit
- the good news is that GTK and Qt have everything we need to make canvas zooming (probably) straightforward!

Ant1r:
wait... recent changes in Pd broke this PR.
After attempting to merge current master to this PR, opening a GOP in a zoomed canvas causes the parent patch to refresh incorrectly.
Switching to draft for now.

millerpuckette:
This is a much better approach to zooming than the existing one.  And meanwhile, although I'm not convinced yet of the need/desirability of moving to GTK, I'm sure that it or any other GUI layer would want to handle zoom/scroll in the GUI and follow the (always changing) conventions about how to interpret scroll wheels and whatnot.  OTOH this has one major fragility (as evidenced in the way I apparently recently broke it) -  it has to know about the exact messages Pd is sending to the GUI so that it can munge them.  I think this will be less of a problem if TCL/TK ends up being phased out, but I'm not 100% sure that will ever really happen....

Ant1r:
Ok, it's working again.
Just had to "scale" the new pdtk_canvas commands (pdtk_canvas_create_rect etc.).
Only the "points" drawing of arrays still behaves wrongly. But I think it can be fixed a bit later.

However: I rebased to current 'master' branch but this PR is targeted to 'develop'; and 'develop' isn't up to date...
Could 'master' be merged to 'develop'? 
Or could someone with admin permission re-target this PR to 'master'?
Or should I make another PR targeted to 'master'?
So the diff would be minimal.

umlaeute:
thanks for the update. i've changed the target branch to `master`

Ant1r:
please hold on again...
I'm exploring other ways to filter incoming commands to the canvas.
If it succeeds, it will be more elegant and hopefully more robust.

Ant1r:
Here we go again ;-)

I built kind of a 'derived class' of the Tk canvas, that transparently scales every canvas command.
It should robustly support any future procedure addition ala `pdtk_canvas_create_rect`.

Moreover it should be more efficient than the previous implementation, because:
- only canvas commands are filtered (previously ALL the messages from the core were tested)
- the messages are now processed as-is, as lists, instead of being converted to strings.

umlaeute:
i found another issue: with Tcl/Tk>=8.7 (that is: 8.7 and 9.x) the mousewheel is scaled differently (at least on X11).
| Tcl/Tk | wheel delta |
|---|---|
| 8.6 | ±1 |
| 8.7a5 | ±120 |
| 9.0 | ±120 |

there's some code in Pd to handle this for scrolling, but apparently this only rescales the scrollbars rather than the delta: https://github.com/pure-data/pure-data/blob/203bc0941e3e13b8f4f1bfa9faae763c8ffc7fa6/tcl/pdtk_canvas.tcl#L147-L149

tcl/pd_canvaszoom.tcl
```diff
@@ -179,20 +179,20 @@ proc ::pd_canvaszoom::zoominit {mytoplevel} {
                 {event generate [focus -displayof %W] <Control-MouseWheel> -delta  1}
             bind all <Control-Button-5> \
                 {event generate [focus -displayof %W] <Control-MouseWheel> -delta -1}
-            bind $c <Control-MouseWheel> {::pd_canvaszoom::delayed_stepzoom %W %D}
```
Ant1r: this actually causes some trouble: the scrollbar is moving in the same time the zoom is changing, so the content is moving under the mouse pointer.

sidenote: the bug that prevents zooming a save-as-open subpatch doesn't exist with TclTk9.0... So this fix isn't needed for 9.0.

Ant1r:
> with Tcl/Tk>=8.7 the mousewheel is scaled differently

This fix in `::pd_canvaszoom::delayed_stepzoom`  seems to work:
```tcl
    if {[expr abs($steps)] >= 120} {set steps [expr $steps / 120]}
```
@umlaeute do you think it's an acceptable fix?

tcl/pd_canvaszoom.tcl
```diff
@@ -0,0 +1,396 @@
...
+    # LATER: probably move this to pd_bindings.tcl
+    switch -- $::windowingsystem {
+        "x11" {
+            bind all <Control-Button-4> \
+                {event generate [focus -displayof %W] <Control-MouseWheel> -delta  1}
+            bind all <Control-Button-5> \
+                {event generate [focus -displayof %W] <Control-MouseWheel> -delta -1}
+            bind ${mytoplevel} <Control-MouseWheel> {::pd_canvaszoom::delayed_stepzoom %W %D}
+            bind ${mytoplevel} <ButtonPress-2> {%W scan mark %x %y}
+            bind ${mytoplevel} <B2-Motion> {%W scan dragto %x %y 1}
```
umlaeute: what does the binding for the middle-button actually do?
afaiu, the idea is that you would be able to set the zoom-level by pressing the middle-button and then dragging, but it seems it doesn't *actually* have any effect here.

or is this about keeping the items under the mouse? (which i think is now broken as well...)

tcl/pd_canvaszoom.tcl
```diff
@@ -179,20 +179,20 @@ proc ::pd_canvaszoom::zoominit {mytoplevel} {
                 {event generate [focus -displayof %W] <Control-MouseWheel> -delta  1}
             bind all <Control-Button-5> \
                 {event generate [focus -displayof %W] <Control-MouseWheel> -delta -1}
-            bind $c <Control-MouseWheel> {::pd_canvaszoom::delayed_stepzoom %W %D}
-            bind $c <ButtonPress-2> {%W scan mark %x %y}
-            bind $c <B2-Motion> {%W scan dragto %x %y 1}
+            bind ${mytoplevel} <Control-MouseWheel> {::pd_canvaszoom::delayed_stepzoom %W %D}
```
Ant1r: this actually causes some trouble: the scrollbar is moving in the same time the zoom is changing, so the content is moving under the mouse pointer.

sidenote: the bug that prevents zooming a save-as-open subpatch doesn't exist with TclTk9.0... So this fix isn't needed for 9.0.

tcl/pd_canvaszoom.tcl
Ant1r: the idea is to *scroll* the canvas with middle-button, as in Gimp (which was my reference behavior).

umlaeute:
> ~~~tcl
>     if {[expr abs($steps)] >= 120} {set steps [expr $steps / 120]}
> ~~~
>
> @umlaeute do you think it's an acceptable fix?

how about this:
- let's just use `120` "everywhere" and scale it down in in `::pd_canvaszoom::stepzoom`
- on Windows, this is the default anyhow
- afaik, on X11 `<MouseWheel>` only ever generates events for Tk>=8.7 (where it reports `120`)
- when using the `<Button-4>` and `<Button-5>` fallback for older TclTk on X11, we can just report `120` instead of `1`
- on macOS, we would need to upscale

I tested this (and it seems to work), but my testing on macOS was rather casual (the room with a mac is currently in use, and it seems that the scrollball on the mouse was a bit broken; also, it seems that `<MouseWheel>` did not report anything on macOS/TclTk-9.x)

tcl/pd_bindings.tcl
```diff
@@ -88,11 +88,13 @@ proc ::pd_bindings::global_bindings {} {
...
-    bind all <$::modifier-Key-minus>       {::pd_menucommands::scheduleAction menu_send_float %W zoom 1}
-    bind all <$::modifier-Key-KP_Add>      {::pd_menucommands::scheduleAction menu_send_float %W zoom 2}
-    bind all <$::modifier-Key-KP_Subtract> {::pd_menucommands::scheduleAction menu_send_float %W zoom 1}
+    bind all <$::modifier-Key-equal>       {::pd_canvaszoom::stepzoom %W 1}
+    bind all <$::modifier-Key-plus>        {::pd_canvaszoom::stepzoom %W 1}
+    bind all <$::modifier-Key-minus>       {::pd_canvaszoom::stepzoom %W -1}
+    bind all <$::modifier-Key-KP_Add>      {::pd_canvaszoom::stepzoom %W 1}
+    bind all <$::modifier-Key-KP_Subtract> {::pd_canvaszoom::stepzoom %W -1}
+    bind all <$::modifier-Key-0>           {::pd_canvaszoom::setzoom %W 0}
+    bind all <$::modifier-Key-KP_0>        {::pd_canvaszoom::setzoom %W 0}
```
umlaeute: i haven't actually tested, but I guess that this will result in double-zooming on macOS. (that is: `::pd_canvaszoom::stepzoom %W 1` is executed twice in a row thus going directly from e.g. `100%` to `131%`, with no chance to reach the intermediate `114%` (which we have on X11 and Windows))

umlaeute: the solution is of course to use `::pd_menucommands::scheduleAction` resp. `::pd_canvaszoom::delayed_stepzoom`.

---

speaking of which: I wonder why we cannot use `::pd_menucommands::scheduleAction` in the first place (well, `delayed_stepzoom` also handles the accumulation, so that's one reason...)

Ant1r: why do you say it will be triggered twice? (I'll test on  a Mac ASAP).

Ant1r: do you mean both `bind all <$::modifier-Key-plus>` and `bind all <$::modifier-Key-KP_Add>` will trigger?

Ant1r:
I think most bugs have been fixed.
@umlaeute please confirm.

Ant1r:
> most bugs have been fixed

Once again... On linux, `<Control-MouseWheel>` was zooming 2 steps at a time.
That's because both the window and the canvas were bound to this event.
This is now fixed.

Ant1r:
@millerpuckette: with the help of @umlaeute we fixed last bugs and cleaned the code.

I've entirely rewritten the history of this PR, so that it's based on top of `master` and the number of commits is drastically reduced from >60 down to 5.

There are still a few bugs about canvas not getting focus sometimes, that will be automatically fixed once `develop` is merged.

Ant1r:
Hey it was merged, that's excellent news!
I'm about to prepare a PR that will remove all the old zoom system from the gui objects.

PR #1664, opened 2022-05-17
unqueue gatoms on free

### LINKED ISSUE #1663: deleting env~ with textbox in complicated patch crashes pd
Steps to reproduce (Pop_OS/Debian, Pd 0.52) 
1. While running a very large patch, create a patch with a noise~ object.
2. Feed noise~ into env~ or snapshot~ (activated with bang~).
3. Show the output with a number box.
4. Turn audio on.
5. Delete the cluster of objects or just the number box.
6. Pd will sometimes crash with this combination of events, freezing and requiring the user to kill wish.

sebshader:
also remove redundant gatom_redraw. closes #1663

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -780,7 +780,6 @@ void gatom_key(void *z, t_symbol *keysym, t_floatarg f)
             rtext_activate(t, 0);
         x->a_grabbed = 0;
         gatom_reborder(x);
-        gatom_senditup(x);
         gatom_redraw(&x->a_text.te_g, x->a_glist);
     }
     else if (c == '\n')
@@ -1171,6 +1170,7 @@ static void gatom_free(t_gatom *x)
         pd_unbind(&x->a_text.te_pd,
             canvas_realizedollar(x->a_glist, x->a_symfrom));
     gfxstub_deleteforkey(x);
+    sys_unqueuegui(x);
 }
 
 static void gatom_properties(t_gobj *z, t_glist *owner)


Lucarda:
tested on Windows and it works fine.

I used the `many_gatom.pd` from #1663 and the crash is gone.

Ant1r:
I'm not sure it's a good idea to call `sys_unqueuegui()` from `gatom_free()`. The crash only happens when the gatom is visible, so it doesn't make sense to change the behavior for non-visible gatoms.

The fact is gatom is also a `t_text`, and when a visible gatom is freed, `text_vis()` is called before `gatom_vis()`.
Here I could fix the crash with adding `sys_unqueuegui()` to both `text_vis()` is called before `gatom_vis()`. I think it's a better idea.

Can you test, and update your PR if it's good?

sebshader:
yes, and I assume you mean put it after the text_vis call in gatom_vis..

sebshader:
on second thought, I wonder if it's actually better to move it to gatom_vis. If we leave it on the queue the callback will trigger, but the list won't have to be searched for every element. (and the callback will check if gl_editor exists so it won't send anything over the pipe). It might be cheaper to just leave it on if it becomes unvised, since the likelihood of it becoming vised again in a very short time is low

sebshader:
maybe we should check in the gatom_free function to see if it's visible though

Ant1r:
The callback only originates from visible and reactive objects (i.e visible GUIs).
So `sys_unqueuegui()` should only be called when destroying a GUI that was visible.

I agree that general `t_text` objects shouldn't call `sys_unqueuegui()` when destroyed.
And I keep thinking it shouldn't be called either for an invisible object (as it would unnecessarily search the queue for this object).

That's why I think it's better to call it from `gatom_vis` than from `gatom_free`.

The following `gatom_vis()` fixes the crash, without any other modification:

``` C
static void gatom_vis(t_gobj *z, t_glist *glist, int vis)
{
    t_gatom *x = (t_gatom*)z;
    text_vis(z, glist, vis);
    if (*x->a_label->s_name)
    {
        if (vis)
        {
            int x1, y1;
            gatom_getwherelabel(x, glist, &x1, &y1);
            sys_vgui(
                "pdtk_text_new .x%lx.c {%lx.l label text} %f %f {%s } %d %s\n",
                glist_getcanvas(glist), x,
                (double)x1, (double)y1,
                canvas_realizedollar(x->a_glist, x->a_label)->s_name,
                gatom_fontsize(x) * glist_getzoom(glist), "black");
        }
        else sys_vgui(".x%lx.c delete %lx.l\n", glist_getcanvas(glist), x);
    }
    if(!vis) sys_unqueuegui(z); /* added line */
}
```

sebshader:
@Ant1r but searching the queue when an object is unvised would also search the queue unnecessarily.
If there are multiple objects on the queue at the time, the operation will be O(n^2) in the worst case I think. If the objects are left on the queue, they will simply be removed when the callback happens, which is linear as the queue gets stepped through.

Ant1r:
I think if the callback from GUI arrives after the object has been destroyed, it will be sent to a memory chunk that has been freed, so it will trigger the crash we are trying to fix.

That's why it's important to flush the gui queue from messages destinated to the object, before the object is destroyed.

sebshader:
@Ant1r yes destroyed, not unvised. That's why we can put it in the free method, not he unvis method. The callback can handle cases where the object isn't visible afaict (but not cases where it is nonexistent)

and if it is unqueued in the free method it will work because it happens in the same thread as the queue is traversed on

Ant1r:
Well what I'm trying to avoid is calling `sys_unqueuegui` when deleting invisible objects: imagine you have thousands of invisible gatoms, nested into an abs (or sub);  you don't want to search the queue for each of them when you delete the abs, whereas you know that none of them is likely to crash.

What you want to avoid is calling `sys_unqueuegui` when hiding objects that stay alive. While I think it's less common to hide a large number of visible gatoms than to delete a large number of hidden ones, you're right of course, this is not necessary either to try to unqueue them.

Since it's not possible from inside `gatom_vis` to know whether the gatom is going to be destroyed or not, I think you're right, there's no other place than gatom_free for unqueuing. But then you should test `if (glist_isvisible(glist))` before.

sebshader:
yeah, I was thinking something like that but wasn't sure if we need gobj_shouldvis or can get away with glist_isvisible. bc gobj_shouldvis seems kind of intensive itself (tho it gets checked every time something gets put on the queue in the 1st place).
I guess glist_isvisible seems like the best tradeoff, as long as it can catch GOP canvasses.

sebshader:
but also I'm worried about some weird condition occurring where an object that is on the queue and is non-vis at the time of deletion and its glist becomes visible (has a gl_editor) when the callback happens.
This could only happen if it was triggered while visible, then deleted shortly after it becomes non-vis, then the canvas it is on becomes visible again so it is a very narrow possibility but I still have to wrap my head around how canvas creation fits into this scheduling logic

Ant1r:
Hmm that's right...

That couldn't happen if unqueuing was done by gatom_vis... Note that it is what iemguis do btw.

sebshader:
it seems like if it was done in unvis the same amount of operations would happen as if it were in free because if it is in unvis it would have already been called in the past for objects when they were unvised if they are not currently visible, and called when the object is deleted if they are. so it seems like the same or fewer amount of total unqueues than if it is in unvis, even if we keep unqueueing non-visible objects when they're deleted.

Ant1r:
> in unvis the same amount of operations would happen as if it were in free

not for gatoms that have never been drawn (i.e most of them, usually).

sebshader:
my thinking was  that people vis and unvis objects much more than they delete objects. plus deleting objects requires other non-realtime-safe operations anyways.

Ant1r:
OK, I think now your solution is the best one. Unqueuing can't take so long most of the time anyway...
So I think this PR should be labelled as a bug fix, and why not merged to the current develop branch; what do you think @umlaeute ?

PR #1672, opened 2022-05-25
readsf~/writesf~ fixes

Spacechild1:
* fix regression with `[readsf~]`: sending an `[open(` message while a stream is running causes the next stream to fail silently. This didn't happen in Pd 0.51 and before.
* fix possible race condition in `readsf_perform` and `writesf_perform`
* minimize critical section in EOF branch in `readsf_perform`
* cosmetic fixes

The first issue has bitten me recently in a rehearsal. While looking at the code I found a few other possible race conditions, hence this PR.

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -1919,12 +1919,12 @@ static void *readsf_child_main(void *zz)
         {
             if (sf.sf_fd >= 0)
             {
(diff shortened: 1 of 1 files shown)


danomatika (approved):
Seems fine to me. I tried not to change the algorithm while doing the update, so I didn't honestly dig into understanding it in detail.

Issue #1674, opened 2022-05-28
route still outputs a list selector even if the output is a float


porres:
this bit me cause I have an object with a list and a float method but if I have something like 

"2 3" into [route 2], what comes out is "list 3" and the object doesn't think it is a float.

or am i doing something wrong?

Spacechild1:
A float message is really the same thing as a list message with a single float atom. 

> this bit me cause I have an object with a list and a float method 

This doesn't sound right. If you have a list method, why would you need a float method? The list method already handles float messages. I mean, you *can* also have a float method, but you shouldn't attempt to differentiate between float messages and single-element-lists!

porres:
I wondered if this happened to me before and we ended up discussing it here, but I couldn't/can't find the issue... now that you replied this my belief is stronger...

> A float message is really the same thing as a list message with a single float atom.

how so? Why then I need to pass the output of [route] into [float] so I can get a proper float output into my external (that's the solution I've found)? To me this means both messages are different.

porres:
moreover, would you oppose to having [route] output a 'proper float'?

Spacechild1:
> Why then I need to pass the output of [route] into [float] so I can get a proper float output into my external

because you're external is broken (as I tried to explain above).

Spacechild1:
For example, `[route]` treats a list with a single float atom as a float message. Try this:

![grafik](https://user-images.githubusercontent.com/16126632/170897162-4214ecf0-19ec-4f72-b8e0-fbecdab43171.png)

> moreover, would you oppose to having [route] output a 'proper float'?

I don't mind, but I don't see the need. For any well-behaving object it wouldn't make a difference.

porres:
what you seem to say to me is that a float message is in fact different from a list message with a single float element, hence the list and float methods will pick them up separately, but we, as developers, should know that and work it out...

I can't see how a list message with one element is the same as a float message

anyway, this also seems like something to be cleared out in the documentation somewhere

porres:
> I don't mind, but I don't see the need. For any well-behaving object it wouldn't make a difference.

I also don't see why it should output a list with a single element, as if there's a use case for it. It looks like it's a misbehave to me, even though we should watch for it...

Spacechild1:
> what you seem to say to me is that a float message is in fact different from a list message with a single float element, hence the list and float methods will pick them up separately, but we, as developers, should know that and work it out...

[...]

PR #1677, opened 2022-05-29
fix float output for [route]

porres:
closes https://github.com/pure-data/pure-data/issues/1674
also closes https://github.com/pure-data/pure-data/issues/1678


porres:
I will reboot this and try to make it less confusing

Issue #1682, opened 2022-05-30
an empty list clears the list box instead of working as a bang


porres:
as discussed in https://github.com/pure-data/pure-data/issues/1681#issuecomment-1141497537

The code for list box intentionally clears the list because of a comment in 

https://github.com/pure-data/pure-data/blob/18c9695ba4de543c51c2c013b45b99b8f1c35a19/src/g_text.c#L744

At least that's how I interpreted it... there's the 'real' bang method for gatoms but an empty list (which would be the same as a bang) clears the list box instead of behaving like a bang. I can see that and how it would be useful to report us it received an empty list. I can see that as a feature.

@Spacechild1 says this is a bug and should never happen, even if it was intended by @millerpuckette , in other words, this would be an illegal feature that needs to be prevented from happening (aka: a bug). I can also see that, so I will go ahead and make a PR to treat empty lists as bangs and let's see where this goes...

Spacechild1:
Just to reiterate why I think this is a bug:

Generally, Pd objects are supposed to treat empty lists just like bangs. I don't see why the list gatom should be an exception. After all, users can't really control if an object outputs an actual "bang" message or an empty list, as this is an implementation detail.

PR #1683, opened 2022-05-30
fix empty list input for list box (should be treated as a bang)

porres:
This closes https://github.com/pure-data/pure-data/issues/1682

Currently, if you send an empty list to a list box it has a distinct behaviour than a bang as it sets and outputs an empty list. This was considered a bug as discussed in https://github.com/pure-data/pure-data/issues/1682


porres:
closing for a new one as this has conflicts

umlaeute:
> closing for a new one as this has conflicts

it would be better to fix conflicts (which is a useful skill when working with git)

porres:
hmmm, I guess I finally learned how to do it :)

PR #1693, opened 2022-06-13
[WIP] GUI abs(traction) layer 🏋

danomatika:
This is research/brainstorming PR for refactoring the communication between the Pd core and the GUI.

_Note: This is **not** a full, open rewrite away from Tcl/Tk but a refactor and abstraction to allow for implementing custom GUIs without worrying about Tcl-specifics. Also, this current work maintains the editing logic within the core._

**Please keep the discussion focused on technical considerations. This is not meant to be a GUI feature wish list!**

Goals:
1. https://github.com/pure-data/pure-data/discussions/1694
    Refactor away (or at least centralize) Tcl/Tk string formatting within the core
3. https://github.com/pure-data/pure-data/discussions/1695
    Move GUI API drawing specifics out of the core into the GUI (ex: "draw a bng" as opposed to "draw a box, draw a circle, draw an inlet, draw an outlet, ...") 
5. Develop/implement a communication API (which could also be used by externals)
6. Allow for message communication via callback as well as network for use in libpd
7. Maintain modularity for the layer (ie. refactor into separate file that could extended or compiled optionally?) 
8. Accomplish the points above without breaking anything... 🤭

Possible workflow (step by step without large changes at once):
* [ ] Refactor TCL-specific string handling & escaping into single function / interface 
* [ ] Refactor *one* GUI object or drawing routine, ie. something simple like "bng"
* [ ] Finalize communication specs
* [ ] Refactor other GUI objects
* [ ] Create an example GUI external?
* [ ] ...

References:
* [Purr Data GUI messaging spec](https://git.purrdata.net/jwilkes/purr-data#gui-messaging-specification) (Jonathan Wilkes, et al.)
  - Ex: bng core g_bang.c: https://git.purrdata.net/jwilkes/purr-data/-/blob/master/pd/src/g_bang.c#L60
  - Ex: bng gui pdgui.js: https://git.purrdata.net/jwilkes/purr-data/-/blob/master/pd/nw/pdgui.js#L3639
* ["3 commits as a PoC of the sort of changes for a GUI-independent backend"](https://github.com/giuliomoro/pure-data-1/tree/guin) (Guilo Moro)

Questions / Considerations (so far):
* What about dialogs?
* We should provide a set of basic drawing routines for data structures and externals.
* How to optimize array draw communication, etc?


Spacechild1:
@danomatika I see!

Spacechild1:
I agree with the goals and workflow suggestion!

As someone who wants to write a (fast) alternative desktop GUI, I would like to emphasize the importance of goal number 2:
> Move GUI API drawing specifics out of the core into the GUI (ex: "draw a bng" as opposed to "draw a box, draw a circle, draw an inlet, draw an outlet, ...")

With GUI frameworks like Qt you would not draw Pd objects by issuing individual canvas drawing commands. Instead, you would instantiate a single (custom) Qt widget that knows how to draw itself. This would be much more efficient because a) there's only a single command to send to the GUI (= less traffic) and b) there will be far fewer objects living in the canvas (`QGraphicsScene`).

umlaeute:
this pseudo-PR got accidentally closed as a side-effect of merging #1696 (:tada: for that).

it's probably better to use github [discussions](https://github.com/pure-data/pure-data/discussions)  and [projects](https://github.com/pure-data/pure-data/projects) instead of such a meta PR.

PR #1633, opened 2022-04-20
iemgui: merge radio and slider code

umlaeute:
this merges the implementation of sliders and radiobuttons.

the vertical/horizontal variants of these widgets are almost identical (from an implementation POV), but having them in two distinct files each, makes bug-fixing a brittle task (in general you keep forgetting doing the same thing in both files).

so this PR get's rid of a lot of duplicate code in the implementation of hslider/vslider and hradio/vradio, by merging them into `g_slider.c` and `g_radio.c` (no more `g_*dial.` :stuck_out_tongue: )

since the merge changed a lot of lines (mostly renaming stuff, but also adding code for the horizontal resp. vertical cases), this breaks some parts of #1628.
in return i have started to implement the Tcl/Tk `canvas move` semantics for faster displacements of the iemguis (similar to @porres solution, this breaks the ABI of the iemgui objects, but (unlike #1628) it keeps API compatibility. i think there's only a single external out there that uses `g_all_guis.h` (@tmusil's *iemgui*), but i really think that these headers should never have been made public in the first place).

this PR was inspired by https://github.com/pure-data/pure-data/issues/1344#issuecomment-937001815 (@millerpuckette 's request to unify sliders and radiobuttons)

- Closes: #632 
- Closes: #920 
- Closes: #992 
- Closes: #1021 
- Closes: #1022 
- Closes: #1042 
- Closes: #1043 
- Closes: #1121 
- Closes: #1583 
- Closes: #1629 
- Closes: #1637 
- Closes: #1638 
- Closes: #1646

PR #1632, opened 2022-04-19
stable development branch

umlaeute:
(just merged, and here it is again)

this is the permanent branch develop that only gets curated, small, no-brainer changes that can be easily merged into master at any time.
(as a general rule-of-thumb we shouldn't include changes to Pd-files in this PR, as it is just too easy to end up with unresolvable file conflicts)

it supersedes https://github.com/pure-data/pure-data/pull/1586

---

fixes so far (this list should be manually updated whenever new fixes are pushed):

- GUI-regressions
  - Closes: #1751 crasher on Windows
  - fix regression that broke "clicking on an error message to find the culprit" functionality (see https://github.com/pure-data/pure-data/pull/1696#issuecomment-1236074692)
  - Closes: https://github.com/pure-data/pure-data/issues/1740
  - Closes: #1742 
  - Closes: #1743 
  - Closes: #1744
  - Closes: #1745
- GUI
  - fix missing context-menu for tables created via `[array define]` (Closes: #1705)
  - Closes: #1737
  - Closes: #1747
  - Closes: https://github.com/pure-data/pure-data/issues/1702
  - Closes: #85
  - Closes: https://github.com/pure-data/pure-data/issues/1708 (and it's accompanying regression Closes: https://github.com/pure-data/pure-data/issues/1754)

- Core/extra
  - Closes: https://github.com/pure-data/pure-data/issues/1720
  - Closes: https://github.com/pure-data/pure-data/pull/1728
- build-system
  - Closes: https://github.com/pure-data/pure-data/issues/1722

<details>
  <summary>fixes that have already been merged via other PRs</summary>

parts of the `develop` branch where also in other PRs (notably #1696), so they have already been closed.
i keep this legacy list here just for completeness sake...

the `develop` branch still contains some goodies :-)

- GUI
  - Fix segfault when deleting all data via a scalar's property window
    Closes: #1595 
  - Unqueue gatoms on free (Closes: #1663)
  - [vh]radio: fix errors when changing the number of cells on invisible canvas
    Closes: #1583 
  - fix regression that broke "clicking on an error message to find the culprit" functionality (see https://github.com/pure-data/pure-data/pull/1696#issuecomment-1236074692)
  - deselect gatoms when creating new objects (Closes: #1656)
  - fix autopatching offset when zoomed in (Closes: #1646)
  - fix missing context-menu for tables created via `[array define]` (Closes: #1705)
- Core:
  - properly handle `[open(` messages in `[readsf~]`/`[writesf~]` while stream is running
[...]

PR #1696, opened 2022-06-14
Refactor TCL-specific string handling into pdgui_vmess()

umlaeute:
this is a first draft.

it introduces new functions in the `pdgui_*` namespace for communicating with the Pd-GUI.

## `pdgui_vmess`

`pdgui_vmess` is a replacement for `sys_vgui` used to to generate (and send) messages  from core to GUI, taking care of string escaping suitable for Tcl/Tk.

e.g:

```C
void pdgui_vmess(const char*dest, const char*format, ...);
/* ... */
pdgui_vmess("guiproc", "is", 3, "foo bar");
```

will create the string `guiproc 3 {foo bar}` which is a valid Tcl/Tk command.

the function is deliberately called `pdgui_vmess`, in order to not clash with PurrData's `gui_vmess` (which has the same signature, but the type-ids are not compatible)

### destination

the 1st argument is a "receiver" on the GUI side. for Tcl/Tk this is typically a `proc`.
The assumption is that this receiver is constant, which doesn't map nicely to Tk's object-orientation (`.foo configure -width 3`).
For now, this is solved by passing `NULL` as the first argument (which will skip it in the generated string), like so:

```C
pdgui_vmess(0, "crr iiii ri",
                glist_getcanvas(glist), "create", "rectangle",
                10, 10,  50, 50,
                "-width", 1);
```

on the long run, i think using `NULL` as the destination should be avoided, and instead fixed-named `proc`s should be used:

```C
pdgui_vmess("::pd_canvas::create", "cr iiii ri",
                glist_getcanvas(glist), 
                "rectangle",
                10, 10,  50, 50,
                "-width", 1);
```

### type-IDs
the 2nd argument is a format-specifier (similar to the one used by `binbuf_addv()`, but *NOT* the same).
the type-ID determines how the following varargs are interpreted.

valid type-IDs so far are:

| id | args | description |
|---|---|---|
|`\0`| (none) | terminates the parsing of the varargs |
| ` `| (none) | space can be used to group the format string; it has no effect on the output formatting
|`i` | `int` | **i**nteger number (due to the nature of C varargs, all ints (e.g. `char`) are automatically promoted to `int`)
|`f` | `double` | **f**loating point number (due to the nature of C varargs, all floats are automatically promoted to `double`)
|`F` | `int` `t_float*` | *a*rray of **f**loat point values (unlike `f` this has a type of `t_float`!) |
|`s`| `const char*`| a **s**tring (that might need escaping)  |
|`S`| `int` `char**`| a **s**tring *a*rray (where each string needs escaping)  |
|`r`| `const char*`| a **r**aw string (that will never be escaped)  |
|`R`| `int` `char**`| an *a*rray of **r**aw strings (that will not be escaped)  |
|`p`| `int` `const char*` | a string that is not `\0`-terminated (and instead has an extra size-field); aka **p**ascal-string |
|`a`|`int` `t_atom*` | a list of atoms (so the GUI recevies them as multiple arguments)
|`A`|`int` `t_atom*` | an *a*rray of **a**toms
|`w`|`int` `t_word*` | a list of float**w**ords  (so the GUI receives them as multiple arguments) |
|`W`|`int` `t_word*` | an *array* of float**w**ords |
|`k`| `int` | a 24bit color |
|`m` | `t_symbol*s`, `int`, `t_atom*` | a Pd-message (for the GUI to send back) |
|`c`| `t_canvas*`| the ID of a TclTk canvas associated with the given canvas |
|`C`| `int` `t_canvas*`| an *array* for canvases |
|`^`| `t_canvas*`| the ID of the (**t**oplevel) window associated with the given canvas|
|`o`|`t_object`| the ID of the **o**bject |

- arrays are capitalized.
- there are both "arrays" and "lists": *arrays* are sent to the GUI as a single argument containing all the values, whereas *lists* are a shorthand for a variable number of arguments of the same type. e.g.
  | invocation in C | translates to Tcl/Tk | note |
  |---------------|----------|-------|
  | `pdgui_vmess(foo, "w", 3, 1, 2, 3)` | `foo 1 2 3` | list |
  | `pdgui_vmess(foo, "W", 3, 1, 2, 3)` | `foo {1 2 3}` | array | 

- the `^` (toplevel window) ID is currently required by how Pd talks to Tcl/Tk.
in the future (#1695), this should not be needed any more and instead be replaced with a reference to the `c`anvas that is contained by the window (thus the use of the `^` type-id is a good indicator about where there's work to be down)

- the `o` (Pd-object) ID is currently only used by `logpost` and friends (and I would like to keep it like that); it probably should go away.

- Purr-Data has a generic `x` type-id for pointers. we cannot use that right now, as we need to format the pointer differently based on the context (mostly: window vs canvas).

##### ideas for more type-ids (none of them implemented)

|description | id | notes |
|---|---|---|
| float32 array | `1` | might be interesting for libpd (so the app can stay oblivious of `t_float`) |
| float64 array | `2` | see above |
| ~symbol, symbolarray~ |~`y`, `Y`~ | ~so the caller doesn't have to do all the `->s_name`; always escaped~ | 
| ~binbuf~ | ~`b`, `B`~ | ~alternative to the *atom array*~ |
| format continuation | `+` | only  valid at the *end* of the format string; indicates that there's another format string with associated types coming (so you can write `pdgui_vmess(0, "ii+", 1, 2, "ii" 3, 4);` as an equivalent to `pdgui_vmess(0, 'iiii', 1, 2, 3, 4);`

## `pdgui_stub_vnew`/`pdgui_stub_deleteforkey`

These are replacements for the `gfxstub_new` resp `gfxstub_deleteforkey` functions.
`pdgui_stub_vnew` adopts the type-ID syntax of `pdgui_vmess` (and shuffles the arguments to indicate the argument order on the GUI side). the `pdgui_stub_deleteforkey` is practically identical to the original `gfxstub_...` function (it's only there for  a consistent naming scheme):

```C
void pdgui_stub_vnew(t_pd *owner, const char* destination, void *key, const char* fmt, ...);
/* ... */
pdgui_stub_vnew(&x->gl_pd, "pdtk_array_dialog", x, "siii", arraybuf, 100, 3, 1);
/* ... */
pdgui_stub_deleteforkey(x);
```

- destination: "receiver" on the GUI-side (as with `pdgui_vmess`)
- fmt: list of type-IDs (as with `pdgui_vmess`)

### issues so far
- needs more testing
- ~not optimized at all :-)~
- ~array handling tries to guess the required size to simplify reallocation~
  - ~the guess tries to be conservative, so we are probably over-allocating~
  - ~if the guess underestimates the final size, this will fail badly~
- ~no escaping of whacko strings has been implemented yet~

~regarding the escaping, i wonder whether we should provide different type-IDs for strings that need escaping (beyond encapsulating them in curlies) and those that don't.
it just seems overkill to escape constant strings...~

### regressions
(regressions that are fixed will get a  :ballot_box_with_check: )

- [x] the editmode-indicator in the window-title has become `\[edit\]` (with backslashes)
- [x] first MIDI-device get's munged in the dialog
       (this is actually caused by some refactoring i did before switching to `pdgui_vmess()`: i just couldn't stand the https://github.com/pure-data/pure-data/blob/18c9695ba4de543c51c2c013b45b99b8f1c35a19/src/s_midi.c#L711-L713 so i turned this into a single list-assignment, forgetting the leading `none` device :facepalm: 
      on the long run, we should just get rid of the `none` device  and use a list of available devices.
- [x] dialogs that take floats will now show the full number with commas (e.g. `1.0000000` instead of `1`)
      this is not very nice, but i'd rather have the GUI take care of formatting the numbers correctly.
      (solved for now by switching to `%g` for formatting single floats)
- [x] Pd-messages (such as sent to `pdtk_check`) are escaped overzealously, and not un-escaped before sending the message back to Pd, resulting in errors like `canvas: no method for '{dirty}'` (canvas has only a method of `dirty` not `{dirty}`)
     (solved for now by not escaping the atoms before sending them (as it used to be); on the long run, the GUI should properly unescape these messages)
- [x] Pd-patches that have visible (opened) subpatches (on disk), throw a Tcl/Tk error when opened.

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -482,7 +482,7 @@ void iemgui_label(void *x, t_iemgui *iemgui, t_symbol *s)
     pdgui_strnescape( lab_escaped, MAXPDSTRING, iemgui->x_lab->s_name, strlen(iemgui->x_lab->s_name) );
 
     if(glist_isvisible(iemgui->x_glist) && iemgui->x_lab != old)
-        sys_vgui(".x%lx.c itemconfigure %lxLABEL -text [::pdtk_text::unescape \"%s \"] \n",
+        sys_vgui(".x%lx.c itemconfigure %lxLABEL -text [::pdtk_text::unescape \"%s\"] \n",
                  glist_getcanvas(iemgui->x_glist), x,
                  strcmp(s->s_name, "empty")?lab_escaped:"");
 }
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -525,9 +525,17 @@ void iemgui_label_font(void *x, t_iemgui *iemgui, t_symbol *s, int ac, t_atom *a
         f = 4;
     iemgui->x_fontsize = f;
     if(glist_isvisible(iemgui->x_glist))
-        sys_vgui(".x%lx.c itemconfigure %lxLABEL -font {{%s} -%d %s}\n",
-                 glist_getcanvas(iemgui->x_glist), x, iemgui->x_font,
-                 iemgui->x_fontsize*zoom, sys_fontweight);
+    {
+        char tag[128];
+        t_atom fontatoms[3];
+        sprintf(tag, "%lxLABEL", x);
+        SETSYMBOL(fontatoms+0, gensym(iemgui->x_font));
+        SETFLOAT (fontatoms+1, -iemgui->x_fontsize*zoom);
+        SETSYMBOL(fontatoms+2, gensym(sys_fontweight));
+        pdgui_vmess(0, "crs rA",
+            glist_getcanvas(iemgui->x_glist), "itemconfigure", tag,
+            "-font", 3, fontatoms);
+    }
 }
 
 void iemgui_size(void *x, t_iemgui *iemgui)
@@ -857,4 +865,3 @@ external GUI object uses obsolete Pd function iemgui_all_colfromload()");
         iemgui->x_lcol = iemgui_color_hex[bflcol[2]];
     }
 }
-
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -482,9 +482,13 @@ void iemgui_label(void *x, t_iemgui *iemgui, t_symbol *s)
     pdgui_strnescape( lab_escaped, MAXPDSTRING, iemgui->x_lab->s_name, strlen(iemgui->x_lab->s_name) );
 
     if(glist_isvisible(iemgui->x_glist) && iemgui->x_lab != old)
-        sys_vgui(".x%lx.c itemconfigure %lxLABEL -text [::pdtk_text::unescape \"%s\"] \n",
-                 glist_getcanvas(iemgui->x_glist), x,
-                 strcmp(s->s_name, "empty")?lab_escaped:"");
+    {
+        char tag[128];
+        sprintf(tag, "%lxLABEL", x);
+        pdgui_vmess("pdtk_text_set", "css",
+                  glist_getcanvas(iemgui->x_glist),
+                  tag, strcmp(s->s_name, "empty")?s->s_name:"");
+    }
 }
 
 void iemgui_label_pos(void *x, t_iemgui *iemgui, t_symbol *s, int ac, t_atom *av)
@@ -493,10 +497,14 @@ void iemgui_label_pos(void *x, t_iemgui *iemgui, t_symbol *s, int ac, t_atom *av
     iemgui->x_ldx = (int)atom_getfloatarg(0, ac, av);
     iemgui->x_ldy = (int)atom_getfloatarg(1, ac, av);
     if(glist_isvisible(iemgui->x_glist))
-        sys_vgui(".x%lx.c coords %lxLABEL %d %d\n",
-                 glist_getcanvas(iemgui->x_glist), x,
-                 text_xpix((t_object *)x, iemgui->x_glist) + iemgui->x_ldx*zoom,
-                 text_ypix((t_object *)x, iemgui->x_glist) + iemgui->x_ldy*zoom);
+    {
+        char tag[128];
+        sprintf(tag, "%lxLABEL", x);
+        pdgui_vmess(0, "crs ii",
+            glist_getcanvas(iemgui->x_glist), "coords", tag,
+            text_xpix((t_object *)x, iemgui->x_glist) + iemgui->x_ldx*zoom,
+            text_ypix((t_object *)x, iemgui->x_glist) + iemgui->x_ldy*zoom);
+    }
 }
 
 void iemgui_label_font(void *x, t_iemgui *iemgui, t_symbol *s, int ac, t_atom *av)
@@ -646,19 +654,10 @@ void iemgui_newzoom(t_iemgui *iemgui)
 
 void iemgui_properties(t_iemgui *iemgui, t_symbol **srl)
 {
-    char label[MAXPDSTRING];
-
     srl[0] = iemgui->x_snd;
     srl[1] = iemgui->x_rcv;
-
-    strcpy(label, iemgui->x_lab->s_name);
-    pdgui_strnescape(label, MAXPDSTRING,
-                    iemgui->x_lab->s_name, strlen(iemgui->x_lab->s_name));
-    srl[2] = gensym(label);
-
-    iemgui_all_dollar2raute(srl);
+    srl[2] = iemgui->x_lab;
     iemgui_all_sym2dollararg(iemgui, srl);
-    iemgui_all_put_in_braces(srl);
 }
 
 int iemgui_dialog(t_iemgui *iemgui, t_symbol **srl, int argc, t_atom *argv)
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -152,6 +152,7 @@ pd_SOURCES_core = \
     m_sched.c \
     s_audio.c \
     s_inter.c \
+    s_inter_gui.c \
     s_loader.c \
     s_main.c \
     s_net.c \
--- a/src/d_resample.c
+++ b/src/d_resample.c
@@ -103,7 +103,7 @@ t_int *upsampling_perform_linear(t_int *w)
 
 void resample_init(t_resample *x)
 {
-  x->method=0;
+  x->method=-1;
 
   x->downsample=x->upsample=1;
 
@@ -140,7 +140,7 @@ void resample_dsp(t_resample *x,
       return;
     }
     switch (method) {
-    default:
+    default: /* always zero padding */
       dsp_add(downsampling_perform_0, 4, in, out, (t_int)(insize/outsize), (t_int)insize);
     }
 
@@ -151,10 +151,10 @@ void resample_dsp(t_resample *x,
       return;
     }
     switch (method) {
-    case 1:
+    case 1: /* sample and hold */
       dsp_add(upsampling_perform_hold, 4, in, out, (t_int)(outsize/insize), (t_int)insize);
       break;
-    case 2:
+    case 2: /* linear interpolation */
       if (x->bufsize != 1) {
         t_freebytes(x->buffer, x->bufsize*sizeof(*x->buffer));
         x->bufsize = 1;
@@ -162,7 +162,7 @@ void resample_dsp(t_resample *x,
       }
       dsp_add(upsampling_perform_linear, 5, x, in, out, (t_int)(outsize/insize), (t_int)insize);
       break;
-    default:
+    default: /* zero padding */
       dsp_add(upsampling_perform_0, 4, in, out, (t_int)(outsize/insize), (t_int)insize);
     }
   }
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -494,7 +494,7 @@ static void garray_arrayviewlist_fillpage(t_garray *x,
     if(page < 0)
         page = 0;
 
-    sys_vgui("::dialog_array::listview_setpage {%s} %d %d %d\n",
+    pdgui_vmess("::dialog_array::listview_setpage", "s iii",
         x->x_realname->s_name,
         page, maxpage+1, pagesize);
 
@@ -509,7 +509,7 @@ static void garray_arrayviewlist_fillpage(t_garray *x,
     }
     sys_vgui("\n");
 
-    sys_vgui("::dialog_array::listview_focus {%s} %d\n",
+    pdgui_vmess("::dialog_array::listview_focus", "si",
              x->x_realname->s_name,
              topItem);
 }
@@ -537,7 +537,7 @@ static void garray_arrayviewlist_new(t_garray *x)
 static void garray_arrayviewlist_close(t_garray *x)
 {
     x->x_listviewing = 0;
-    sys_vgui("pdtk_array_listview_closeWindow {%s}\n",
+    pdgui_vmess("pdtk_array_listview_closeWindow", "s",
              x->x_realname->s_name);
 }
 /* } jsarlo */
@@ -783,7 +783,7 @@ void garray_redraw(t_garray *x)
     else
     {
       if (x->x_listviewing)
-        sys_vgui("pdtk_array_listview_fillpage {%s}\n",
+          pdgui_vmess("pdtk_array_listview_fillpage", "s",
                  x->x_realname->s_name);
     }
     /* } jsarlo */
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -512,7 +512,6 @@ static void garray_arrayviewlist_fillpage(t_garray *x,
 
 static void garray_arrayviewlist_new(t_garray *x)
 {
-    char cmdbuf[200];
     int size=0;
     t_word*data=0;
 
@@ -521,11 +520,11 @@ static void garray_arrayviewlist_new(t_garray *x)
         return;
     }
     x->x_listviewing = 1;
-    sprintf(cmdbuf,
-            "pdtk_array_listview_new %%s {%s} %d\n",
-            x->x_realname->s_name,
-            0);
-    gfxstub_new(&x->x_gobj.g_pd, x, cmdbuf);
+
+    pdgui_stub_vnew(&x->x_gobj.g_pd,
+        "pdtk_array_listview_new", x,
+        "si",
+        x->x_realname->s_name, 0);
 
     garray_arrayviewlist_fillpage(x, 0, 0);
 }
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -265,7 +265,7 @@ static void garray_fittograph(t_garray *x, int n, int style)
             glist_redraw(gl);
         }
             /* close any dialogs that might have the wrong info now... */
-        gfxstub_deleteforkey(gl);
+        pdgui_stub_deleteforkey(gl);
     }
 }
 
@@ -347,21 +347,21 @@ void canvas_menuarray(t_glist *canvas)
 {
     t_glist *x = (t_glist *)canvas;
     int gcount;
-    char cmdbuf[200], arraybuf[80];
+    char arraybuf[80];
     for (gcount = 1; gcount < 1000; gcount++)
     {
         sprintf(arraybuf, "array%d", gcount);
         if (!pd_findbyclass(gensym(arraybuf), garray_class))
             break;
     }
-    sprintf(cmdbuf, "pdtk_array_dialog %%s array%d 100 3 1\n", gcount);
-    gfxstub_new(&x->gl_pd, x, cmdbuf);
+    pdgui_stub_vnew(&x->gl_pd,
+        "pdtk_array_dialog", x, "siii",
+        arraybuf, 100, 3, 1);
 }
 
     /* called from graph_dialog to set properties */
 void garray_properties(t_garray *x)
 {
-    char cmdbuf[200];
     t_array *a = garray_getarray(x);
     t_scalar *sc = x->x_scalar;
     int style = template_getfloat(template_findbyname(sc->sc_template),
@@ -371,14 +371,12 @@ void garray_properties(t_garray *x)
 
     if (!a)
         return;
-    gfxstub_deleteforkey(x);
-        /* create dialog window.  LATER fix this to escape '$'
-        properly; right now we just detect a leading '$' and escape
-        it.  There should be a systematic way of doing this. */
-    sprintf(cmdbuf, "pdtk_array_dialog %%s {%s} %d %d 0\n",
-            x->x_name->s_name, a->a_n, x->x_saveit +
-            2 * filestyle);
-    gfxstub_new(&x->x_gobj.g_pd, x, cmdbuf);
+    pdgui_stub_deleteforkey(x);
+    pdgui_stub_vnew(&x->x_gobj.g_pd,
+        "pdtk_array_dialog", x,
+        "siii",
+        x->x_name->s_name,
+        a->a_n, x->x_saveit + 2 * filestyle, 0);
 }
 
     /* this is called back from the dialog window to create a garray.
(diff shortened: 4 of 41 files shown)


umlaeute:
i don't think i will convert the iemgui's to the ´pdgui_vmess()` syntax (except for the `cnv` that i did as an exercise).
it will clash heavily with #1633, and i will try to only do this work once.

also: maybe somebody dislikes the current approach and has a better idea...

src/g_all_guis.c
```diff
@@ -517,9 +525,17 @@ void iemgui_label_font(void *x, t_iemgui *iemgui, t_symbol *s, int ac, t_atom *a
...
     if(glist_isvisible(iemgui->x_glist))
-        sys_vgui(".x%lx.c itemconfigure %lxLABEL -font {{%s} -%d %s}\n",
-                 glist_getcanvas(iemgui->x_glist), x, iemgui->x_font,
-                 iemgui->x_fontsize*zoom, sys_fontweight);
+    {
+        char tag[128];
+        t_atom fontatoms[3];
+        sprintf(tag, "%lxLABEL", x);
+        SETSYMBOL(fontatoms+0, gensym(iemgui->x_font));
+        SETFLOAT (fontatoms+1, -iemgui->x_fontsize*zoom);
```
Lucarda, line 533: i think this line should be `SETFLOAT (fontatoms+1, iemgui->x_fontsize*zoom);`

there is a "minus" sign `-iemgui->...` I think is a typo

umlaeute: It's indeed intentional, as negative numbers are Tcl/Tk's way to express absolute (non-scalable) fontsizes.
the original code added the `-` in the format-string...

umlaeute:
plan for renaming type-ID's:


| old | new | description |
|---|---|---|
|`p` | `x` | the "pd-object" is currently only used for `logpost`, where the only relevant thing is that it can be associated with an object
| | `p` | non-`\0`-terminated strings
|`c` | ??? | canvas
| | `c` | colors
|`w` | ??? | window
| (`W`)   | `w` | list of floatwords
|    | `W` | array of floatwords

giuliomoro:
how does the non-`\0`-terminated string work? Needs a size specifier as well?

umlaeute:
> how does the non-`\0`-terminated string work? Needs a size specifier as well?

yes. like arrays (or lists).

the problem is, that there's a number of times where we want to transmit substrings of pre-existing strings (so we cannot just put in `\0` all over the place), and i think that creating a secondary buffer for a 0-terminated version of the string is just overkill and inefficient.

umlaeute:
@giuliomoro (or anybody else, really) if you have any ideas for alternative type-IDs for *canvas*, *window* (and *pd-object*, as i'm not sure whether i really want to degrade that to `x`), resp. for *color*, *floatwords* and non-0-terminated strings, i'd be very thankful.

umlaeute:
after converting the Pd-codebase to use the new `pdgui_vmess()` and `pdgui_stub_vnew()` here's some usage-statistics:

| type | #numuses | notes |
|---|---|---|
| raw string | 739 |
| integer | 679 |
| string | 296 |
| canvas | 210 |
| window | 31 |
| string array | 37 (~28~) |
| atom array | 26 | most uses are for simplifying fontspecs; this i basically the catchall if you need an array |
| float | 21 |
| raw string array| 0  (~9~)| all uses are for tags (which is wrong, as tags should be ordinary strings; fixed)
| floatword list | 4 | once, a `t_word*` array is accessed directly; the other three times this is only used to be able to send a list instead of of an array |
| pd-object | 2 | `logpost()` |
| atom list | 1 | makes the implementation of `glob_plugindispatch()` very elegant (but if need be, the GUI could just accept a list instead of arguments)

danomatika:
I haven't tried this yet, but it's looking good so far.

umlaeute:
@danomatika has changed the type-ID for "float" from `f` to `g` in the PR-description.
- the type-ID is **not** the `sprintf`-specifier, it is the `pdgui_vmess`-specifier (they conceptually overlap, but are not the same)
- what's the advantage of `g` vs `f`? (`f` maps nicely to the words "**f**loating point number" and `A_FLOAT`; `g` only makes sense with `sprintf` in mind)
- i'd like the PR-description (the toplevel post) to be a documentation of the PR as it currently is. (at the time of writing, the implementation uses `f` for floating-point numbers)

danomatika:
Ok sorry. I saw your commit and assumed you just forgot to update the spec in the PR post.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
ah i see. thanks for the attempt.

giuliomoro:
I had a pass at looking at the code and it all seemed good. As I don't trust myself with Pd code one bit, I thought I'd put together some way of testing this automatically.

The result is [pure-data-gui-testing](https://github.com/giuliomoro/pure-data-gui-testing).

This tests for consistency of both the Pd->GUI messages and the generated svg screenshot. One thing that makes such comparison arduous, even when running the same build of Pd repeatedly, is that the IDs are based on the pointers used by Pd and are therefore instance-dependent. I work around that by using regexps to [identify pointers and replace them](https://github.com/giuliomoro/pure-data-gui-testing/blob/main/pd-gui-parser/__main__.py#L180-L187) with instance-independent identifiers. (YES, this could be improved knowing a bit more about how the format of these IDs, in order to avoid false positives. Currently the only tweaking is on the expected length of these pointers).
With this testing system in place, I can verify that I get consistent results when running the same Pd binary twice and also when comparing develop(8ebd453cd60101af7f8933e073ca928ba2b043c3) and master(18c9695ba4de543c51c2c013b45b99b8f1c35a19). 

When I started comparing [umlaeute:feature/gui-mess](https://github.com/umlaeute/pure-data/tree/feature/gui-mess) with develop, I noted that in some places you tweaked whitespaces, curly braces and replaced the `[list ...]` syntax with `{...}`, so I [take that into account](https://github.com/giuliomoro/pure-data-gui-testing/blob/main/pd-gui-parser/__main__.py#L22-L96): tokenise everything in order to achieve consistent spacing and replacing the `[list ...]` syntax with `{}` everywhere. This got me a much more readable diff, but some differences remained.

One is between, e.g.: these lines (note that the `_%05d_` are the IDs I generate to replace the original pointers:

```diff
+ .x_00000_.c create rectangle 676 156 696 158 -fill #000000 -width 0 -tags { plot_00024_ plot_00024__array_00025__onset-1-1+0 array}
- .x_00000_.c create rectangle 676 156 696 158 -fill #000000 -width 0 -tags { plot_00024_ array}
```
(introduced  in 1834a3566a411ee24b4f8e2c9f399f024c11e93a), so I [worked around that](https://github.com/giuliomoro/pure-data-gui-testing/blob/main/pd-gui-parser/__main__.py#L194), too.

After all the above translations and workarounds we got to a point where I only see two changes between now the only changes I see between develop(8ebd453cd60101af7f8933e073ca928ba2b043c3) and this branch (480462e115faca3678102552a21bac7507e0adf5) are:

1) something like this towards the beginning:

```
+ ::pdwindow::logpost 0x0 4
- ::pdwindow::logpost .x0 4
```

2) and something in setting paths and startup flags:

```
+set ::sys_searchpath { /Users/giulio/Documents/Pd/externals /Users/giulio/Documents/Pd/externals/context /Users/giulio/Documents/Pd/patch2svg-plugin}
+set ::sys_temppath
+set ::sys_staticpath { /Users/giulio/Library/Pd /Library/Pd /Users/giulio/pure-data/extra}
+set ::startup_flags -verbose
-set ::tmp_path
 -lappend ::tmp_path /Users/giulio/Documents/Pd/externals
 -lappend ::tmp_path /Users/giulio/Documents/Pd/externals/context
 -lappend ::tmp_path /Users/giulio/Documents/Pd/patch2svg-plugin
-set ::sys_searchpath $::tmp_path
-set ::tmp_path
-set ::sys_temppath $::tmp_path
-set ::tmp_path
-lappend ::tmp_path /Users/giulio/Library/Pd
-lappend ::tmp_path /Library/Pd
-set ::sys_staticpath $::tmp_path
-set ::startup_flags [subst -nocommands -verbose ]
```

Also note that the the `svg` files are perfectly equivalent. The above results were obtained using [this test patch](https://github.com/giuliomoro/pure-data-gui-testing/blob/main/test-gui.pd) and the full files I obtained are attached
[here](https://github.com/pure-data/pure-data/files/9106542/Archive.zip)
 (the `-tr` files are after the "translation" operated by `pd-gui-parser`, the others are before such transation) . These were obtained running 

```
python3 ~/pure-data-gui-testing/pd-gui-tester --tmp /tmp/pd-gui-tester/ --svg --log --ptr-len 12  ~/pure-data-develop/src/pd ~/pure-data-gui-mess/src/pd -send "autostart 1" -send "autoquit 1" -send "svgSavePath {}.svg" ~/pure-data-gui-testing/test-gui.pd
```
on macos.

========================

To summarise:
- the patch looks good to me
- my understanding is that these commits shouldn't substantially affect the messages sent to the GUI, but some refactoring is happening, so I wrote some tests to verify that we get equivalent results (if not verbatim)
- the minimal test coverage highlights a couple of places where the results have diverged substantially, as outlined above. If @umlaeute can confirm that the differences I still see (most importantly the `::logpost 0x0` vs `:logpost .x0`)  are expected, then I am happy.
- I would also like to get suggestions on how to improve my test pd patch to increase coverage.

umlaeute:
thanks for taking the trouble.

the `logpost` differences shown are "don't care", as these are really just *any invalid* address (so that when you click on the message, you are not taken to the wrong object).
however, it is a reminder to manually check ´logpost` messages with a *valid* pointer, and see whether they still take us to the correct object.

the refactoring: yes, this was somewhat necessary. the idea of this entire branch is to get rid of Tcl-strings on the core side: rather than trying to come up with some wrapper to abstract away some Tcl-hacks, i decided to just clean them up while being there:
- the entire path setting (e.g. `set  ::sys_searchpath`) was rewritten (on the core-side) to use a single helper function (rather than repeating the same ugly pattern again and again); while there i ditched the temporary `tmp_path` variable (as being too Tcl-ish), as i cannot imagine what this is actually good for
- i pondered long about the `[subst -nocommands %s ]` (for setting `startup_flags`): i couldn't come up with a valid case when `subst -nocommands` should be used (and iirc, I wrote this line!), so I concluded that this was just some awry attempt to get spaces and quotes correct. my local tests showed that the current escaping seems to work as expected (that is: we can set broken startup-flags like `-font-face "Yikes [exit]}` and still can use the startup dialog (e.g. to fix the flags).

i also have some local test-patches that are informed by coverage testing, to make sure that all code-paths are triggered (though my evaluation of the results was minimal). i can/should probably submit them to your test repository.

giuliomoro:
>the refactoring: yes, this was somewhat necessary

OK, I updated the test to take care of these cases and it now passes against this branch. As long as we agree that the fixups [here](https://github.com/giuliomoro/pure-data-gui-testing/blob/ced61a4c937abeb0dc2332c4c6af3dea5b411a8e/pd-gui-parser/__main__.py#L221-L291) are well motivated and addressed, then we can claim that - as far as the test patch goes - this branch is equivalent to the `develop` it branched off.

I am eager to find out about more features that can be tested so that we can increase test coverage.

umlaeute:
@millerpuckette any comments on this? there are a couple of other GUI-related PRs in the queue (by @porres , @Ant1r , @sebshader , myself and probably many others) that are affected by this changeset.

millerpuckette:
So far I have two substantive comments and two stylistic ones...  first, try opening a patch that has a viaible subpatch:

#N canvas 499 102 450 300 12;
#N canvas 1023 231 450 300 (subpatch) 1;
#X restore 160 113 pd;
... I got a TK error.
Second - the inlet~ upsampling method now defaults to zero-padding, which has wild consequences.  Enjoy this example:

#N canvas 499 102 450 300 12;
#N canvas 1023 231 450 300 (subpatch) 0;
#X obj 152 98 block~ 128 1 2;
#X obj 69 100 osc~;
#X obj 74 66 inlet~;
#X obj 61 155 outlet~;
#X obj 169 160 print~;
#X obj 206 135 bng 15 250 50 0 empty empty empty 17 7 0 10 #fcfcfc #000000 #000000;
#X connect 1 0 3 0;
#X connect 2 0 1 0;
#X connect 2 0 4 0;
#X connect 5 0 4 0;
#X restore 160 113 pd;
#X obj 160 87 sig~ 440;
#X obj 159 139 sigmund~;
#X floatatom 159 164 5 0 0 0 - - - 0;
#X obj 86 115 osc~ 440;
#X obj 86 141 sigmund~;
#X floatatom 86 166 5 0 0 0 - - - 0;
#X connect 0 0 2 0;
#X connect 1 0 0 0;
#X connect 2 0 3 0;
#X connect 4 0 5 0;
#X connect 5 0 6 0;

--------------------
stylistic comments: how about using "g_" as the prefix instead of "pdgui_"?  Seems like this is going to appear ubiquitously int he code, so cleaner is better.

And... how important is colorized TTY output?  Personally, I find it enraging.  Perhaps others disagree :)

umlaeute:
Thanks for the review.

I'll check the to error.

The resampling is of course a regression, which I thought I had taken care not to run into...will fix.

I picked "pdgui_" precisely because I think it is clearer than a cryptic "g_" prefix (now that I'm older, I tend to prefer readable multiline code, instead of fringe to cram as much logic as possible into a single line...)


As for coloring the output, this is hopefully well hidden: "pd -d 3" should give you the traditional uncolored output. You have to explicitly switch on the "color-bit" in the debug flags to enable the eye sore

giuliomoro:
```
#N canvas 499 102 450 300 12;
#N canvas 1023 231 450 300 (subpatch) 1;
#X restore 160 113 pd;
```

This should fix it?

```
diff --git a/tcl/pdtk_canvas.tcl b/tcl/pdtk_canvas.tcl
index c5a20d15..12188284 100644
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -439,7 +439,7 @@ proc ::pdtk_canvas::pdtk_canvas_setparents {mytoplevel args} {
     set parents {}
     foreach parent $args {
         if { [catch {set parent [winfo toplevel $parent]}] } {
-            if { [file extension $parent] eq .c } {set parent [file rootname $parent]}
+            if { [file extension $parent] eq ".c" } {set parent [file rootname $parent]}
         }
         lappend parents $parent
         addchild $parent $mytoplevel
```

umlaeute:
@millerpuckette  i think i've "addressed" all your concerns:
- the Tk-error should be gone
- the default upsampling method is now sample-and-hold (rather than zero-pad) even for `[inlet~]`s
- i still think that `pdgui_` is clearer than `g_`
- you get colored debugging output *only* if you use a debug-level of 5 (core->gui), 6 (gui->core) or 7 (core<->gui) (that is: the debug-level is ORed with `4`)

danomatika:
How about a “ui_” prefix? I suppose it breaks the single char prefix convention.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

millerpuckette:
I like the ui_ idea... it's more specific than g_ but less verbose than pdgui_ ...

A bit of a nitpick but I still try to hold to the 80-chars-horizontal rule without
requiring too many line breaks :)

danomatika:
I do 80-char too. :)

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
> I do 80-char too. :)

we all do (at least I try to).

but i'm honestly baffled about this discussion, and I think I missing the point.

afaiu, the remaining issue we are having now is that 
- a single line of code should not exceed 80 chars
- single line statements are preferred over multi line statements
- the `ui_` prefix is preferred, as it would allow us to have more single line statements that fit into 80 chars

now, in practice the `ui_` prefix is *3* ("three") letters shorter than `pdgui_`. which means that there will only be a gain for lines that have 80-82 chars with the long prefix. shorter lines will fit into the 80-chars limit regardless of which of the two prefixes; longer lines won't fit anyhow.

so I checked how long the actual lines invoking `pdgui_vmess`  are (with linebreaks removed).
here's some statistics:
| data | value |
|-------|-------|
| number of invocations | 177 |
| mean(linelength) | 100.35 |
| stddev(linelength) | 60.26 | 
| median(linelength) | 76 |
| number of lines with 80<=length<83| **4** |

the linelength calculation leaves the indentation intact.
also note that most iemguis have not been converted to `pdgui_vmess` yet (but the 4 lines in question are from other parts of the code anyhow).


so please tell me what this is really about, because 4 lines seem hardly worth it.

---

but why am i fighting for the `pdgui_` prefix, if it isn't worth it?
that's mainly because I really think that the direction of Pd is away from the single application towards re-usable code in the form of libpd.

if the code only affected a single application (Pd), then I couldn't care less about the actual prefix (even `g_` would be fine :yum: ).
but when building my own application that happens to use libpd, i'd rather have control over generic prefixes like `ui_` without having to fear to step on the tows of some helper-library.

umlaeute:
but in any case: if @millerpuckette really wants the `ui_` prefix, how about running the following script before merging:

```sh
sed -e 's|pdgui_|ui_|g' -e 's|ui_\(strnescape\)|pdgui_\1|g' -i src/*.c src/*.h
git commit src -m "Fix function prefix for Pd/Gui communication (leaving the pre-existing pdgui_strnescape intact)"
```

Spacechild1:
Personally, I find the `pdgui` prefix much clearer. It immediately tells me that the function is about the Pd GUI. I think we should prefer readability over terseness.

giuliomoro:
I concur with @Spacechild1 and @umlaeute on the naming and I find that @umlaeute point about potential pseudo-namespace conflict is particularly relevant.

Btw, this is largely replacing something that used to be called `sys_vgui()`, which was neither clear nor "terse" (in the sense of 1-char-prefix-terse).

porres:
> Btw, this is largely replacing something that used to be called sys_vgui(), which was neither clear nor "terse" (in the sense of 1-char-prefix-terse).

good point :)

sebshader:
we already know that it is a message from pd though.

giuliomoro:
> we already know that it is a message from pd though. 

This is about it being a message _for_ Pd. When using this in a libpd context, it may not be clear that `ui_vmess()` if _for_ Pd, whereas `pdgui_vmess()` is clearer. Sure, libpd _could_ provide all needed wrappers so that Pd's own symbols don't have to be visible at all, but that's a different task.

>  and eventually the ui client doesn't necessarily have to be graphical..

are you suggesting `pdui_` ? 

I guess the conversation about the prefix to be used could go on forever.  In the meantime, it would be great to hear more about the bulk of the PR, i.e.: the logic, implementation and content of the changes.

danomatika:
Gotta say that discussing such things seems trivial except that we will be stuck with whatever is chosen for the next 20 years, so I appreciate potentially not having to refactor a libpd wrapper and client code down the road at the cost of the conversation “going on forever.” In the end, it’s up to Miller but I think it’s good to look at the different points of view.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
@millerpuckette is this deadlocked?

(what i tried saying above with my `sed` script was: if you really want to have such names, please go ahead and change them; it's just that i'd prefer to not do the change myself - call it snobbish if you like :-))

millerpuckette:
I think given all teh pros and cons I'm content to just leave it as pdgui_ ... but now I need to take a deeper look at the code which I haven't made time to do yet.

danomatika:
> I think given all the pros and cons I'm content to just leave it as pdgui_

As an aside, there is precedence in the pdinstance_ naming already. 😁

umlaeute:
@giuliomoro wrote:
> the `logpost` differences shown are "don't care", as these are really just _any invalid_ address (so that when you click on the message, you are not taken to the wrong object).

:facepalm: 

of course they shouldn't have  been "don't care". i was somewhat distracted by the actual address (`NULL`), which is really a "don't care".
however, the actual formatting pattern also changed for the valid addresses from `.x%x` to `0x%x`, which in turn broke the "click on an error message to be taken to the culprit" functionality.

this has been fixed in 97958558016aace43881cf9e75ac69a51ddde218 ( part of #1632 )

danomatika:
If we bring up namespacing, then it follows most things should follow a convention like pd_ui_… or pd_g_… We do this now with libpd.

My suggestion if ui_ is just a compromise, but if this interface is meant to be a public library API, then it should have a namespace, however we can simply wrap it with libpd ad we do for a lot of the API already. It depends based on if the functions are considered “internal” or not.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

PR #1712, opened 2022-08-03
HTML: include BSD installation notes from INSTALL.txt

Lucarda:
This PR is intended to be merged into `feature/double-extension`. It adds to https://github.com/pure-data/pure-data/pull/1605

### DIFF
--- a/doc/1.manual/index.htm
+++ b/doc/1.manual/index.htm
@@ -171,7 +171,7 @@ can be found at:
 		<li><a href="x6-b.htm"> Windows resources </a></li>
 		<li><a href="x6-b.htm#s6.5.2">  Windows ASIO support </a></li>
 		</ol>
-    <LI> <a href="x6.htm#s6.6"> BDS </A>
+    <LI> <a href="x6.htm#s6.6"> BSD </A>
         <ol>
 		<li><a href="x6.htm#s6.6.1"> FreeBSD </a></li>
         <li><a href="x6.htm#s6.6.2"> OpenBSD </a></li>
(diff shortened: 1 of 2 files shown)


umlaeute:
i'm not entirely sure whether this is worth it.

my assumption is:
- there are *very* few BSD-users out there that want to use Pd
- these users are somewhat tech-savy (e.g. familiar with buliding projects from source)

with these assumptions, I think we can expect them to check the README.txt https://github.com/pure-data/pure-data/blob/573b4a2ee17dc6a7dd168574489db4a0e81a3066/README.txt#L22-L24 and the INSTALL.txt https://github.com/pure-data/pure-data/blob/573b4a2ee17dc6a7dd168574489db4a0e81a3066/INSTALL.txt#L417-L422 ...

and don't need them to rely on HTML documentation.

For the vast majority of users, the addition of the BSD-instruction to the HTML manual will just be additional noise.

Since we already have a section that is titled `Linux & BSD` (albeit it doesn't describe anything BSD related), i would just add a single sentence at the end of *that* section:
> For more information (and how to build on BSD) consult the INSTALL.txt file

alternatively you could create a `x6-c.html` file that bundles the BSD instructions (without cluttering the main page).

(In general I'm not sure I see the point of duplicating the documentation, but ymmv)

umlaeute:
oh and just to add: inserting a new section breaks the anchors...
(using section numbers as anchors is probably stupid anyhow, precisely because you can't insert new sections without breaking things)

Lucarda:
closing this. there's a new https://github.com/pure-data/pure-data/pull/1716

Issue #1713, opened 2022-08-05
External examples: Build failing on darwin arm (M1)


HeckHeckHeckHeck:
Target: Pd-0.52-2
Description:
```bash
$ cd /Applications/Pd-0.52-2.app/Contents/Resources/doc/6.externs/
$ make pd_darwin
```

Expected:
builds the examples, resulting in: "obj<n>.pd_darwin"

Observed:
* `-arch i386` is being used for compile (cc -DPD -O2 -Wall -W -Wshadow -Wstrict-prototypes -Wno-unused -Wno-parentheses -Wno-switch -arch i386 -arch x86_64 -I../../src -o obj1.o -c obj1.c)

* `-arch i386 -arch x86_64` is being used for linking (cc -bundle -undefined suppress -arch i386 -arch x86_64 \
            -flat_namespace -o obj1.pd_darwin obj1.o)

Proposed solution:
* Add make target: `pd_darwin_arm`
* I will submit  a PR ;)

PR #1714, opened 2022-08-05
#1713 - External examples: Build failing on darwin arm (M1)

HeckHeckHeckHeck:
Hya dear PureData community,
please let submit my first little insignificant PR to this project. 
Cheers
Heck

### DIFF
--- a/doc/6.externs/0.README.txt
+++ b/doc/6.externs/0.README.txt
@@ -9,7 +9,7 @@ is in use.  For Windows, this should be Microsoft Visual C (it's also possible
 to compile externs using mingw but this makefile doesn't support that).
 
 To compile, type "make pd_linux", "nmake pd_nt" (for windows), or
-"make pd_darwin" (for Mac OSX).
+"make pd_darwin" (for Mac OSX x86), and "make pd_darwin_arm" (for Mac OSX ARM).
 
 If you have saved these files to a new location (other than the Pd distribution
 itself), it will probably be necessary to customize the makefile by changing the
@@ -17,7 +17,7 @@ definition of "PDNTINCLUDE" (windows) or "LINUXINCLUDE" (linux or Mac)
 to point back to the PD source directory.
 
 Mac users who haven't installed a 32-bit compilation chain will also have to
-edit the makefile to delete the phrase "-arch i386" in two places.
+edit the makefile to delete the phrase "-arch i386" in two places. (x86 macs only)
 
 Windows users will most likely have to update the makefile to point correctly to
 the current version of Microsoft Visual C.
--- a/doc/6.externs/makefile
+++ b/doc/6.externs/makefile
@@ -1,5 +1,5 @@
 current:
-	echo make pd_linux, pd_linux32, pd_nt, pd_darwin
+	echo make pd_linux, pd_linux32, pd_nt, pd_darwin, pd_darwin_arm
 
 clean: ; rm -f *.pd_linux *.o
 
@@ -60,7 +60,7 @@ LINUXINCLUDE =  -I../../src
 	strip --strip-unneeded $*.l_ia64
 	rm $*.o
 
-# ----------------------- macOS -----------------------
+# ----------------------- macOS x86 -----------------------
 
 pd_darwin: obj1.pd_darwin obj2.pd_darwin \
      obj3.pd_darwin obj4.pd_darwin obj5.pd_darwin dspobj~.pd_darwin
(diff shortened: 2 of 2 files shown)


HeckHeckHeckHeck:
sorry, its bad, got better solution. soon...

Spacechild1:
I think the `-arch` flags should be provided by the user. There is no need to have a specific makefile target for ARM. What we need is a way for the user to pass custom C flags.

Generally, the makefile is pretty outdated (no mingw support!) and I'm wondering why we don't just use pd-lib-builder instead? Yes, pd-lib-builder doesn't support MSVC, but we could still use a seperate makefile for that.

danomatika:
There is no reason other than that these examples are old and not often updated.

Spacechild1:
Actually, I hit this problem with https://github.com/pure-data/pure-data/pull/1357 because it adds another example object. I couldn't compile with MinGW, so I made my own makefile with pd-lib-builder. I can make a PR if you think this would be reasonable.

HeckHeckHeckHeck:
IMO, the user having to pass the build flags himself defeats the purpose of a Makefile. Since, what ppl usually want is to just type 'make' and it works.
Also, i just had quick look into the Makefiles for the PD build itself and i wondered if that even works for ARM macs. 
My intention is to just deliver a quick little patch that cant hurt and fixes the problem only for this tiny example in the docs. 
Just a tiny improvement :)
It doesnt even seem to need the whole 'pd-lib-builder' overhead (which is a great tool though)
So, my plan is to detect arch (x86/ARM) for the pd_darwin target and set the flags appropriately. 
That would be trivial and fixes this little incident.
Though, on the other hand i wondered if the -arch flag is even neccessary at all, since the CC being used will default to the correct one anyways? 
But i cannot interfere with platforms i cannot test, that only leaves me the 'arch' detect & switch option.

danomatika:
The -arch flag should not be needed on mac since the default behavior of Xcode & the command line tools is to simply build for the current system architecture. In this case, the makefile should build for arm64 on an M1 machine without requiring any additional flags.

HeckHeckHeckHeck:
Sorry for the noise, its not worth the hassle.
Just ignore me.
i will submit a PR and just hope it gets merged. Wont break anything for sure.

HeckHeckHeckHeck:
@danomatika exactly. But i cannot test. So, i cant touch.

danomatika:
@Spacechild1 This solution is fine for now. In the future we should really move to pd-lib-builder as it handles all of this automatically.

@HeckHeckHeckHeck If you are not aware of it, check out [pd-lib-builder](https://github.com/pure-data/pd-lib-builder) to avoid rolling your own makefile. There is also an [example external project](https://github.com/pure-data/helloworld) which uses it.

Spacechild1:
> IMO, the user having to pass the build flags himself defeats the purpose of a Makefile. Since, what ppl usually want is to just type 'make' and it works.

Well, no. You often have to pass options/arguments to makefiles.

Anyway, as @danomatika said, on Mac the makefile should just work *without* any `-arch` flags. In that case, it would just build for the host architecture. You would only have to pass `-arch` flags if you want to build for *another* architecture - or build a fat binary. So the solution is simply to just remove the `-arch` flags from the `pd_darwin` target. Since this makefile is really only for testing the objects, this should be sufficient.

danomatika:
@HeckHeckHeckHeck BTW you *can* test by invoking make with Rosetta ala:

   arch -x86_64 make

danomatika:
>So the solution is simply to just remove the -arch flags from the pd_darwin target. Since this makefile is really only for testing the objects, this should be sufficient.

Yes.

HeckHeckHeckHeck:
ok then someone just removes the -arch flags then. 
Its really not worth wasting any more of our time for just this little insignificant problem.
btw.
"@HeckHeckHeckHeck If you are not aware of it, check out [pd-lib-builder](https://github.com/pure-data/pd-lib-builder) to avoid rolling your own makefile." - i did already mention it, so it seems you just type and dont read.
"Anyway, as @danomatika said, on Mac the makefile should just work without any -arch flags. "
that was my suspicion initially, so you dont seem to read, either.
sigh...

danomatika:
Nope, I skimmed the last replies. It's 2 AM and I am trying to finish something else before needing to wake up agin early tomorrow. If you take that as we don't care, well that's your perspective but we are both responding to you. This is also why I phrased my response with "If you are not aware of it..." implying that I assume you *may* know of it, but maybe not.

> ok then someone just removes the -arch flags then.
> Its really not worth wasting any more of our time for just this little insignificant problem.

It's not insignificant. It's a valid problem are our responses are not "oh, your wrong, don't waste our time" but simply a back and forth of trying to discuss what a technical solution would be. If you create a second PR that removes the flag, we'd gladly take it into the develop branch.

HeckHeckHeckHeck:
Yes, sure. you are right. okay. will do that. thanks to both of you.

danomatika:
Thank you for finding this and opening the initial PR. It's often that there may be one solution, but after talking about it someone finds an alternate way which all agree might be better. In either case, we work together to find a solution. Personally, I have never tried building these examples, so I would have never found this myself... ;)

HeckHeckHeckHeck:
Cheers. Absolutely my opinion, as well.
Removing the -arch flags is definitely the elegant solution.
(i suspect they were just there for 32bit compat in the first place, anyways)

danomatika:
> Cheers. Absolutely my opinion, as well.
> Removing the -arch flags is definitely the elegant solution.
> (i suspect they were just there for 32bit compat in the first place, anyways)

Ja, these makefiles are pretty crusty and have not been updated in a while. This was likely added during the 64 bit transition....

HeckHeckHeckHeck:
https://github.com/pure-data/pure-data/pull/1715
Cheers

PR #1715, opened 2022-08-06
External examples: Improve Makefile for darwin arm (and make clean)

HeckHeckHeckHeck:
tiny improvement...
cant do any harm.
Cheers
Heck

### DIFF
--- a/doc/6.externs/makefile
+++ b/doc/6.externs/makefile
@@ -68,11 +68,11 @@ pd_darwin: obj1.pd_darwin obj2.pd_darwin \
 .SUFFIXES: .pd_darwin
 
 DARWINCFLAGS = -DPD -O2 -Wall -W -Wshadow -Wstrict-prototypes \
-    -Wno-unused -Wno-parentheses -Wno-switch -arch i386 -arch x86_64
+    -Wno-unused -Wno-parentheses -Wno-switch
 
 .c.pd_darwin:
 	cc $(DARWINCFLAGS) $(LINUXINCLUDE) -o $*.o -c $*.c
-	cc -bundle -undefined suppress -arch i386 -arch x86_64 \
+	cc -bundle -undefined suppress \
             -flat_namespace -o $*.pd_darwin $*.o 
 	rm -f $*.o
 
--- a/doc/6.externs/makefile
+++ b/doc/6.externs/makefile
@@ -1,7 +1,12 @@
 current:
 	echo make pd_linux, pd_linux32, pd_nt, pd_darwin
 
-clean: ; rm -f *.pd_linux *.o
+clean:
+	rm -f *.o
+	rm -f *.pd_linux
+	rm -f *.pd_darwin
+	rm -f *.l_ia64
+	rm -f *.l_i386
 
 # ----------------------- Windows-----------------------
 # note; you will certainly have to edit the definition of VC to agree with


danomatika:
I would maybe add a comment above the pd.darwin makefile target noting how to make alternative arch builds or fat builds ala `-arch i386 -arch x86_64` or `-arch arm64 -arch x86_64`. In this case, people will need to do fat builds manually which I think makes the most sense since there are many different macOS versions which support building for different arch. OTOH if we go for "modern build only" then we would have arm64 & x86_64 but then this would break on older systems, so best to inform and use the system default for now. :)

danomatika:
Ping @umlaeute @millerpuckette @Spacechild1

HeckHeckHeckHeck:
seen.

sebshader:
this is might be kind of off-topic but also you have to pass in empty CFLAGS in order to not have autoconf configure use the default optimization and debugging (-O2 and debugging symbols are on by default)

HeckHeckHeckHeck:
nr. lines of CODE changed: 4
nr. of lines from discussions: about 150
nr. of PR's open: 90

Thats not a 'modus operandi' i believe software can be developed.
Maintainership seems to be broken.
Gatekeepers seems to block the progress.
I have also looked into other PR's, to get a better picture. (discussions seem quite amateur level)

Not interested. Will just fix for myself.
Will just fork, in case.

There are 2 types of Dev's:
1.) Dev's who like to write code
2.) Dev's who just like to discuss.

Projects dominated by type 2, i observe:
* type 1 dev's leave (the more exp, the quicker)
* progress stagnates

On a case-specific note: 
Would the change be about the main build i would definitely understand the "code change : discussion" ratio.
But not for an example in the docs folder, that nobody bothered about for quite a while now.

Best 
Heck

danomatika:
A simple “anyone looked at this yet?” would have sufficed. Your point cuts both ways: I personally don’t feel like spending more time on contributing myself when I don’t feel it’s  skies either…

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

danomatika:
> if we go for "modern build only" then we would have arm64 & x86_64 but then this would break on older systems, so best to inform and use the system default for now. :)

I'm not seeing a response to this, so I will add some info myself after merging.

danomatika:
> this is might be kind of off-topic but also you have to pass in empty CFLAGS to the autoconf configure in order to not have it use the default optimization and debugging (-O2 and debugging symbols are on by default iirc)

THat's a good point and I also see the same behavior using auto tools. I can try just passing an empty CFLAGS at a minimum.

danomatika:
**Required fixes**

I added CFLAGS handling as it's required otherwise passing the arches to make does nothing. I suppose no-one tried this, even though I had provided the make command example ion a previous post.

I have also added info on how to build for multi or non-native archs.

See d7aeada3a938f982395de478f9f79dd47e869264

**Pd's development organization or lack thereof**

Pd's development is largely anarchic and driven by people who have time/interest, for better or worse. If you have suggestions on overall development organizational approaches, you can start a discussion on the pd-dev mailing list: 

https://lists.puredata.info/listinfo/pd-dev

If you want my personal perspective on Pd's long-term development and it's issues, I wrote a long forum reply for a different issue, but the topics are similar. Again, this is my view and I don't write for anyone else:

https://forum.pdpatchrepo.info/topic/13016/why-does-pd-look-so-much-worse-on-linux-windows-than-in-macos/35

In any case, we still welcome issues being pointed out and technical solutions if you have them, even if it takes time to get to integrating or discussing them. Have a good weekend, I intend to. :)

HeckHeckHeckHeck:
@danomatika Okay, thanks for your reply/reaction.
Also, thanks for the tips/insight.
Thats an interesting read.
Forgive me if the post was maybe a bit too homest, but as you know engineering and improvements only work based on criticism. 
Nothing personal, though, there.
I wish you a very good weekend, indeed ;)

PR #1716, opened 2022-08-06
HTML: include BSD installation notes from INSTALL.txt

Lucarda:
This PR is intended to be merged into `feature/double-extension`. It adds to https://github.com/pure-data/pure-data/pull/1605


umlaeute:
not a big deal, but i think in general it's better to start from the branch you intend to merge into (that is: start from `feature/double-extension` rather than `master`). 

(it might have been a glitch with my setup though; if this is the case, please ignore my comment)

Issue #1717, opened 2022-08-08
unique seed values and seed value for [expr]


porres:
It'd be great to have a unique seed value every time you open your patch. By checking things out for my externals, one could use 

```
#include <time.h>
time(NULL)
```

to get a different value every time you open a patch, and also a "static int" variable to add to this value so different objects have different seed values.

Now, we may discuss if this would "break" patches. Maybe in principle, but not really as the seed values for [random] do change depending if you have opened another patch with another [random] object before. If you want reproducibility you gotta use the 'seed' message anyway, so I don't think there's any issue in getting unique seeds every time.

But if it is the case where we don't want to change this, I can propose a more kludgy way to include this functionality while respecting 100% backwards compatibility.

porres:
This would apply to all objects that have a 'seed' method, namely:  [random], [noise~] and [array random]. I don't think there are others.

Now, [expr] has a random function too, but it uses 'rand()'. One can set a seed with "srand" but I don't see how multiple [expr] objects can have a "static" rand() function. It seems that the seed and order of random numbers from one object affects the order of another instance. This is not the case for the other objects where each one keeps its own seed and sequence.

We can maybe add a seed method here here and also use "time(NULL)" for a unique seed. But if we want it to behave like it would in [random] with unique sequences, maybe we need another way to calculate the random values without using "rand()". What do you say?

porres:
Oops, this is a duplicate of an earlier request and discussion of mine, see https://github.com/pure-data/pure-data/issues/1187 I'm closing the old one in favor of this but keeping the reference.

For the record, @Spacechild1 agrees that a unique seed every time you instantiante [random] for the first time makes sense. He also proposes that the 'seed' message without an argument gives us a new unique seed based on system time.

porres:
I guess we can keep [expr] without a setable seed and leave this to the other more flexible and more dedicated objects

claudeha:
> the seed values for [random] do change depending if you have opened another patch with another [random] object before. If you want reproducibility you gotta use the 'seed' message anyway

At least one of my patches uses the reproducibility available by default.  But I also do tricks like launch pd with a message to a loader patch to create a certain number of [random] objects before loading the main patch, to get a different sequence (but still 100% reproducible).

I would prefer a mechanism that allows reproducibility, for example a command line flag (ideally connected to a function that could be called by a libpd-using program) to set the initial seed generator value.

I have no opinion on whether the default is reproducible or not, just that reproducibility is possible.

[...]

PR #1718, opened 2022-08-09
add unique seed according to system time in peudo-random generators

porres:
This closes https://github.com/pure-data/pure-data/issues/1717 and takes care of [random], [array random] and [noise~].

Now the seed is set via system time when you load the patch/object at creation time. A _static int instanceindex_ is used to provide a different seed for each instance.

A seed message without arguments set unique seeds according to system time again.

### DIFF
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -22,12 +22,6 @@
 #include <endian.h>
 #endif
 
-#ifdef _WIN32
-#include <time.h>
-#else
-#include <sys/time.h>
-#endif
-
 #ifdef __MINGW32__
 #include <sys/param.h>
 #endif
@@ -473,21 +467,13 @@ void sigvcf_setup(void)
 /* -------------------------- noise~ ------------------------------ */
 static t_class *noise_class;
 
-static int instanceindex = 0;
-
 typedef struct _noise
 {
     t_object x_obj;
     int x_val;
 } t_noise;
 
-static int makeseed(void)
-{
-    /* seed each instance differently.  Once in a blue moon two threads
-    could grab the same seed value.  We can live with that. */
-    int init = time(NULL) * ++instanceindex;
-    return (init *= 1319);
-}
+int makeseed(){};
 
 static void *noise_new(void)
 {
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -22,6 +22,12 @@
 #include <endian.h>
 #endif
 
+#ifdef _WIN32
+#include <time.h>
+#else
+#include <sys/time.h>
+#endif
+
 #ifdef __MINGW32__
 #include <sys/param.h>
 #endif
@@ -467,19 +473,26 @@ void sigvcf_setup(void)
 /* -------------------------- noise~ ------------------------------ */
 static t_class *noise_class;
 
+static int instanceindex = 0;
+
 typedef struct _noise
 {
     t_object x_obj;
     int x_val;
 } t_noise;
 
+static int makeseed(void)
+{
+    /* seed each instance differently.  Once in a blue moon two threads
+    could grab the same seed value.  We can live with that. */
+    int init = time(NULL) * ++instanceindex;
+    return (init *= 1319);
+}
+
 static void *noise_new(void)
 {
     t_noise *x = (t_noise *)pd_new(noise_class);
-        /* seed each instance differently.  Once in a blue moon two threads
-        could grab the same seed value.  We can live with that. */
-    static int init = 307;
-    x->x_val = (init *= 1319);
+    x->x_val = makeseed();
     outlet_new(&x->x_obj, gensym("signal"));
     return (x);
 }
@@ -505,10 +518,12 @@ static void noise_dsp(t_noise *x, t_signal **sp)
     dsp_add(noise_perform, 3, sp[0]->s_vec, &x->x_val, (t_int)sp[0]->s_n);
 }
 
-static void noise_float(t_noise *x, t_float f)
+static void noise_seed(t_noise *x, t_symbol *s, int argc, t_atom *argv)
 {
-    /* set the seed */
-    x->x_val = (int)f;
+    if(!argc)
+        x->x_val = makeseed();
+    else
+        x->x_val = (int)(atom_getfloat(argv));
 }
 
 static void noise_setup(void)
@@ -517,8 +532,8 @@ static void noise_setup(void)
         sizeof(t_noise), 0, A_DEFFLOAT, 0);
     class_addmethod(noise_class, (t_method)noise_dsp,
         gensym("dsp"), A_CANT, 0);
-    class_addmethod(noise_class, (t_method)noise_float,
-        gensym("seed"), A_FLOAT, 0);
+    class_addmethod(noise_class, (t_method)noise_seed,
+        gensym("seed"), A_GIMME, 0);
 }
 
 /* ----------------------- global setup routine ---------------- */
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -473,7 +473,7 @@ typedef struct _noise
     int x_val;
 } t_noise;
 
-int makeseed(){};
+int makeseed(void);
 
 static void *noise_new(void)
 {
--- a/src/x_array.c
+++ b/src/x_array.c
@@ -713,7 +713,7 @@ typedef struct _array_random   /* any operation meaningful on a subrange */
     unsigned int x_state;
 } t_array_random;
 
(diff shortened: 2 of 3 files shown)


porres:
ping @umlaeute  and @Spacechild1 for revision

src/x_misc.c
```diff
@@ -56,7 +58,7 @@ typedef struct _random
 
 static int makeseed(void)
 {
-    static unsigned int random_nextseed = 1489853723;
+    unsigned int random_nextseed = time(NULL) * ++instanceindex * 151;
     random_nextseed = random_nextseed * 435898247 + 938284287;
```
Spacechild1, line 62: The original implementation of `makeseed` is actually a PRNG by itself. The only thing we have to do is replace the hardcoded initial value with a call to `time(NULL)`. 
```
static int makeseed(void)
{
        /* this is not thread-safe, but it does not matter for our purposes */
    static unsigned int random_nextseed = 0;
    if (!random_nextseed)
        random_nextseed = time(NULL);
    random_nextseed = random_nextseed * 435898247 + 938284287;
    return (random_nextseed & 0x7fffffff);
}
```

---

Sidenote: in C++11 you could simply write the following and it would even be thread-safe:
```
static unsigned int random_nextseed = time(NULL);
```

Spacechild1: Also, we only have to implement `makeseed` in `x_random` and make it non-static, so we can reuse it for `[array random]` and `[noise~]`.

porres: > BTW, it does not really make sense to implement the same function three times. Instead, makeseed() 

it's not exactly the same for [noise~] but it's the same for [random] and [array random]

porres: > The only thing we have to do is replace the hardcoded initial value with a call to time(NULL).

this wouldn't work to create different seed values for each instance and wouldn't work for a 'seed' message without an argument to reset the seed according to current time.

porres: >  Instead, makeseed() in x_random.c could be non-static 

the instanceindex variable would also need to be non-static, right? But anyway, before things were also exactly the same for different objects and it was fine, it just wasn't in a separate function. Actually, it was a separate function in [random] but it was used just once, I then made it be used twice, which makes more sense I think.

Spacechild1: > it's not exactly the same for [noise~] but it's the same for [random] and [array random]

There is no reason why `[noise~]` would need a different seed method.

Spacechild1: > this wouldn't work to create different seed values for each instance

it does work because `makeseed` is really a PRNG on its own and every call yields a different value. the only difference is that we *initialize* the sequence with a pseudo-random number (= `time()`) instead of using a hardcoded value.

>  wouldn't work for a 'seed' message without an argument to reset the seed according to current time.

it does work because `makeseed` returns an unpredictable number.

> the instanceindex variable would also need to be non-static, right?

in my function there is no instance index variable ;-)

Spacechild1: BTW, I just edited my implementation of `makeseed`: `random_nextseed ` is now `PERTHREAD` to make it reentrant.

porres: So, copy your makeseed "static" code into radom's code, call it in [noise~] and [array random], as simple as that, right? Will try it!

porres: btw, what is pd's method for generating pseudo random numbers? Doesn't seem to be "Linear Congruential Generator", but it's similar and maybe that's it?

porres: I'm not succeeding in making the bridge and telling where to look for "makeseed".

I get errors:

Undefined symbols for architecture x86_64:
  "_makeseed", referenced from:
      _noise_new in pd-d_osc.o
      _noise_seed in pd-d_osc.o
      _arrayobj_new in pd-x_array.o
      _array_random_new in pd-x_array.o
      _array_random_seed in pd-x_array.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)

I tried declaring "static int makeseed();" or just "int makeseed();" in the code of noise~/array...

porres: hmm, maybe I got it now

porres:
@Spacechild1 the commit https://github.com/pure-data/pure-data/pull/1718/commits/09e50f61f674572f830bdc52f4f449c92ff16d88 seems to do what you want, please revise it as I don't understand how I made it compile after it was giving errors... :) I was having trouble pointing to the static function.

porres:
Well, it compiles but the code doesn't seem to work! It works fine for [random], but now that I tested [noise~] I see I keep getting the same sample sequence when using "seed" without arguments.

So please check how it should actually work and maybe send me a PR @Spacechild1 ?

src/x_array.c
```diff
@@ -713,20 +713,23 @@ typedef struct _array_random   /* any operation meaningful on a subrange */
     unsigned int x_state;
 } t_array_random;
 
+int makeseed();
```
Spacechild1, line 716: This should be `int makeseed(void);`.

NOTE: in C++ it would be the same as `int makeseed(void);`, but in C this is the old K&R function declaration style without parameters.

src/d_osc.c
```diff
@@ -473,13 +473,12 @@ typedef struct _noise
     int x_val;
 } t_noise;
 
+int makeseed(){};
```
Spacechild1, line 476: This should be `int makeseed(void);` :-)

Spacechild1:
>  as I don't understand how I made it compile after it was giving errors... :)

Tip: if you get compiler errors, try to understand what they *mean* - instead of trying out random things until the error disappears :-)

porres:
thanks for the explanation, I tried all sorts of things trusting my intuition from I see around. I think I also tried using "void", but hey...

porres:
> I think I also tried using "void", but hey...

Yes, I had tried that... and I tried it again and got an error:

```
Undefined symbols for architecture x86_64:
  "_makeseed", referenced from:
      _noise_new in pd-d_osc.o
      _noise_seed in pd-d_osc.o
      _arrayobj_new in pd-x_array.o
      _array_random_new in pd-x_array.o
      _array_random_seed in pd-x_array.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [pd] Error 1

```

same error from https://github.com/pure-data/pure-data/pull/1718#discussion_r941864721

weird... will make a PR anyway so you can see it

porres:
As you can see, many checks were unsuccessful... first one I saw said 

```
d_osc.obj : error LNK2019: unresolved external symbol _makeseed referenced in function _noise_new
x_array.obj : error LNK2001: unresolved external symbol _makeseed
../bin/pd.dll : fatal error LNK1120: 1 unresolved externals

```

Spacechild1:
@porres read again the first paragraph of https://github.com/pure-data/pure-data/pull/1718#issuecomment-1210042419

You have to remove the `static` keyword from `makeseed`, otherwise the function will be static. That's the reason why you got linker errors.

When you're done, can you clean up the commits? Hint: use `git reset HEAD~<N>` to undo the last N commits but keep the changes. Then you can commit all of your actual changes as a single commit and force-push.

Another tip: if you work on a feature branch that nobody else is working on, you do not have to create new commits for every little change. Instead you can ammend the last commit and force push. (For the full super powers, read about "interactive rebasing".)

porres:
I had tried reverting the last commits in my documentation branch with that, but I'm running through some weird authentication issues that I've seen before and fixed by reinstalling github desktop. Anyway, I might just reboot this and create a quite clean one instead

umlaeute:
Note that "reverting" is something (slightly) different: it is adding a new change that undoes a previous one (so your change history records both your "faulty" commit and your attempt to clean it up).

This is great in various cases, eg if your original change created a regression (in something that was not obviously related) and you want to document that you reverted that change on purpose, after that was discovered.

But the most important use cases are:
- some other changes have been committed between the original "faulty" one and *now*
- other people have pulled your branch and (might) have started working in top of your "faulty" commit.

But neither is the case here:
- the faulty commit is still top
- nobody has started to work on top of your commits (this is an assumption, but I think a valid one for fresh feature branches; this is not `master` or `develop`)
- your commits seem to be trying to find a working solution. They don't introduce a regression but instead are just broken by  themselves)

So the suggestion is to change your last commit to a fixed version and force push.

porres:
Cool, I was able to do this in my documentation branch, but for this one I think it's best I just reboot it for a quite cleaner version so it's easier for Miller

PR #1719, opened 2022-08-10
load seeds based on system time for pseudo random number generators

### LINKED ISSUE #1717: unique seed values and seed value for [expr]

porres:
this is a reboot of https://github.com/pure-data/pure-data/pull/1718

This closes #1717 and takes care of [random], [array random] and [noise~].

Now the seed is set via system time when you load the patch/object at creation time.

A seed message without arguments set unique seeds according to system time again.

In the previous PR we discussed and decided to use the same "makeseed" function for all objects - [noise~] did use a different code to generate the first seed, but now this is unified.

### DIFF
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -473,13 +473,12 @@ typedef struct _noise
     int x_val;
 } t_noise;
 
+int makeseed(void);
+
 static void *noise_new(void)
 {
     t_noise *x = (t_noise *)pd_new(noise_class);
-        /* seed each instance differently.  Once in a blue moon two threads
-        could grab the same seed value.  We can live with that. */
-    static int init = 307;
-    x->x_val = (init *= 1319);
+    x->x_val = makeseed();
     outlet_new(&x->x_obj, gensym("signal"));
     return (x);
 }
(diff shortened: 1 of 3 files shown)


sebshader:
I'm wondering if it would be a good idea to expose some of this (makeseed) in the api..

porres:
> I'm wondering if it would be a good idea to expose some of this (makeseed) in the api...

I dont think so, not for this one

Spacechild1:
Looks good to me!

jayrope:
+1

claudeha:
In one project using libpd I modified Pd like this so that I could offset all `random` default seeds based on time+date (for randomness) or user choice (for reproducibility) before loading the patch:

```diff
diff --git a/src/x_misc.c b/src/x_misc.c
index 39732b4a..80ed012c 100644
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -53,9 +53,9 @@ typedef struct _random
 } t_random;
 
 
+unsigned int random_nextseed = 1489853723;
 static int makeseed(void)
 {
-    static unsigned int random_nextseed = 1489853723;
     random_nextseed = random_nextseed * 435898247 + 938284287;
     return (random_nextseed & 0x7fffffff);
 }
```

Issue #1722, opened 2022-08-19
doc/6.externs/ uses '.l_ia64' for Linux/amd64 binaries


umlaeute:
the correct file extension would be `.l_amd64`.

while `.l_ia64` is supported it is https://github.com/pure-data/pure-data/blob/6aaa6f7d2c5133b848e5fcd66d7f030eb79b85fb/src/s_loader.c#L63

we should strive to reduce the "wide use" by not teaching people to use it.

umlaeute:
speaking of doc/6.externs:

the Makefile has a `pd_linux` target (producing files with the notorious `.l_ia64`) and a `pd_linux32` target (producing files with a `.l_i386` target.

However, the sole difference between the two is the use of the `-fPIC` compiler flag.
Now `-fPIC` is *required* for Linux/amd64 binaries, but it doesn't tell the compiler to produce such binaries.

In fact, running `make pd_linux pd_linux32` on a arm-machine (e.g. RaspberryPi) will produce both files like `obj1.l_i386` and `obj1.l_ia64`, claiming to be of the x86 family, while in reality they are both arm binaries:

```
$ file obj1.l_i386 obj1.l_ia64 
obj1.l_i386: ELF 32-bit LSB pie executable, ARM, EABI5 version 1 (SYSV), dynamically linked, stripped
obj1.l_ia64: ELF 32-bit LSB pie executable, ARM, EABI5 version 1 (SYSV), dynamically linked, stripped
```

i suggest to:
- unconditionally add `-fPIC` to `LINUX_CFLAGS`
- use the generic `.pd_linux` extension

umlaeute:
also the `-Werror` flag is bad, as it currently breaks the build completely (recent versions of gcc emit warnings for "cast[s] between incompatible function types", something that is done in the source code of practically every Pd objectclass.
unfortunately the relevant warning-suppression flag ("`-Wno-error=cast-function-typ`") cannot be used with older versions of gcc (and if used, is an error by itself...)

i don't see why need `-Werror` in the first place.

PR #1723, opened 2022-08-19
fix doc/6.externs build for Linux

umlaeute:
- don't pretend to build i386 or x86-64 binaries (as we can't control this)
  - use .pd_linux as a generic file extension
  - drop the "pd_linux32" target
- unconditionally build with "-fPIC" (it works everywhere)
- build with debugging symbols
- drop the "-Werror" flag as it fails builds with newer gccs.

Closes: https://github.com/pure-data/pure-data/issues/1722

### DIFF
--- a/doc/6.externs/makefile
+++ b/doc/6.externs/makefile
@@ -1,12 +1,10 @@
 current:
-	@echo make pd_linux, pd_linux32, pd_nt, or pd_darwin
+	@echo make pd_linux, pd_nt, or pd_darwin
 
 clean:
 	rm -f *.o
 	rm -f *.pd_linux
 	rm -f *.pd_darwin
-	rm -f *.l_ia64
-	rm -f *.l_i386
 
 # ----------------------- Windows-----------------------
 # note: you will certainly have to edit the definition of VC to agree with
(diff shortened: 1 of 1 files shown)


Spacechild1:
Looks good to me!

(Side note: before we spend too much time fixing this ancient makefile, we should really consider using pd-lib-builder for the example externals.)

danomatika:
I agree, but these are really the basics steps towards modernization, so perhaps easier to swallow. pd-lib-builder is great but not if you want to learn how to use a manual makefile for a pd object, however this may be a goal for less people than it did before. Also, I think it would be good to mention pd-lib-builder in the readme here.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

umlaeute:
we also might want to mention pd-lib-builder in a prominent location in the example Makefile itself.

Spacechild1:
> but not if you want to learn how to use a manual makefile for a pd object

I'm not sure if this is something that most people want/need to learn :-) Since these are examples, I think we should show/teach the canonical way (= pd-lib-builder).

PR #1730, opened 2022-08-30
fix overriding of CFLAGS with regard to debug/optimization builds

### LINKED ISSUE #1727: configure appends "-O2" option to `CFLAGS`
Here is my output when I run `configure` (Msys2, Windows 10):
```
CFLAGS:               -ffast-math -funroll-loops -fomit-frame-pointer -O3  -g -O2
```
This is bad because `-O2` will override `-O3`, preventing lots of optimizations.

Note that my `CFLAGS` environment variable is empty (verified by running `echo $CFLAGS`) and I don't see this flags in `configure.ac`, so I guess this is somehow added by automake.

We *prepend* our flags to `CFLAGS`, `CPPFLAGS`, etc. because we want to allow the user to override them, so it is really unfortunate if automake decides to fill `CFLAGS` with some default options.

My current workaround is to explicitly set `CFLAGS` to empty:
`./configure CFLAGS=`
=>
`CFLAGS:               -ffast-math -funroll-loops -fomit-frame-pointer -O3`

umlaeute:
allow the `--enable-debug` resp (the new) `--enable-optimizations` configure-flags to override autoconf's default FLAGS, but user-provided FLAGS trumps them all.

this introduces the new `--enable-optimizations` flag that adds optimizing CFLAGS (rather than triggering this behaviour by checking whether `--enable-debug` is disabled).
i also renamed the `RELEASE_CFLAGS` to `OPTIMIZE_CFLAGS` because that seems to be a more apt name.

for a lengthy discussion of the new behavior see #1727 

Closes: https://github.com/pure-data/pure-data/issues/1727


INSTALL.txt
```diff
@@ -85,6 +85,28 @@ script settings on the commandline:
...
+compiler to optimize for speed more aggressively (even at the cost of incorrect
+math operations).
+
+    # build with speed optimizations
+    ./configure --enable-optimizations
+
+
+If things go wrong and you need to run Pd through a debugger (like gdb), you can
+build Pd with debugging symbols using the "--enable-debug" flag.
+It's possible to specify both the "--enable-debug" and "--enable-optimizations"
```
Spacechild1, line 100: On the other hand, autotools already adds the `-g` flag by default. Even if it didn't, a release build with debug symbols can be easily achieved with `./configure --enable-optimizations CFLAGS="-g"`.

Generally, I'm not sure if `DEBUG_CFLAG` should really come *before* `OPTIMIZATION_CFLAGS`. This only works if optimizations are *disabled* by default (which not all people would agree with). If they were *enabled* by default, then you would always need to do `--disable-optimizations --enable-debug` to get a useable debug build. I think `DEBUG_CFLAGS` should either come *after* `OPTIMIZATION_CFLAGS` - or `--enable-debug` could even discard `OPTIMIZATION_CFLAGS`.

Generally, I think these two options are really mutually exclusive and it does not make much sense to combine them. After all, `--enable-debug` does not only mean "build with debug symbols" but also "build without optimizations".

umlaeute: > On the other hand, autotools already adds the `-g` flag by default.

My idea was to not assume anything about the default flags. I do not see harm in specifying the flag twice.

> ./configure --enable-optimizations CFLAGS="-g"

I'm not sure I can follow (again): if we do that, we could as well ditch the `--enable-optimizations` flag and tell users to just pass the relevant `CFLAGS`...

> This only works if optimizations are disabled by default (which not all people would agree with).

I for one do not agree. I've been getting useful debugging backtraces with full optimization.
You might not be able to access all variables in the stack, but the backtrace itself is still valuable.

> Generally, I think these two options are really mutually exclusive

Well, you can always not combine them (as the documentation tells you), if they don't work for you when used together.

> After all, --enable-debug does not only mean "build with debug symbols", but also "build without optimizations".

For me `--enable-debug` means: "give me a build that is reasonably usable within a debugger".
...which I think the current implementation does

umlaeute: > I've been getting useful debugging backtraces with full optimization.

here's two backtraces triggered by a simple crasher patch in pure Pd, just to illustrate the usefulness:

### fully optimized, no debugging

imagine i want to optimize Pd for my i7/skylake, so I run something akin to:

```sh
../configure --enable-jack --disable-debug --enable-optimizations CFLAGS="-march=native -mtune=native -mfpmath=sse -msse4 -mfma -mavx"
```

<details>
  <summary markdown="span">but when my patch crashes, i get the following backtrace...</summary>

```gdb
(gdb) bt
#0  0x000055555560d468 in outlet_anything ()
#1  0x0000555555606ee0 in binbuf_eval ()
#2  0x00005555555ebc81 in text_click ()
#3  0x00005555555bb5c7 in canvas_doclick.constprop ()
#4  0x000055555560a472 in pd_typedmess ()
#5  0x0000555555606ee0 in binbuf_eval ()
#6  0x0000555555613246 in socketreceiver_read ()
#7  0x000055555561222b in sys_domicrosleep ()
#8  0x00005555556141d1 in sys_pollgui ()
#9  0x000055555560f8e5 in m_mainloop ()
#10 0x00007ffff7a2920a in __libc_start_call_main (main=main@entry=0x55555556d320 <main>, argc=argc@entry=3, argv=argv@entry=0x7fffffffdfb8) at ../sysdeps/nptl/libc_start_call_main.h:58
#11 0x00007ffff7a292bc in __libc_start_main_impl (main=0x55555556d320 <main>, argc=3, argv=0x7fffffffdfb8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdfa8)
    at ../csu/libc-start.c:389
#12 0x000055555556d371 in _start ()
```

</details>

we obviously can still see the public symbols, but that's about it.

### fully optimized, additional debugging

since the backtrace doesn't show much, lets enable debugging:

```sh
../configure --enable-jack --enable-debug --enable-optimizations CFLAGS="-march=native -mtune=native -mfpmath=sse -msse4 -mfma -mavx"
```

<details>
  <summary markdown="span">which gives us this backtrace...</summary>

```gdb
(gdb) bt
#0  outlet_anything (x=<optimized out>, s=0x555555719c60, argc=0, argv=0x7fffffffd920) at ../../src/m_obj.c:613
#1  0x0000555555606ee0 in binbuf_eval (x=<optimized out>, target=target@entry=0x555555776950, argc=argc@entry=1, argv=argv@entry=0x7fffffffd9a0) at ../../src/m_binbuf.c:789
#2  0x00005555555ebc81 in message_float (f=0, x=0x555555776920) at ../../src/g_text.c:381
#3  message_click (alt=<optimized out>, ctrl=0, shift=<optimized out>, ypos=<optimized out>, xpos=<optimized out>, x=0x555555776920) at ../../src/g_text.c:464
#4  text_click (z=0x555555776920, glist=<optimized out>, xpix=<optimized out>, ypix=<optimized out>, shift=<optimized out>, alt=<optimized out>, dbl=0, doit=1) at ../../src/g_text.c:1340
#5  0x00005555555bb5c7 in canvas_doclick (x=0x555555772440, xpos=93, ypos=72, mod=<optimized out>, doit=1, which=<optimized out>) at ../../src/g_editor.c:2262
#6  0x000055555560a472 in pd_typedmess (x=0x555555772440, s=<optimized out>, argc=<optimized out>, argv=<optimized out>) at ../../src/m_class.c:1084
#7  0x0000555555606ee0 in binbuf_eval (x=<optimized out>, target=0x555555772910, target@entry=0x0, argc=argc@entry=0, argv=argv@entry=0x0) at ../../src/m_binbuf.c:789
#8  0x0000555555613246 in socketreceiver_read (fd=10, x=0x5555557609d0) at ../../src/s_inter.c:685
#9  socketreceiver_read (x=0x5555557609d0, fd=10) at ../../src/s_inter.c:617
#10 0x000055555561222b in sys_domicrosleep (microsec=microsec@entry=0) at ../../src/s_inter.c:234
#11 0x00005555556141d1 in sys_pollgui () at ../../src/s_inter.c:1022
#12 0x000055555560f8e5 in m_pollingscheduler () at ../../src/m_sched.c:351
#13 m_mainloop () at ../../src/m_sched.c:432
#14 0x00007ffff7a2920a in __libc_start_call_main (main=main@entry=0x55555556d320 <main>, argc=argc@entry=3, argv=argv@entry=0x7fffffffdfb8) at ../sysdeps/nptl/libc_start_call_main.h:58
#15 0x00007ffff7a292bc in __libc_start_main_impl (main=0x55555556d320 <main>, argc=3, argv=0x7fffffffdfb8, init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdfa8)
    at ../csu/libc-start.c:389
#16 0x000055555556d371 in _start ()
```

</details>


this is much more useful.
as you can see, some values are "optimized out", but most of the stuff is accessible.
to make the "optimized out" values visible, we would have to build Pd with `-O0`, but than my Pd no longer eats 45% of one CPU  but maxes out.

ymmv on other systems. but for me allowing both flags to be enabled at the same time definitely makes sense.

danomatika:
IOhannes is right not to assume -g is always added by autotools. On macOS I usually have to add it myself and it’s good to throw it in with an —enable-debug flag.

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

INSTALL.txt
Spacechild1: > I for one do not agree. I've been getting useful debugging backtraces with full optimization.
You might not be able to access all variables in the stack, but the backtrace itself is still valuable.

But being able to access all local variables is the very purpose of a debug build!

For me - and major build systems seem to agree - a debug build means:
1. debug symbols
2. no optimizations
3. assertions enabled

The mere presence of debug symbols does not make something a debug build. In fact, CMake offers `RelWithDebInfo`, which is a modestly optimized release build with debug symbols.

We're free to define the meaning of our `--enable-debug` option as we wish, but we have to be aware of the implications.

Now, we could tell users to do `--enable-debug --disable-optimizations` for a *real* debug build (in case optimizations are enabled by default). I find it a bit silly, but it wouldn't be the end of the world.

---

The general problem we're facing is that we're trying to implicitly create various build type configurations by combination of several options. Another solution could be to have more fine grained options (e.g. `--enable-debug-symbols`), but I'm not sure I would go there...

umlaeute: > Now, we could tell users to do --enable-debug --disable-optimizations for a real debug build (in case optimizations are enabled by default). I find it a bit silly, but it wouldn't be the end of the world.

but optimizations (other than the default `-O2`) are *not* enabled by default.

umlaeute: i'd probably just strip away the documentation section that tells people that they can enable both options at the same time.

> We're free to define the meaning of our --enable-debug option as we wish, but we have to be aware of the implications.

yes sure.
but `--enable-debug` will give you a build with **no** optimization.
It's only if you *also* specify `--enable-optimizations` you will get a build that has both debug info and optimizations (quite what i would expect if i typed `--enable-debug --enable-optimizations`)

Spacechild1: > but optimizations (other than the default -O2) are not enabled by default.
> but --enable-debug will give you a build with no optimization.
It's only if you also specify --enable-optimizations you will get a build that has both debug info and optimizations (quite what i would expect if i typed --enable-debug --enable-optimizations)

Sure. I have been talking about the *possibility*. People have already expressed concern with disabling optimizations by default. My point is that we should find a solution that works nicely regardless the (current) default values.

Spacechild1:
@danomatika That's interesting!

Personally, I wouldn't be opposed to always adding `-g`. If you want to save disk space, you can easily strip with `make install-strip` resp. `make app`.

millerpuckette:
I can merge this but I'm not sure how to fix conflicts in configure.ac...

umlaeute:
i've updated the PR so it applies cleanly on the latest master (5c104d7a01d42f22d29554286b491489fc1bb6aa)

umlaeute:
i have no idea why:
- this PR was not closed automatically (presumably because there was still an open "review request")
- there's a "pages  build and deployment" job

anyhow, closing.

PR #1732, opened 2022-08-31
add 'numsymbols' method to [pdcontrol]

### LINKED ISSUE #1731: request: get the current number of created symbols
It would be nice to be able to know how many symbols have been created since Pd started, for debugging and optimizing purpose.
For example we could add a message `[numsymbols(` to `[pdcontrol]`.

Ant1r:
closes #1731 

This allows to query the total number of symbols that have been created since Pd started, by sending `[numsymbols(` to `[pdcontrol]`.

A `pd_numsymbols` integer field has been added at the end of `t_pdinstance` struct. Could this break some compatibility?
It could have been added to `t_instancestuff` instead, but currently it's not possible, since `pd_stuff` is allocated after the first symbols are created.

### DIFF
--- a/doc/5.reference/pdcontrol-help.pd
+++ b/doc/5.reference/pdcontrol-help.pd
@@ -1,9 +1,9 @@
-#N canvas 557 44 727 512 12;
+#N canvas 557 44 727 554 12;
 #X obj 37 19 pdcontrol;
-#X text 483 477 updated for Pd version 0.50.;
-#X obj 42 257 pdcontrol;
-#X obj 42 314 print;
-#X msg 280 324 isvisible;
+#X text 483 517 updated for Pd version 0.50.;
+#X obj 42 247 pdcontrol;
+#X obj 42 304 print;
+#X msg 165 351 isvisible;
 #N canvas 568 591 287 200 subpatch 0;
 #X obj 99 26 inlet;
 #X obj 99 78 pdcontrol;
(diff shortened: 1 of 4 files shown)


Spacechild1:
> A pd_numsymbols integer field has been added at the end of t_pdinstance struct. Could this break some compatibility?

AFAICT, this should be fine since you add it to the end of the struct and I can't imagine any external that would somehow depend on the *size* of `t_pdinstance`.

umlaeute:
what would the performance impact be, if you just counted the entries in the symbol-table instead? (that is: don't use  caching variable).

this obviously will be much slower, but would that be a problem?

Ant1r:
yes, I thought about that.
I don't think it would be a huge problem; it could be enough to interrupt the audio flow though (in a patch of mine, the count could easily grow up to 10e5...).

OTOH a single int is not that expensive...

umlaeute:
an int is not expensive, but it clutters the struct with "unimportant" data and the number needs to be maintained (updated).

otoh, how bad is an audio glitch in this context?
you mentioned "debugging" and "optimizing" as possible use-cases for that info, but that would not need to be exposed in a real-time context (you probably could live with a drop-out in the debugging/optimization phase; and during performance, afterwards you wouldn't need that info any more)

so personally i would probably go for the non-realtime-safe method in a first iteration, and only if it turns out that the non-realtime-safety is a problem in practice, use a cache for that value.

Ant1r:
Ok I agree with you. I'll rewrite it, counting the hashtab directly.

Ant1r:
superseded by  #1735.

PR #1735, opened 2022-09-01
add 'runtimeinfo' and 'canvasinfo' methods to [pdcontrol]

### LINKED ISSUE #1731: request: get the current number of created symbols

Ant1r:
closes #1731

This PR adds 2 methods to [pdcontrol]:
- `[runtimeinfo <selectors...> (` : outputs runtime information, currently:
    - `numsymbols NUMSYMBOLS`: total number of symbols that have been created since Pd started.
    - `numobjects OBJECTS CANVASES`: total number of objects and canvases.
    - `pdversion MAJOR MINOR BUGFIX`: print Pd version numbers.
    
    Optional `selectors` allow to query only desired information;
    it can be any combination of "numsymbols", "numbojetcs" and "pdversion" selectors.

- canvasinfo <level> - outputs information about the canvas, currently:
    - `numobjects OBJECTS CANVASES`: number of objects and canvases in the canvas and its children.
    
    Optional `level` selects the target canvas (0: this patch, 1: its owner, and so on) for counting objects and canvases.

pdcontrol-info

note: the following comments before today (2026/04/15) are outdated and should probably be hidden.


Spacechild1:
Personally, I would rather have dedicated methods for each property. Some properties, such as the symbol table size, might be expensive to calculate - which is fine -, but I wouldn't want to pay this cost for every other unrelated property. @umlaeute might disagree, though.

umlaeute:
>  @umlaeute might disagree, though.

not necessarily :-) (hey, i'm not against *everything*).
do you have any specific parameter in mind?

(and one could always add filtering caps to the `[runtimeinfo(` message...

umlaeute:
> Another idea: `runtimeinfo` could have optional arguments

yes, that's what i meant with "one could always add filtering cap[abilitie]s"

> But I'm not really sure what this would buy us

a.g. a straightforward way to query what parameters could be requested.

e.g. if `[runtimeinfo(` outputs `[numsymbols 1234567, numobjects 10(`, you know that you could call it as `[runtimeinfo numobjects(`.
you could of course also consult the help-patch :yum: 
and the initial run would probably take quite long (as it has to gather all the stuff you are not interested in).

i think what it really buys us is a bit of structure for `[pdcontrol]`.

Ant1r:
I've added `numobjects #objects #canvases` output to `[runtimeinfo(`.

There's an optional float parameter to select the parent-level for counting the objects and canvases:
- -1 : all patches
- 0 : this patch
- 1 : the parent patch
- 2 : parent's parent
- (...)

Capture du 2022-09-06 17-02-47

I've not added "filtering" yet; I quite like it, but I'm not sure it's necessary for now ; it's rather quick to query the numbers, even if it produces a noticeable audio dropout at 15ms with big patches (mine was about ~ 30 000 symbols, 300 000 objects, 30 000 canvases).

Ant1r:
Output filtering is now implemented.
Also, there's a new `pdversion` output.
Other idea(s)?

porres:
I think getting the version is cool and useful, but I already have a PR/commit for that, see https://github.com/pure-data/pure-data/pull/828/commits/8ed7ee60a019f88ab69d80b24b3fb953975c518b

Ant1r:
Oh! Sorry I missed it.
Otoh don't you think it's better to bring `pdversion` into the response to `[runtime info(`, than having a dedicated message?

porres:
I would say I like it better in a dedicated message, but while you are adding this I may have to give it in...

umlaeute:
i think we ought to distinguish between general information of the current Pd-runtime (`numsymbols`, `pdversion`,...) and information about the a given canvas (`numobjects`).

Ant1r:
not that I'm against it, but `numobjects` can also query the total number of objects (and canvases) across all canvases, so it looks like "runtime" information.
Or we could have:
`runtimeinfo [numsymbols] [pdversion] [numobjects]`
and
`canvasinfo [numobjects] [?] `

Ant1r:
@umlaeute did you mean something like that?

umlaeute:
i'm not sure i understand your suggestion in detail.
i'm less worried about the *output* than about the *input* (which messages to send to `[pdcontrol]` to query for info about the runtime resp info about the current patch.

Ant1r:
Sorry it wasn't clear. I propose something like this:

`[runtimeinfo(` sent to `[pdcontrol]` would return: 
`[numsymbols #numsymbols , totalobjects #numobjects #numcanvases , pdversion #major #minor #bugfix (`

and `[canvasinfo #optional-parent-level(` sent to `[pdcontrol]` would return: 
`[numobjects #numobjects #numcanvases(`

Ant1r:
@millerpuckette I think this would be a good help for debugging or optimizing our patches!
see [this discussion](https://lists.iem.at/hyperkitty/list/pd-list@lists.iem.at/thread/ND4OKT4N7VBDFYRJW2OH6CPJOK4PRCSC/#ESKPQ7K4R5Q2VZQQIN4MINR5DVDFZCQK) for example, where the user's problem may well be caused by the symbol hash table becoming overfilled.

umlaeute:
today i wonder why i favoured a generic `[runtimeinfo <details>(` message over dedicated selectors (as proposed in https://github.com/pure-data/pure-data/pull/1735#issuecomment-1234189798) 

> i think what it really buys us is a bit of structure for `[pdcontrol]`.

... probably that. anything that keeps the object from turning into a catch-all bag of all things would be good.

PR #1738, opened 2022-09-05
Feature/knb3

Ant1r:
This is the 3rd reboot of #938.

closes https://github.com/pure-data/pure-data/issues/911
Here is a proposal to add a new knob GUI, associated to "Shift+Ctrl+K" key accelerator.

Originally imported from moonlib/mknob, it has now gained several features, such as:

-    optional arc drawing
-    configurable angular area
-    optional ticks
-    removable cursor
-    +/- mode
-   angular, horizontal, vertical or horizontal+vertical movement control

Capture du 2022-09-05 19-27-13


Ant1r:
It seems we cannot target to `master` anymore.
- can someone redirect this PR to `develop` (I don't seem to have permission to do so), or should I close and open another one?
- is `develop` the right branch to target now?

porres:
hi, it seems that when we change anything in the properties, the square outline and inlet/outlet disappears

porres:
also, if you have two knob objects in your patch and you edit anything in both of them (and then both do not draw square in/outlet), now if you click and drag in any f these two, the other one also responds.

porres:
when changing the 'angle', the 'ticks' should also be updated.

porres:
if you have a range setting that is not simmetrical, say -0.1 to 1.0, the negative range will have the same amount of ticks as the positive range, this doesn't look good - should be evenly spaced

Screen Shot 2022-11-04 at 12 21 17

porres:
hmmm, this already has some conflicts by the way :)

porres:
hope you don't mind suggestions at this point, but instead of start and end angle I think it makes more sense to provide an arc range in angles and an offset for the 360º point. By default the offset would be -90 (or 270). An offset of - or +180 would be to the left and so on.

porres:
> if you have a range setting that is not simmetrical, say -0.1 to 1.0, 

whenever you have negative and positive values, you also get an extra tick for the 0 value, right? The behaviour changes a bit with the number of ticks. You can't have a tick number of 1, for instance. With 2 you get 3 ticks... with 3 ticks you get FIVE! So it seems both positive and negative range are divided by 3... similarly, 4 ticks gives us 7 ticks and so on... this doesn't seem right to me but you may have thought this through and might convince me :)

porres:
> you may have thought this through and might convince me :)

I would say it makes sense to have ticks behave the same no matter the range is

Ant1r:
thanks a lot Alex for testing!
indeed, knb seems to need a little more work...

> when we change anything in the properties, the square outline and inlet/outlet disappears

> two knob objects in your patch and you edit anything in both of them (and then both do not draw square in/outlet), now if you click and drag in any f these two, the other one also responds.

these two are the same, the bug is: when opening properties dialog, a space is appended a the end of the label and send/receive symbols... So if send/receive symbols were initially empty, now they become " ", the outline disappears, and both knobs send and receive to " " channel.
The properties dialog needs additional work anyway, to conform to the latest iemgui code changes.



> when changing the 'angle', the 'ticks' should also be updated.

absolutely...



> this already has some conflicts by the way :)

yep! I'll fix this.

> the negative range will have the same amount of ticks as the positive range, this doesn't look good - should be evenly spaced

> I would say it makes sense to have ticks behave the same no matter the range is

i agree, I'll make this simpler.

> hope you don't mind suggestions at this point

of course not ;-) however I'm not sure about this:
> instead of start and end angle I think it makes more sense to provide an arc range in angles and an offset for the 360º point. By default the offset would be -90 (or 270).

why [range, offset] representation would be better than [min, max] ? Both are equivalent after all.

Anyway the current default is representing the "classical" 270° analog potentiometer, which is [min=-135°, max=135°] (the current convention is that 0° = upward) , or [range=270°, offset = -135°] ; thus I think default offset can't be -90°.

porres:
> why [range, offset] representation would be better than [min, max] ? Both are equivalent after all.

they are equivalent of course, I just feel it's easier to work and think in terms of range and offset

umlaeute:
It's not a problem (unless you would pass `w` as the first arg to `IEMGUI_SETDRAWFUNCTIONS`)

umlaeute:
> I keep the [angle min; angle max] representation for now

This is consistent with how sliders and knob set their *range*.

Just my 2¢

porres:
[knb] is missing an undo function for setting properties, like other iemguis currently have for reverting back from properties changes.

Ant1r:
> compilation warnings could be avoided by simply casting 'x' to 'unsigned long'

done (for [knb]; other GUIs could do it as well).

> [knb] is missing an undo function for setting properties

done too (thanks for watching Alex!)

porres:
So here's the deal, people in plugdata world can't wait to have a knob, I guess I'll work on a variant based on this, but will make it different and incompatible (and named differently). I will still watch this, test it and lobby in favor of its inclusion though. Worst case scenario, Vanilla has a knob, and I will also offer yet another knob external, and PlugData will come with 2 knobs...

Now, tried compiling this as a single object and tested it against 0.53-2 and it seems I can't delete the object. It works fine in 0.53-0. Maybe something changed. Can you try and sync with the current master so we can test how this works?

I'm just gonna go ahead and work on my variant which will completely remove all dependencies from "iemgui" code and make stand on its own.

cheers

porres:
see comment for fix here --> https://github.com/porres/pd-knob/pull/4#issue-1646360496

porres:
Now that I'm working on this, here's more ideas:

change the wiper size according to the knob size, same with the 'ticks' and also adjust sensitivity according to the object size.

also, I see there's a if(glist_isvisible(x->x_glist)) conditional, why not also include "&& gobj_shouldvis((t_gobj *)x, x->x_glist))"?

Ant1r:
I'm not sure I want to put more work in this project without more indication of what it will become.
I don't think @millerpuckette ever commented on this, so I really don't know if:
- there was no time to consider this proposal
- there is no intention to add a round knob to Pd
- this knob's design doesn't look consistently with other GUIs, or suffers from feature creep
- the coding style is bad

So, without further indication, I think I'll close this PR in the near future, before it gets obsolete again and needs more and more work just to keep being tryable.

porres:
I see, I am also frustrated this has been going back and forth for 3 years now and fixes are being needed just to keep up with the code.

The current fix for deleting the object is easy and I can send you a PR. As for keep working on making it better with the suggestions i made to adjust wiper size to object size I can see that it's probably best to have this first included or some positive feedback from Miller.

I guess and hope the main issue was not time to consider this, and with all the multichannel stuff on the way I am afraid this will also fall behind more crucial priorities for the next update.

Even though I am working on an external myself right now, I would really like us to have this in Vanilla, it is an important built-in object in my opinion, and the one I am working on right now is not compatible in style with other iemguis and way more feature creep than this one...

porres:
Ok, I don't know how to fork from your fork and send you a PR @Ant1r, github complains I already have a Pd fork, but the fix is really trivial, just replaces "%lx" from your tags to "%p"

for instance:

 sprintf(tag_object, "%lxOBJ", (unsigned long)x);

becomes:

 sprintf(tag_object, "%pOBJ", x);

if you go ahead and do this, this PR and object is mergeable and workable and we can pressure @millerpuckette to consider this in the next update and get a proper feedback.

I say that with all the tests and things we've been doing over the years grants that this object is well proof tested and ready for inclusion. The only thing that keeps getting in the way are changes to the Pd core and the luck would have it that we went through a major GUI overhaul recently... if it weren't for this, there wouldn't such headaches.

Ant1r:
Thanks Alex. That was a really quick fix this time ;-)

Spacechild1:
I strongly endorse this PR! A knob is such a basic GUI element for audio elements and honestly it seems strange that Pd still doesn't have one.

Issue #1344, opened 2021-06-24
feature request: get rid of number2 and incorporate it into gatom


umlaeute:
i find it ridiculous that we have two separate numberboxes.

instead, make nbx2 look (and behave) the very same as the ordinary gatom by default

i don't have any plan yet how to do the migration.

porres:
> i find it ridiculous that we have two separate numberboxes.

me too

> instead, make nbx2 look (and behave) the very same as the ordinary gatom by default.

Not sure I get it. So when you create an atom number box, you get [nbx] instead? What happens with patches that use the legacy style? And how would you be able to call/force the legacy style? 

> symbol boxes want styling was well...

Yeah => https://github.com/pure-data/pure-data/issues/932

umlaeute:
> Not sure I get it. So when you create an atom number box, you get [nbx] instead?

well, we only want one.
i suggest to use nbx2 as the prototype, as it has more features. e.g. it is stylable.
i think the biggest issue with nbx2 is, that it wastes a lot of space with that weirdo triangle on the left-hand side.

so the plan is to:
- make nbx2 look more like an ordinary gatom by default (same size, no left-hand triangle, active-indicator,...)
- replace gatom by nbx2

> What happens with patches that use the legacy style? And how would you be able to call/force the legacy style?

i think i answered this a-priori with "i don't have any plan yet how to do the migration."

nicolasdanet:
FYI: In my fork i kept the gatom, and replaced the UI of nbx2 only (not the behaviour).
Instead of a number box it is showned as a dial.

![Dial](https://user-images.githubusercontent.com/25415812/123379269-27848280-d58e-11eb-86c2-a802cb59d07a.png)

porres:
seems reasonable to just add a dial knob, by the way => https://github.com/pure-data/pure-data/issues/911

umlaeute:
i'm sure that there's a bunch of stuff that anybody could want or do.
but this issue is about **removing a duplicate element** only.

sebshader:
I think there's some utility to having 2 different objects. gatoms seem to be suited more for placing directly on the canvas whereas the number2 is better for GOP guis since they can have their backgrounds individually set and customized like all of the other iemguis.
Specifically, patches have been built around number and symbol boxes being transparent, and number2 boxes not being transparent (and each having a specific look)
Perhaps the triangle could be removed by default but there would be an option for it.
It would be nice to have a "symbol2" object though

porres:
[...]

Issue #947, opened 2020-04-17
iemguis issues


porres:
This issue is grouping issues related to iemguis just to keep track of them in an organised way.

-------------
**BUGS:**
-------------

Bugs with proposed PRs:

- [vu] has no "send symbol" and its outlets can't be hidden https://github.com/pure-data/pure-data/issues/632 - proposed fix at https://github.com/pure-data/pure-data/pull/654

- number box [nbx] issues with printing 'nan' and 'inf' https://github.com/pure-data/pure-data/issues/629 - proposed fix at https://github.com/pure-data/pure-data/pull/631

- [nbx] clipping values https://github.com/pure-data/pure-data/issues/633 - proposed fix at https://github.com/pure-data/pure-data/pull/631

- inlet/outlet, font and object sizes => https://github.com/pure-data/pure-data/issues/920 - there's a fix at https://github.com/pure-data/pure-data/pull/946

- weird characters like "[" (as label) hang iemgui properties => https://github.com/pure-data/pure-data/issues/880 and a similar (considered duplicate) one => https://github.com/pure-data/pure-data/issues/417 - There's a fix at https://github.com/pure-data/pure-data/pull/948

- '$/#' substitutions on iemguis => https://github.com/pure-data/pure-data/issues/558 - fixed by #953 

- space to underline substitution => https://github.com/pure-data/pure-data/issues/960 - also fixed by #953 

- Slider fixes, no issue reported, but there's a PR https://github.com/pure-data/pure-data/pull/933

- radio's init not working - https://github.com/pure-data/pure-data/issues/730 - fixed at https://github.com/pure-data/pure-data/pull/831

-----------------

Bugs with no PRs yet:

- literal '\' freeze iemguis => https://github.com/pure-data/pure-data/issues/622

- atom number box and number box inconsistency: https://github.com/pure-data/pure-data/issues/851

- vu color inconsistency https://github.com/pure-data/pure-data/issues/699

- non proper names for bang's flash properties https://github.com/pure-data/pure-data/issues/976

-------------
**FEATURES:**
-------------

Features with PRs: 

- add a knob, requested at https://github.com/pure-data/pure-data/issues/911 - proposed by https://github.com/pure-data/pure-data/pull/938

- save colors as hex symbols, as intended, see https://github.com/pure-data/pure-data/pull/937

- allow setting colors individually, requested at https://github.com/pure-data/pure-data/issues/941 - proposed at https://github.com/pure-data/pure-data/pull/942

-----------------

[...]

PR #1749, opened 2022-09-09
indicate nbx activation as with gatom

### LINKED ISSUE #1690: [nbx]: hardcoded red color to show 'activity' is bad in some cases, for instance, you can't have a red background.
[nbx] has a hardcoded color of red for the number when the box is activated, and this is bad if you have a red background as you can't see the number!

I propose we the triangle as the area that indicates the object is activated, we can have it changing the color instead. In fact, as it is, this triangle is rather useless, it seems the functionality is to differentiate it from an atom box, but this would give it a 'higher' purpose. 

The triangle could then be hardcoded to a white background when not active and to the same red color when active. It would also fill to blue when the object is selected. along with the font.

Proof of concept:

nbx-2

nbx-3

nbx-4

nbx-5

nbx-1

### LINKED ISSUE #1753: inconsistency displaying too big numbers in gatom and [nbx]
I really hate how [nbx] collapses to "+" or "-" when it can't display the full number. This also used to be the case in the number atom box but it got fixed: now it shows ">" at the end to indicate truncation. 

Anyway, just saying we should use the same logic and code from atom into [nbx]

### LINKED ISSUE #1755: nbx: remove extra space on the right side
for unknown reasons, the iemgui nbx adds a lot of unused space to the right side (the ordinary number gatom does not).

the space is proportional to the "width" field (a nbx with 10 digits has much more extra space than one with 3 digits), indicating a bug in the text-width calculation.

nbx-padding

i hate it.

umlaeute:
### activation indicator

this simply uses the border width to indicate whether an iemgui `[nbx]` is active or not, similar as the ordinary gatoms do it (which is nice as it doesn't clash with any color concepts)

active-nbx

- Closes: #1690 

### streamlined number shortening algorithm

at also uses the same float-shortening algorithm as the number gatom (so if the number is too big, it displays a trailing `>` rather than replacing the entire number with a `+`

nbx-shortener

- Closes: #1753 

### improved fontwidth calculation
the new calculation uses the system fontspecs (as other objects/gatoms do), getting rid of the empty space on the right hand of the nbx.
for backwards compatibility, we make sure that the new width is not bigger than it used to be (so objects on GOPs will always be visible). this shouldn't be an issue with the default fonts, but might be one, if a weird non-standard font is used).
it's possible to force the old behaviour by setting the compat-level to `<0.53`

nbx-width

- Closes: #1755

### reduce anti-aliasing artifacts with odd heights

previously, the triangle on the left-hand would show an ugly artifact if the nbx2-height was odd. this is now fixed:

image

### use `±inf` as unbound limits

originally, the object used some "very large numbers" (namely `±1e37`) for "virtually no limit".
such large numbers are seldomly encountered in practice.
nevertheless, the limit is arbitrary, and when using Pd64 it is far from the actual minimum/maximum numbers (doubles are in the range of `±1e308`).
i think the proper fix is to use `±inf` to indicate that there is no upper (or lower) limit.

### DIFF
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -28,17 +28,33 @@ static void my_numbox_clip(t_my_numbox *x)
         x->x_val = x->x_max;
 }
 
+int sys_nearestfontsize(int fontsize);
 static void my_numbox_calc_fontwidth(t_my_numbox *x)
 {
-    int w, f = 31;
+    int w;
+    float f = x->x_gui.x_fontsize;
+    int nearest_fontsize =  sys_nearestfontsize(x->x_gui.x_fontsize);
+    int nearest_fontwidth = sys_zoomfontwidth(nearest_fontsize, 1, 1);
+    float fontwidth = (float)nearest_fontwidth * f / (float)nearest_fontsize;
 
-    if(x->x_gui.x_fsf.x_font_style == 1)
-        f = 27;
-    else if(x->x_gui.x_fsf.x_font_style == 2)
-        f = 25;
+        /* legacy font calculation */
+    switch(x->x_gui.x_fsf.x_font_style) {
+    default:
+        f *= 31./36.;
+        break;
+    case 1:
+        f *= 27./36.;
+        break;
+    case 2:
+        f *= 25./36.;
+        break;
+    }
 
-    w = x->x_gui.x_fontsize * f * x->x_numwidth;
-    w /= 36;
+        /* make sure to not exceed the legacy font-size */
+    if ((f < fontwidth) || (pd_compatibilitylevel < 53))
+        fontwidth = f;
+
+    w = (int)(fontwidth * x->x_numwidth);
     x->x_gui.x_w = (w + (x->x_gui.x_h/2)/IEMGUI_ZOOM(x) + 4) * IEMGUI_ZOOM(x);
 }
 
@@ -456,7 +472,7 @@ static void my_numbox_set(t_my_numbox *x, t_floatarg f)
     if (memcmp(&ftocompare, &x->x_val, sizeof(ftocompare)))
     {
         x->x_val = ftocompare;
-        if (pd_compatibilitylevel < 53)
+        if (pd_compatibilitylevel < 56)
             my_numbox_clip(x);
         sys_queuegui(x, x->x_gui.x_glist, my_numbox_draw_update);
     }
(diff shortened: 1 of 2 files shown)


reduzent:
Looks definitely nicer without the unnecessary whitespace on the right. However, it breaks many layouts:

image

What makes matters worse is that there is no way to make the new nbx fit into the box size it had before. At least, it is not breaking functionality as this change _reduces_ size only.

On a related note: When using an odd height, the triangle cannot be rendered with exact 45 degree lines, which looks ugly:

image

This is _not_ a regression of this MR, rather it is something that might be addressed with the same MR.

porres:
I'm following up on gui related stuff in https://github.com/pure-data/pure-data/issues/947 and I'm checking existing PRs with conflicts. Would you be able to solve them for this one @umlaeute ?

umlaeute:
i was thinking about resurrecting this.

@reduzent (or anybody else): do you think that it's ok to break the layout with the reduced amount of whitespace?

millerpuckette:
I have long dreamed of phasing out numbox2 altogether - leaving it in its current form for back compatibility but leaving it off the "put" menu.  Of course there are probably things that it does that the number box doesn't which would have to be added to numbox (if they're actually useful :).

cheers
M

porres:
> I have long dreamed of phasing out numbox2 altogether 

we've been talking about that in here :)  https://github.com/pure-data/pure-data/issues/1344

reduzent:
> i was thinking about resurrecting this.
> 
> @reduzent (or anybody else): do you think that it's ok to break the layout with the reduced amount of whitespace?

Go ahead

umlaeute:
i've updated the PR (and rebased it on current `master`, which is e8b68d43478479deaabfb1f2b310a7a57906226d)

@millerpuckette i understand the wish to only have a single numberbox gatom. however:

the reasoning for this PR is: until the floatatom has grown feature-parity with nbx2, we can remove the papercuts of either of them.
as such, this PR doesn't add any new features; instead it just tries to make the nbx2 experience more consistent with the floatatom experience.

umlaeute:
@millerpuckette regardless of whether numbox2 will go away: it would be great if we could merge this before more work is done on the nbx2 code: I guess, that for Pd-GTK to be useable, nbx2 needs to be ported to the new drawing routines, so now would be a good time for merging :-))

millerpuckette:
on second thought, I don't much care if stuff gets added to nbx that's already in the number box since if we enlarge number box to take over nbx we'll already have it there...

Issue #1739, opened 2022-09-06
fudiparse over TCP/IP


porres:
so, we can use [fudiformat] to send messages over the network via UDP or TCP/IP. I want to confirm the use case of using TCP/IP and I am assuming it only makes sense to send it via binary output, as in [netsend -b].

The problem is that [netreceive] gets a stream of bytes and [fudiparse] doesn't handle it as it only converts message by message. If there is a use case for using TCP/IP and raw FUDI messages, I think [fudiparse] should be able to reconstruct the message by listening to the packet separator. Maybe we need an argument for that?

Spacechild1:
> so, we can use [fudiformat] to send messages over the network via UDP or TCP/IP. I want to confirm the use case of using TCP/IP and I am assuming it only makes sense to send it via binary output, as in [netsend -b].

In danger of stating the obvious, but if you just want to send/receive FUDI over the network, you would simply use `[netsend]`/`[netreceive]` (TCP) resp. `[netsend -u]`/`[netreceive]`....

> I think [fudiparse] should be able to reconstruct the message by listening to the packet separator. Maybe we need an argument for that?

That would be a nice feature nevertheless. Manually assembling the bytes is not too difficult, but it's still a PITA.

porres:
> In danger of stating the obvious: if you just want to send/receive FUDI over the network, you would simply use [netsend]/[netreceive] (TCP) resp. [netsend -u]/[netreceive -u] (UDP)....

yup, that's how I see it. So I still don't get the need or use case for [fudiformat]/[fudiparse]. They seem to be used more for converting bytes to ASCII characters and vice versa, but I'm sure there's gotta be a need for it. It would be nice to make it clear why and put that information on the help file maybe.

Spacechild1:
> So I still don't get the need or use case for [fudiformat]/[fudiparse]

I think the mean use case is sending/receiving FUDI messages with external (low-level) networking objects, such as `iemnet`.

Other use cases:
* send/receive FUDI messages over a serial connection
* assemble/manipulate Pd messages on the byte level, similar to `[list fromsymbol]` and `[list tosymbol]`, but with whole messages instead a single symbols
* embed Pd messages in another protocol, e.g. as an OSC blob in an OSC messages

I think the documentation for these objects could be improved indeed. Especially the following sentence is very misleading:
> The fudiformat object makes FUDI objects suitable for sending over the network via netsend (in UDP mode).
[...]

PR #1750, opened 2022-09-09
html: add doc about FUDI

Lucarda:
This PR comes in response to https://github.com/pure-data/pure-data/issues/1739#issuecomment-1242153884

It just add a new html file to `doc/8.topics/` with contents from https://en.wikipedia.org/wiki/FUDI


Lucarda:
closing this as it has been merged to `porres:Documentation`

Issue #1751, opened 2022-09-10
current master segfaults on startup


Spacechild1:
I compiled latest master (3990e8883511889e6094642b95e223caf4b29f83) on Windows 7 and when I try to run `pd.exe` it would segfault immediately. The culprit seems to be `pdgui_vmess`. Here's a backtrace:
```
#0  addmess (v=...) at s_inter_gui.c:284
#1  pdgui_vamess (
    message=message@entry=0x674b4c93 <__FUNCTION__.3+7507> "set",
    format=format@entry=0x674b4c90 <__FUNCTION__.3+7504> "rS",
    args=args@entry=0x22ef70 "▒LKg") at s_inter_gui.c:384
#2  0x00000000673e92fb in pdgui_vmess (
    message=message@entry=0x674b4c93 <__FUNCTION__.3+7507> "set",
    format=format@entry=0x674b4c90 <__FUNCTION__.3+7504> "rS")
    at s_inter_gui.c:412
#3  0x00000000673ea56e in namelist2gui (
    name=0x674b4c97 <__FUNCTION__.3+7511> "::sys_searchpath",
    namelist=<optimized out>) at s_main.c:1517
#4  0x00000000673edebb in sys_set_searchpath () at s_main.c:1526
#5  0x00000000673e7561 in sys_do_startgui (
    libdir=libdir@entry=0x1ce2030 "C:/github/pure-data/pd-0.52.2")
    at s_inter.c:1425
#6  0x00000000673e7ed0 in sys_startgui (
    libdir=0x1ce2030 "C:/github/pure-data/pd-0.52.2") at s_inter.c:1611
#7  0x00000000673ed591 in sys_main (argc=1, argv=<optimized out>)
    at s_main.c:379
#8  0x00000000004013c1 in ?? ()
#9  0x00000000004014d6 in ?? ()
#10 0x00000000779559cd in KERNEL32!BaseThreadInitThunk ()
   from C:\Windows\system32\kernel32.dll
#11 0x0000000077b8a561 in ntdll!RtlUserThreadStart ()
   from C:\Windows\SYSTEM32\ntdll.dll
#12 0x0000000000000000 in ?? ()
```
Can anyone confirm this? I can have a look myself, but I don't have too much time at the moment...

Ant1r:
no segfault on Linux...

Lucarda:
I can confirm on Win11

```
Thread 1 received signal SIGSEGV, Segmentation fault.
[...]

PR #1752, opened 2022-09-11
pdgui_vmess() fixes and improvements

Spacechild1:
The first commit (10a2e9d756e0179b1c94c066b8dfa479d5a75a1e) fixes a critical crasher bug (https://github.com/pure-data/pure-data/issues/1751).

The remaining 3 commits are mere improvements.

### DIFF
--- a/src/s_main.c
+++ b/src/s_main.c
@@ -1518,7 +1518,7 @@ static void namelist2gui(const char*name, t_namelist*namelist)
     }
     pdgui_vmess("set", "rS",
               name,
-              i, names);
+              (int)i, names);
     freebytes(names, namesize*sizeof(const char*));
 }
 
--- a/src/s_inter_gui.c
+++ b/src/s_inter_gui.c
@@ -91,7 +91,7 @@ static const char* str_escape(const char*s, int size)
 
 typedef struct _val {
     int type;
-    size_t size;
+    int size;
     const char* string;
     union
     {
--- a/src/s_inter_gui.c
+++ b/src/s_inter_gui.c
@@ -306,7 +306,7 @@ static int addmess(t_val v)
     return 0;
 }
 
-static int va2value(const char fmt, va_list args, t_val*v) {
+static int va2value(const char fmt, va_list *args, t_val*v) {
     int result = 1;
     v->type = fmt;
     v->size = 1;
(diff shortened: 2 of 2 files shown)


src/s_main.c
```diff
@@ -1518,7 +1518,7 @@ static void namelist2gui(const char*name, t_namelist*namelist)
     }
     pdgui_vmess("set", "rS",
               name,
-              i, names);
+              (int)i, names);
```
umlaeute, line 1521: good catch.

but i think the fix is wrong. instead, `i` should just be made an `int` (which I did with a8a245d3752af4026a17bb2c1ecb5d7c1a18606e)

umlaeute (commented):
good catch for the `va_list` issue.

i've merged that into `develop` (with my small amendment of a450c160726b24db79205a2ba3c6dd11fffd518d

umlaeute:
> good catch for the `va_list` issue.

hmm. actually it now crashes on Linux :-(

the issue seems to be that on Linux and macOS, `va_list` is implemented as an array (rather than just a pointer).
see https://stackoverflow.com/a/57805088/1169096

hopefully i've submitted a fix with dbcaf3c6b40fe9866ba0d663dd2f91c10d0fab46 that works on all platforms.
this time tested on:
- Linux/amd64
- macOS/amd64
- Wine/amd64
- Wine/i386

Spacechild1:
That's wild! I have read the explanation in https://stackoverflow.com/a/57805088/1169096 several times now and still don't really understand it :-/ How is taking the address of a local variable (`&ap`) different from taking the address of a function argument (`&ap_`)? But as long as it works... :-)

Spacechild1:
Anyway, I can confirm that https://github.com/pure-data/pure-data/commit/dbcaf3c6b40fe9866ba0d663dd2f91c10d0fab46 works fine also on Windows.

Spacechild1:
> How is taking the address of a local variable (&ap) different from taking the address of a function argument (&ap_)? 

Now I get it! An `va_list` array argument actually decays to a pointer to the first argument, so the `&` operator yields a pointer to a pointer to the first argument. The `va_list` array on the stack does not decay, so `&` just refers to the array.

C++ would have caught this at compile time (incompatible pointer types)... *sigh*

Issue #1703, opened 2022-06-30
switching audio  backends with callbacks enabled freezes Pd


reduzent:
When `Callbacks` are enabled in the audio settings, switching audio backends between `jack` and `standard (portaudio)` freezes Pd. GUI objects become immutable, dialogs stop popping up when called. Once in this situation, it can only be recovered by restarting Pd. With `Callbacks` unchecked, switching audio backends works fine.

Test env:
  * Pd 0.52.2
  * macOS 12.1 (M1)
  * JACK 1.9.21 (macOS-universal)

reduzent:
I experience similar behaviour on/with:

  * Pd 0.52.2
  * Ubuntu 22.04
  * JACK 1.9.20

reduzent:
This is addressed by #1756

Issue #1554, opened 2022-01-20
fix callback scheduler (= "Callbacks" in the audio settings)


Spacechild1:
Currently, the callback scheduler has a few issues:

* cannot stop DSP resp. switch to the polling scheduler
* with portaudio, attempting to quit Pd while DSP is running causes a dead lock

We should make sure that we do not directly close the stream within the callback itself but instead signal the main thread. With proper thread synchronization, it should also be able to turn DSP off or switch to the polling the scheduler.

I want to do a PR with several scheduler fixes and this will be part of it. I just add this here for reference (and to remind myself :-)

reduzent:
This is addressed by #1756

Issue #550, opened 2019-01-31
Opening audio settings freeze Pd on windows using callbacks


mxa:
With ASIO (via portaudio) and callbacks active, switching DSP on, then off and then opening the audio settings a will freeze Pd on (this?) MS Windows.

Windows 10
Soundcards: Fireface 400 / Realtek HDA

Lucarda:
Could it be that using "Callbacks" is still experimental?

From the "release notes" on 0.41:
https://github.com/pure-data/pure-data/blob/master/doc/1.manual/x5.htm#L568


I get crashes if I switch DSP off and on with "Use Callbacks" selected.

Windows 8.1 
Pd 32-bit.
Pd 64-bit.
Tascam 4x4.

Spacechild1:
With "Use Callbacks", the scheduler runs inside the low level audio callback, which is not a good idea, especially if you want low latency... So you probably shouldn't use it anyway. But yes, I've also experienced crashes when trying it out.

Lucarda:
@mxa 

Could you include "**using callbacks**" in the title of this issue?

reduzent:
This should be addressed by #1756

Issue #10, opened 2016-03-15
static code analysis


umlaeute:
i've configured [Travis CI](https://travis-ci.org/pure-data/pure-data) and [Coverity Scan](https://scan.coverity.com/projects/pure-data-pure-data) to automatically build and analyse the code-base of Pd.

Currently the configuration is only active in the `travis-ci` branch, but eventually i will create a pull-request to merge it into `master`.

If anybody wants access to the [Coverity Scan dashboard](https://scan.coverity.com/projects/pure-data-pure-data) to see the actual bugs/vulnerabilities/issues, drop me an email.

danomatika:
This is working well so far. Thanks.

danomatika:
Can Travis also run `make distcheck` in addition to `make`?

umlaeute:
sure, see cdd8451

danomatika:
Seems to be working now. Thanks!

PR #1756, opened 2022-09-13
improve scheduler

Spacechild1:
General:
* `glob_exit` breaks from the scheduler loop, instead of calling `exit`, and thus allows for proper cleanup in `sys_main`.
* `sys_main` calls a new function `pd_term` (counterpart to `pd_init`) before returning. Currently, `pd_term` just destroys the root canvases, but LATER it should free *all* resources. Ideally, Pd should not leak any memory on shutdown.
* fix `audio_shouldkeepopen` for the case where no audio backend has actually been opened (yet). Now `pd -jack` immediately connects to the Jack server; before you had to turn on DSP first.
* if the user switches backends or applies audio settings, only open audio if DSP is on or audio should be kept open
* only derive sleep grain from `sys_schedadvance` if we are in the polling scheduler and DSP is turned on, otherwise use a fixed (default) sleep grain of 1 ms.

Callback scheduler:
* allow to toggle DSP
* allow to switch to the polling scheduler
* check if audio device got stuck
* fix high CPU usage with portaudio

Polling scheduler:
* don't sleep if we already slept in `sys_send_dacs` (= `SENDDACS_SLEPT`)

Jack/portaudio
* enable `THREADSIGNAL` by default
* use a semaphore instead of a condition variable to get rid of unnecessary locking
* only sleep on the semaphore when really idle

Portaudio
* rewrote `pa_send_dacs` to match the much simpler `jack_send_dacs`
* show audio I/O error if we have been late and the audio callback had to drop samples (just like the Jack backend already does)
* remove unused code (`FAKEBLOCKING` not defined) to improve readability

Jack
* gracefully handle server shutdown
* log Jack errors with `PD_DEBUG` to avoid spamming the console

MIDI
* only delay outgoing MIDI messages if audio is delayed (= polling scheduler with DSP on)

---

`sys_reopen_audio` and `sys_close_audio` now only notify the scheduler, therefore they can be safely called anywhere, including the audio callback. The actual implementation has been moved to *private* functions (`sys_do_reopen_audio` and `sys_do_close_audio`)

At least here on Windows, the (new) `THREADSIGNAL` implementation allows for significantly lower latency.

- Closes https://github.com/pure-data/pure-data/issues/550
- Closes https://github.com/pure-data/pure-data/issues/1554 
- Closes https://github.com/pure-data/pure-data/issues/1703
- Closes https://github.com/pure-data/pure-data/issues/1930

This branch has been successfully tested with the following configurations:

* Windows / portaudio (ASIO) / polling scheduler + callback scheduler
* macOS / portaudio (CoreAudio) / polling scheduler + callback scheduler
* macOS / Jack / polling scheduler + callback scheduler
* Linux / Jack / polling scheduler + callback scheduler

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -1459,14 +1459,17 @@ void glob_dsp(void *dummy, t_symbol *s, int argc, t_atom *argv)
         int newstate = atom_getfloat(argv);
         if (newstate && !THISGUI->i_dspstate)
         {
-            sys_set_audio_state(1);
+                /* if audio should be kept open, we don't reopen the device,
+                unless it really has been closed (for whatever reason) */
+            if (!audio_shouldkeepopen() || !audio_isopen())
+                sys_reopen_audio();
             canvas_start_dsp();
         }
         else if (!newstate && THISGUI->i_dspstate)
         {
             canvas_stop_dsp();
             if (!audio_shouldkeepopen())
-                sys_set_audio_state(0);
+                sys_close_audio();
         }
     }
     else post("dsp state %d", THISGUI->i_dspstate);
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -1443,20 +1443,20 @@ void canvas_update_dsp(void)
 }
 
 /* the "dsp" message to pd starts and stops DSP computation, and, if
-appropriate, also opens and closes the audio device.  On exclusive-access
+appropriate, also opens and closes the audio device. On exclusive-access
 APIs such as ALSA, MMIO, and ASIO (I think) it's appropriate to close the
 audio devices when not using them; but jack behaves better if audio I/O
-simply keeps running.  This is wasteful of CPU cycles but we do it anyway
+simply keeps running. Also, we want to preserve any connections made between
+Pd and other Jack clients. This is wasteful of CPU cycles but we do it anyway
 and can perhaps regard this is a design flaw in jack that we're working around
-here.  The function audio_shouldkeepopen() is provided by s_audio.c to tell
-us that we should elide the step of closing audio when DSP is turned off.*/
+here. The function audio_shouldkeepopen() is provided by s_audio.c to tell us
+that we should elide the step of closing audio when DSP is turned off.*/
 
 void glob_dsp(void *dummy, t_symbol *s, int argc, t_atom *argv)
 {
-    int newstate;
     if (argc)
     {
-        newstate = atom_getfloatarg(0, argc, argv);
+        int newstate = atom_getfloat(argv);
         if (newstate && !THISGUI->i_dspstate)
         {
             sys_set_audio_state(1);
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -304,6 +304,7 @@ void pd_anything(t_pd *x, t_symbol *s, int argc, t_atom *argv)
 }
 
 void mess_init(void);
+void sched_init(void);
 void obj_init(void);
 void conf_init(void);
 void glob_init(void);
@@ -347,6 +348,7 @@ void pd_term(void)
 void pd_init_systems(void)
 {
     mess_init();
+    sched_init();
     sys_lock();
     obj_init();
     conf_init();
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -327,7 +327,25 @@ void pd_init(void)
     pd_init_systems();
 }
 
-void pd_init_systems(void) {
+void pd_term(void)
+{
+    t_glist *c;
+    for (c = pd_getcanvaslist(); c; c = c->gl_next)
+        canvas_closebang(c);
+#if 0
+        /* Canvases may be slow to close and as a workaround people
+        may want Pd to shutdown quickly. Conversely, others might
+        prefer it if canvases (and all their containing objects) are
+        always freed properly. For now let's exit quickly and LATER
+        figure out a way to handle this. */
+    while ((c = pd_getcanvaslist()))
+        pd_free((t_pd *)c);
+#endif
+    pd_term_systems();
+}
+
+void pd_init_systems(void)
+{
     mess_init();
     sys_lock();
     obj_init();
@@ -337,9 +355,9 @@ void pd_init_systems(void) {
     sys_unlock();
 }
 
-void pd_term_systems(void) {
-    sys_lock();
-    sys_unlock();
+void pd_term_systems(void)
+{
+        /* TODO free resources */
 }
 
 t_canvas *pd_getcanvaslist(void)
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -23,6 +23,7 @@
 #define SYS_QUIT_QUIT 1
 #define SYS_QUIT_RESTART 2
 static int sys_quit;
+static int sys_exitcode;
 extern int sys_nosleep;
 
 int sys_usecsincelastsleep(void);
@@ -211,6 +212,11 @@ void dsp_tick(void);
 
 static int sched_useaudio = SCHED_AUDIO_NONE;
 static double sched_referencerealtime, sched_referencelogicaltime;
+void sys_exit(int status)
+{
+    sys_exitcode = status;
+    sys_quit = SYS_QUIT_QUIT;
+}
 
 void sched_reopenmeplease(void)   /* request from s_audio for deferred reopen */
 {
@@ -422,6 +428,10 @@ static void m_callbackscheduler(void)
 
 int m_mainloop(void)
 {
+        /* open audio and MIDI */
+    sys_reopen_midi();
+    if (audio_shouldkeepopen())
+        sys_reopen_audio();
     while (sys_quit != SYS_QUIT_QUIT)
     {
         if (sched_useaudio == SCHED_AUDIO_CALLBACK)
@@ -437,17 +447,15 @@ int m_mainloop(void)
             }
         }
     }
-    return (0);
+
+    sys_close_audio();
+    sys_close_midi();
+    return (sys_exitcode);
 }
 
 int m_batchmain(void)
 {
     while (sys_quit != SYS_QUIT_QUIT)
         sched_tick();
-    return (0);
-}
-
-void sys_exit(void)
-{
-    sys_quit = SYS_QUIT_QUIT;
+    return (sys_exitcode);
 }
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -9,7 +9,12 @@
 #include "s_stuff.h"
 #ifdef _WIN32
 #include <windows.h>
+#include <sys/timeb.h>
+#else
+#include <sys/time.h>
 #endif
+#include <errno.h>
+#include <pthread.h>
 
     /* LATER consider making this variable.  It's now the LCM of all sample
     rates we expect to see: 32000, 44100, 48000, 88200, 96000. */
@@ -21,8 +26,14 @@
     ((int)(STUFF->st_dacsr /(double)STUFF->st_schedblocksize))
 
 #define SYS_QUIT_QUIT 1
-#define SYS_QUIT_RESTART 2
+#define SYS_QUIT_REOPEN 2
+#define SYS_QUIT_CLOSE 3
 static int sys_quit;
+static pthread_cond_t sched_cond;
+static pthread_mutex_t sched_mutex;
+static int sched_useaudio = SCHED_AUDIO_NONE;
+static double sched_referencerealtime, sched_referencelogicaltime;
+
 static int sys_exitcode;
 extern int sys_nosleep;
 
@@ -210,19 +221,38 @@ void glob_fastforward(void *dummy, t_floatarg f)
 
 void dsp_tick(void);
 
-static int sched_useaudio = SCHED_AUDIO_NONE;
-static double sched_referencerealtime, sched_referencelogicaltime;
+    /* ask the scheduler to quit; this is thread-safe, so it
+       can be safely called from within the audio callback. */
 void sys_exit(int status)
 {
+    pthread_mutex_lock(&sched_mutex);
     sys_exitcode = status;
     sys_quit = SYS_QUIT_QUIT;
+    pthread_cond_signal(&sched_cond);
+    pthread_mutex_unlock(&sched_mutex);
 }
 
-void sched_reopenmeplease(void)   /* request from s_audio for deferred reopen */
+    /* ask the scheduler to (re)open the audio system; thread-safe! */
+void sys_reopen_audio(void)
 {
-    sys_quit = SYS_QUIT_RESTART;
+    pthread_mutex_lock(&sched_mutex);
+    if (sys_quit != SYS_QUIT_QUIT)
+        sys_quit = SYS_QUIT_REOPEN;
+    pthread_cond_signal(&sched_cond);
+    pthread_mutex_unlock(&sched_mutex);
 }
 
+    /* ask the scheduler to close the audio system; thread-safe! */
+void sys_close_audio(void)
+{
+    pthread_mutex_lock(&sched_mutex);
+    if (sys_quit != SYS_QUIT_QUIT)
+        sys_quit = SYS_QUIT_CLOSE;
+    pthread_cond_signal(&sched_cond);
+    pthread_mutex_unlock(&sched_mutex);
+}
+
+    /* called by sys_do_reopen_audio() and sys_do_close_audio() */
 void sched_set_using_audio(int flag)
 {
     sched_useaudio = flag;
@@ -231,13 +261,6 @@ void sched_set_using_audio(int flag)
         sched_referencerealtime = sys_getrealtime();
         sched_referencelogicaltime = clock_getlogicaltime();
     }
-        if (flag == SCHED_AUDIO_CALLBACK &&
-            sched_useaudio != SCHED_AUDIO_CALLBACK)
-                sys_quit = SYS_QUIT_RESTART;
-        if (flag != SCHED_AUDIO_CALLBACK &&
-            sched_useaudio == SCHED_AUDIO_CALLBACK)
-                post("sorry, can't turn off callbacks yet; restart Pd");
-                    /* not right yet! */
 
     pdgui_vmess("pdtk_pd_audio", "r", flag ? "on" : "off");
 }
@@ -335,6 +358,8 @@ static void m_pollingscheduler(void)
 {
     sys_lock();
     sys_initmidiqueue();
+        /* NB: we don't need to lock the scheduler mutex because sys_quit
+        will only be modified from this thread */
     while (!sys_quit)   /* outer loop runs once per tick */
     {
         sys_addhist(0);
@@ -351,6 +376,8 @@ static void m_pollingscheduler(void)
             sched_referencelogicaltime = pd_this->pd_systime;
             continue;
         }
+            /* do at least one GUI update per DSP tick, so that Pd stays responsive
+             * if the scheduler can't keep up with the audio callback */
         sys_pollgui();
         sys_pollmidiqueue();
         sys_addhist(2);
@@ -392,8 +419,11 @@ static void m_pollingscheduler(void)
     sys_unlock();
 }
 
+static volatile int callback_inprogress;
+
 void sched_audio_callbackfn(void)
 {
+    callback_inprogress = 1;
     sys_lock();
     sys_addhist(0);
     sched_tick();
@@ -403,55 +433,100 @@ void sched_audio_callbackfn(void)
     sys_unlock();
     (void)sched_idletask();
     sys_addhist(3);
+    callback_inprogress = 0;
 }
+
+    /* callback scheduler timeout in seconds */
+#define CALLBACK_TIMEOUT 2.0
+
+int sys_try_reopen_audio(void);
 
 static void m_callbackscheduler(void)
 {
+    sys_lock();
     sys_initmidiqueue();
+    sys_unlock();
+        /* wait in a loop until the audio callback asks us to quit. */
+    pthread_mutex_lock(&sched_mutex);
     while (!sys_quit)
     {
-        double timewas = pd_this->pd_systime;
-#ifdef _WIN32
-        Sleep(1000);
-#else
-        sleep(1);
-#endif
-        if (pd_this->pd_systime == timewas)
+        int wasinprogress;
+            /* get current system time and add timeout */
+        double timewas, timeout = CALLBACK_TIMEOUT;
+        struct timespec ts;
+    #ifdef _WIN32
+        struct __timeb64 tb;
+        _ftime64(&tb);
+            /* add fractional part to timeout */
+        timeout += tb.millitm * 0.001;
+        ts.tv_sec = tb.time + (time_t)timeout;
+        ts.tv_nsec = (timeout - (time_t)timeout) * 1000000000;
+    #else
+        struct timeval now;
+        gettimeofday(&now, 0);
+            /* add fractional part to timeout */
+        timeout += now.tv_usec * 0.000001;
+        ts.tv_sec = now.tv_sec + (time_t)timeout;
+        ts.tv_nsec = (timeout - (time_t)timeout) * 1000000000;
+    #endif
+            /* wait for semaphore (with timeout) */
+        timewas = pd_this->pd_systime;
+        wasinprogress = callback_inprogress;
+        if (pthread_cond_timedwait(&sched_cond, &sched_mutex, &ts) == ETIMEDOUT)
         {
-            sys_lock();
-            (void)sys_pollgui();
-            sched_tick();
-            sys_unlock();
+                /* check if the schedular has advanced since the last time
+                   we checked (while it was not in progress) */
+            if (!sys_quit && !wasinprogress && (pd_this->pd_systime == timewas))
+            {
+                pthread_mutex_unlock(&sched_mutex);
+                    /* if the scheduler has not advanced, but the callback is
+                       still in progress, it just blocks on some Pd message.
+                       Otherwise, the audio device got stuck or disconnected. */
+                if (!callback_inprogress && !sys_try_reopen_audio())
+                    return;
+                pthread_mutex_lock(&sched_mutex);
+            }
         }
-        if (sys_idlehook)
-            sys_idlehook();
     }
+    pthread_mutex_unlock(&sched_mutex);
 }
+
+void sys_do_reopen_audio(void);
+void sys_do_close_audio(void);
 
 int m_mainloop(void)
 {
+    pthread_mutex_init(&sched_mutex, 0);
+    pthread_cond_init(&sched_cond, 0);
+
         /* open audio and MIDI */
     sys_reopen_midi();
     if (audio_shouldkeepopen())
         sys_reopen_audio();
+
+        /* run the scheduler until it quits. */
     while (sys_quit != SYS_QUIT_QUIT)
     {
+            /* check if we should close/reopen the audio device. */
+        if (sys_quit != 0)
+        {
+            int reopen = sys_quit == SYS_QUIT_REOPEN;
+            sys_quit = 0;
+            sys_do_close_audio();
+            if (reopen)
+                sys_do_reopen_audio();
+        }
         if (sched_useaudio == SCHED_AUDIO_CALLBACK)
             m_callbackscheduler();
-        else m_pollingscheduler();
-        if (sys_quit == SYS_QUIT_RESTART)
-        {
-            sys_quit = 0;
-            if (audio_isopen())
-            {
-                sys_close_audio();
-                sys_reopen_audio();
-            }
-        }
+        else
+            m_pollingscheduler();
     }
 
-    sys_close_audio();
+    sys_do_close_audio();
     sys_close_midi();
+
+    pthread_mutex_destroy(&sched_mutex);
+    pthread_cond_destroy(&sched_cond);
     return (sys_exitcode);
 }
 
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -216,7 +216,10 @@ static float sched_fastforward;
 
 void glob_fastforward(void *dummy, t_floatarg f)
 {
-    sched_fastforward = TIMEUNITPERMSEC * f;
+    if (sched_useaudio == SCHED_AUDIO_CALLBACK)
+        pd_error(0, "'fast-forward' does not work with 'callbacks' (yet)");
+    else
+        sched_fastforward = TIMEUNITPERMSEC * f;
 }
 
 void dsp_tick(void);
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -222,6 +222,18 @@ void glob_fastforward(void *dummy, t_floatarg f)
         sched_fastforward = TIMEUNITPERMSEC * f;
 }
 
+void sched_init(void)
+{
+    pthread_mutex_init(&sched_mutex, 0);
+    pthread_cond_init(&sched_cond, 0);
+}
+
+void sched_term(void)
+{
+    pthread_mutex_destroy(&sched_mutex);
+    pthread_cond_destroy(&sched_cond);
+}
+
 void dsp_tick(void);
 
     /* ask the scheduler to quit; this is thread-safe, so it
@@ -513,9 +525,6 @@ void sys_do_close_audio(void);
 
 int m_mainloop(void)
 {
-    pthread_mutex_init(&sched_mutex, 0);
-    pthread_cond_init(&sched_cond, 0);
-
         /* open audio and MIDI */
     sys_reopen_midi();
     if (audio_shouldkeepopen())
@@ -545,8 +554,6 @@ int m_mainloop(void)
     sys_do_close_audio();
     sys_close_midi();
 
-    pthread_mutex_destroy(&sched_mutex);
-    pthread_cond_destroy(&sched_cond);
     return (sys_exitcode);
 }
 
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -372,14 +372,16 @@ static void m_pollingscheduler(void)
                 }
                 timeforward = (lateness > 0 ? SENDDACS_YES : SENDDACS_NO);
             }
-            else timeforward = sys_send_dacs();
+            else
+                timeforward = sys_send_dacs();
             sys_addhist(3);
                 /* test for idle; if so, do graphics updates. */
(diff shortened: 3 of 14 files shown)


Spacechild1:
I have been using this branch for several weeks in an opera project with no crashes and no dropouts. My setup was: Thinkpad P14s, Windows 10, RME MADI-Face, 32 input channels, 36 output channels, 256 samples block size, 10 ms latency.

I will make some quick tests on macOS (portaudio) and Linux (Jack), but since I'm mostly running Pd on Windows, it would be great if others (@umlaeute, @porres, @danomatika, @reduzent) could also give it a try.

danomatika:
Here's a Pd 0.53-1 arm64 build for newer mac systems which includes the updated portaudio 19.7.0 and the contents of this branch. It opens and produces audio for the test patch on my macOS 11.7 system. I have not texted further yet.

http://docs.danomatika.com/pdbuilds/0.53/Pd-0.53-1-arm64-schedulerfix.zip

Spacechild1:
@danomatika Thanks!

> It opens and produces audio for the test patch on my macOS 11.7 system

That's good news :-)

danomatika:
@millerpuckette This may be a *tentative* fix for the macOS 13 issue. It probably needs to be looked over and tested on Linux.

danomatika:
Also @millerpuckette , to test this I pulled Christof's branch via:

    git fetch origin pull/1756/head:scheduler_fix
    git checkout scheduler_fix
    git merge master

This merged cleanly, then I could build with the `scheduler_fix` branch changes applied.

millerpuckette:
This is promising but... am I not right that just updating portaudio was enough to fix the Pd hanging problem on MACOS 13?  If so I should imiedialtely push that one out and hold off on this one for a "feature" release.

Spacechild1:
> am I not right that just updating portaudio was enough to fix the Pd hanging problem on MACOS 13?

I think so. This PR is not specific to macOS 13. It just happened that one user was accidentally using the "callbacks" option and wondered why Pd got stuck when stopping DSP. I just mentioned as a side note that this PR would fix all the "callback" related problems - which are not confined to macOS.

> and hold off on this one for a "feature" release.

Definitely!

Spacechild1:
@millerpuckette As a side note: may I go ahead and remove the non-`FAKEBLOCKING` parts in `s_audio_pa.c` (`Pa_WriteStream` and `Pa_ReadStream`)? They have never been used (and most likely never will) and just clutter the code...

Spacechild1:
Seems like it works fine on all configurations. I have removed the [WIP] label and updated the PR description.

reduzent:
> It just happened that one user was accidentally using the "callbacks" option and wondered why Pd got stuck when stopping DSP.

I was using callbacks on purpose and wondered why Pd uses 100% of a core (not knowing that those two are related). This seems to be specific for CoreAudio/maOS. There is no callback option in ALSA and on JACK there is no problem with using callbacks.

Spacechild1:
> This seems to be specific for CoreAudio/maOS. 

It was really a problem with Pd's portaudio backend (`s_audio_pa.c`). I also got the high CPU usage when using ASIO on Windows with "callbacks" enabled.

(For others: this issue is also fixed with this PR.)

reduzent:
I found some crasher conditions that I didn't spot before, either because of sloppy testing or because things have changed.  This is on Ubuntu 22.04 with current scheduler_fix (b5b448f0e5).

### Switching from Jack with callbacks to Jack without callbacks

Switching to using callbacks seems to always work, but when I uncheck 'Use callbacks' in the audio settings dialog and hit [Apply]  a few times, I can reliably crash Pd. Here is the stack trace:

```
gdb -ex=run --args pd -jack -channels 2 -callback
```

```
Thread 16 "pd" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff6c7e640 (LWP 372388)]
__new_sem_post (sem=0x0) at ./nptl/sem_post.c:35
35	./nptl/sem_post.c: No such file or directory.
(gdb) btwatchdog: signaling pd...

#0  __new_sem_post (sem=0x0) at ./nptl/sem_post.c:35
#1  0x0000555555682f89 in sys_semaphore_post (sem=<optimized out>) at s_audio_paring.c:382
#2  0x000055555567c2ca in jack_polling_callback (nframes=128, unused=<optimized out>) at s_audio_jack.c:114
#3  0x00007ffff7d6ab0f in ?? () from /lib/x86_64-linux-gnu/libjack.so.0
#4  0x00007ffff7d6a7d8 in ?? () from /lib/x86_64-linux-gnu/libjack.so.0
#5  0x00007ffff7d87910 in ?? () from /lib/x86_64-linux-gnu/libjack.so.0
#6  0x00007ffff7a94b43 in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:442
#7  0x00007ffff7b26a00 in clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:81
 ```

BTW: I'm using JACK from Ubuntu which probably isn't compiled with debug symbols. Do you need this?

### Switching to ALSA always crashes

When running Pd with JACK, I cannot switch to ALSA. Steps to reproduce the crash:

1. run: `gdb -ex=run --args pd -jack -channels 2 -callback`
2. Go to `Media` -> `ALSA`
3. Hit `Apply`

Backtrace:

```
Program received signal SIGHUP, Hangup.
__futex_abstimed_wait_common64 (private=975940096, cancel=true, abstime=0x7fffffffdce0, op=393, expected=0, futex_word=0x5555556bfe28 <sched_cond+40>) at ./nptl/futex-internal.c:57
57	./nptl/futex-internal.c: No such file or directory.
(gdb) 
(gdb) watchdog: signaling pd...
btwatchdog: signaling pd...

#0  __futex_abstimed_wait_common64 (private=975940096, cancel=true, abstime=0x7fffffffdce0, op=393, expected=0, 
    futex_word=0x5555556bfe28 <sched_cond+40>) at ./nptl/futex-internal.c:57
#1  __futex_abstimed_wait_common (cancel=true, private=975940096, abstime=0x7fffffffdce0, clockid=7442, expected=0, 
    futex_word=0x5555556bfe28 <sched_cond+40>) at ./nptl/futex-internal.c:87
#2  __GI___futex_abstimed_wait_cancelable64 (futex_word=futex_word@entry=0x5555556bfe28 <sched_cond+40>, expected=expected@entry=0, 
    clockid=clockid@entry=0, abstime=abstime@entry=0x7fffffffdce0, private=private@entry=0) at ./nptl/futex-internal.c:139
#3  0x00007ffff7a93f1b in __pthread_cond_wait_common (abstime=0x7fffffffdce0, clockid=0, mutex=0x5555556bfdc0 <sched_mutex>, 
    cond=0x5555556bfe00 <sched_cond>) at ./nptl/pthread_cond_wait.c:503
#4  ___pthread_cond_timedwait64 (cond=cond@entry=0x5555556bfe00 <sched_cond>, mutex=mutex@entry=0x5555556bfdc0 <sched_mutex>, 
    abstime=abstime@entry=0x7fffffffdce0) at ./nptl/pthread_cond_wait.c:652
#5  0x00005555555feb34 in m_callbackscheduler () at m_sched.c:474
#6  m_mainloop () at m_sched.c:516
#7  0x0000555555608f45 in sys_main (argc=5, argv=<optimized out>) at s_main.c:393
#8  0x00007ffff7a29d90 in __libc_start_call_main (main=main@entry=0x55555556e6a0 <main>, argc=argc@entry=5, argv=argv@entry=0x7fffffffdeb8)
    at ../sysdeps/nptl/libc_start_call_main.h:58
#9  0x00007ffff7a29e40 in __libc_start_main_impl (main=0x55555556e6a0 <main>, argc=5, argv=0x7fffffffdeb8, init=<optimized out>, 
    fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdea8) at ../csu/libc-start.c:392
#10 0x000055555556e6d5 in _start ()

```

**Comment by Spacechild1** [comment]
> Switching to ALSA always crashes
> When running Pd with JACK, I cannot switch to ALSA. Steps to reproduce the crash:

AFAICT, the actual issue is trying to switch from an API with callbacks *enabled* to another API that does not support callbacks. I had the same issue trying to switch from ASIO (with callbacks) to MMIO. This is caused by a smal oversight in `glob_audio_setapi`. I just pushed a fix, can you give it a try?

**Comment by Spacechild1** [comment]
> Switching from Jack with callbacks to Jack without callbacks

> Switching to using callbacks seems to always work, but when I uncheck 'Use callbacks' in the audio settings dialog and hit [Apply] **a few times**, I can reliably crash Pd. 

Just to be sure: does "a few times" only refer to "hit [Apply]", or to the whole sequence? Also, are you sure that the crash really happens when going from "callbacks" to "no callbacks" - and not the other way around?

**Comment by reduzent** [comment]
> > Switching to using callbacks seems to always work, but when I uncheck 'Use callbacks' in the audio settings dialog and hit [Apply] **a few times**, I can reliably crash Pd.
> 
> Just to be sure: does "a few times" only refer to "hit [Apply]", or to the whole sequence of actions?

Once the checkbox is unchecked, I can crash it by hitting `Apply` once or multiple times.

**Comment by reduzent** [comment]
Here a report from macOS 12.6.3 (amd64) with b5b448f0e5007. Steps to reproduce:

1. Start Pd in debugger: ` lldb -- /Users/user/Desktop/Pd-scheduler_fix.app/Contents/Resources/bin/pd -jack -channels 2 -callback`
2. Open Dialog `Media` -> `jack`
3. Uncheck `Use callbacks`
4.  Hit `Apply`

I immediately get: 

```
Process 8855 stopped
* thread #7, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007ff81d293882 libsystem_pthread.dylib`pthread_mutex_lock + 4
libsystem_pthread.dylib`pthread_mutex_lock:
->  0x7ff81d293882 <+4>:  cmpq   $0x4d55545a, (%rdi)       ; imm = 0x4D55545A
    0x7ff81d293889 <+11>: jne    0x7ff81d2938fe            ; <+128>
    0x7ff81d29388b <+13>: movl   0xc(%rdi), %eax
    0x7ff81d29388e <+16>: movl   %eax, %ecx
Target 0: (pd) stopped.
```

Backtrace:

```
(lldb) bt
* thread #7, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007ff81d293882 libsystem_pthread.dylib`pthread_mutex_lock + 4
    frame #1: 0x00000001000dd7ae pd`sys_semaphore_post(sem=0x0000000000000000) at s_audio_paring.c:384:5 [opt]
    frame #2: 0x00000001000db249 pd`jack_polling_callback(nframes=64, unused=<unavailable>) at s_audio_jack.c:114:5 [opt]
    frame #3: 0x00000001005ed975 libjack.0.1.0.dylib`___lldb_unnamed_symbol437 + 229
    frame #4: 0x00000001005ed9aa libjack.0.1.0.dylib`___lldb_unnamed_symbol438 + 10
    frame #5: 0x00000001005fd232 libjack.0.1.0.dylib`___lldb_unnamed_symbol899 + 130
    frame #6: 0x00007ff81d2984e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #7: 0x00007ff81d293f6b libsystem_pthread.dylib`thread_start + 15
```

**Comment by Spacechild1** [comment]
> Once the checkbox is unchecked, I can crash it by hitting Apply once or multiple times.

Aha. Does it really happen when you hit Apply only *once*? Or rather when you hit it multiple times? In case of the latter, I would already have an explanation.

**Comment by reduzent** [comment]
> > Once the checkbox is unchecked, I can crash it by hitting Apply once or multiple times.
> 
> Aha. Does it really happen when you hit Apply _once_? Or rather when you hit it multiple times? In case of the latter, I would already have an explanation.

Sometimes it is sufficient to hit `Apply` exactly once (in a fresh run of Pd). More often, it requires more times.

**Comment by Spacechild1** [comment]
> I'm pretty confident (but also not absolutely sure) that this worked before. From what I remember, checking if disabling/enabling callbacks works smoothly was the point of my last testing session.

I does work fine with portaudio here on Windows, so it seems to be a Jack problem (on Linux and macOS) - which is not surprising considering that I have never actually tested the Jack implementation myself :-) Will investigate.

---

Does switching from Jack (with callbacks) to ALSA work now?

**Comment by Spacechild1** [comment]
> Here a report from macOS 12.6.3 (amd64) with https://github.com/pure-data/pure-data/commit/b5b448f0e5007b042d798006f47e285830164b49. Steps to reproduce:

Hmmm, I can't reproduce this on macOS 13.1. I start Pd with `./pd -jack -channels 2 -callbacks` and I can toggle between "callbacks" and "no callbacks" just fine...

**Comment by Spacechild1** [comment]
Actually, b5b448f0e5007b042d798006f47e285830164b49 is outdated. (I have force pushed in the meantime.) Can you try with the current HEAD (f89ae374923a05076d6f5067579b97a76fe986ea)?

**Comment by reduzent** [comment]
> Actually, [b5b448f](https://github.com/pure-data/pure-data/commit/b5b448f0e5007b042d798006f47e285830164b49) is outdated. (I have force pushed in the meantime.) Can you try with the current HEAD ([f89ae37](https://github.com/pure-data/pure-data/commit/f89ae374923a05076d6f5067579b97a76fe986ea))?

I get the same crashes with f89ae37. To make sure that I am using JACK in a well-known state, I re-installed this [package](https://github.com/jackaudio/jack2-releases/releases/download/v1.9.22/jack2-macOS-universal-v1.9.22.tar.gz). The current version is 1.9.22. I was probably using 1.9.21 before. Now, I get a similar behavior to Linux: I have to hit `Apply` multiple times sometimes in order to crash Pd.

**Comment by reduzent** [comment]
BTW: are you testing on an Intel or Arm Mac?

**Comment by Spacechild1** [comment]
I am using an Intel Mac with the same Jack package you have linked (Jack 1.9.22 Universal installer). Can you verify that the crash does not happen with portaudio (CoreAudio)?

**Comment by reduzent** [comment]
Another way to crash Pd (f89ae374923a0) on macOS:

1. Start Pd: `lldb -- /Users/user/Desktop/Pd-scheduler_fix.app/Contents/Resources/bin/pd -pa`
2. Do _not_ turn DSP on
3. Open the Dialog `Media` -> `jack`
4. Crash

```
Process 1907 stopped
* thread #7, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
    frame #0: 0x00007ff80514f882 libsystem_pthread.dylib`pthread_mutex_lock + 4
libsystem_pthread.dylib`pthread_mutex_lock:
->  0x7ff80514f882 <+4>:  cmpq   $0x4d55545a, (%rdi)       ; imm = 0x4D55545A
    0x7ff80514f889 <+11>: jne    0x7ff80514f8fe            ; <+128>
    0x7ff80514f88b <+13>: movl   0xc(%rdi), %eax
    0x7ff80514f88e <+16>: movl   %eax, %ecx
Target 0: (pd) stopped.
```
Backtrace:

```
(lldb) bt
* thread #7, stop reason = EXC_BAD_ACCESS (code=1, address=0x0)
  * frame #0: 0x00007ff80514f882 libsystem_pthread.dylib`pthread_mutex_lock + 4
    frame #1: 0x00000001000dd67e pd`sys_semaphore_post(sem=0x0000000000000000) at s_audio_paring.c:384:5 [opt]
    frame #2: 0x00000001000db119 pd`jack_polling_callback(nframes=128, unused=<unavailable>) at s_audio_jack.c:114:5 [opt]
    frame #3: 0x00000001005ed975 libjack.0.1.0.dylib`___lldb_unnamed_symbol437 + 229
    frame #4: 0x00000001005ed9aa libjack.0.1.0.dylib`___lldb_unnamed_symbol438 + 10
    frame #5: 0x00000001005fd232 libjack.0.1.0.dylib`___lldb_unnamed_symbol899 + 130
    frame #6: 0x00007ff8051544e1 libsystem_pthread.dylib`_pthread_start + 125
    frame #7: 0x00007ff80514ff6b libsystem_pthread.dylib`thread_start + 15
```

**Comment by reduzent** [comment]
Things are different between using Pd's menus and using the mediasettings library. With your branch, Pd doesn't hang anymore when switching backends, even when `use callbacks` is enabled. However, when using the mediasettings' libraries [audiosettings] object to switch backends, Pd still hangs when `callbacks` are enabled. Is there some magic in the Tk menus  to make sure callbacks are off before switching?  If so, should I make sure programming-wise that callbacks get disabled _before_ swtiching backends? I tried that before with Pd from master and I couldn't avoid Pd hanging because it looked like it takes some time before it was safe to switch and I wasn't able to reliably detect how much time was necessary. It still would hang from time to time after switching, even when I waited a few seconds.

**Comment by Spacechild1** [comment]
> However, when using the mediasettings' libraries [audiosettings] object to switch backends, Pd still hangs when callbacks are enabled. 

Can you provide an example patch?

**Comment by Spacechild1** [comment]
@reduzent I finally found the error! I accidentally created the semaphore *after* opening the stream which created a race condition (that was only triggered with very low block sizes).

Now everything should work fine. Can you pull and double check?

**Comment by reduzent** [comment]
> @reduzent I finally found the error! 

Sweet!

> I accidentally created the semaphore _after_ opening the stream which created a race condition (that was only triggered with very low block sizes).

I indeed was running jackd with blocksizes of 64 and 128 (setup from latency testing). This shows how many details are important while reporting. 

> Now everything should work fine. Can you pull and double check?

I confirm I can't crash Pd anymore by hammering the `Apply` button in the `Media` -> `jack` dialog on Linux. Will do some more thorough testing this afternoon, also on macOS.

**Comment by Spacechild1** [comment]
Nice! Please also test with the mediasettings library.

**Comment by reduzent** [comment]
Ok, I did some more testing and
  * switching between JACK and CoreAudio works fine on macOS
  * switching between JACK and ALSA works fine on Linux
regardless of the current callback setting.

Switching between backends does not respect the previous callback setting, they are always off after switching. It's just an observation, I don't think it is necessarily bad.

**Comment by reduzent** [comment]
Hammering `Apply` in the `jack` dialog does not crash anymore.

**Comment by reduzent** [comment]
Switching between backends with the [audiosettings] (from [mediasettings](http://deken.puredata.info/library/mediasettings)) external locks Pd, when callbacks are enabled. It seems any audio processing is stopped and Pd doesn't react to GUI interactions. Switching works fine while callbacks are off. I tried to avoid locks by switch off callbacks with each `driver` change with a message like `params @callback 0, driver $1`  but it still locks Pd. 

See attached patch.
[audiosettings-test.zip](https://github.com/pure-data/pure-data/files/10943703/audiosettings-test.zip)

**Comment by reduzent** [comment]
@Spacechild1 One other thing: When I load netpd-sessions in the `scheduler_fix` build, I experience some glitches. For instance, One synth goes into fully overblown mode after the third loop, like if an oscillator would be amplified by a large value. I had similar issues when work on multi-channel stuff was going on. Current master is fine, though. Could it be that `scheduler_fix` is not on par with current master regarding mutlichannel changes?  I was reluctant to merge current `master` into your `scheduler_fix` branch because I wanted to make sure we test on the exact same thing and wanted to avoid another dimension of complexity. But maybe it would be good to rebase `scheduler_fix` onto current `master`? Or what do you think?

**Comment by Spacechild1** [comment]
> Switching between backends does not respect the previous callback setting, they are always off after switching. It's just an observation, I don't think it is necessarily bad.

This is expected. When you change the audio backends, most audio settings are reset to the default.

> But maybe it would be good to rebase scheduler_fix onto current master?

Done.

> Switching between backends with the [audiosettings] (from [mediasettings](http://deken.puredata.info/library/mediasettings)) external locks Pd

Thanks, will have a look!

**Comment by Spacechild1** [comment]
The problem is really in the `[audiosettings]` object:

https://git.iem.at/pd/mediasettings/-/blob/master/audiosettings.c#L665-667

This will deadlock when called from within the audio callback, i.e. if callbacks are enabled.

The general problem is that the `[audiosettings]` library makes assumptions about Pd internals. For the `[params(` method it actually just sends a Pd message, which in turn will do the right thing. Unfortunately, it can't do the same for the `[driver(` method because the `"audio-setapi"` message would open the audio dialog. As a workaround, it mimicks the (old) implementation of `glob_audio_setapi`, but this does not work with the callback scheduler. At the moment, it is not possible for `[audiosettings]` to change backends once your in the callback scheduler.

Here's a solution: we could add a second (optional) argument to the `"audio-setapi"` message to surpress the audio dialog. Then the `[audiosettings]` object could use that message and wouldn't have to worry about Pd scheduler internals.

I am curious what @umlaeute thinks.

**Comment by Spacechild1** [comment]
> I tried to avoid locks by switch off callbacks with each driver change with a message like params @callback 0, driver $1 but it still locks Pd.

This does not work because the `[driver ...(` message is still sent from within the audio callback. You need to add some delay so that Pd has a chance to exit the callback scheduler and enter the polling scheduler. Adding a 1 second delay between `[params @callback 0(` and `[driver 3(` seems to do the trick - at least here on Windows when switching from Portaudio with callbacks to MMIO. Needless to say, this is just a workaround and not a real solution.

**Comment by reduzent** [comment]
> > But maybe it would be good to rebase scheduler_fix onto current master?
> 
> Done.

Cool. The problem with some session producing "wrong" audio is gone. I'm glad it  was not related to scheduler_fix stuff.

**Comment by reduzent** [comment]
> > I tried to avoid locks by switch off callbacks with each driver change with a message like params @callback 0, driver $1 but it still locks Pd.
> 
> This does not work because the `[driver ...(` message is still sent from within the audio callback. You need to add some delay so that Pd has a chance to exit the callback scheduler and enter the polling scheduler. Actually, a `[del 0]` between `[params @callback 0(` and `[driver ...(` already seems to do the trick. (Tested on Windows by switching from Portaudio + callbacks to MMIO.) Needless to say, this is just a workaround and not a real solution.

Ok, I didn't test with `[del 0]` between the `callback 0` and the `driver $1` message. That would be a totally viable solution for me. Before `scheduler_fix` the time required was totally non-deterministic and in the range of seconds, so it was really hard to not run into a lock from time to time. 

Now with 6b0d66b, Pd locks always when setting driver to JACK with `driver 5` to `[audiosettings]`. Switching backends through the menu is still fine, though. Tested on macOS and Linux.

**Comment by Spacechild1** [comment]
The problem is that `[audiosettings]` directly calls `sys_reopen_audio` from a Pd message, so Pd would be locked. `jack_open_audio` sets a callback when the sample rate changes. Strangely, that callback is invoked immediately. The samplerate callback locks Pd before updating the DSP graph - which will in turn dead lock because Pd is already locked.

Actually, now that I think of it, `sys_reopen_audio` and `sys_close_audio` should probably just set a flag and notify the scheduler, i.e. take the role of `sched_reopenaudio` and `sched_closeaudio`. The actual implementation could be moved to (private) helper functions. This would allow externals to safely call `sys_reopen_audio` and `sys_close_audio` inside Pd messages and the `[audiosettings]` object would work as expected. I will give it a shot.

**Comment by Spacechild1** [comment]
@reduzent With [e48d2e3](https://github.com/pure-data/pure-data/pull/1756/commits/e48d2e37d1df3520ab83e875ee9ff5606821b55f) the `[audiosettings]` object should work again. Now you can even change backends, with callbacks enabled, without any workarounds!

`sys_close_audio` and `sys_reopen_audio` are now 100% safe, as they just notify the scheduler. (They essentially replace `sched_reopenaudio` and `sched_closeaudio`.) The actual implementation has been moved to *private* functions `sys_do_reopen_audio` and `sys_do_close_audio`.

Maybe that's the better design after all: make the public functions safe and move the risky stuff into private functions.

**Comment by reduzent** [comment]
> @reduzent With e48d2e3 the `[audiosettings]` object should work again. Now you can even change backends, with callbacks enabled, without any workarounds!

Sweet potato! Tested on Linux and it works smoothly through menu and `[audiosettings]`. That's great!

**Comment by Spacechild1** [comment]
> Sweet potato! Tested on Linux and it works smoothly through menu and [audiosettings]. That's great!

Yay! I have now integrated the last commit into the previous commits for a cleaner history. Looks like I'm done for now. Thanks for all the testing!

**Comment by reduzent** [comment]
> > @reduzent With [e48d2e3](https://github.com/pure-data/pure-data/commit/e48d2e37d1df3520ab83e875ee9ff5606821b55f) the `[audiosettings]` object should work again. Now you can even change backends, with callbacks enabled, without any workarounds!

Now, I tested also on macOS and it works all well. While I write this, I see new commits being pushed, so maybe I should mention that I tested specifically e48d2e37.

@millerpuckette The handling of back-ends of this PR is a clear improvement over what we had until now. This is well tested. I hope this PR is considered for merging.

**Comment by Spacechild1** [comment]
> While I write this, I see new commits being pushed, so maybe I should mention that I tested specifically https://github.com/pure-data/pure-data/commit/e48d2e37d1df3520ab83e875ee9ff5606821b55f.

I just squashed some commits. I double checked and everything still works on my side, so it should be ok.

I think this PR is now ready for merging.

**Comment by danomatika** [comment]
> Maybe that's the better design after all: make the public functions safe and move the risky stuff into private functions.

In general, I would say YES. ;)

**Comment by Spacechild1** [comment]
Aaand some more commits.

**Comment by reduzent** [comment]
Now, this branch has conflicts with master. Hopefully, this gets merged soon and doesn't require multiple rounds of conflict fixes.

BTW: I've been using a Pd from this branch since March on Linux and macOS and I haven't encountered any new issues. It's a clear improvement, I can't stretch that enough.

**Comment by millerpuckette** [comment]
I think these changes will be very useful... however, the original m_sched was largely system-independent, with things such as locking and thread management quarantined in (the very ugly) s_inter.c . This is important for people wanting to use libpd in bare-metal applications. An example is espd (msp.ucsd.edu/ideas/espd), which uses m_sched.c but not s_inter.c.

Perhaps it would be possible to move all the dreadful semaphores and system clock functions to a new file, perhaps s_audio_(something).c?

**Comment by danomatika** [comment]
Maybe something like s_inter_sched.c with a common internal header? Then you could build Pd with an alternate dummy file with stubs for a minimal scheduler?

enohp ym morf tnes
-----------
Dan Wilcox
danomatika.com
robotcowboy.com

**Comment by millerpuckette** [comment]
This is important but will probably take a while to work through so I suggest we plan on fixing and adopting this for 0.55 rather than 0.54.

**Comment by Spacechild1** [comment]
I have rebased on top of 0.54 and added a 0.55 milestone.

> Perhaps it would be possible to move all the dreadful semaphores and system clock functions to a new file, perhaps 

Regarding `m_sched.c` I think we only need to move the clock stuff from `m_sched.c` to a dedicated file (e.g. `m_clock.c`). Then we would only build `m_sched.c` for the actual Pd app. But can we do that as a second step *after* this PR has been merged? The change set is already pretty large...

[I think we should also think about breaking up `s_inter.c` as it contains a whole bag of unrelated functionality. For example, on bare metal I probably won't need the GUI communication code, but I still might want to register poll functions and do network I/O. There probably should be a larger discussion about restructuring some of the codebase for better modularity, but please not in this PR :)]

---

BTW, I just tested again today and here on Windows 10 I can now run Pd at 64 samples with 3 ms delay :)

**Comment by reduzent** [comment]
I often turn DSP off during operations that would otherwise trigger multiple DSP graph calculations. It appears that with this branch, turning DSP off and on quickly enough makes Pd lose sound. DSP is still calculated, but soundcard is silent. Running another cycle of DSP off and on brings audio back.

This does not happen when using JACK, but it happens with the other backends. I tested:
  * Linux / ALSA
  * macOS / "standard (portaudio)" (CoreAudio)
  * Windows (I only got reports, didn't test myself)

Check attached patch. Make sure that audio is running and you hear the sine. When you hit the bang, audio is turned off. The [env~] indicates that DSP is still running. Another click on the bang, brings audio back. 

Expected behaviour: Audio is always on. 

[test-dsp-off-on.zip](https://github.com/pure-data/pure-data/files/13513608/test-dsp-off-on.zip)

**Comment by Spacechild1** [comment]
Good catch! Should be fixed with the latest commit. Please try with Jack and portaudio.

**Comment by millerpuckette** [comment]
I'd like to adopt this but I simply disagree with not having quit mean really quit - I've seen too many situations where Pd hangs trying to quit cleanly.  (For example, patches with lots of copies of the same abstraction can take a long time to close; a workaround is simply to quit Pd without closing the patch.  I've also seen Pd hang trying to close audio and/or MIDI on various platforms.

Meanwhile, I want to change the startup order in Pd to open patches and handle "-send" startup args before waiting for the startup message from the GUI (glob_initfromgui) to fix an ugly race condition in pd~  - but I think that change will conflict with this PR.  What do you think of my trying to cherry-pick from this one (everything except the clean-shutdown bit) before trying to do that?

**Comment by Spacechild1** [comment]
> What do you think of my trying to cherry-pick from this one (everything except the clean-shutdown bit) before trying to do that?

I think manually cherry-picking all these commits is a bit risky :) If really needed, I can just remove that one commit and force push. But first consider the following:

>  I've also seen Pd hang trying to close audio and/or MIDI on various platforms.

Here's the original code for `glob_exit`:
```
void glob_exit(void *dummy, t_float status)
{
    sys_exit();
    sys_close_audio();
    sys_close_midi();
    if (sys_havegui())
    {
        sys_closesocket(INTER->i_guisock);
        sys_rmpollfn(INTER->i_guisock);
    }
    exit((int)status);
}
```
As you can see, this first closes the audio devices, then closes the GUI and finally exits.

This is logically equivalent to my commit. In fact, I think my code is *less* likely to hang with the callback scheduler, because it doesn't try to quit it on the audio thread (which may indeed block).

The only difference at the moment is that I also close all the parent canvases for a graceful shutdown. This is mainly done for 2 reasons:

1. by closing all canvases before exiting, we make sure that the external destructors are run and custom "closebang" actions are triggered. (For example, I have patches where the config is autosaved on closing.)

2. `pd_term_systems` is still empty, but the idea is that this is place where we properly free all global resources.
    * IMO Pd should not leak memory on shutdown, so we have a chance to catch *actual* memory leaks!
    * It would also allow us to properly free all the classes, which would also call the user-provided `t_classfreefn`; at the moment, `class_free` is never called!
       (The class free function would be very handy for external libraries that really need to be properly terminated; I have seen VST plugin DLLs segfault on exit because we have no chance to properly deinitialize it.)

>  For example, patches with lots of copies of the same abstraction can take a long time to close; a workaround is simply to quit Pd without closing the patch

That seems to be your actual issue. Personally, I don't like that closing the Pd app has a different behavior than subsequently closing all windows. This is just not very intuitive. See point 1. above.

I mean, if a patch takes a long time to close, than it probably also takes a lot of time to open anyway... This is seems to be a bit of an edge case, and I am not sure it outweights the gains. In theory, we could just add some force quit message and/or menu entry that bypasses the graceful shutdown.

If you *really* don't want Pd to close the canvases, I would suggest that I just wrap the relevant code in a `#if 0` clause, so at least I (and others) can easily activate it :)

**Comment by millerpuckette** [comment]
The most recent problem I've had is that abstractions stupidly bind themselves to "pd-myabstraction" style symbols, and unbinding them is order n^2.  If you have tens of thousands of them this makes patches VERY slow to close.  I want to figure out a good fix for that someday.  But meanwhile, what if quitting Pd called all the closebangs but didn't actually take the time to close all the patches?  Or even better, what if that were the default and there were a startup flag or something to ask that patches really get closed (as you prefer)?

**Comment by millerpuckette** [comment]
Meanwhile I'll email my current, not fully tested, s_main.c and s_inter.c so you can see if there are going to be conflicts... github doesn't want me to attach them here :)

**Comment by Spacechild1** [comment]
> The most recent problem I've had is that abstractions stupidly bind themselves to "pd-myabstraction" style symbols, and unbinding them is order n^2. If you have tens of thousands of them this makes patches VERY slow to close. 

:-D

> But meanwhile, what if quitting Pd called all the closebangs but didn't actually take the time to close all the patches? Or even better, what if that were the default and there were a startup flag or something to ask that patches really get closed (as you prefer)?

That's a good idea! I would suggest that I just change
```
void pd_term(void)
{
    t_glist *c;
    while ((c = pd_getcanvaslist()))
    {
        canvas_closebang(c);
        pd_free((t_pd *)c);
    }
    pd_term_systems();
}
```
to
```
void pd_term(void)
{
    t_glist *c;
    while ((c = pd_getcanvaslist()))
    {
        canvas_closebang(c);
#if 0
        pd_free((t_pd *)c);
#endif
    }
    pd_term_systems();
}
```
That would at least call the closebangs. LATER we can figure out how to handle the actual canvases. What do you think?

Spacechild1:
>  I want to figure out a good fix for that someday.

I would be curious to know how slow it would be with https://github.com/pure-data/pure-data/pull/1760. This implementation uses arrays instead of linked lists for the bindlists, so I would expect unbinding to be much faster.

Spacechild1:
I just updated `pd_term` to not free the canvases by default and force pushed.

Spacechild1:
@millerpuckette Looks like it still merges cleanly, even after your recent changes. Let's take the opportunity :)

millerpuckette:
hmm... this breaks libpd - which depends on sched_tick always advancing by one DSP tick.  Is there a way to get this to work and preserve the fact that sched_tick always advances time?

Spacechild1:
>  Is there a way to get this to work and preserve the fact that sched_tick always advances time?

I didn't touch `sched_tick` at all... What exactly isn't working?

millerpuckette:
I think it's that sys_quit keeps getting set - which causes sched_tick to bail without ticking.

Spacechild1:
Good catch! I think I found it. Can you try with the latest commit?

Spacechild1:
PS: the CI only fails because of a trailing whitespace in `s_inter.c` on current master :)

danomatika:
Thanks for checking libpd BTW. This reminds me to make a PR with the current libpd C sources for 0.55.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Apr 8, 2024, at 4:26 PM, Christof Ressi ***@***.***> wrote:﻿
PS: the CI only fails because of a trailing whitespace in s_inter.c on current master :)

Spacechild1:
Hmmm... I'll have a look.

Spacechild1:
Ok, the crash with `-nosound` should be fixed. I also caught a stupid regression in my recent `pd_term` fixup that would keep Pd in an infinite loop 🤦 

Is the libpd issue resolved?

porres:
@Spacechild1  is this that thing you mentioned in that SC talk?

Spacechild1:
@millerpuckette Do you see any more problems?

Spacechild1:
@millerpuckette Awesome, thanks a lot!

danomatika:
Great! I think this helps with a lot of things.

@Spacechild1 Is there anything to change in lined for this?

Spacechild1:
> @Spacechild1 Is there anything to change in libpd for this?

I don't think so!

PR #1759, opened 2022-09-14
improve stack overflow protection

Spacechild1:
The current overflow protection only works for a single recursion path. For more than one path (e.g. several outlets feeding back to the same inlet), Pd would end up in an infinite loop.

Solution: once the stacklimit is reached, I set an overflow flag to prevent any further messaging until the stack got completely unwound.


umlaeute:
infloop

```
#N canvas 0 0 450 300 12;
#X obj 101 106 t f;
#X obj 142 105 t f f;
#X msg 143 73 0;
#X connect 0 0 1 0;
#X connect 1 0 0 0;
#X connect 1 1 0 0;
#X connect 2 0 1 0;
```

patch looks good to me.

Spacechild1:
@umlaeute's Pd patch is indeed a minimal demonstration of the issue; without this fix, Pd will end up in an infinite loop.

PR #1760, opened 2022-09-14
make `pd_unbind` safe

### LINKED ISSUE #2087: [garray] rename crash
When calling the rename function on a garray, there is a potential crash due to pd_unbind being called while we are looping over a linked list in bindlist_anything.

https://github.com/pure-data/pure-data/blob/c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c/src/m_pd.c#L105-L111

When pd_typedmess is called with an argument of "rename", it will call garray_rename() -> pd_unbind(), which will then change the linked list structure and can deallocate several elements from the linked list. Since the deallocation of the linked list elements is very recent, it will often not lead to a crash, but it's dangerous nonetheless.

I've tried storing the linked list elements in an array before looping over them, but that does not work since pd_unbind can deallocate multiple elements from the linked list, not just change the list structure.

Spacechild1:
Currently, it is unsafe to dynamically unbind symbols because some other object could be sending to them.

The solution is to simply make a copy of the bind list before iterating over it. A counter keeps track of the total number of items. We use stack allocation up to a certain limit and then switch to heap allocation to avoid a stack overflow.

(In a way, this not much different than how `[list]` has to output a copy of the stored list to prevent other objects from modifying the list while iterating over it.)

To make copying as fast as possible, I changed the bind list implementation from a linked list to a dynamically sized array. My benchmarks show that the new implementation is even a little bit faster overall - despite the additional copy (and even if we trigger heap allocation by having more than 1024 elements). Here's the benchmark patch:
[bindlist-benchmark.zip](https://github.com/pure-data/pure-data/files/9570465/bindlist-benchmark.zip)

This PR would finally make it possible to safely implement dyamically settable `[receive]` objects (https://github.com/pure-data/pure-data/pull/604), and make existing similar objects safe, for example IEM GUI's `[receive(` method or the `[rename(` method for graphical arrays.

Fixes https://github.com/pure-data/pure-data/issues/2087

### DIFF
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -4,6 +4,7 @@
 
 #include "m_pd.h"
 #include "m_imp.h"
+#include "m_private_utils.h"
 #include "g_canvas.h"   /* just for LB_LOAD */
 
 #include <string.h>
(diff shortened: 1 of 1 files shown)


src/m_pd.c
```diff
@@ -50,64 +64,99 @@ void gobj_save(t_gobj *x, t_binbuf *b)
...
 /* deal with several objects bound to the same symbol.  If more than one, we
 actually bind a collection object to the symbol, which forwards messages sent
-to the symbol. */
+to the symbol.
+CHR: We make a copy of the bindlist before messaging, so we can safely
+unbind symbols while iterating over the bindlist. We allocate on the stack
+up to a certain limit where we switch to heap allocation instead */
 
-static t_class *bindlist_class;
+static PERTHREAD int stackcount = 0;
```
Ant1r, line 72: Shouldn't `stackcount` be PERINSTANCE rather than PERTHREAD?
(Of course PERINSTANCE doesn't exist unfortunately... I mean adding it to instance stuff)

Spacechild1: That's not necessary. Recursive Pd message calls always stay in the same thread. Since the stack count always goes back to 0, there is no state that has to be preserved. See also https://github.com/pure-data/pure-data/blob/ade0160ca291f7fb010ad15c973ed7fce7e1aae5/src/m_obj.c#L353

Ant1r: Yes I also saw the `stackcount` in `m_obj.c`, and I had the same interrogation ;-)
Now I understand that it's not persistent data, so there's no risk two pd instances in the same thread could use the same data concurrently. Thanks  for the explanation!

Spacechild1:
Updated for Pd 0.56.1.

porres:
ping @millerpuckette 

this is an old one, and something I'd like to see it happening because of https://github.com/pure-data/pure-data/pull/604

could we consider it for 0.57?

millerpuckette:
Yeah, it's time to go ahead with this :)

PR #1765, opened 2022-09-21
toolkit-agnostic Core->GUI Communication (Draft B)

umlaeute:
this is just another draft, in addition to #1763 to resolve #1695 

i started with a simple GUI-object (namely `cnv`, as it doesn't have any iolets), and implemented:
- a common framework for how to create "widgets" on the GUI side
 - this mostly establishes a pseudo-API for common tasks, like creating/deleting/moving/selecting/reconfiguring widgets
- an implementation of `cnv` that lives entirely on the GUI side
- simplify g_mycanvas.c to do that...


tcl/pd_widget.tcl
```diff
@@ -80,9 +80,9 @@ proc ::pd::widget::_call {behaviour obj args} {
 }
 
 #
-proc ::pd::widget::create {type obj cnv} {
+proc ::pd::widget::create {type obj cnv posX posY} {
```
giuliomoro: What's the use of `posX` `posY` if the width/height are unspecified? See `bang::create()` : you are left with a zero-width/height element placed somewhere else than `(0, 0)`, but that's not much more useful than having it at `(0, 0)`... If `::create()` just becomes a shorthand for also initializing an object's properties, then why not giving it full opt parsing capabilities like `config`? Or take `args` that can be type-specific?

umlaeute: to quote from the commit message:
> add initial position to ::pd::widget::create
>
>    - this matches the object-creation on-disk more closely ("#X obj 100 100 foo ...")
 >   - we need to set an initial position anyhow...
>    - saves us a line on the core :-)

my vague idea is:
- Pd itself knows where the object is to be positioned, for all types of "widgets".
  i'd like to think that this constructor could be called by Pd, rather than the object's "widgetbehaviour".
  (in reality this is of course not true; for Pd a `[bng]` object looks very much like a `[pack]` object - both are of type `obj`)
- all the other properties seem to be really objectclass-specific, so it makes sense to keep these two apart
- the (GUI)constructor is typically called only "once", whereas the properties might change more often. rather than duplicating the arguments in the calls to both the `create` and the `config` proc (from a Pd-core POV), it seemed to me nicer to just call the `config` in addition to `create`.
- i also have this vague idea that a single GUI-widget can live on multiple canvases at the same time. while their config (and relative movement) can be shared across the canvases (i think), this is not true for their initial position (see GOP: on the parent canvas, your numberbox might live on +100+100, whereas on the subpatch it might live on +200+532)

giuliomoro:
running  5d47903d95214cbc652224d096419ed78b791919 this  with
```
#N canvas 684 56 596 539 12;
#X obj 98 265 bng;
```
patch gives me:
```
(Tcl) INVALID COMMAND NAME: invalid command name "::pd::canvas::add_object"
    while executing
"::pd::canvas::add_object $cnv $obj"
    (procedure "::pd::widget::create" line 7)
    invoked from within
"::pd::widget::create bang 0x7fe76c807400 .x7fe76bd1fc50.c 98 265 "
    ("uplevel" body line 2)
    invoked from within
"uplevel #0 $docmds"set cursor to runmode_nothing
```

umlaeute:
> running https://github.com/pure-data/pure-data/commit/5d47903d95214cbc652224d096419ed78b791919 with

try again with 41ac6a6c6ca309c3cceb7d5515b2c5d998a280b6.

sorry for the bogus push.

giuliomoro:
thanks bng seems to work fine now. 
Tried out 8e71645219266bc2f400ebdd6f169d65bdae41e3 and, besides the msg editing not working, I found the following:
If you save a `{h,v}radio` with a bg color and open the patch, the bg color won't be displayed till you open the properties dialog and press OK. 
This seems to be related to some interdependencies between `radio::config` with `-number` and `radio::activate` (because . Calling them both twice fixes it but it's clearly not the best solution ...

```
diff --git a/src/g_radio.c b/src/g_radio.c
index afb19f34..1cc7cabf 100644
--- a/src/g_radio.c
+++ b/src/g_radio.c
@@ -45,6 +45,7 @@ static void radio_draw_new(t_radio *x, t_glist *glist)
     pdgui_vmess("::pd::widget::create_inlets" , "o i", x, 0);
     pdgui_vmess("::pd::widget::create_outlets", "o i", x, 0);

+    (*x->x_gui.x_draw)(x, x->x_gui.x_glist, IEM_GUI_DRAW_MODE_CONFIG);
     (*x->x_gui.x_draw)(x, x->x_gui.x_glist, IEM_GUI_DRAW_MODE_CONFIG);
     (*x->x_gui.x_draw)(x, x->x_gui.x_glist, IEM_GUI_DRAW_MODE_IO);
 }
```

umlaeute:
yes. i think this is the same issue as described under "regressions" in the top-posting.
editing is not working, because i still have to think about it :-)

umlaeute:
### lessons learned
- editing is a mess :-/ text,rtext,editor,canvas? my head swims.
- handling the connections on the GUI side (a la `connect obj1 outlet1 obj2 outlet2`) is not that hard to do, but will break existing GUI externals. see #1767
- the idea that an object tracks all the canvases it has been created on is somewhat neat. but adds a lot of boilerplate. see #1768
- SOMETHING QUITE IMPORTANT (which I forgot; it will come back though)
- creating namespaces on the GUI side is fun, but it's easy to just create a new namespace for anything (which doesn't really improve anything)
- there's no real need to put each widget into a separate file

giuliomoro:
Glad to see editing is now mostly working on the latest commit ... one little bug: when you paste into a msg/obj/comment, you get one extra space at the end.

umlaeute:
### more lessons learned

- as of now, this ***significantly** slows down* building the GUI
  e.g. on my computer a patch like the one found at https://github.com/pure-data/pure-data/issues/1663#issuecomment-1128172356 takes about 2 seconds to load with Pd-0.52.2, but 4 seconds with this PR (781c6c3dffcf350ad124780fbc55f5899e086eb7)
  :thinking:

giuliomoro:
>as of now, this significantly slows down building the GUI
😞 I guess this is down to the increased complexity of the tcl/tk code? From what I could see, you didn't significantly increase the amount of data sent to the GUI or added extra ping-pong of packets between gui and server.

umlaeute:
yes. from Pd'-cores POV, there is somewhat less work to do, but the GUI  now has to do some more lifting (and there's a couple of indirections built into my current system, e.g. the command dispatcher.

now I dearly love the command dispatcher (e.g. allowing us to use a single `::pdwidget::displace $obj 10 0` to move any object, without knowing its implementation details) and would hate to replace it with something more direct.

umlaeute:
so time for some profiling.
the test-patch is a stripped down version of the one mentioned above, containing 32*24 numberboxes without iolets.
the test setup just draws these numberboxes entirely on the GUI side (the Pd-core doesn't even know about it), basically replaying what Pd would send and timing the replay.
This is on Linux, i7-7700 (amd64), Tcl/Tk-8.6.12

### reference

| test          | time (ms) | notes 
|---------------|-----------|---|
| reference  |        21 | this just mimicks how Pd draws the nbx by issuing the raw Tk calls |

21ms for 768 objects is not that bad.

### this implementation (with some hacks)

| test          | time (ms) | notes
|---------------|-----------|---|
|rewrite | 1066 | current state of this PR |
| condensed | 619 | instead of multiple calls to `::pdwidget::config`, merge them all into the `::pdwidget::create` call
| no-label | 334 | drop the (useless) setup of the gatom-label (as the test patch doesn't have any labels)
| no-label + no-select | 251 | also drop the proc that tells newly created objects that they are "not-selected".

the "no-label" and "no-select" options are only here for comparision. in the end, we cannot just omit them (although there *is* room for optimization. in any case: text is obviously expensive)


there are two obvious ways to optimize the current implementation:
1. on the Tcl side: reduce the number of function indirections ("getting rid" of the behaviour dispatcher)
2. on the Tk side: simplify the tagging (the current implementation uses orthoginal tags, e.g. `-tags [list ${obj} text]` and then joins the tags (e.g. `${obj}&&text`) to filter items; Pd uses explicit tags (e.g. `-tags ${obj}T`) which should make the filtering much simpler; i have no idea how Tcl/Tk implements tagging, but i guess checking the hash of a single tag is always simpler as checking two tags and then adding some boolean algebra on top of it).

### bypassing the widgetbehavior dispatcher

| test          | time (ms) |
|---------------|-----------|
| condensed |       505 |
|  no-label  |       211 |
| no-label + no-select |       206 |



### using explicit tags

| test          | time (ms) |
|---------------|-----------|
| rewrite  |       726 |
| condensed  |       368 |
| no-label  |       229 |
|  no-label + no-select  |       179 |

### using explicit tags *and* bypassing the widgetbehavior dispatcher

| test          | time (ms) |
|---------------|-----------|
| condensed |       283 |
| no-labels |       138 |
| no-label + no-select |       130 |

as can be seen there's plenty of potential to optimize, but even in the best case (condense the configuration into the constructor; explicit tags; bypassing the proc dispatcher), we are still about 10 times slower than what is currently happening...

Ant1r:
Hey, these are pretty bad news...
So it seems TclTk is indeed very slow... OTOH this confirms that it's a good idea to find a way to get rid of it in the end!...

Maybe there's a possibility to continue in this way (which is to make pd core agnostic from the GUI toolkit, while keeping TclTk at first, before people can try implementing the GUI with another toolkit):

how about writing a temporary C (or C++...) gateway between the core and the GUI, that would do all the widget abstraction layer, and would send only digested commands to TclTk, which would only have to do the final display job? Then we should expect similar performance as the "reference" Pd.

Maybe we would need to run the gateway in a secondary thread, to avoid impacting the real-time requirements with GUI-related processing and dynamic memory allocation.

Thanks a lot for all your time on this anyway!

giuliomoro:
I tried to use the `tcl` profiler package and got some results that I am unsure about how to interpret (I don't have a neat way of visualising them yet), but they seem to indicate a lot of time is spent in `_do_create_iolets` (even if the iolets are all hidden). 

Sample output for that function
```
Profiling information for ::pdwidget::_do_create_iolets
============================================================
            Total calls:  2057
    Caller distribution:
  ::pdwidget::_create_inlets:  2
  ::pdwidget::_create_outlets:  3
  ::pdwidget::_refresh_iolets:  2052
           Compile time:  183209
          Total runtime:  795693929
        Average runtime:  386822
          Runtime StDev:  272634
         Runtime cov(%):  70.5
  Total descendant time:  34299887
Average descendant time:  16674
Descendants:
  ::pd::canvas::get_zoom:  2057
  ::pdwidget::base_tag:  2057 
```

Full output: 
[pd-profiler.txt](https://github.com/pure-data/pure-data/files/9778104/pd-profiler.txt)

------------
### steps to reproduce

Use this Pd patch (32*24 floatatoms without iolets plus a loadbang-delayed-message):
[many-gatoms.pd.zip](https://github.com/pure-data/pure-data/files/9778125/many-gatoms.pd.zip)

Apply these changes:
```
diff --git a/src/m_glob.c b/src/m_glob.c
index a2fa0120..936d6808 100644
--- a/src/m_glob.c
+++ b/src/m_glob.c
@@ -34,6 +34,7 @@ void glob_start_startup_dialog(t_pd *dummy, t_floatarg flongform);
 void glob_startup_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_ping(t_pd *dummy);
 void glob_plugindispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
+void glob_guidispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_watchdog(t_pd *dummy);
 void glob_loadpreferences(t_pd *dummy, t_symbol *s);
 void glob_savepreferences(t_pd *dummy, t_symbol *s);
@@ -104,6 +105,12 @@ void max_default(t_pd *x, t_symbol *s, int argc, t_atom *argv)
     endpost();
 }

+void glob_guidispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
+{
+	if(1 == argc && argv[0].a_type == A_SYMBOL) //surely can be done better
+		pdgui_vmess(0, "r", argv[0].a_w.w_symbol->s_name);
+}
+
 void glob_plugindispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
 {
     pdgui_vmess("pdtk_plugin_dispatch", "a", argc, argv);
@@ -179,6 +186,8 @@ void glob_init(void)
         gensym("perf"), A_FLOAT, 0);
     class_addmethod(glob_pdobject, (t_method)glob_compatibility,
         gensym("compatibility"), A_FLOAT, 0);
+    class_addmethod(glob_pdobject, (t_method)glob_guidispatch,
+        gensym("gui-dispatch"), A_GIMME, 0);
     class_addmethod(glob_pdobject, (t_method)glob_plugindispatch,
         gensym("plugin-dispatch"), A_GIMME, 0);
     class_addmethod(glob_pdobject, (t_method)glob_helpintro,
diff --git a/tcl/pd-gui.tcl b/tcl/pd-gui.tcl
index bcaca58a..a98cc1ad 100755
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -8,6 +8,9 @@
 # "." automatically gets a window, we don't want it.  Withdraw it before doing
 # anything else, so that we don't get the automatic window flashing for a
 # second while pd loads.
+package require profiler
+::profiler::init
+
 if { [catch {wm withdraw .} fid] } { exit 2 }

 # This is mainly for OSX as older versions only
@@ -829,6 +832,17 @@ proc load_startup_plugins {} {
     }
 }

+proc do_dump_profile {} {
+	# dump all kind of stuff, although data is repeated
+	# machine-readable
+	puts [::profiler::dump]
+	# human-readable
+	puts [::profiler::print]
+	# sorted by runtime. Last entry will be the more busy one.
+	# consider using other keys instead of totalRuntime
+	puts [::profiler::sortFunctions totalRuntime]
+}
+
 # main
 proc main {argc argv} {
```

Note: the Wish.app 8.6 that is installed in my macos didn't have the `profiler` package and the `wish8.6` I installed via `brew` required me to manually edit `tcllib1.20/profiler/pkgIndex.tcl` to fix the error `Error in startup script: attempt to provide package profiler 0.4 failed: package profiler 0.5 provided instead`

giuliomoro:
I tried to use the `tcl` profiler package and got some results that I am unsure about how to interpret (I don't have a neat way of visualising them yet), but they seem to indicate a lot of time is spent in `_do_create_iolets` (even if the iolets are all hidden). 

Sample output for that function
```
Profiling information for ::pdwidget::_do_create_iolets
============================================================
            Total calls:  2057
    Caller distribution:
  ::pdwidget::_create_inlets:  2
  ::pdwidget::_create_outlets:  3
  ::pdwidget::_refresh_iolets:  2052
           Compile time:  183209
          Total runtime:  795693929
        Average runtime:  386822
          Runtime StDev:  272634
         Runtime cov(%):  70.5
  Total descendant time:  34299887
Average descendant time:  16674
Descendants:
  ::pd::canvas::get_zoom:  2057
  ::pdwidget::base_tag:  2057 
```

Full output: 
[pd-profiler.txt](https://github.com/pure-data/pure-data/files/9778104/pd-profiler.txt)

------------
### steps to reproduce

Use this Pd patch (32*24 floatatoms without iolets plus a loadbang-delayed-message):
[many-gatoms.pd.zip](https://github.com/pure-data/pure-data/files/9778125/many-gatoms.pd.zip)

Apply these changes:
```
diff --git a/src/m_glob.c b/src/m_glob.c
index a2fa0120..936d6808 100644
--- a/src/m_glob.c
+++ b/src/m_glob.c
@@ -34,6 +34,7 @@ void glob_start_startup_dialog(t_pd *dummy, t_floatarg flongform);
 void glob_startup_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_ping(t_pd *dummy);
 void glob_plugindispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
+void glob_guidispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_watchdog(t_pd *dummy);
 void glob_loadpreferences(t_pd *dummy, t_symbol *s);
 void glob_savepreferences(t_pd *dummy, t_symbol *s);
@@ -104,6 +105,12 @@ void max_default(t_pd *x, t_symbol *s, int argc, t_atom *argv)
     endpost();
 }

+void glob_guidispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
+{
+	if(1 == argc && argv[0].a_type == A_SYMBOL) //surely can be done better
+		pdgui_vmess(0, "r", argv[0].a_w.w_symbol->s_name);
+}
+
 void glob_plugindispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
 {
     pdgui_vmess("pdtk_plugin_dispatch", "a", argc, argv);
@@ -179,6 +186,8 @@ void glob_init(void)
         gensym("perf"), A_FLOAT, 0);
     class_addmethod(glob_pdobject, (t_method)glob_compatibility,
         gensym("compatibility"), A_FLOAT, 0);
+    class_addmethod(glob_pdobject, (t_method)glob_guidispatch,
+        gensym("gui-dispatch"), A_GIMME, 0);
     class_addmethod(glob_pdobject, (t_method)glob_plugindispatch,
         gensym("plugin-dispatch"), A_GIMME, 0);
     class_addmethod(glob_pdobject, (t_method)glob_helpintro,
diff --git a/tcl/pd-gui.tcl b/tcl/pd-gui.tcl
index bcaca58a..a98cc1ad 100755
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -8,6 +8,9 @@
 # "." automatically gets a window, we don't want it.  Withdraw it before doing
 # anything else, so that we don't get the automatic window flashing for a
 # second while pd loads.
+package require profiler
+::profiler::init
+
 if { [catch {wm withdraw .} fid] } { exit 2 }

 # This is mainly for OSX as older versions only
@@ -829,6 +832,17 @@ proc load_startup_plugins {} {
     }
 }

+proc do_dump_profile {} {
+	# dump all kind of stuff, although data is repeated
+	# machine-readable
+	puts [::profiler::dump]
+	# human-readable
+	puts [::profiler::print]
+	# sorted by runtime. Last entry will be the more busy one.
+	# consider using other keys instead of totalRuntime
+	puts [::profiler::sortFunctions totalRuntime]
+}
+
 # main
 proc main {argc argv} {
```

Note: the Wish.app 8.6 that is installed in my macos didn't have the `profiler` package and the `wish8.6` I installed via `brew` required me to manually edit `tcllib1.20/profiler/pkgIndex.tcl` to fix the error `Error in startup script: attempt to provide package profiler 0.4 failed: package profiler 0.5 provided instead`

giuliomoro:
I notice here on macos that if I remove the
```
cnv lower "${tag}&&${iotag}" "${tag}&&iolets"
cnv raise "${tag}&&${iotag}" "${tag}&&iolets"
```
from the bottom of `_do_create_iolets()`, the function's runtime decrease greatly, so my results are surely affected by the above mentioned overall sloppiness of modern tk on macos. Hopefully on your system you'll get more meaningful results.

umlaeute:
@Ant1r just for the record, i don't think all is lost. it was pretty easy sailing so far, but no the sea is getting rough. i'm sure we will come up with some acceptable solution.

@giuliomoro thanks for the additional tests. i have `profiler` (0.6) available in my Tcl/Tk-8.6 installation (on Debian), so that's not a problem. I already used it a bit (to get a quick overview), but found a simple `time` invocation to be more meaningful for now.
i'm not sure i understand the need for `glob_guidispatch` in your patch (is this just a leftover that accidentally made it into the patch?)

as for `_do_create_iolets`: yes that's one of these uses of boolean tag compounds that i mentioned (and which i started to replace with "explicit tags" (e.g. `${tag}_iolets` instead of `${tag}&&iolets`. as you can see in my timing tables, this reduced the time by a factor of ~1.5.

i've also completely disabled the call to `::pdwidget::refresh_iolets` for now, as this *is* a performance hog

giuliomoro:
It looks like we cannot patch second inlets/outlets in this branch. This is because  in  `_getclosest()` in `src/g_editor.c`
```
static int _getclosest(int numiolets, int posX, int x1, int x2)
{
    int width = x2-x1, closest;
    if(!width)
        return 0;

    switch(numiolets) {
    case 1:
        closest = 0;
        break;
    default:
        closest = ((posX-x1) * (numiolets-1) + width/2)/width;
        break;
    }
    if(closest >= numiolets)
        closest = numiolets - 1;
    if(closest < 0)
        closest = 0;
    return closest;
}
```

`width` turns out to be negative which in turns makes `closest` always 0. Here's a suggested fix:
https://github.com/giuliomoro/pure-data-1/commit/80880290c06a0b1586be86d5a03c73e8f907daa9

umlaeute:
> Also, @umlaeute wondering what you think the next steps for this branch should be.

that's a good question.
we need to find some solution for the performance degradation.

i think the benchmark is the current implementation: the performance of this rewrite must about the same (or faster obviously).

this most likely means that the API will need to be a little less abstract (i guess the part that eats most resources is the iolet/connection thingy. so we probably have to ditch the idea for now, that the GUI can handle these on its own).

i also have the feeling that the proc calls (and proc calls from these) add overhead, which shouldn't be there (afaik,Tcl byte-compiles procs, which should make calling a proc that calls multiple TclTk commands faster than calling those commands manually; but it seems this is not the case here)

Spacechild1:
Optimization rule nr. 1: measure, measure, measure, i.e. run a profiler on the core and GUI process to find the actual hotspots.

@umlaeute Do you have a test patch that exemplifies the performance problems?

> afaik,Tcl byte-compiles procs, which should make calling a proc that calls multiple TclTk commands faster than calling those commands manually; but it seems this is not the case here

Yes, this seems really strange to me. I would also expect performance to be *better*, since we send fewer messages to the GUI. We have to profile to find the actual problem.

umlaeute:
> @umlaeute Do you have a test patch that exemplifies the performance problems?

i guess that would be https://github.com/pure-data/pure-data/pull/1765#issuecomment-1275667141

Spacechild1:
> i guess that would be https://github.com/pure-data/pure-data/pull/1765#issuecomment-1275667141

Thanks, will give it a try.

---

One thing that came to my mind: I would be useful if there was a way to designate GUI *transactions*. Currently, we send lots of individual drawing commands, even if they just form one big operation (e.g. redrawing a canvas). If we could (optionally) wrap such commands in some kind of "begin" and "end" messages, this would enable huge optimization opportunities for the GUI backend, such as batching requests, temporarily disabling redrawing, etc.

For example, in `canvas_map`:
```
pdgui_vmess("::pd::canvas::begin", "c", x);
for (y = x->gl_list; y; y = y->g_next)
    gobj_vis(y, x, 1);
pdgui_vmess("::pd::canvas::end", "c", x);
```

This is a pretty common pattern. Two examples that spring to my mind: https://github.com/pac-dev/protoplug/blob/master/Frameworks/vstsdk2.4_minimal/pluginterfaces/vst2.x/aeffectx.h#L305-L306 and https://steinbergmedia.github.io/vst3_doc/vstinterfaces/classSteinberg_1_1Vst_1_1IComponentHandler.html#a8456ad739430267a12dda11a53fe9223.

Spacechild1:
To expand on the previous post:

In naive canvas implementations, every change to the canvas may trigger a full redraw. (I *think* that Tk belongs to this category.) If you issue many such commands in a row, every command will implicitly repeat all previous commands, leading to O(N2) complexity. Theoretically, we could patch the Tk library and add commands to temporarily disable any redrawing. Together with "::canvas::begin" and "::canvas::end" messages, this could improve patch loading times dramatically. But this is just wild speculation :-)

Now, imagine a less naive implementation where a command does not immediately trigger a redraw, but rather schedules a redraw event. Even in this case, "::canvas::begin" and "::canvas::end" would be huge benefit because we could batch commands, so that several commands only cause a single redraw event.

Finally, the core communicates with the GUI over a TCP socket; although commands are *sent* consecutively, they might arrive with delays and consequently get scattered over several event loop ticks. "::canvas::begin" and "::canvas::end" would allow the GUI backend to wait for a sufficient amount of commands before dispatching them.

umlaeute:
> This is a pretty common pattern. Two examples that spring to my mind

that's really just one example (the "Steinway").

i'm not opposed though...

Spacechild1:
Here's another example from Pd itself: `canvas_suspend_dsp` and `canvas_resume_dsp` are used to enclose multiple actions that might update the DSP graph. The purpose is exactly to avoid redundant calculations and quadratic complexity. `::canvas::begin` and `::canvas::end` would apply the same idea to GUI rendering.

giuliomoro:
Just found a bug on this branch: load a patch with a ticked checkbox and it won't show the checkmark upon loading. As expected, the state is actually stored, so you need to click on it twice to see the tick appear.  Example patch:

```
#N canvas 219 208 977 392 12;
#X obj 157 133 tgl 19 1 empty empty empty 0 -10 0 12 #fcfcfc #000000 #000000 1 1;
```

umlaeute:
I think this is what is described in the top post under the heading "regression"

danomatika:
Maybe doing all the command parsing and string handling to draw on the Tcl side is simply slower because the original implementation just sent the command directly to Tk as opposed to additional Tcl proc layers. I feel like that could be a bottleneck and could use some finer-grained profiling.

danomatika:
If calling procs and string handling is the slowdown on the Tcl/TK side, one approach could be to write a Tcl plugin to do it in C then send it back to Tk.

umlaeute:
i always assume that the tcl guys know what they are doing (language wise): after all interpreting strings is at the core of the language itself, and afaik there's even some jit-compiler hidden somewhere.

maybe we are bypassing this by when reading the commands from the socket.

danomatika:
> i always assume that the tcl guys know what they are doing (language wise): after all interpreting strings is at the core of the language itself, and afaik there's even some jit-compiler hidden somewhere.
> 
> maybe we are bypassing this by when reading the commands from the socket.

Yeah, I did too, otherwise it never would have been usable on the original hardware it was developed for in the 90s. The only thing I can think of is sort sort of memory allocation on the fly, but that shouldn't;t be anything we have to deal with directly. Strange.

PR #1766, opened 2022-09-29
load accompanying GUI-plugin for library

### LINKED ISSUE #1555: Pd GUI doesn't search for GUI plugins recursively
Generally, GUI plugin files must be named `*.plugin.tcl`.

Unfortunately, `load_startup_plugins` in `pd-gui.tcl` doesn't search for plugins recursively. Instead it assumes that plugin files are either located on the top level or in a *direct* subfolder `*-plugin`: https://github.com/pure-data/pure-data/blob/bad155d26cceead85e845da4f7c3d41498ea0574/tcl/pd-gui.tcl#L785-L790

This makes it unpractical for externals to ship GUI plugins, e.g. to implement custom Tcl/Tk views. As a workaround, they typically assemble the Tcl/Tk code in the setup function and send it to the GUI. Here's an elaborate example: https://github.com/porres/pd-else/blob/master/Classes/Source/keyboard.c#L863

Is there a reason for the current behavior or can we change it?

umlaeute:
This PR simplifies distributing libraries with accompanying GUI-plugins.

Whenever a Pd-library is loaded, Pd-core signals the GUI to load a GUI-plugin of the same name and in the same path as the loaded external.

### example
1. user loads `zexy`
2. Pd-core finds it in `~/.local/lib/pd/extra/` (the real path being `~/.local/lib/pd/extra/zexy/zexy.pd_linux`) and loads it
3. Pd-core tells Pd-GUI to load a GUI-plugin `zexy` in `~/.local/lib/pd/extra/`
4. The GUI finds the GUI-plugin as `~/.local/lib/pd/extra/zexy/zexy-plugin.tcl` and loads it.
5. zexy can now use all the nifty features provided by its GUI-plugin

### why is this useful?
some libraries come with a considerable amount of GUI code.
E.g. `else` comes with a lot of GUI objects, and implements them as tons of Tcl/Tk code that is compiled into the external itself (and then sent to the GUI).
Others, like `patcherize` ship both a compiled object and a GUI-plugin, and go through hoops and loops to automatically load the external when the GUI-plugin is autoloaded (abusing hidden canvases and what not).

Neither solution is good: `else`'s GUI-objects have hardcoded their Tcl/Tk dependency (and we want to get rid of it - at least theoretically); and the user has no way to *not* load `patcherize` once it's installed.

With this PR, the GUI-part of a library is separated from the core (and theoretically the library could provide multiple GUI-implementations - one for each GUI-frontend), and the user has full control over what they load (using the trusted library loading features)

Closes: https://github.com/pure-data/pure-data/issues/1555

### DIFF
--- a/src/s_loader.c
+++ b/src/s_loader.c
@@ -473,6 +473,9 @@ int sys_loadlib_iter(const char *path, struct _loadlib_data *data)
     for(q = &loaders; q; q = q->next)
         if ((ok = q->loader(data->canvas, data->classname, path)))
             break;
+    if(ok)
+        pdgui_vmess("::pdgui::load_plugin", "ss", data->classname, path);
+
     /* if all loaders failed, try to load as abstraction */
     if (!ok)
         ok = sys_do_load_abs(data->canvas, data->classname, path);
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -739,11 +739,12 @@ proc check_for_running_instances { } {
 # ------------------------------------------------------------------------------
 # load plugins on startup
 
-proc load_plugin_script {filename} {
+set ::pdgui::_loaded_plugins_fullpath {}
+proc load_plugin_script {filename {check_loaded 1}} {
     global errorInfo
 
     set basename [file tail $filename]
-    if {[lsearch $::loaded_plugins $basename] > -1} {
+    if { $check_loaded && [lsearch $::loaded_plugins $basename] >= 0} {
         ::pdwindow::post [ format [_ "'%1\$s' already loaded, ignoring: '%2\$s'"] $basename $filename]
         ::pdwindow::post "\n"
         return
@@ -762,6 +763,7 @@ proc load_plugin_script {filename} {
         ::pdwindow::error "\n-----------\n"
     } else {
         lappend ::loaded_plugins $basename
+        lappend ::pdgui::_loaded_plugins_fullpath $filename
     }
 }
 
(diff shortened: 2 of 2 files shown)


Spacechild1:
> Pd-core tells Pd-GUI to load a GUI-plugin zexy in ~/.local/lib/pd/extra/

I guess you meant `~/.local/lib/pd/extra/zexy`?

Anyways, I really, really like this! In particular:

> With this PR, the GUI-part of a library is separated from the core (and theoretically the library could provide multiple GUI-implementations - one for each GUI-frontend)

Spacechild1 (approved):

umlaeute:
> I guess you meant `~/.local/lib/pd/extra/zexy`?

Actually that's the tricky part, and indeed I meant what I wrote (*without* the trailing `zexy` component).

The question is, who decides whether to load the GUI external: is it Pd, or some object "loader".

Pd's standard external loader is just one in a list of loaders (though typically the only one), and this loader is given a path like `~/.local/lib/pd/extra` to load "zexy". As a special property of the "external loader" it will look up both "zexy.dll" and "zexy/zexy.dll" and load whatever it finds. Pd (outside the "external loader") doesn't really know about this double lookup.

So, if the "external loader" triggers the loading of the GUI-plugin, it can tell it where it found the binary (and thus include the trailing library component if needed). But that means, that GUI-plugin loading only works for loaders that explicitly support GUI plugins.
OTOH if Pd's generic loading infrastructure (which calls the various loaders) can be made responsible for initiating the GUI-plugin loading, then it works for all loaders... But then we don't know the actual path the loader used to load, just the one we passed into it.
This is what I ended up doing 

An alternative (unifying both) would be to move the "library/library.dll" expansion from the "external loader" to the loading infrastructure... I'm not entirely sure whether this will have unexpected consequences.

umlaeute:

chr15m:
> currently no tcl-plugin is loaded when loading an abstraction. don't know whether this would be useful.

It would!

umlaeute:
for a reference discussion see
- https://lists.iem.at/hyperkitty/list/pd-list@lists.iem.at/thread/5M7WHO3PZADXDDPKMLLYR6XKCS252UHQ/
- and the follow-up (on pd-dev) https://lists.iem.at/hyperkitty/list/pd-dev@lists.iem.at/thread/5NBI7555BI74T45R4AB2ZTYGWVOAFFO4/

to summarize i think a less intrusive alternative to the solution proposed in this PR would be to move the responsibility to load GUI plugins to the actual consumers (e.g. an external library; or an abstraction)
- add a global message (to Pd-core) to load plugins on the GUI side
  e.g. `[; pd plugin-load /path/to/my/plugin(`
- provide an API for externals to query their own path in the `setup` function, e.g.
   ```C
   /* returns the load path of the external (or NULL otherwise)
       only valid within the setup callback
   */
   const char* sys_get_current_externalpath(void);
   ```

umlaeute:
given that `const char* sys_get_current_externalpath(void);` is already available as `class_gethelpdir()`, i've implemented the rest in #2481

PR #1772, opened 2022-10-16
i18n: translations

umlaeute:
this PR is for keeping translations on `master` uptodate.
no functionality is added, just the translation files for the interface.
as such it should be always safe to merge this into `master`.

while i've updated the help for German for the upcoming 0.53 release, help is needed from other native speakers (@Lucarda , @porres , @Ant1r , @mxa , and whoever feels like it).

it would be great if we could find a more automatic workflow...


porres:
I tried creating a PR and it failed, can you help me with that?

umlaeute:
Weird. It shouldn't be any different from creating any other PR. If you post a link to your branch, I could have a look

Ant1r:
What "failed"? #1774

mxa:
Should we perhaps use https://www.transifex.com/open-source/ ?

eighthave:
Hey all!  At this point, I would recommend using https://hosted.weblate.org. 
Its free software!

po/es.po
```diff
@@ -311,9 +309,7 @@ msgid "Pd window"
 msgstr "Ventana de Pd"
 
 msgid "Match whole word only"
-msgstr ""
-"Comparar solo con\n"
```
Lucarda, line 365: this `\n` was intentional as it fits better in the button.

I'll try to reintroduce it via `weblate.org`

umlaeute: i thought so, but didn't really check.

maybe you can add a note on weblate

Lucarda: done

PR #1798, opened 2022-10-24
translation updates

umlaeute:
so far this PR only fixes the language code for "Chinese (traditional)", which should be `zh_tw` (Chinese as spoken/written in Taiwan), rather than `tw` (Taiwanese, which does not exist).

this PR will be updated (or resurrected, if necessary) once there are any real translation updates.


umlaeute:
That would probably be "Hakka".
In any case, afaik Hakka also uses the "zh" abbreviation (being *a* Chinese language).

The important part is actually to map language codes as provided by the operating system to the correct mapping (and while I don't think any OS provides `tw`, I have to admit that I haven't tested what a Taiwanese MSW or Darwin installation actually provides; I checked the existing "valid" codes on my Debian system, and this lists `zh_tw` but not ~`tw`~

umlaeute:
Also my understanding of the Chinese language systems is close to nil, but I think that both Hakka and Mandarin would use the same transcriptions, so the translation should be valid for both (at least for people that read traditional Chinese; we still need a translation for simplified Chinese)

umlaeute:
note to myself: according to https://en.wikipedia.org/wiki/Zh-TW, a possible language code for "Taiwanese Hokkien" might be `nan-TW` (that is: `nan_tw`) if we ever need to distinguish it from `zh_tw`.

Issue #1799, opened 2022-10-24
pipewire-jack-audio-connection-kit not recognised as jack


muziker:
When compiling pure data and specifying --enable-jack, on a pipewire system like a fedora, the configuration script doesn't recognise this as a proper jack library. 

pure-data 0.52-2

umlaeute:
i don't know about the fedora package, so: which (JACK-related) header files does it install?

e.g. on Debian, the `pipewire-jack` package only installs the replacement `libjack.so` library, and no headers at all.

for compiling Pd with JACK support, you *must* have JACK-headers installed.

umlaeute:
while waiting for feedback, i've found out the following:
- `pipewire-jack-audio-connection-kit` is obviously not enough
- the package required to install is `pipewire-jack-audio-connection-kit-devel`
- this still is not enough, as the above installs the `libjack.so` file into `/usr/lib64/pipewire-0.3/jack`
- this location is announced via `pkg-config --libs jack`, but Pd does not use this

muziker:
Yes I just found out where the pipewire jack .so files are located. The configure script doesn't use pkg-config to do it's jack_set_error_function test. I copied the .so files for a quick fix to this.

umlaeute:
urgh, copying files around is usually the wrong apporach.

a better and simpler approach, is to just tell configure where to look for the libaries:

```sh
./configure --enable-jack LDFLAGS="-L/usr/lib64/pipewire-0.3/jack"
```

muziker:
Too true ! But in some instances of other projects, LDFLAGS from the environment gets clobbered somehow during the build process.

PR #1804, opened 2022-10-26
stable development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/1797

---

### GUI
- Closes: https://github.com/pure-data/pure-data/issues/1805 (the never-ending story of placing windows on odd desktop layouts)
- Closes: https://github.com/pure-data/pure-data/issues/1826 (crashing regression when setting iemgui receivers)
- Closes: https://github.com/pure-data/pure-data/issues/1827 (regression that prohibits unsetting iemgui send/receive)
- Closes: https://github.com/pure-data/pure-data/issues/1831 (another iemgui regression that would ignore `[set(` messages while dragging sliders)

### build system

- Closes: https://github.com/pure-data/pure-data/issues/1799 (building against pipewire-jack on Fedora)
- Update the non-autotools `libpd` buildsystem

### Documentation
- Updated notes for macOS/arm64

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -206,7 +206,8 @@ need. For example, to install the autotools & gettext using Homebrew:
 By default, Pd is built for the current system architecture, usually 64 bit. If
 you want to override this you can use the --enable-universal configure option
 which allows you to specify the desired architecture(s) when building Pd. For
-Intel/AMD processors, 32 bit is called "i386" and 64 bit is "x86_64". By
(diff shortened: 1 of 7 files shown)


porres:
I see a few important regression fixes taken care of, isn't it worth discussing a 53-1 bugfix already? Do you have other known regressions you're about to attack @umlaeute ? I just fixed a regression in one of the data structures examples by the way.

ping @millerpuckette

PR #1810, opened 2022-10-30
Fixes typos in x4.htm

idonotfollowyou:

PR #1811, opened 2022-10-31
Documentation typos

idonotfollowyou:
This re-applies #1809 and #1810 
I will add any further work to this PR


idonotfollowyou:
Removed. Now using #1812 which is using the new htmldocs branch.

PR #1825, opened 2022-11-08
m_binbuf: use strtol's end pointer to skip walking through dollar string

myQwil:
`strtol` is the same as `atol`. The latter makes a call to the former, but `atol` leaves out additional parameters, such as `strtol`'s end pointer, which we can use to more quickly and efficiently determine the length of the dollarsym's index.


Spacechild1:
@millerpuckette I think this is a great little improvement!

src/m_binbuf.c
```diff
@@ -487,24 +487,16 @@ int canvas_getdollarzero(void);
...
-    }
-
-    if (cs==s)      /* invalid $-expansion (like "$bla") */
+    if (argno < 0 || cs == s)      /* invalid $-expansion (like "$bla") */
     {
         sprintf(buf, "$");
         return 0;
     }
-    else if (argno < 0 || argno > ac) /* undefined argument */
+    else if (argno > ac) /* undefined argument */
```
myQwil, line 499: It's been a while, so I don't recall why I messed around with these 'if' statements. It's possible they can be reverted back to what they originally were.

Spacechild1: That's good question. `$-1` can be considered either an "invalid $-expansion" or an "undefined (out-of-range) argument". When in doubt, I'd say just keep the original logic :)

myQwil: After a bit of debugging, it seems like it's basically impossible to specify a truly negative index. If a dash comes right after the dollar symbol, it will never get to this function. The only way argno < 1 is if a ridiculously large number is specified, at which point, I guess `strtol` returns -1 to indicate that it failed to fully parse the number.

myQwil: I might be thinking too much about what technically ends up happening instead of what should happen in principle. Maybe it should be, or will be possible for negative indexes to reach this function in the future, or maybe there are edge cases I haven't thought of where it can still happen, so I'll change the if statements back to what they were.

PR #1839, opened 2022-11-25
Turn inletunion into an anonymous union

myQwil:
Anonymous structs and unions were added in the C11 standard.


Spacechild1:
Pd is a C89 code base. What is the motivation for this PR, btw?

myQwil:
My bad, I wasn't aware of that.
The main reason was to start adopting conventions from a newer standard.

Issue #1850, opened 2022-12-05
sigmund~ input list also works when not in '-t' mode


porres:
The documentation says that list input should only be used when invoked with the '-t' flag for analyzing tables, but this also works when not invoked with this flag.

I think that including a simple mode check in list mode should be included to prevent list input when not in '-t' mode.

PR #1851, opened 2022-12-05
sigmund fixes

porres:
closes https://github.com/pure-data/pure-data/issues/1850


millerpuckette:
the loud flag is optional these days - it does nothing for the moment and doesn't need documenting.

PR #1876, opened 2023-01-25
fix scalar-scalar version of [+~] and [-~]

Spacechild1:
Currently, the scalar-scalar perform routines for `[+~]` and `[-~]` are broken, see [plus-minus-bug.zip](https://github.com/pure-data/pure-data/files/10498473/plus-minus-bug.zip)

I don't understand the reasoning behind `sp[0]->s_length == 1`, so maybe I am missing something? Anyway, with this PR, the example above works as intended.

### DIFF
--- a/src/d_arithmetic.c
+++ b/src/d_arithmetic.c
@@ -86,18 +86,18 @@ t_int *scalarplus_perf8(t_int *w)
 static void dsp_add_multi(t_sample *vec1, int n1, t_sample *vec2,
     int n2, t_sample *outvec, t_perfroutine func, t_perfroutine func8)
 {
-    int i, reps, blocksize, thisblock;
+    int i;
     if (n1 > n2)
         for (i = (n1+n2-1)/n2; i--; )
     {
-        int blocksize = (n2 < n1 - i*n2 ?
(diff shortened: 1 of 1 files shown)


Spacechild1:
Ah, wait, now I see the actual problem! Will rework and reopen.

millerpuckette:
OTOH something else is broken now - large patches are crashing pd.  I've
got some more work to do...

Spacechild1:
BTW, [unpack~] is missing a call to signal_setchansout!

millerpuckette:
That should be fixed in master commit 9f0bc915aa4a7df625b9e10ab17712d930d8b43f, no?
But there's another memory crash I'm getting when loading large patches.

Spacechild1:
> That should be fixed in master commit https://github.com/pure-data/pure-data/commit/9f0bc915aa4a7df625b9e10ab17712d930d8b43f, no?

Ah, right!

millerpuckette:
OK, at least it runs again.  It's in an inelegant state though so more tweaking might be desirable.

Spacechild1:
I think the null signal solution is quite elegant (and obvious in hindsight). BTW, have a look at https://github.com/pure-data/pure-data/pull/1877

PR #1877, opened 2023-01-27
A few multi-channel bugfixes and improvements

Spacechild1:
1. fix scalar-scalar versions of `[+~]` and `[-~]`

2. fix DSP flags for `scalarminus` class to prevent bogus output

3. small fixes in `ugen_doit` (parenthesis and a wrong calc size)

4. use a dedicated free-list for scalar signals.
Currently, scalar signals are allocated from the vector free-list and returned to the "borrowed" free-list... Of course, it should really use the same free-list for both operations.
We can't use the vector free-list because it can cause memory leaks. (We overwrite `s_vec` - which might point to an actual 1-sample buffer!)
We *could* use the "borrowed" list, but I'm not sure about the implications. This can be potentially dangerous because both scalar and "borrowed" signals mess with `s_vec` in their own ways.
I think the safest - and clearest - solution is to use a dedicated free-list for scalar signals.

6. rename `signal_setchansout` to `signal_setoutchans`.
Maybe it is just me, but IMO `signal_setchansout` looks really strange. Also, I think "set out(put) channels" makes more sense than "set channels out(put)". @millerpuckette If you don't want this, just merge up to the previous commit :-)

### DIFF
--- a/src/d_arithmetic.c
+++ b/src/d_arithmetic.c
@@ -86,18 +86,18 @@ t_int *scalarplus_perf8(t_int *w)
 static void dsp_add_multi(t_sample *vec1, int n1, t_sample *vec2,
     int n2, t_sample *outvec, t_perfroutine func, t_perfroutine func8)
 {
-    int i, reps, blocksize, thisblock;
+    int i;
     if (n1 > n2)
         for (i = (n1+n2-1)/n2; i--; )
     {
-        int blocksize = (n2 < n1 - i*n2 ?
+        t_int blocksize = (n2 < n1 - i*n2 ?
             n2 : n1 - i*n2);
         dsp_add((blocksize & 7 ? func : func8), 4,
             vec1 + i * n2, vec2, outvec + i * n2, blocksize);
     }
     else for (i = (n1+n2-1)/n1; i--; )
     {
-        int blocksize = (n1 < n2 - i*n1 ?
+        t_int blocksize = (n1 < n2 - i*n1 ?
             n1 : n2 - i*n1);
         dsp_add((blocksize & 7 ? func : func8), 4,
             vec1, vec2 + i*n1, outvec + i*n1, blocksize);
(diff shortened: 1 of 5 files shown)


porres:
@millerpuckette there's an important fix for [send~] in https://github.com/pure-data/pure-data/pull/1873 as well

umlaeute:
@porres miller hand-applied my fix as c0a775ef73d89171453c6a6d3c32e6051c35c8b7 rather than merging in the `develop` branch.

millerpuckette:
I haven't managed to study this yet but in advance of that, I think the problem is that signals in the 'reusable' list aren't truly reusable yet - one can't yet bash the "s_vec" field, for instance, until the DSP method has finished using the signal.  So the correct fix would be to have a reuse-after-dsp-routine list (which can be an automatic variable in the ugen_doit() routine) that then gets made "reusable" after calling the dsp method.

Meanwhile, I disagree with both signal_setoutchans and the existing one... how about signal_multiout(&s, channels)?

millerpuckette:
... or, so that there's a verb in the name, "signal_prepmultiout()"

umlaeute:
A verb is nice, but... "prep" is awfully generic

millerpuckette:
I hadn-added the first 2, still thinking about 3 and 4.  As to the name, perhaps signal_setmultiout()...?  I don't know what's bothersome about "setchansout" but perhaps it's not a problem with "setmultiout".

Spacechild1:
> signal_setmultiout()

That's my favourite so far.

millerpuckette:
I attempted a fix that's an alternative to comit 3 here - I stuck it on a new branch so you can test before comitting to it...

branch:  https://github.com/pure-data/pure-data/tree/delay-signal-reusable

commit:
https://github.com/pure-data/pure-data/commit/d3b939bb12ae2b3ec236d09e39b9115099e6b367

Spacechild1:
Closing as partly merged, partly superseded by https://github.com/pure-data/pure-data/tree/delay-signal-reusable

PR #1879, opened 2023-01-28
Fixed regression: Apple Silicon compilation

timothyschoen:
By replacing getfn with mess1, it will compile again on macOS/iOS.


timothyschoen:
Guess this is already fixed then!

Issue #1794, opened 2022-10-22
console font defaults to tiny '10' size


porres:
I think this has been happening for some time but I only remember when installing a new version like these current new test releases. I think it should default to 12 like the other font sizes.

porres:
funny, the code also says it should use font size from pd-gui.tcl

https://github.com/pure-data/pure-data/blob/master/tcl/pdwindow.tcl#L450

porres:
> the code also says it should use font size from pd-gui.tcl

I don't really see a default font size in  pd-gui.tcl though... and I remember I was the onde who also proposed and changed default font size from 10 to 12, which is set here --> https://github.com/pure-data/pure-data/blob/master/src/s_main.c#L190

I think we can just change the number 10 to 12 and be over with it... unless there is a way to have a single declared value for both the default font size on the "c" side and the "tcl/tk" side...

porres:
hmmm, there's also this setting, but I don't think it is effective, at least not for the pdwindow... https://github.com/pure-data/pure-data/blob/master/tcl/dialog_font.tcl#L5

porres:
@umlaeute I have this PR but I'm thinking maybe you can merge this into your develop branch?

umlaeute:
@porres wrote:
> there's also this setting, but I don't think it is effective

the `::dialog_font::fontsize` variable has actually nothing to do with the default font-size of the Pd-console. instead it is used as a cache when opening the "Font Dialog" to see if the user requested a font-size change or not (in the latter case, Pd is not notified about the non-font-change in order to avoid a useless redraw).

no harm is done by setting this value to e.g. `12` (apart from the potential confusion of your future self when you again wonder why changing this value to e.g. `24` doesn't do anything useful)

> funny, the code also says it should use font size from pd-gui.tcl

no.
the code says it should use `pd_font_$size` (which would expand to something like `pd_font_12`), and I *think* this means would be a variable that holds the actual full font-specification for the font of a given size (e.g. `{DejaVu Sans Mono} bold 12`).
in any case, there's no such variable any more (as it was removed with db2b8403f486fd432f2279210084752a0b034a5a), and the comment should probably be removed.

> unless there is a way to have a single declared value for both the default font size on the "c" side and the "tcl/tk" side...

right. unfortunately there seems to be none.

umlaeute:
i wonder if the default width of the Pd-console should also change along with the font-size... after all, the text needs more space if we use a larger font.

PR #1919, opened 2023-03-16
change def size to 12

porres:
same def size used by Pd, closes https://github.com/pure-data/pure-data/issues/1794


umlaeute:
wouldn't it be even better to use the actual font-size of Pd by default? (as in `pd -font-size ...`)

porres:
Not sure what you mean. I couldn't find a default setting in pdgui.tcl like I said

Issue #1918, opened 2023-03-14
Only redraw data structures scalars when necessary


reduzent:
The current mechanism to redraw all scalars is too eager and triggers a redraw when it is actually not necessary. 

**Current Behavior**: When ever a _new_ template is loaded or when an existing template is modified, all scalars are redrawn. From what I understand, this is done with `canvas_redrawallfortemplate` function when creating new templates.

**Desired Behaviour**: When ever a template is modified, a redraw should be triggered, but not when simply loading a template. 

**Reasoning**: The current behavior scales badly and blocks any GUI interaction during the redraw. The situation gets worse the more scalars are drawn. 

Maybe the redraw can be skipped while not being in Edit mode? I'm not yet sure what the best solution for this is, but it seems we need either a way to suppress redraws or make the mechanism less eager so that redraw do not happen when not editing templates.

More details about the problems this cause are described in #1913 and on the [mailing list](https://lists.puredata.info/pipermail/pd-list/2023-03/132050.html).

reduzent:
Please add `improvement`-tag

millerpuckette:
I found a relevant comment for the function glist_redrawall() - currently, all scalars whatever are redrawn when any template at all is changed.  This needs fixing.  I think just detecting edit mode is both fragile and anyway only prevents some unnecessary redraws.

reduzent:
I don't see any redraws anymore in 0.54-0test2 (49daa9b428945bc028664b0342f2e6b9e1f3c77d). I somehow missed which commit fixed it. Should this be closed?

reduzent:
Pd 0.54 doesn't show redraws anymore. I close this isse

PR #1933, opened 2023-03-24
only redraw all scalars when template's owning canvas is in edit mode

reduzent:
This is an attempt to address #1918. 

Apparently, redraws are only necessary when editing templates so that changes to the templates are reflected in all scalars. As long as the template is not edited, redrawing all scalars seems superfluous.

This PR suppresses redraws normally and enables them for structs in canvasses that are in edit mode.  I couldn't spot any regression after testing all kinds of patches using data structures.


reduzent:
fixes #1918

reduzent:
Pd 0.54 has fixed the issue that this PR is trying to address.

PR #1935, opened 2023-03-25
Do not save template structs at the head of patch files (fixes #772)

### LINKED ISSUE #772: [struct] in abstraction taints parent patch
When an abstraction using [struct] objects is used in patch and the patch is saved, it gets "tainted" with additional lines in the resulting patch file:

before saving:
```
#N canvas 164 99 541 380 12;
#X obj 101 97 abs-with-struct;
#X text 40 33 This is patch contains only this comment and an abstraction
with a struct: abs-with-struct.pd;
```

after saving:
```
#N struct polygon float x float y;
#N canvas 164 99 541 380 12;
#X obj 101 97 abs-with-struct;
#X text 40 33 This is patch contains only this comment and an abstraction
with a struct: abs-with-struct.pd;
```

There was a similar issue with [declare] objects in abstraction tainting the parent patch when saved. This got resolved. 

Check with attached files to reprodruce the problem. Simply load main.pd and save it.

[abs-tainting-parent.zip](https://github.com/pure-data/pure-data/files/3704572/abs-tainting-parent.zip)

reduzent:
This addresses #772.

The function `canvas_savetemplatesto`  scans current and all child canvases recursively and writes the structs it finds to the head of the patchfile. This "taints" the current canvas with structs from all its children with no apparent reason. Templates still load fine without the `#N struct ` lines in the beginning of Pd files.

### DIFF
--- a/src/g_readwrite.c
+++ b/src/g_readwrite.c
@@ -750,46 +750,6 @@ static void canvas_collecttemplatesfor(t_canvas *x, int *ntemplatesp,
     }
 }
 
-    /* save the templates needed by a canvas to a binbuf. */
-static void canvas_savetemplatesto(t_canvas *x, t_binbuf *b, int wholething)
-{
-    t_symbol **templatevec = getbytes(0);
-    int i, ntemplates = 0;
-    canvas_collecttemplatesfor(x, &ntemplates, &templatevec, wholething);
-    for (i = 0; i < ntemplates; i++)
-    {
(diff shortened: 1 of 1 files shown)


reduzent:
The mentioning of this issue happened in error.  It actually wanted to refer to 1933

millerpuckette:
The trouble begins when two abstractions are using incompatible structs with the same name - the only way I could find to resolve that was to have the owning patch lay the law down.  If there's a better way I'm open to it :)

PR #1936, opened 2023-03-25
set vector length of a new throw~ to 64

### LINKED ISSUE #1901: [throw~] is not properly initialized before DSP is turned on
When sending `set öpfel` to a `[throw~ ]` object before DSP is turned on, it triggers an error message:

```
throw~ öpfel: my vector size 1 doesn't match catch (64)
```

Neither of the objects is re-blocked, so it looks like the vector size of `[throw~] ` is only initialized after turning on DSP.  Attached patch shows the error message.

I suspect this got introduced with the new multi-channel stuff.  This is tested with 4baa03e4a4edb5. The message doesn't appear with 0.53-1.
[throw-not-initializing-vector-size.zip](https://github.com/pure-data/pure-data/files/10884377/throw-not-initializing-vector-size.zip)

reduzent:
If we're going to initialize it to some random value, then at least to one that doesn't trigger error messages in most cases.

fixes #1901

### DIFF
--- a/src/d_global.c
+++ b/src/d_global.c
@@ -307,7 +307,7 @@ static void *sigthrow_new(t_symbol *s)
     t_sigthrow *x = (t_sigthrow *)pd_new(sigthrow_class);
     x->x_sym = s;
     x->x_whereto  = 0;
-    x->x_length = 1;
+    x->x_length = 64;
     x->x_f = 0;
     return (x);
 }


umlaeute:
i merged a slightly modified version into `develop`:
- initalize the vector length to an *invalid* value (namely: 0)
- only emit an error message if the vector length is valid (and non-matching, obviously)

this basically defers any error-message until the DSP is turned on (which really is the only time we know the vector length for sure)

Issue #1942, opened 2023-04-08
Allow externals to use 'dbl' inside GOP


porres:
Pd has a restriction where it prevents GOP to handle 'double click' and I have an external that uses it that is broken in GOP because of this.

This also affects list boxes mostly, see -> https://github.com/pure-data/pure-data/issues/1661

This issue is somewhat related to the one above, but is more general and concerns the fact that no external can use double clicks, whereas the other issue deals with specific problems regarding vanilla's gui boxes.

The fix for the general case couldn't be more trivial, just add 'dbl' in the place of '0' here in line '1097'

https://github.com/pure-data/pure-data/blob/master/src/g_graph.c#L1097

This still doesn't fix the other list box issue, but I am sending a PR for this particular issue here soon.

porres:
it seems this can't be solved on its own, so I'm closing and we can just focus on the more complex https://github.com/pure-data/pure-data/issues/1661 issue

Issue #1661, opened 2022-05-08
Bug: Double-click to edit a gatom in a GOP


ludnny:
Hello!

I just discovered that you can edit a symbol box by double-clicking. 

Unfortunately, this "double-click to edit" doesn't work in a Graph On Parent
![38](https://user-images.githubusercontent.com/46849877/167317453-4471a729-cac7-43ec-955f-edb694a3b67e.gif)

Tested on 0.52.2 Win64

umlaeute:
just for completeness sake: your example shows a `list box` rather than a `symbol box`.

however, the problem appears with all gatoms (`number box`, `symbol box`, `list box`)

Ant1r:
also: when `rtext_activate()` sets the `editor.e_textedfor` field, maybe the GOP canvas is used, while it should be the parent canvas:
https://github.com/pure-data/pure-data/blob/ade0160ca291f7fb010ad15c973ed7fce7e1aae5/src/g_rtext.c#L547-L548
(I'm not sure to fully understand the overall process though...)

umlaeute:
>  the GOP canvas is used, while it should be the parent canvas:

yes. however it's much more convoluted.

millerpuckette:
It looks like I was never able to figure out how to get this working so just suppressed passing the double-click into the GOP object.

porres:
> just for completeness sake: your example shows a list box rather than a symbol box.

And for yet more completeness, it seems the restriction is in the Pd guts for whatever any other external that implements a double click feature (that we get from 'w_clickfn' in 't_widgetbehavior') - and I just wrote an external with such a feature and found the same restriction. It's funny, cause other things like shift+click and more work fine, is double click the only odd one out of the picture? What's so troublesome about it?

I am trying to implement a knob box with the same features of VCV's knob and work on some abstractions inspired by modular systems and VCV. In it, you can double click to go to a pre-defined value, like the middle of the knob which can be '0db' for a gain adjustment knob. I want to also add a feature from VCV that you can display the number that the knob is producing when dragging, and double clicking on the number would allow you to also edit the knob value manually by typing a specific number. Since the whole point is to provide external abstractions, it makes the whole deal pointless :( 

[...]

PR #1943, opened 2023-04-08
Allow support for double click inside GOP

porres:
This removes the restriction for handling double click events inside GOP (Graph on Parent).

I have implemented an external that makes use of this and this PR closes https://github.com/pure-data/pure-data/issues/1942

### DIFF
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -1094,7 +1094,7 @@ static int graph_click(t_gobj *z, struct _glist *glist,
                 /* check if the object wants to be clicked */
             if (canvas_hitbox(x, y, xpix, ypix, &x1, &y1, &x2, &y2)
                 &&  (clickreturned = gobj_click(y, x, xpix, ypix,
-                    shift, alt, 0, doit)))
+                    shift, alt, dbl, doit)))
                         break;
         }
         if (!doit)


umlaeute:
I've converted this to `draft` so it doesn't get accidentally merged.

as mentioned in https://github.com/pure-data/pure-data/issues/1661#issuecomment-1245121922 i *think* that there are issues with this naive fix (when I last checked (though I don't think anything has changed since then)).
unfortunately i do not remember these issues, but reconstructing what i thought back then, i think it was
- non-trivial
- hard to explain (that 's why i didn't write it down in the first place)

porres:
Hi @umlaeute I need to ask you again why do you think this is a work in progress and 'naive'. It seems to me you got things confused. This PR only aims at the https://github.com/pure-data/pure-data/issues/1942 issue, but you're bringing up another issue instead.

I already have my [knob] external out in the wild and it's missing its double clicking feature inside GOP. Another external of mine ([numbox~]) also has such a feature and I intend to have yet more externals that use it.

I don't think or can't understand why there'd be a restriction to allow double clicking in GOP subpatches and I think that this fix is very very trivial and I was even wondering that it could be made part of the develop branch...

Can't stress this enough, this has nothing to do with double clicking on number/symbol/list boxes and its issue, but it would also be needed as a first step if you want to fix that later.

I wonder if you don't want to allow this before fixing vanilla's boxes issues, but I don't think we should be prevented from implementing such a feature in externals just because vanilla objects have bugs.

I really hope this makes it into 0.54 and I'm also pinging @millerpuckette to check on this.

thanks

porres:
I wouldn't mind if this means you're attacking this with priority, which implies it'll be fixed before the next release ;) I consider the issue of not being able to double click and edit boxes in GOP serious anyway and it's a bug that should have been fixed in the 0.52 cycle. So, what do you say? Are you on it?

But anyway, I also haven't really understood how simply allowing double clicking could or should be a problem.

cheers

millerpuckette:
I tried it months ago and something broke... I can't remember what.  The code is too fragile to touch right now.

porres:
Hi @millerpuckette and @umlaeute I really need to be able to use double clicking on my GUI externals and it's bad I can't do this inside a GOP abstraction since I'm now about to release an abstraction collection with GUIs.

I have to say I still don't understand the issue in allowing this. The fix is trivial, it's just changing  **"shift, alt, 0, doit"** to **"shift, alt, dbl, doit"** and get the double click value in the graph. I thought it was trivial enough for the 'develop' branch actually.

I wonder if there's a confusion with the other issue with gui boxes, that I can understand could be a complex issue to solve, and that may require changes, but I can't see how changes could affect me just being able to grab double clicking in a GOP.

Double clicking is only used right now for the gui boxes and it's also bad that we can't use it inside a GOP. I really can't see how these are connected and if that needs to be sorted before allowing this fix I'd like to understand please. And if it comes to it, well, since this is a bug that needs fixing anyway, I hope we can try and fix this bug which is kinda overdue now as well.

But again, do we need it? I hope you can help me understand.

thanks

millerpuckette:
No, that change introduces bugs. I 'm trying to figure out how to do 
this correctly but it's a big change - it means

having one text edit buffer globally for the Pd instance instead of one 
per window...

cheers

Miller

porres:
Ok, I would like more details so I could get it, but I'll just take it. Are you trying this for the next update? Should be good to have https://github.com/pure-data/pure-data/issues/1661 fixed for good anyway.

porres:
well @millerpuckette since this is no good, I'm closing this and the issue I referred to, so we just keep https://github.com/pure-data/pure-data/issues/1661 opened

Issue #1904, opened 2023-03-05
Hidden Application after installation on Windows 11 


umlaeute:
I've just installed Pd-0.53-1 on a "Windows 11 Home" machine (that was just converted from "Windows 11 S-Mode" and while *that* installation was almost a year old, no significant modding happened (as is the case with S-Mode), so I consider the machine almost "pristine").

After installation (with the default exe-installer), Pd is launched just fine.

However, afterwards I can no longer find the Pd application launcher.
I can find the `pd.exe` of course, but there is
- no Application shortcut on the Desktop
- no Entry in any "App" menu of the system

of course I am a total noob when it comes to Windows (and even more so in a crippled version like "Home"), so I might just be looking in the wrong places.

Sidenote: file associations work fine.
It just seems to be a problem with registering Pd as an "App" to the system.

Lucarda:
I had just tested the installer on `Windows 11 pro` and I get the descktop shortcut and the entry on the App menu:

 
![win11-post-install](https://user-images.githubusercontent.com/25239522/222991192-8497bdd0-a106-4197-98ea-390d260d3020.png)


I think that on `Win11 home` you have to run the installer with `admin rights`.

you can first uninstall the app from:

`Control Panel\All Control Panel Items\Programs and Features` (paste this on a file explorer address)

re-run the installer (right-click on the installer -> Run as administrator)

Note: This is one of many issues that makes one consider upgrading from `home` to `pro` :)

umlaeute:
thanks for confirming.

> I think that on *Win11 Home* you have to run the installer with **admin rights**.

I think I did.
I'm *pretty* sure that I had to enter the admin-password for the installer (but probably I did not "Run as Administrator").
I'm going to try again in the next few days.

a quick question about your screenshot: in the "Pinned" section it says `pd` (all lower-case), rather than `Pure Data (64bit)`.
Is this by your doing, or something automatic (e.g. because the actual executable is called `pd.exe` rather than `Pure Data (64bit).exe`)?
And does this launch the pd-core or the pd-gui? (see #1905)


> Note: This is one of many issues that makes one consider upgrading from home to pro :)

it's not that I'm actually going to use either *Home* nor *Pro*.
However, I would like users be able to run Pd, even if they somehow decide that they *Home* is their thing.

Lucarda:
> a quick question about your screenshot: in the "Pinned" section it says pd (all lower-case), rather than Pure Data (64bit) (as in the "Recommended" section).
[...]

PR #1944, opened 2023-04-10
nsis: allow the installer to run on non-admin users

Lucarda:
this PR is targeted to `develop` branch.

this rewrite handles 'default writable target dirs' and 'registry settings' depending if the user is admin or not. 

if it's an admin the installation is done for all users.


Lucarda:
@umlaeute here is an artifact to test:

https://nc.nubegris.com.ar/index.php/s/jFwoCmZrbEFjK2n

everything is working Ok and faster (because the older installer printed to the console each file that gets written).

The only minor issue is that we don't get the app listed in the "new" part of the win11 start-menu. (the folder and shortcuts get created)

Lucarda:
here is an artifact to test (non-admin users + icons&art):

https://nc.nubegris.com.ar/index.php/s/b6pXiy7j62oWMs4

umlaeute:
i haven't checked your builds yet, but my local (cross-build) ones look cool... nice!

> microsoft recommends to not provide start-menu shortcuts for un-installers

fair enough.

> The only minor issue is that we don't get the app listed in the "new" part of the win11 start-menu.

any idea what causes this?
or is does this only happen when installing as an ordinary user (as opposed to an admin), as discussed in #1904?

Lucarda:
> any idea what causes this?

not really. I could hack it via d462142 by redefining `SetShellVarContext` in the startmenu section.

I think this PR is done and I'm un-drafting it. I think is Ok to go (i had already tested it enough).

msw/build-nsi.sh
```diff
@@ -190,11 +190,11 @@ case $nsis_exit in
...
-cp "${SCRIPTDIR}/installer-art/big.bmp" ${OUTDIR}/big.bmp
-cp "${SCRIPTDIR}/installer-art/small.bmp" ${OUTDIR}/small.bmp
-cp "${SCRIPTDIR}/installer-art/small-un.bmp" ${OUTDIR}/small-un.bmp
-cp "${SCRIPTDIR}/installer-art/pd-un.ico" ${OUTDIR}/pd-un.ico
-cp "${SCRIPTDIR}/../tcl/pd.ico" ${OUTDIR}/pd.ico
+cp ${SCRIPTDIR}/installer-art/big.bmp ${OUTDIR}/big.bmp
+cp ${SCRIPTDIR}/installer-art/small.bmp ${OUTDIR}/small.bmp
+cp ${SCRIPTDIR}/installer-art/small-un.bmp ${OUTDIR}/small-un.bmp
+cp ${SCRIPTDIR}/installer-art/pd-un.ico ${OUTDIR}/pd-un.ico
+cp ${SCRIPTDIR}/../tcl/pd.ico ${OUTDIR}/pd.ico
```
umlaeute, line 197: that doesn't look right.
why are the quotes removed? (was that me??)
there should *also* be quotes around the `${OUTDIR}/...`

msw/pd.nsi
```diff
@@ -144,6 +147,12 @@ SectionGroup /e "${PRODUCT_NAME}"
     SectionEnd
 
     Section "Create Startmenu entry" StartMenu
+        ; ugly hack so the app shows up in "recently added"
+        ${If} $CONTEXT == "all"
+            SetShellVarContext all
+        ${Else}
+            SetShellVarContext current
```
umlaeute, line 154: can't we use `SetShellVarContext $CONTEXT` (also in L60 resp L65)

msw/build-nsi.sh
Lucarda: i did a `force push` that didn't change anything (my bad) and 1minute after you `force push`ed the fix. while I was trying to fix my bad I got lost in history. I accidentally tried to fix conflicts but something went wrong.

msw/pd.nsi
Lucarda: I did try to use `SetShellVarContext $CONTEXT` (that was the initial intencion) but it didn't worked for me. I might have it wrong but it looks the compiler does not allow the expansion there? 

will investigate.

Lucarda: not sure why but we can't:

```
Usage: SetShellVarContext all|current
Error in script "C:/msys64/tmp/tmp.qMn9ZCGmH5/pd.nsi" on line 156 -- aborting creation process
```

if i pop a messegebox i know `$CONTEXT` expands to `all`. the msgbox pops up almost at the end of the "copy files" part.

umlaeute:
Don't worry. NSIS syntax is arcane...

msw/pd.nsi
Lucarda: > Don't worry. NSIS syntax is arcane...

good.

msw/build-nsi.sh
Lucarda: now fixed via 687c970

umlaeute:
i did a quick test on a VM, and it looks very cool.

two tiny remarks:
- when installing as a unprivileged user, I was nevertheless asked for an admin password (after giving the admin password, the installation proceeded to per-user setup).
  the admin password popup appeared to be in the context of "Unsafe application downloaded from the internet", so this might just be another stab of M$ to prevent innocent users from installing random malware. is this normal?
-  the icon with the red toggle for uninstalling is nice, though it raised a slight irritation in me (but nothing i couldn't live with)

and since we are there:
some other applications show the "Estimated installed size" in the `Add/Remove Programs` applet.
should we add something like this as well?

it's really lowest priority, but just in case: https://superuser.com/questions/570929/

Lucarda:
> when installing as a unprivileged user, I was nevertheless asked for an admin password (after giving the admin password, the installation proceeded to per-user setup).
the admin password popup appeared to be in the context of "Unsafe application downloaded from the internet", so this might just be another stab of M$ to prevent innocent users from installing random malware. is this normal?

I don't understand why is this happening. on my win11 *pro* dev machine i have


|user   |with pass  |prompts for password   |installs to|
|:---   |:---:      |:---:                  | ---:      |
|admin  |no pass    |yes                    |all users  |
|user1  |no pass    |no                     |user       |
|user2  |pass       |no                     |user       |

and all works as expected.

note that same file downloaded from the internet on the "admin" user the icon gets an added shield:

nsis-shield

while on the other unprivileged users not

nsis-no-shield



- Are you using a Win *home* edition?

- does your non-privileged user has a password?
  - what happens if you provide the *user* pass instead of the *admin* pass

Note that at the nsis code level there's nothing else to do (i think).

umlaeute:
- OS: `Windows 10 Enterprise Evaluation`
- the non-privilged user has a password (the admin has a password as well)
- before the install, i downloaded Pd as the user who was to perform the installation.

(in the meantime i have reset the VM, so I cannot check for any icon shield)

Lucarda:
there might be some peculiarity I don't know on *enterprise*.

not sure if this helps: https://learn.microsoft.com/en-us/windows/security/identity-protection/user-account-control/how-user-account-control-works

umlaeute:
This is what I get when running as the ordinary (non-admin, but password-protected) user:

smartscreen

i get the same when running as an ordinary (non-admin) *password-less* user.

there's no way to get past this screen without entering the password of an administrator.

i figure that the [SmartScreen](https://en.wikipedia.org/wiki/Microsoft_SmartScreen#SmartScreen_in_Windows) can be configured to be less strict, and my Win10/Enterprise probably has strict settings.

reasons that i could think of why you are not seeing this:
- Win11/Pro has less strict settings
- you manually relaxed your SmartScreen settings
- you build the installer on the computer you are testing (and Windows recognized the file as "produced here" and thus whitelisted it)


in any case: after allowing the installation with the admin password, the binary is marked as "good", and afterwards I can run the installer again (either as a normal user or as admin) without being prompted any more.

umlaeute:
and here's *yet another* request (i do appreciate everything you put into this!).

why does the installer-art has this weird *brownish* color-gradient?
I don't remember having seen this color in the context of Pd ever.
could we pick a more Pdish color? (e.g. tcl-grey; or one of those pesky pastell colors that iemguis love so much; or just plain white without a gradient?)

Lucarda:
> why does the installer-art has this weird brownish color-gradient?

now using grey background.

preview: https://nc.nubegris.com.ar/index.php/s/EgsHCQFYQnSKxng

Lucarda:
here is an artifact for 671f385 :https://nc.nubegris.com.ar/index.php/s/pzRNEdjnRx9MGsn

i think this PR is done so i'm un-drafting it.

Lucarda:
artifact for cf1d133 : https://nc.nubegris.com.ar/index.php/s/M4M9xTF3Dd4aZj9

Lucarda:
artifact for a5f4385 : https://nc.nubegris.com.ar/index.php/s/npLTNp7jPS9bA6D

Lucarda:
@umlaeute i think this is ready to go.

When 0.54 is out i would like to add it to [winget]( https://learn.microsoft.com/en-us/windows/package-manager/winget/) the Windows package manager (works out-of-the-box on Win11 or on up-to-date-Win10). So you can do on a terminal:

`winget install puredata`

if you do now `winget search puredata` you get 0.51. Someone has uploded the `Manifest`: 

https://github.com/microsoft/winget-pkgs/tree/master/manifests/m/MillerPuckette/PureData

Anyway: is there something else to be done before you merge?

umlaeute:
@Lucarda now that the branch has been merged, you could delete it from your repository :-)
https://github.com/lucarda/pure-data/branches

Lucarda:
done

Issue #1959, opened 2023-05-27
Record window state within top level patch


pmallard:
Hello everyone,

Currently PD does not save the window state property (state "zoomed" that occur when maximizing a window or "normal" for not maximized but visible windows). 

It may then open patches on unwanted position (i.e. the user save in the maximize state but the window load on the former position, before the maximized state).

I propose to record the windows state within pd top level patches in order to be able to open the patch in the maximize state.

This change leads to a modification of the .pd file format to add the window state on the top level canvas line.
When the window state is missing it would default to normal state (i.e. neither maximized or minimized).

This may be at the very first line of the pd file as follows for a maximized window:
#N canvas 86 117 1600 817 10 zoomed;

Another example for a normal one :
#N canvas 86 117 1600 817 10 normal;

Unfortunately this change will impact the pd message 'setbounds' and may leads to regression with other window manager than the default tcl/tk one.
 
Regards,

umlaeute:
see my notes in the PR why i think that this is a bad idea.

PR #1960, opened 2023-05-27
Add wm_state in pd files updated by setbounds

pmallard:
This pull request intends to add the wm state (zoomed / normal) in the pd file as set when user save its pd file.
Reading a pd file without this state set it to normal.
See #1959 

I tested it on mingw64 host with both save/load of new .pd file and former ones.

A clear limitation of this patch is that it may leads to regression with other gui than the default tcl tk one (i.e. modification of pd communication protocol on "setbounds" command) .

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -685,7 +685,6 @@ static void canvas_dosetbounds(t_canvas *x, t_symbol * wm_state, int x1, int y1,
         pd_error(0, "Error recording wm_state");
     }
 
-
     if (!glist_isgraph(x) && (x->gl_y2 < x->gl_y1))
     {
             /* if it's flipped so that y grows upward,
@@ -1177,7 +1176,7 @@ static void canvas_relocate(t_canvas *x, t_symbol *canvasgeom,
             we just suppress that here. */
     if (cw > 5 && ch > 5)
         canvas_dosetbounds(x, gensym("normal"), txpix, typix,
-                           txpix + cw, typix + ch);
+            txpix + cw, typix + ch);
 }
 
 void canvas_popabstraction(t_canvas *x)
@@ -2039,9 +2038,9 @@ void g_canvas_setup(void)
         /* we prevent the user from typing "canvas" in an object box
         by sending 0 for a creator function. */
     canvas_class = class_new(gensym("canvas"), 0,
-                             (t_method)canvas_free, sizeof(t_canvas),
-                             CLASS_NOINLET | CLASS_MULTICHANNEL, 0);
-        /* here is the real creator function, invoked in patch files
+        (t_method)canvas_free, sizeof(t_canvas),
+            CLASS_NOINLET | CLASS_MULTICHANNEL, 0);
+            /* here is the real creator function, invoked in patch files
             by sending the "canvas" message to #N, which is bound
             to pd_camvasmaker. */
     class_addmethod(pd_canvasmaker, (t_method)canvas_new, gensym("canvas"),
@@ -2051,7 +2050,7 @@ void g_canvas_setup(void)
     class_addmethod(canvas_class, (t_method)canvas_coords,
         gensym("coords"), A_GIMME, 0);
 
-        /* -------------------------- objects ----------------------------- */
+/* -------------------------- objects ----------------------------- */
     class_addmethod(canvas_class, (t_method)canvas_obj,
         gensym("obj"), A_GIMME, A_NULL);
     class_addmethod(canvas_class, (t_method)canvas_msg,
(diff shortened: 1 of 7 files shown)


src/g_canvas.c
```diff
@@ -649,24 +660,32 @@ int glist_isgraph(t_glist *x)
 
     /* This is sent from the GUI to inform a toplevel that its window has been
     moved or resized. */
-static void canvas_setbounds(t_canvas *x, t_float left, t_float top,
-                             t_float right, t_float bottom)
+static void canvas_setbounds(t_canvas *x, t_symbol * wm_state, t_float left, t_float top,
+                             t_float right, t_float bottom )
```
umlaeute, line 664: i wouldn't change the interface for the canvas' `setbound` message like this.
probably just use `A_GIMME` and do the typechecking yourself
(this also removes the confusion of the order of arguments)

src/g_editor.c
```diff
@@ -1900,10 +1905,10 @@ void canvas_vis(t_canvas *x, t_floatarg f)
                 sprintf(winpos, "+%d+%d", (int)(x->gl_screenx1), (int)(x->gl_screeny1));
             }
 
-            pdgui_vmess("pdtk_canvas_new", "^ ii si", x,
+            pdgui_vmess("pdtk_canvas_new", "^ ii si s", x,
                 (int)(x->gl_screenx2 - x->gl_screenx1),
                 (int)(x->gl_screeny2 - x->gl_screeny1),
-                winpos, x->gl_edit);
+                        winpos, x->gl_edit, x->gl_wm_state->s_name);
```
umlaeute, line 1911: please do not introduce any trailing whitespace.
(even better: actively remove any trailing whitespace in the files you touch; use a separate commit for whitespace-fixes)

tcl/pdtk_canvas.tcl
```diff
@@ -95,7 +95,7 @@ proc pdtk_canvas_place_window {width height geometry} {
 #------------------------------------------------------------------------------#
 # canvas new/saveas
 
-proc pdtk_canvas_new {mytoplevel width height geometry editable} {
+proc pdtk_canvas_new {mytoplevel width height geometry editable wm_state} {
```
umlaeute, line 98: that could be made an optional argument with `{wm_state {}}`

src/g_canvas.c
```diff
@@ -2019,9 +2039,9 @@ void g_canvas_setup(void)
         /* we prevent the user from typing "canvas" in an object box
         by sending 0 for a creator function. */
     canvas_class = class_new(gensym("canvas"), 0,
-        (t_method)canvas_free, sizeof(t_canvas),
-            CLASS_NOINLET | CLASS_MULTICHANNEL, 0);
-            /* here is the real creator function, invoked in patch files
+                             (t_method)canvas_free, sizeof(t_canvas),
+                             CLASS_NOINLET | CLASS_MULTICHANNEL, 0);
+        /* here is the real creator function, invoked in patch files
```
umlaeute, line 2044: the Pd-codebase indents comments by an extra 4 spaces (there's multiple occurences where you changed that comment-indentation for no obvious reason)

umlaeute (changes requested):

umlaeute:
hmm, i'm not entirely sure about the purpose of this PR.

as this is mostly using Tcl/Tks `wm state` proc, i figure that when you say "zoomed / normal" you do *not* talk about Pd's built-in zoom, but about some feature of your window manager. is this correct?

now the [docs](https://www.tcl.tk/man/tcl/TkCmd/wm.html#M62) mention that there is a "zoomed" state - but only on macOS and Windows.
and indeed, doing something like `wm state .pdwindow zoomed` on my Linux/XFce4 system, just throws me an error.
that would basically make patches built on Windows/macOS incompatible with a Linux runtime, which I think we want to avoid.
also, the other (cross-platform) states are `iconic` and `withdrawn`, which seems not very useful to me.

finally: I do not like the idea of encoding some Tcl/Tk keywords in the Pd patch.
we are trying to remove TclTk-idiosyncracies, rather than add them.

src/g_editor.c
pmallard: My bad I missed that one, my editor default was to align argument list to function parenthesis, I corrected it. Will add a new commit for fixing this one.

src/g_canvas.c
pmallard: I have missed this change, sorry. I will correct that in the next commit.

pmallard:
The idea behing this PR is to have the user the ability to work its patch layout completly. To me recording the geometry in the patch is not enough since it does not reflect the maximize state.
Indeed when using the maximize button of a window, the window geometry is left unchanged (at least on mingw).

If this is of any interest, maybe we could instead of recording the window state zoomed (that is not portable) using just a boolean in pd core code letting the tcl part with the interpretation of this boolean.
This would require however to have the tcl part being specific.

src/g_canvas.c
pmallard: Yes I realize that symbol argument needed to be passed at the beginning while my first choice was to pass it at the end.
Maybe if we move to a boolean (see my comment below) there would be a better/simpler type candidate for that (A_FLOAT ??)

umlaeute:
currently there appears to be no way to programmatically get the maximization state in a cross-platform way.
afaict, there are systems (like mine) that do not have a way to *set* the maximization state.

finally, even if all systems supported getting and setting the maximization of a window, the result will be *very* specific to the actual hardware setup: on your 4480×2520 screen a maximized windows means something very different than on my 1024x768 screen.

i think the suggested feature makes patches less portable.

pmallard:
Ok in line with your comment... Closing this PR then.

Issue #1961, opened 2023-05-27
Build a debug version on windows mingw


pmallard:
Hi,

It is is currently not possible to build a debug version (with symbols embedded) on windows due to the call of msw-app.sh that strip binaries in a way not recovered by libtool --mode=execute.

To be able to debug gdb on windows (at least msys64) -n shall be added to msw-app.sh call.

Another strange thing is the -s added to EXTERNAL_LDFLAGS for mingw host not present anywhere else in configure.ac.

I propose a patch that add the -n flag on msw-app.sh call and remove the -s flag of EXTERNAL_LDFLAGS when configure is called with --enable-debug. I also add extra comment in msw/README.txt

However the removal of the -s flag of EXTERNAL_LDFLAGS has not been tested on my side, and I really have no idea of its impact, I may propose an updated version that would revert this change. Let me know.

This issue is linked some how to #1957 that aims to correct more deeply the compile time problem due to the installation steps.

Spacechild1:
Closed via https://github.com/pure-data/pure-data/pull/1962

PR #1962, opened 2023-05-27
Feature/msw compile for debug

pmallard:
This pull request propose to be able to build on mingw64 in debug mode (i.e. with symbols findable by gdb) as discussed in #1961  .
It adds the -n flag on the call to msw-app.sh when configure is called with --enable-debug and also remove -s on the EXTERNAL_LDFLAGS.

Note that the modification around EXTERNAL_LDFLAGS is not tested and should be removed from the patch (I can correct it on a subsequent commit).

Also add some details in the README.txt

I tested this sequence on mingw64 :

```
mkdir build
cd build
../configure --enable-debug CFLAGS=-g
make
make app
libtool --mode=execute gdb --args pd-0.53.1/bin/pd.exe
```

I have been able to test modification proposal of #1960 with it.

Thx for your attention.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -59,7 +59,6 @@ PD_LDFLAGS=""
 PD_CHECK_IPHONE(IPHONEOS=yes, IPHONEOS=no, AC_MSG_ERROR([iOS SDK not available]))
 PD_CHECK_ANDROID(ANDROID=yes, ANDROID=no, AC_MSG_ERROR([Android SDK not available]))
 
-
 AS_CASE([$host],
 [*darwin*],[
     AS_IF([test "x${IPHONEOS}" = "xno"],[
@@ -134,6 +133,7 @@ AS_CASE([$host],
 
     # externals are dynamically linked to pd.dll in their individual automake files
     EXTERNAL_CFLAGS="-mms-bitfields"
+    EXTERNAL_LDFLAGS="-Wl,--enable-auto-import -no-undefined -lpd"
     EXTERNAL_EXTENSION=dll
 
     # workaround for rpl_malloc/rpl_realloc bug in autoconf when cross-compiling
@@ -274,6 +274,7 @@ AC_SUBST([ALSA_LIBS])
 AC_SUBST([JACK_CFLAGS])
 AC_SUBST([JACK_LIBS])
 
+# Whether to strip symbol from binaries (msw_strip="") or not (msw_strip="-n") on windows. Used in msw/Makefile
 AC_SUBST([msw_strip])
 
 # pass include paths down to all Makefiles
@@ -292,10 +293,8 @@ AS_IF([test x$debug = xyes],[
 
 AS_IF([test "x${debug}" = "xyes" -a "x${WINDOWS}" = "xyes" ],[
     msw_strip="-n"
-    EXTERNAL_LDFLAGS="-Wl,--enable-auto-import -no-undefined -lpd"
     ],[
     msw_strip=""
-    EXTERNAL_LDFLAGS="-s -Wl,--enable-auto-import -no-undefined -lpd"
     ])
 
 ##### macOS version min #####
--- a/configure.ac
+++ b/configure.ac
@@ -59,6 +59,7 @@ PD_LDFLAGS=""
 PD_CHECK_IPHONE(IPHONEOS=yes, IPHONEOS=no, AC_MSG_ERROR([iOS SDK not available]))
 PD_CHECK_ANDROID(ANDROID=yes, ANDROID=no, AC_MSG_ERROR([Android SDK not available]))
 
+
 AS_CASE([$host],
 [*darwin*],[
     AS_IF([test "x${IPHONEOS}" = "xno"],[
@@ -133,7 +134,6 @@ AS_CASE([$host],
 
     # externals are dynamically linked to pd.dll in their individual automake files
     EXTERNAL_CFLAGS="-mms-bitfields"
-    EXTERNAL_LDFLAGS="-s -Wl,--enable-auto-import -no-undefined -lpd"
     EXTERNAL_EXTENSION=dll
 
     # workaround for rpl_malloc/rpl_realloc bug in autoconf when cross-compiling
@@ -274,6 +274,8 @@ AC_SUBST([ALSA_LIBS])
 AC_SUBST([JACK_CFLAGS])
 AC_SUBST([JACK_LIBS])
 
+AC_SUBST([msw_strip])
+
 # pass include paths down to all Makefiles
 AC_SUBST([AM_CPPFLAGS], [$AM_CPPFLAGS])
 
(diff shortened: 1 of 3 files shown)


umlaeute:
is the `CFLAGS=-g` really needed?
as discussed at length in #1727, the `-g` -flag is already present in the default CFLAGS.

or is this merely to get rid of the `-O2` optimization?

configure.ac
```diff
@@ -133,7 +134,6 @@ AS_CASE([$host],
 
     # externals are dynamically linked to pd.dll in their individual automake files
     EXTERNAL_CFLAGS="-mms-bitfields"
-    EXTERNAL_LDFLAGS="-s -Wl,--enable-auto-import -no-undefined -lpd"
```
Spacechild1, line 136: hmmm... wouldn't it make sense to keep the Windows specific flags here and then just append `-s` later in the `AS_IF([test "x${debug}"` clause?

Also, `-s` should be added to a debug build for *all* platforms, not only on Windows. (In practice, the default CFLAGS may already contain the `-s` flag, but it wouldn't hurt to set it explicitly. After all, that's what you want when you ask for a debug build...)

Spacechild1, line 136: I would keep the platform-specific link flags here. Can't we just remove the `-s` all together? I think the question of stripping should be confined to `make app`.

msw/README.txt
```diff
@@ -142,3 +142,17 @@ Pd font.
 Currently, pdfontloader.dll is pre-built and included within the pdprototype.tgz
 tarball. To build pdfontloader, see https://github.com/pure-data/pdfontloader
 source.
+
+## debug
+
+To avoid stripping symbols out of libtool pre-binaries and debug pd,call configure with --enable-debug and CFLAGS=-g :
```
Spacechild1, line 148: If you manually set `CFLAGS` to `-g`, this will implicitly remove optimization flags. This is typically fine for debug builds, but maybe it would be better to make this explicit, e.g.:

```
If you intend to debug Pd, you probably want an unoptimized build with debug symbols. This can be achieved with the following flags:

./configure --enable-debug CFLAGS="-g -O0"
make
make app
```

(Technically, `CFLAGS=-g` would be sufficient, as it would remove the default CFLAGS, but I don't think that this is obvious.)

msw/README.txt
```diff
@@ -142,3 +142,17 @@ Pd font.
...
+
+./configure --enable-debug CFLAGS=-g
+make
+make app
+
+This will call msw-app.sh with the -n flag (through msw_strip variable)
```
Spacechild1, line 154: Maybe better:

> This will call msw-app.sh with the -n flag, preventing it from striping the debug symbols

msw/README.txt
```diff
@@ -142,3 +142,17 @@ Pd font.
...
+
+The call to gdb shall be perform using libtool mode execute :
+libtool --mode=execute gdb --args pd-0.53.1/bin/pd.exe
```
Spacechild1, line 157: What does `libtool --mode=execute` do? I always just do `gdb --args pd-0.x.x/binpd.exe`.

Spacechild1:
I agree that `make app` should not strip the binaries if configured with `--enable-debug`. Don't know why I never had the idea... 

I just have a couple of small questions/comments. After these are sorted out, I think we can merge this into `develop`.

Thanks!

msw/README.txt
pmallard: It's ok for me but when you call it that way, resulting CFLAGS is -O0 -g -O0 since -O0 is prepended in configure script.
It is harmless but redundant ...

As @umlaeute mentions in #1727 -O2 is append when CFLAGS are not set explicitly. This is such an odd behavior to me.
Another option would be to force a CFLAGS reset (with CFLAGS=) and somehow add -g in configure script when --enable-debug is defined, but I think this is discussed in details in #1727 .

Spacechild1: > -O2 is append when CFLAGS are not set explicitly. This is such an odd behavior to me.

What really happens is that if CFLAGS are not set explicitly, our build system uses the default CFLAGS (that are set by the build environment). On Msys2 this happens to be `-O2`. Now, `--enable-debug` does add `-O0`, but as you note correctly, it is *prepended* to the default CFLAGs, which makes it pretty useless...

But yeah, this is already discussed at length in #1727 and out of scope for this PR. The current behavior is weird anyway, so it probably does not matter whether you suggest `CFLAGS=-g` or `CFLAGS=-g O0`.

---

Side note:

IMO, the underlying problem is that autotools has no notion of build profiles. CMake, on the other hand, provides several useful build profiles out of the box (`Debug`, `Release`, `RelWithDebInfo` and `MinSizeRel`) which just do The Right Thing®.

pmallard: I do like CMake also. Moving to it or any up to date build system could be a great idea.

umlaeute: this appears to be very OT for this PR.

msw/README.txt
pmallard: I have double check and you are totally right, the call to libtool is not needed, one can call gdb directly.

I have found this libtool mode when I was still looking for my symbol but it is useless and maybe not appropriate at all in our context => Removed from README.txt

msw/README.txt
pmallard: Ok updated

configure.ac
pmallard: If I understand correctly what you suggest is removing the EXTERNAL_LDFLAGS "-s" completly.
I really have no idea of the effect since I was much more focus on core than on external in the development for now.
What I can propose is to :
1) Leave the EXTERNAL_LDFLAGS unchanged and remove all that modify EXTERNAL_LDFLAGS from my PR
2) remove the -s only from this line and prepend it in the AS_IF debug section later if debug is yes.

msw/README.txt
Spacechild1: Sorry, I couldn't resist :-)

configure.ac
Spacechild1: > If I understand correctly what you suggest is removing the EXTERNAL_LDFLAGS "-s" completly

Yes. There is no reason why we should strip the binaries by default. Stripping only makes sense if you want to distribute the binaries and save some space, and for that you would need `make app` anyway. Hence my suggestion that stripping should only be done by `make app` (if requested).

> remove the -s only from this line and prepend it in the AS_IF debug section later if debug is yes.

`-s` means "strip". You certainly don't want to do that for a debug build :-)

Spacechild1: (Actually, I don't even understand how the `-s` got there in the first place... After all, the Pd core is *not* stripped by default.)

pmallard: Ok will remove the -s then from EXTERNAL_LDFLAGS

pmallard: Should be ok now but strangely the review is not mark Outdated...
Could you please have a check and resolve this one after ?

Spacechild1: Yeah, that's weird.

Spacechild1:
Looks good now! I tested it locally and it works fine. Thanks for the PR, I find it very useful!

@umlaeute If you don't object, I would squash the commits and merge them into develop.

Spacechild1:
Merged into develop. Thanks!

PR #1963, opened 2023-06-04
stable development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/1907

---

### Core
- fix return function value in `pd~` (this is *so* embarrassing; i pinged yesterday about the inclusion of #1605 and made sure to fix this issue *before* the ping - but then forgot to push my changes)
- some very minor code cleanup

### GUI
- prevent certain versions of Tcl/Tk on macOS (that are known to crash otherwise) to use tabbed preferences (and fallback to a single-page preference)
- improve dialog for selecting the interface language

### build-system
- Closes: https://github.com/pure-data/pure-data/pull/1962


Spacechild1:
Hmmm, looks like https://github.com/pure-data/pure-data/pull/1962 introduced a blank line which makes the CI fail. Strangely, the CI always seemed to pass in the PR itself, that's why I didn't notice...

umlaeute:
> Strangely, the CI always seemed to pass in the PR itself, that's why I didn't notice...

that's because the CI is not run at all for first-time contributors (as a safety measure to prevent malicious parties to collect secrets exposed within the pipeline).

in any case, I've fixed the trailing whitespace...

PR #1978, opened 2023-06-10
Some missing multichannel objects

Spacechild1:
* `[print~]`: print multichannel signals:
    ```
    <symbol>:
    channel 1:
    <data>
    channel 2:
    <data>
    ...
    ```
* `[snapshot~]`: handle multichannels signals by outputting a list

* `[sig~]`: multiple creation arguments generate a multichannel signal; each scalar value gets a float inlet.

### DIFF
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -15,31 +15,57 @@ static t_class *sig_tilde_class;
 typedef struct _sig
 {
     t_object x_obj;
-    t_float x_f;
+    t_atom *x_vec;
+    int x_n;
 } t_sig;
 
-static void sig_tilde_float(t_sig *x, t_float f)
+static void sig_tilde_float(t_sig *x, t_floatarg f)
 {
-    x->x_f = f;
+    x->x_vec[0].a_w.w_float = f;
 }
 
 static void sig_tilde_dsp(t_sig *x, t_signal **sp)
 {
-    dsp_add_scalarcopy(&x->x_f, sp[0]->s_vec, (t_int)sp[0]->s_n);
+    int i;
+    signal_setmultiout(sp, x->x_n);
+    for (i = 0; i < x->x_n; i++)
+        dsp_add_scalarcopy(&x->x_vec[i].a_w.w_float,
+            sp[0]->s_vec + i * sp[0]->s_n, (t_int)sp[0]->s_n);
 }
 
-static void *sig_tilde_new(t_floatarg f)
+static void *sig_tilde_new(t_symbol *s, int argc, t_atom *argv)
 {
+    int i;
     t_sig *x = (t_sig *)pd_new(sig_tilde_class);
-    x->x_f = f;
+    if (argc > 0)
+    {
+        x->x_vec = (t_atom *)getbytes(argc * sizeof(*x->x_vec));
+        for (i = 0; i < argc; i++)
+            SETFLOAT(x->x_vec + i, atom_getfloat(argv + i));
+        x->x_n = argc;
+    }
+    else
+    {
+        x->x_vec = (t_atom *)getbytes(sizeof(*x->x_vec));
+        SETFLOAT(x->x_vec, 0);
+        x->x_n = 1;
+    }
+    for (i = 1; i < x->x_n; i++)
+        floatinlet_new(&x->x_obj, &x->x_vec[i].a_w.w_float);
     outlet_new(&x->x_obj, gensym("signal"));
     return (x);
 }
 
+static void sig_tilde_free(t_sig *x)
+{
+    freebytes(x->x_vec, x->x_n * sizeof(*x->x_vec));
+}
+
 static void sig_tilde_setup(void)
 {
-    sig_tilde_class = class_new(gensym("sig~"), (t_newmethod)sig_tilde_new, 0,
-        sizeof(t_sig), 0, A_DEFFLOAT, 0);
+    sig_tilde_class = class_new(gensym("sig~"),
+        (t_newmethod)sig_tilde_new, (t_method)sig_tilde_free,
+            sizeof(t_sig), CLASS_MULTICHANNEL, A_GIMME, 0);
     class_addfloat(sig_tilde_class, (t_method)sig_tilde_float);
     class_addmethod(sig_tilde_class, (t_method)sig_tilde_dsp,
         gensym("dsp"), A_CANT, 0);
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -388,52 +388,80 @@ static t_class *snapshot_tilde_class;
 typedef struct _snapshot
 {
     t_object x_obj;
-    t_sample x_value;
     t_float x_f;
+    int x_n;
+    t_atom *x_vec;
 } t_snapshot;
 
 static void *snapshot_tilde_new(void)
 {
     t_snapshot *x = (t_snapshot *)pd_new(snapshot_tilde_class);
-    x->x_value = 0;
-    outlet_new(&x->x_obj, &s_float);
+    x->x_vec = getbytes(sizeof(t_atom));
+    SETFLOAT(x->x_vec, 0);
+    x->x_n = 1;
     x->x_f = 0;
+    outlet_new(&x->x_obj, &s_float);
     return (x);
+}
+
+static void snapshot_tilde_free(t_snapshot *x)
+{
+    freebytes(x->x_vec, x->x_n * sizeof(t_atom));
 }
 
 static t_int *snapshot_tilde_perform(t_int *w)
 {
     t_sample *in = (t_sample *)(w[1]);
-    t_sample *out = (t_sample *)(w[2]);
-    *out = *in;
-    return (w+3);
+    t_atom *out = (t_atom *)(w[2]);
+    int nchans = (int)(w[3]);
+    int n = (int)(w[4]), i;
+    for (i = 0; i < nchans; i++)
+        SETFLOAT(out + i, in[i * n]);
+    return (w+5);
 }
 
 static void snapshot_tilde_dsp(t_snapshot *x, t_signal **sp)
 {
-    dsp_add(snapshot_tilde_perform, 2, sp[0]->s_vec + (sp[0]->s_n-1),
-        &x->x_value);
+    int i, nchans = sp[0]->s_nchans;
+    if (nchans != x->x_n)
+    {
+        x->x_vec = (t_atom *)resizebytes(x->x_vec,
+            x->x_n * sizeof(t_atom), nchans * sizeof(t_atom));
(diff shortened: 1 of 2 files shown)


porres:
> "start channel" should really be the start channel, like in dac~, and not a zero-based offset

I agree this is better 

> in multichannel mode, the "set" message should really be "set [nchans] [start channel]"

how is it now? I thought it only worked for non multichannel and didn't even document this

Spacechild1:
> how is it now? I thought it only worked for non multichannel and didn't even document this

As I described it: "set [nchans] [start channels]", i.e. the same as the creation arguments.

millerpuckette:
I'm not sure about offset versus starting-channel in the "set" message - I could go one way or the others.  OTOH I want to avoid dynamically changeable channel counts for now - that will be much easier to add later than to remove if they tirn out to be problematic in the future.

Spacechild1:
I removed the `[clone]` flag changes again.

> I'm not sure about offset versus starting-channel in the "set" message - I could go one way or the others

I think the "set" message, if supported, should mirror the creation arguments. IMO it's clear that "start channel" should be, well, the start channel :) I find it very weird that `[adc~ -m 2 1]` would output channels 2 and 3, while `[snake~ in 2] -> [dac~ 1]` would write to channels 1 and 2.

> OTOH I want to avoid dynamically changeable channel counts for now

That ship has sailed already ;-) `[resize <n>( -> [clone -do]`

Generally, I don't see the issue with dynamically changeable channel counts at all. What kind of problems do you anticipate? Note that the DSP graph is updated whenever the channel count changes. Seems 100% safe to me...

Spacechild1:
To give another argument: why is it ok to dynamically change the blocksize, resampling factor or overlap of a canvas (via the "set" message of `[block~]`), but dynamically changing the channel should be somehow problematic?

Spacechild1:
@millerpuckette I have seen that you have fixed the "start channel" creation argument - thanks!

I have now split the "set" message part into two commits and moved them to the end.

The first commit (bd83549) only allows to set the start channel, basically the same as in `[dac~]`.

The second commit (0714131) adds an optional second argument that allows to change the number of channels.

If you are still hesitant about changing channel counts with messages, you can just merge up to bd83549, leaving 0714131 for later. Either way we would have a reasonable behavior for the "set" message in multichannel mode.

millerpuckette:
I've been conflating two ideas in my head... in fact, I think I have nothing against having messages to change the number of channels output by a receive~/catch~/adc~ - but I'm scared of having these channel counts automatically changed because of the existence of "send~", etc., objects.  Another consideration is that it's late in the game to be adding features - I'm trying to get things stabilized so we can start finding the dozens of problems I've already introduced for the next release :)

Spacechild1:
> but I'm scared of having these channel counts automatically changed because of the existence of "send~", etc., objects.

Nothing to be scared of :)

Anyway, you are right that this already a huge release and there will enough bugs to tackle already. The "channel" count message for `[send~]` and `[catch~]` can always be added later. I was more worried about the (IMO) redundant "channels" argument in `[receive~]`, but now I realized that we can make it optional again later:

a) if the argument is provided, use a static channel count (and warn about mismatching send channel count)

b) if omitted, get the channel count automatically from the send object (and update the DSP graph if necessary)

millerpuckette:
I think "adc set" should take a list, so that if you've got a wierd situation you can do funny stuff.  For example, some 6-channel guitars are 1-2-3-4-5-6 and others are 6-5-4-3-2-1.  So I think it's best for "set" to take a whole list.  OTOH this would be inappropriate as creation arguments because it would be messy to make the number variable in that way.
But there's also a question here - should the "set" message then also change the number of channels of adc~ (as we're thinking catch~ etc could be?)  or should that be a separate "channels" message to match the other ones?

Spacechild1:
Sorry, I got terribly confused. For some reason I thought that `[set( -> [dac~]` would change the offset, but it rather sets the individual channels. So at least the behavior is the same as with `[adc~]`.

>  For example, some 6-channel guitars are 1-2-3-4-5-6 and others are 6-5-4-3-2-1.

Ok, that's a valid use case! Of course, the channels might not be consecutive.

Conversely, when I want to get channels 2-128 from `[adc~]`, I don't want to type/generate a huge list...

Also, I don't want the number of channels to be fixed at creation time.

I think we really need both behaviors:
a. set a list of individual channels
b. set a range of channels with start + count

a. is covered by `[set(`, so we can leave it as is. Sorry for the noise.

We would probably need another message for b. I would suggest something like `[channels <start> [<nchans>](` or `[channels <nchans> [<start>](` for `[adc~]` and `[start <f>(` for `[dac~]`. This can be added later, of course.

---

The remaining question, which you have already raised, is whether `[set(` for `[adc~]` should be able to change the number of channels. My hunch would be "yes". Alternatively, you *could* do `[channels(` followed by `[set(`, but this feels a bit convoluted to me.

millerpuckette:
Yep... I think this is bad and stupid.  Probably there should be a "set" message that works as in adc and a whole other message that just sets to consecutive channels with a selected channel count and offset... sigh...

Spacechild1:
We can now move the discussion to https://github.com/pure-data/pure-data/issues/1984 :)

umlaeute:
i think `[readsf~]` and `[writesf~]` could profit from multichannel capabilities as well.

millerpuckette:
I think sig~, snapshot~, readsf~< and writesf~ all fall into the that-would-be-nice category and I'm impatient to get features frozen and start putting out a test release.
Meanwhile, the differences between dac~ and adc~ set messages needs to get resolved because I don't want to freeze a stupid design - So I want ot get that worked out and stop there for now.

Spacechild1:
> i think [readsf~] and [writesf~] could profit from multichannel capabilities as well.

Definitely! How could I forget about these? However, there are a few open design questions, so I would rather make a separate PR for that, see https://github.com/pure-data/pure-data/issues/1990.

All objects in this PR, on the other hand, are trivial - at least I believe so :)

Anyway, both can wait for later.

> it seems [dac~] is fine.

@porres Nope, see https://github.com/pure-data/pure-data/issues/1984. Miller is working on it.

porres:
checking your [sig~] changes, I don't see it being significantly different than [snake~ in]

I was expecting something that convert lists to mc signals, like max's [list~], so [sig~] would keep its single inlet for either a float or list and change the number of channels accordingly.

I am trying to test a change to propose this instead but it doesn't really work, I have to force a new DSP check so the values and number of signals get updated. Is it just impossible?

Spacechild1:
> checking your [sig~] changes, I don't see it being significantly different than [snake~ in]

The main difference is that `[snake~ in]` takes the number of channels while `[sig~]` takes a list of scalar values. `[sig~]` is typically used with constant values, e.g.:
```
[sig~ 1 1.5 2]   [r~ foo]
 |                |
 |               /
[/~ ]
```
---

> I see MAX's [mc.list~] cannot change number of channels dinamically. I mean, you can, but you have to restart DSP so I guess that dealing with multichannel might not be so easy...

Yes, if the number of channels change, the DSP graph needs to updated. The same is true for the `[channes(` message for `[send~]` and `[catch~]`.

Generating a multi-channel signal from a (variable-length) list can be certainly useful, but I think this should be a `[snake~]` method.

porres:
> I think the [sig~] version is both shorter and clearer.

I see it and I agree, I also clearly don't oppose, but I can see the "that would be nice" aspect of it.

> Generating a multi-channel signal from a (variable-length) list can be certainly useful, but I think this should be a [snake~] method.

that would be awesome, but it sees it would require the DSP reboot, right? If so, I don't like that :(

And I think one could adventure into dynamic patching for recreating a [sig~] or [snake~ in] object with the correct number of channels according to the list length, that'd work in the worst case scenario, right? Too bad you gotta all that way for it... and all it does is fire a DSP graph update, right? So I ask, couldn't we have a an internal message to force the update without being dsp off/on?

Spacechild1:
> that would be awesome, but it sees it would require the DSP reboot, right? If so, I don't like that :(

Well, again, the `[channels(` message might also update the DSP graph. The point is that you *can* do it if required, but you probably shouldn't do while sound is playing. (In fact, there have always been messages that trigger DSP graph updates, e.g. `[set( -> [adc~]` or resizing a table that is used in DSP.)

> And I think one could adventure into dynamic patching for recreating a [sig~] or [snake~ in] object with the correct number of channels according to the list length, that'd work in the worst case scenario, right?

That would be 100x more complicated and still trigger a DSP graph update...

> So I ask, couldn't we have a an internal message to force the update without being dsp off/on?

I'm not sure what you mean. A DSP graph update is the same as turning DSP off and on... The fact is: if an object wants to change its number of output channels, the DSP graph has to be updated. There is absolutely no way around this.

porres:
> I'm not sure what you mean. 

I meant we could have a simpler internal message, something like "update-dsp", so we could us in patches (instead of "dsp 0, dsp 1", and something we could also call from withing the code of compiled objects, so we don't need to print an error and tell the user that turning dsp off and on is needed so the channels actually change. We can still advise people not to do it while playing audio.

On that note, should I say something like that in the help file of adc~_dac~?

Spacechild1:
> and something we could also call from withing the code of compiled objects, 

You mean `canvas_suspend_dsp`, `canvas_resume_dsp` and `canvas_update_dsp`?

> so we don't need to print an error and tell the user that turning dsp off and on is needed so the channels actually change. 

If you are talking about `[send~]`, this is a glitch that should be fixed with 84abd53594a978aea1fbc2ae58e24103a160e666 (develop branch). The `[channel(` message for `[catch~]` does not give any such error and does not require the user to manually turn DSP on and off.

(Miller explicitly decided to print an error for `[receive~]`'s "set" method if the channel count differs from the previous send, but this is technically not necessary and I am not really fond of it. Maybe this error message will be removed in the future.)

In general, objects just suspend/resume resp. update the DSP graph internally if required; there shouldn't ever be a *need* for the user to do that.

porres:
> In general, objects just suspend/resume DSP resp. update the DSP graph internally if required; there shouldn't ever be a need for the user to do that.

so I guess I can create a mc.list~ like object and have it update the dsp with 'canvas_update_dsp'  to change the number of channels, right?

porres:
> so I guess I can create a mc.list~ like object and have it update the dsp with 'canvas_update_dsp' to change the number of channels, right?

for th rectod, I tested it and it works... I think sig~ could be like that but another method to [snake~] would also be good

porres:
@millerpuckette what do you think about at least cherry picking cos~ mc support from https://github.com/pure-data/pure-data/pull/1978/commits/1f8d45cdf1186582e569edfb99f18e459b213580 ?

I will then get ready to update the documentation

Spacechild1:
Miller cherry-picked the `cos~` commit by hand, so I have removed it from this branch again.

PR #1979, opened 2023-06-10
send~/catch~: allow to dynamically set channel count

Spacechild1:
This PR allows to dynamically set the channel count of the `[send~]` and `[catch~]` object with a `[channels(` message.

~~Also, `[receive~]` now automatically takes the channel count from the corresponding `[send~]` object. If the user changes the send symbol (via the "set" message) *and* the channel count differs from the previous `[send~]` object, we need to update the DSP graph.~~ EDIT: implemented via 5a00d3655201c0464d95b35de2d86e0ba7a19ee4

### DIFF
--- a/src/d_global.c
+++ b/src/d_global.c
@@ -49,6 +49,13 @@ static t_int *sigsend_perform(t_int *w)
     return (w+4);
 }
 
+static void sigsend_channels(t_sigsend *x, t_float fnchans)
+{
+    x->x_nchans = fnchans >= 1 ? fnchans : 1;
+    x->x_length = 1; /* trigger update via sigsend_fixbuf */
+    canvas_update_dsp();
+}
+
 static void sigsend_fixbuf(t_sigsend *x, int length)
 {
     if (x->x_length != length)
@@ -87,6 +94,8 @@ static void sigsend_setup(void)
         A_DEFSYM, A_DEFFLOAT, 0);
     class_setdspflags(sigsend_class, CLASS_MULTICHANNEL);
     CLASS_MAINSIGNALIN(sigsend_class, t_sigsend, x_f);
+    class_addmethod(sigsend_class, (t_method)sigsend_channels,
+        gensym("channels"), A_FLOAT, 0);
     class_addmethod(sigsend_class, (t_method)sigsend_dsp,
         gensym("dsp"), A_CANT, 0);
     class_sethelpsymbol(sigsend_class, gensym("send-receive-tilde"));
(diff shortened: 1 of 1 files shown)


millerpuckette:
For the moment I decided to make all the channel counts static, that is, not settable by messages, just to prevent situations where you don't know whether to print a warning message out or not.  I think that in the great majority of cases it's good enough to be able to set all channel counts by creation arguments.  So I'd like to hold off adding this feature right now (which would then have to be part of Pd forever in favor of waiting to see if there's a real need for it.

millerpuckette:
P.S. my threshold for adding capabilities to Pd isn't just "this would be useful" but the stricter "this is needed in one or more ongoing projects" -- so lots of stuff that people think would be naturally good to have aren't in there yet.

Spacechild1:
> "this is needed in one or more ongoing projects"

All my feature PRs are motivated by my own personal projects, so you can be sure that there will always be at least 1 use case. (I am not saying that all my PRs are reasonable or should be merged on that basis alone :)

> just to prevent situations where you don't know whether to print a warning message out or not.

Then why doesn't `[throw~]` print an error message if the input channel count does not match the `[catch~]` output channel count?

> in favor of waiting to see if there's a real need for it.

Oh, there is a real need for it! How else would I accomodate changing channel counts in upstream objects, e.g. via `[resize( -> [clone -mo]`? A more concrete example: I am thinking about adding multichannel support for the next version of `[vstplugin~]`, allowing users to set output channel count dynamically. For example, you could then change the ambisonics order in the IEM `StereoEncoder` and all downstream objects would adapt. Or you could dynamically change the output channel count of `AllRADecoder` based on the number of speakers and pipe it right into `[dac~]` as a multichanneld signal. This would be pretty powerful.

Spacechild1:
Actually, now I think that `[receive~]` may just take the channel count from of its matching `[send~]`, hence making the error message completely obsolete.

I know that the static channel count was motivated by the fact that you can dynamically change the corresponding send symbol, but the solution is simple:  if the channel count has changed between send objects, just update the DSP graph! This way you can still switch dynamically between several sends without audio dropouts, as long as they have the same channel count; at the same time, you can dynamically change channel counts if required.

IMO, this would make the behavior of the two object families send~/receive~ and throw~/catch~ more symmetric and practical.

Will push a commit soon as a POC.

Spacechild1:
@millerpuckette Check out the last commit! I think that's a big improvement. Personally, I would find it really awkward if we had to explicitly provide a channel count for `[receive~]`... (To be clear: this is orthogonal to the "channels" message for `[send~]` and `[catch~]`)

porres:
> I think that in the great majority of cases it's good enough to be able to set all channel counts by creation arguments.

as someone who designs lots of externals as abstractions, being able to set them is crucial, otherwise we have to abuse and suffer with dynamic patching

Spacechild1:
To give another concrete example: I want to change the ambisonics order of my patch. Currently, I would need to track down every `[send~]`, `[receive~]` and `[catch~]` object involved and manually change the creation arguments. With this PR, everything can be done with a single message. This is a real practical issue!

millerpuckette:
I ended up throwing out the receive~ channel count argument - it can always get added later but should't be taken out once Pd 54 is finalized.

Spacechild1:
Merged to master (bbe61ea24a5685543fa5ea154f4a380d8864b5a3)

Issue #1994, opened 2023-06-16
request, allow setting [clone]'s "d" flags as messages


porres:
Hi, this would be really useful for abstractions that use [clone] where I could give it arguments to have it output either a multichannel connection or a summed bus.

I was thinking of other strategies, like having a new [snake~ mix] method for summing the output later, but I still wouldn't know how to route this as there are other challenges involved. Like, how would one be able to send to the outlet~ either one or the other?

porres:
@Spacechild1 what do you think of this? I made a simple change and was able to set the 'packout' variable (the 'do' flag) but I need to restart the dsp chain so it actually works and I'm not sure what's the best way of doing this. I tried canvas suspend and restart (something I saw on the code) but it didn't work.

porres:
cool, I tried and it works and I know now I was doing something stupid and the other method should have worked too, but I guess this way is better anyway, but what do you think about this? Shouldn't be of any concern to allow this right?

I just still don't like the terms we use, for now I have a 'do <float>' message.

porres:
Ok, a PR is up, changes to the documentation should come if accepted, I'd love some input from you @Spacechild1 thanks

PR #2004, opened 2023-06-19
allow 'd' flags as messages

porres:
closes https://github.com/pure-data/pure-data/issues/1994

### DIFF
--- a/src/g_clone.c
+++ b/src/g_clone.c
@@ -148,6 +148,37 @@ static void clone_in_fwd(t_in *x, t_symbol *s, int argc, t_atom *argv)
 
 static void clone_setn(t_clone *, t_floatarg);
 
+static void clone_in_do(t_in *x, t_floatarg f)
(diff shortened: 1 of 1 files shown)


Spacechild1:
Looks good to me. Personally, I don't really see a use case, but that doesn't say much :-)

As you said yourself, the method names are not exactly pretty. Maybe we could use the full names, but then again I don't even know what "-d" and "-do" actually stand for...

porres:
I have a special use case at least for 'do'.

You see, I am designing this multi channel oscillator bank abstraction. It takes a multichannel signal for setting the frequencies of the oscillators and I am using clone for each oscillator in the bank.

Now, I want the user to be able to set if is desired that the output is a single channel or a multichannel, but for now I can only hardcode this in the [clone]'s creation arguments.

I hope @millerpuckette sees this as something that is actually important to have

PR #2031, opened 2023-06-27
Update translations

umlaeute:
Minor translation updates for the 4 or so new strings as found in #2015.

languages updated so far:
- English (Canada,...)
- French
- German
- Spanish


umlaeute:
note that there are some pending translations (e.g. Portuguese) that have not yet been added to this PR.
they should show up in the next 12 hours or so, but if you want to merge earlier, I can try to force an update (the webinterface for the translations tries to accumulate translations before committing the changes)

PR #2032, opened 2023-06-28
nsis: handle "Pd64" installation

Lucarda:
This PR in intended to be merged into `develop` branch.

if `pd64.exe` is found give the installer different default installation path, app name, and start menu entry.

### DIFF
--- a/msw/build-nsi.sh
+++ b/msw/build-nsi.sh
@@ -250,7 +250,8 @@ cp "${SCRIPTDIR}/../tcl/pd.ico" "${WORKDIR}/pd.ico"
 
 # run the build
 mkdir -p "${OUTDIR}"
-if makensis -DPDVER="${PDVERSION}" -DWISHN="${WISHNAME}" -DARCHI="${PDARCH}" "${NSIFILE}" 1>&2
+if makensis -DPDVER="${PDVERSION}" -DWISHN="${WISHNAME}" -DARCHI="${PDARCH}" \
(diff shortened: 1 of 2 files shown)


umlaeute:
i guess, ideally we would take the `${PDEXE}`, strip the extension and path components, and uppercase the first letter.

but looking at the NSIS documentation, this seems to involve a lot of busywork...
so assuming that we won't see anything except the pd.exe and pd64.exe for a long time, the current implementation is probably good enough :-)

Lucarda:
> but looking at the NSIS documentation, this seems to involve a lot of busywork...

we can do it in `.sh` script and pass, may be, one more `-Dfoo` but it's roughly the same.

one thing i'm unsure:

what we like most to see in the **startup menu**:

- Pure Data
- Pd64

or 

- Pd
- Pd64

umlaeute:
i honestly don't know.
historically, we have had `Pure Data` (but i don't remember for how long), and i like this as it is a bit more prominent than just `Pd` (though I do usually appreciate understatements).

Lucarda:
Ok, lets keep it *as is* for now. If we change it or *when* we change it the installer will no longer warn you to uninstall the previous *app* name

Lucarda:
closing this as completed as Miller has merged it via 640007d

umlaeute:
@Lucarda now that the branch has been merged, you could delete it from your repository :-)
https://github.com/lucarda/pure-data/branches

Lucarda:
done

PR #2049, opened 2023-07-07
multichannel support for delay objects

Spacechild1:
`[delwrite~]` and `[delread~]` are conceptually very similar to `[send~]` and `[receive~]` resp. `[tabsend~]` and `[tabreceive~]`, so it would make sense if they also had multichannel support.

---

`[delwrite~]` now has an optionional third creation argument that sets the number of channels. This can also be changed dynamically with the `[channels(` message (requiring a DSP graph update), similar to `[send~]`.

`[delread~]` and `[delread4~]` automatically take the channel count from the corresponding `[delwrite~]` object, similar to `[receive~]`.

### DIFF
--- a/src/d_delay.c
+++ b/src/d_delay.c
@@ -14,6 +14,7 @@ static t_class *sigdelwrite_class;
 typedef struct delwritectl
 {
     int c_n;
+    int c_nchans;
     t_sample *c_vec;
     int c_phase;
 } t_delwritectl;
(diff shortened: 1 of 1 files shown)


tmm88:
nice job

Spacechild1:
@millerpuckette Do you see an issue with this?

The reason for the "channels" message for `[delwrite~]` is exactly the same as with `[send~]`: the readers might be reordered before the writer in the DSP graph, so we can't simply take the channel count from the input signal, yet we want to be able to change it at runtime.

Spacechild1:
Hooray!

PR #2051, opened 2023-07-07
tabread4~: make onset inlet signal rate

Spacechild1:
This PR makes the onset inlet for `tabread4~` signal rate. For example, this would be particularly useful for sample-accurate granular synthesis, where you could use `[vline~]` for the onset. (Currently, I have to use the [undocumented!] `tabread4~~` object in `zexy` for my granular synthesis abstractions, which I find a bit silly.)

I think there is no real reason to restrict the onset inlet to control rate - except for efficiency. However, the overhead of the additional (cached!) memory access should be tiny compared to the actual interpolation math.

If efficiency is indeed a concern, I could add a scalar version (using `CLASS_NOPROMOTESIG`)

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -471,24 +471,23 @@ typedef struct _tabread4_tilde
     t_object x_obj;
     t_arrayvec x_v;
     t_float x_f;
-    t_float x_onset;
 } t_tabread4_tilde;
 
 static void *tabread4_tilde_new(t_symbol *s, int argc, t_atom *argv)
 {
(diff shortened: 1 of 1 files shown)


umlaeute:
or even better: use Pd64 instead.

Spacechild1:
> or even better: use Pd64 instead.

Since sample accurate table onsets would be my only use case for Pd64, I'd rather keep using `[zexy/tabread~~]` and save my precious SIMD registers.

porres:
can I see your granular synth patch? :)

PR #2052, opened 2023-07-07
multichannel support for readsf~ and writesf~

### LINKED ISSUE #1990: multichannel readsf~ and writesf~
`[readsf~]` and `[writesf~]` should be multi-channel aware. However, there are few open questions that need to be sorted out.

---

`[readsf~]`

Different soundfiles can have different number of channels, so the object might to need to change the number of output channels - which requires a DSP graph update. I see two general options:

1. `[open(` causes a DSP graph update on demand, i.e. if the channel count between soundfiles has changed
2. have a dedicated `[channels(` method - similar to `[send~]` and `[catch~]` -, so that the number of channels can be set explicitly and independently from the `[open(` method.

Either way, `[readsf~]` would require a `-m` flag, but in the case of 2. it would also take an optional initial channel count.

---

`[writesf~]`

Some things to watch out for:

1. What happens if we send multichannel signals to `[writesf~ 2]`? Do we append the second signal - or ignore it (and warn)?
2. Without creation arguments, the object assumes it will write a soundfile with 1 channel. Only in the "dsp" method, it will detect a multichannel signal and adapt accordingly. So what should happen when I send the `[open(` method *before* DSP is turned on? Maybe we might need an `-m` flag, so we are able to detect and warn about this case?
3. The channel count might change while writing is in progress, e.g. if an upstream object changes its channel count. I think that the object would just continue with the channel count at the time of the `[open(` message (`t_soundfile.sf_nchannels`). For that we would probably need to change the `soundfile_xferout_sample` method to also include the *actual* channel count.

---

Did I miss somthing?

Spacechild1:
`[readsf~]` now takes an optional `-m` flag before the `channels` argument which creates a multichannel signal outlet instead of individual signal outlets. The number of channels can also be set dynamically with the `[channels(` message, similar to other multichannel objects.

---

`[writesf~]` simply accepts multichannel input signals, similar to `[dac~]`, and the `[open(` takes the current number of channels. If the input channel count changes while the object is streaming, it will print a warning to the console.

There is one caveat: if the input signal is multichannel, you must send the `[open(` message *after* DSP is turned on, otherwise it would create a singlechannel file and you get a warning (as described above).

---

Closes #1990

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -433,7 +433,7 @@ int open_soundfile_via_canvas(t_canvas *canvas, const char *filename,
 static void soundfile_xferin_sample(const t_soundfile *sf, int nvecs,
     t_sample **vecs, size_t framesread, unsigned char *buf, size_t nframes)
 {
-    int nchannels = (sf->sf_nchannels < nvecs ? sf->sf_nchannels : nvecs), i;
+    int i, nchannels = (sf->sf_nchannels < nvecs ? sf->sf_nchannels : nvecs);
     size_t j;
     unsigned char *sp, *sp2;
     t_sample *fp;
@@ -786,16 +786,15 @@ static void soundfile_finishwrite(void *obj, const char *filename,
     object_sferror(obj, "soundfiler write", filename, errno, sf);
 }
 
-static void soundfile_xferout_sample(const t_soundfile *sf,
+static void soundfile_xferout_sample(const t_soundfile *sf, int nvecs,
     t_sample **vecs, unsigned char *buf, size_t nframes, size_t onsetframes,
     t_sample normalfactor)
 {
-    int i;
+    int i, nchannels = (sf->sf_nchannels < nvecs ? sf->sf_nchannels : nvecs);
     size_t j;
     unsigned char *sp, *sp2;
     t_sample *fp;
-    for (i = 0, sp = buf; i < sf->sf_nchannels; i++,
-        sp += sf->sf_bytespersample)
+    for (i = 0, sp = buf; i < nchannels; i++, sp += sf->sf_bytespersample)
     {
         if (sf->sf_bytespersample == 2)
         {
@@ -892,18 +891,21 @@ static void soundfile_xferout_sample(const t_soundfile *sf,
             }
         }
     }
+        /* zero out other channels */
+    for (i = nchannels; i < sf->sf_nchannels; i++, sp += sf->sf_bytespersample)
+        for (j = 0, sp2 = sp; j < nframes; j++, sp2 += sf->sf_bytesperframe)
+            memset(sp2, 0, sf->sf_bytespersample);
 }
 
-static void soundfile_xferout_words(const t_soundfile *sf, t_word **vecs,
-    unsigned char *buf, size_t nframes, size_t onsetframes,
+static void soundfile_xferout_words(const t_soundfile *sf, int nvecs,
+    t_word **vecs, unsigned char *buf, size_t nframes, size_t onsetframes,
     t_sample normalfactor)
 {
-    int i;
+    int i, nchannels = (sf->sf_nchannels < nvecs ? sf->sf_nchannels : nvecs);
     size_t j;
     unsigned char *sp, *sp2;
     t_word *wp;
-    for (i = 0, sp = buf; i < sf->sf_nchannels;
-         i++, sp += sf->sf_bytespersample)
+    for (i = 0, sp = buf; i < nchannels; i++, sp += sf->sf_bytespersample)
     {
         if (sf->sf_bytespersample == 2)
         {
@@ -1000,6 +1002,10 @@ static void soundfile_xferout_words(const t_soundfile *sf, t_word **vecs,
             }
         }
     }
+        /* zero out other channels */
+    for (i = nchannels; i < sf->sf_nchannels; i++, sp += sf->sf_bytespersample)
+        for (j = 0, sp2 = sp; j < nframes; j++, sp2 += sf->sf_bytesperframe)
+            memset(sp2, 0, sf->sf_bytespersample);
 }
 
 /* ----- soundfiler - reads and writes soundfiles to/from "garrays" ----- */
@@ -1520,7 +1526,7 @@ size_t soundfiler_dowrite(void *obj, t_canvas *canvas,
         ssize_t byteswritten;
         thiswrite = (thiswrite > bufframes ? bufframes : thiswrite);
         datasize = sf->sf_bytesperframe * thiswrite;
-        soundfile_xferout_words(sf, vectors, (unsigned char *)sampbuf,
+        soundfile_xferout_words(sf, argc, vectors, (unsigned char *)sampbuf,
             thiswrite, wa.wa_onsetframes, normfactor);
         byteswritten = write(sf->sf_fd, sampbuf, datasize);
         if (byteswritten < 0 || (size_t)byteswritten < datasize)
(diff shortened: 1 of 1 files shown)


porres:
I think [writesf~] could be like [adc~] with the '-m' flag, you tell the object that it'll take a multichannel and also set number of channels to write. If the input has more channels, extra are ignored, if it has less, extra ones are silenced. You could have messages to set different channels as in [adc~]

Spacechild1:
> I think [writesf~] could be like [adc~] with the '-m' flag, you tell the object that it'll take a multichannel and also set number of channels to write. If the input has more channels, extra are ignored, if it has less, extra ones are silenced. You could have messages to set different channels as in [adc~]

I am bit conflicted about it. It really isn't necessary because we can deduce the number of channels from the input signal. The only caveat (that I've already mentioned in the PR description) is that for multichannel inputs the `[open(` message must be sent before DSP is turned on resp. after the DSP graph has changed. I actually post a warning if the channel count has changed.

Now, the `-m` flag + the `[channels(` message would be more explicit, but generally we want to avoid setting channels explicitly unless it is really necessary.

Here's another solution: `[open(` could take the channel count as *optional* second argument; this would allow to explicitly set the number of channels, instead of deducing it from the input channels. This way we can have our cake and eat it too.

charlesneimog:
Hi @Spacechild1 and @umlaeute, I wanted to check in and see if there are any updates on this PR. It would be incredibly helpful for switching ambisonics setups and rendering other configurations. Thanks!

porres:
I say it's just up to @millerpuckette  to take this, hopefully for 0.56-0

Spacechild1:
I agree that it would be incredible useful! Hopefully we will see this - and other - multichannel improvements in Pd 0.56.

porres:
@Spacechild1  I'm testing writesf~ and I getting an extra 3rd channel when trying to record 2 with a multichannel inpput

porres:
As for the design, I'd rather we had a -m flag like the others and the possibility to change the number of input channels dinamically for [writesf~]

Spacechild1:
>  I'm testing writesf~ and I getting an extra 3rd channel when trying to record 2 with a multichannel inpput

Can't reproduce this. Just tested with the following patch:

grafik

This creates a stereo file, as expected. Can you share your patch?

> As for the design, I'd rather we had a -m flag like the others and the possibility to change the number of input channels dinamically for [writesf~]

As I've already explained above (https://github.com/pure-data/pure-data/pull/2052#issuecomment-1984430457), we generally only use the `-m` flag if 
a. the object cannot deduce the channel count, e.g. `[adc~]` or
b. the object cannot know whether it should operate in multi-channel mode `[clone]`.
Neither is the case here.

Regarding dynamically changing the channel count: the channel count is simply taken from the input, so I'm not sure what you're asking here. If you want to only record a subset of channels, you would simply do that by splitting the multichannel signal. (Of course, we still need a built-in `[snake~]` method for that.)

In the linked comment I also floated the idea of adding a second (optional) argument to `[open(` that would force the channel count. If you think that would be a good idea, please open a feature request.

Anyway, the current implementation seems fine to me as it is.

Spacechild1:
@porres I just had a look at your documentation updates (thanks for that!) and now I see the issue:

grafik

You've accidentally created a second inlet for `[writesf~]`, that's why you are getting three channels in total :)

Note the typo in "spreass". I also spotted a typo in readsf~ ("dinamically")

Side note: I'm not sure I would understand "spreads the channels through the inlets". What about something like: "If the input is a multi-channel signal, its channels are written consecutively, just like a series of single-channel inputs." Just an idea.

porres:
I thought we had to define the number of channels :) 

well, what happens then if we set the number of channels? And can different buses of input just merge many multichannel or mono connections? That's quite cool actually :)

Spacechild1:
> I thought we had to define the number of channels :)

I see! That's a misunderstanding. The channel count is simply deduced from the input signal. Now I also see why you've requested a "channels" message.

> well, what happens then if we set the number of channels?

It actually sets the number of inlets, which in turn can either hold single-channel or multi-channel signals.

> And can different buses of input just merge many multichannel or mono connections? That's quite cool actually :)

Yes, that's what's happening. It's very similar to `[dac~]` btw.

porres:
> Yes, that's what's happening. It's very similar to [dac~] btw.

Cool, i have to fix the docs then

porres:
@Spacechild1 max number of channels is still 64 right?

what if I have 64 inputs with 64 multichannel inputs each? :)

porres:
anyway, I suspect it stays with the same limit of 64 channels, but I also wonder if we can enjoy the opportunity and increase this arbitrary limit.

I see the maximum number of channels in a wave file is actually 65535 :) 

C'mon, let's go crazy!

Spacechild1:
>  but I also wonder if we can enjoy the opportunity and increase this arbitrary limit.

Good idea! Will give it a try. There is certainly no reason for keeping the rather arbitrary 64 channel limit.

Spacechild1:
I don't think there is a limit beside the ones imposed by the file format itself. As you correctly noted, WAV files can contain up to 65535 channels.

porres:
and so it seems aiff and caf have no real limit!

umlaeute:
no idea why you think there's no limit. 

afair, AIFF uses a 16bit field for storing the number of channels, so it's limited to 65536 channels, just like WAV.

CAF uses a 32bit field, so the number of channels is much higher, but still bound to an upper limit.

umlaeute:
> because we live on a golden age of internet where knowledge is easily accessible

but then it says:
> Visão geral gerada por IA

so rather than relying on what some random AI hallucinates, it would be wise to do some *actual* research in the wide wild internet.
then you would eventually stumble upon something like http://midi.teragonaudio.com/tech/aiff.htm, which clearly states:

```

typedef struct {
  ID             chunkID;
  long           chunkSize;

  short          numChannels;
  unsigned long  numSampleFrames;
  short          sampleSize;
  extended       sampleRate;
} CommonChunk;
```

my original source of truth (because despite saying "afair", i was really doing some research first) was [`libsndfile`](https://github.com/libsndfile/libsndfile/blob/52b803f57a1f4d23471f5c5f77e1a21e0721ea0e/src/aiff.c#L125-L127): https://github.com/libsndfile/libsndfile/blob/52b803f57a1f4d23471f5c5f77e1a21e0721ea0e/src/aiff.c#L125-L127

porres:
yeah, I was being ironic, I know we can't trust those, hence the "so it seems" - and at least it got it right about .wav 

I tried a bit harder and opened new links, but as you might be aware, google is pretty useless these days and some 2/3 other links were also saying that "in theory", this was unlimited. That felt weird and I wanted to get more to the bottom, but it was pretty late at night and I gave up...

I also tried researching about the next format, with no much luck either, but I'm assuming it is a pretty old format that won't likely support a high limit, we also need to check that.

As for the real limit, if wav and aiff/caf can handle up to 65536 channels, I say we get wild and let that happen :) and just have a check to clip and not let it exceed this llimit in the code. 

I suspect though that this would require another check for 'next' formats, though I believe no one is using Pd to record these format these days :)

umlaeute:
we don't actually have to look very far:
https://github.com/pure-data/pure-data/blob/691bbf8bf0e60807ae46b415d4578460a44f78d3/src/d_soundfile_next.c#L50-L57

porres:
I'd have to ask AI what this actually means, and it'd probably spit out garbage :) are you saying it could take 4294967295 channels?

umlaeute:
yes, most like.y

porres:
65535 seems like a reasonable maximum limit then :) 

I know few people with more speakers than that

porres:
Oh, clearly, that would also need to be applied for soundfiler and readasf~ 

I'll open a new issue

porres:
actually, as I remembered, we already have https://github.com/pure-data/pure-data/issues/2337

umlaeute:
>I know few people with more speakers than that
the implicit assumption, that a soundfile channel maps directly to a speaker, has caused me endless pain in the past. 

please do not make this a part of your argument.

porres:
My idea and argument is that it could be as high as possible, and I understand that the "wall" is 65535! So... be it ;) and I joked about speakers, but behind the joke is an understanding that is is also probably good enough for whatever we wanna do.

PR #2054, opened 2023-07-07
Add signal comparison/logical operators

Spacechild1:
Maybe the new multichannel feature is a good excuse to finally(!) add the missing signal comparison/logical operators:

`[==~]`, [`!=~]`, `[>~]`, `[<~]`, `[>=~]`, `[<=~]`, `[&&~]`, `[||~]`

IMO, it is a bit silly that we currently need externals (`zexy`, `cyclone`) or `[expr~]` for these basic math operations, even more so as the former are (still) single channel.

---

As a first step, I have added more convenience functions to `d_arithmetic.c` which remove a lot of the code duplication. In total, this PR only adds 350 lines (700 added, 350 removed).

Side note: it would be possible to further reduce the code size - drastically! - by implementing the perform routines as macros, taking the binary operator as an argument, but this might be a bit controversial, so I haven't done it.

### DIFF
--- a/src/d_arithmetic.c
+++ b/src/d_arithmetic.c
@@ -101,30 +101,26 @@ static void any_binop_scalar_dsp(t_signal **sp, t_sample *g,
         4, sp[0]->s_vec, g, sp[1]->s_vec, bign);
 }
 
-
-/* ----------------------------- plus ----------------------------- */
-static t_class *plus_class, *scalarplus_class;
-
-typedef struct _plus
+typedef struct _binop
 {
     t_object x_obj;
     t_float x_f;
-} t_plus;
+} t_binop;
 
-typedef struct _scalarplus
+typedef struct _scalarbinop
 {
     t_object x_obj;
     t_float x_f;
-    t_float x_g;            /* inlet value */
-} t_scalarplus;
+    t_float x_g;
+} t_scalarbinop;
 
-static void *plus_new(t_symbol *s, int argc, t_atom *argv)
+static void *any_binop_new(t_symbol *s, int argc, t_atom *argv,
+    t_class *binop_class, t_class *scalarbinop_class)
 {
-    if (argc > 1)
-        post("+~: extra arguments ignored");
-    if (argc)   /* argument implies we'll do a scalar add as in "+~ 1" */
+    if (argc > 1) post("%s: extra arguments ignored", s->s_name);
+    if (argc)
     {
-        t_scalarplus *x = (t_scalarplus *)pd_new(scalarplus_class);
+        t_scalarbinop *x = (t_scalarbinop *)pd_new(scalarbinop_class);
         floatinlet_new(&x->x_obj, &x->x_g);
         x->x_g = atom_getfloatarg(0, argc, argv);
         outlet_new(&x->x_obj, &s_signal);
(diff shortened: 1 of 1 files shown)


porres:
If this is taken, I think it would make sense to have this objects call a new help file named **binops-other-tilde-help.pd**.

For the record, I see most of the objects from **binops-other-help.pd** are here, all but [&], [|], [<<], [div] and [mod~] and yeah, I don't think they'd be worth including.

millerpuckette:
well, I'm mostly OK with this, but not so much && and ||, which are integer operations and signals are floating-point.  I think that's dangerous mission creep (you'd want "%" for example - I have no idea where it would end).

The comparison operators, although I think they're not really necessary, sem to be so popular a request that I'd be OK with adding them.

Spacechild1:
> well, I'm mostly OK with this, but not so much && and ||, which are integer operations and signals are floating-point.

Pd only has floats and no true integers. By this logic, we should have no logical operators at all :)

On a more serious note: people do use (pseudo)binary signals to do audio-rate logic. For example, consider the following patch:

grafik


Now, you could also implement the OR like this:

grafik


However, this is not exactly obvious. IMO `[||~]` would be the natural way to implement this and the intent is perfectly clear to the reader.

> (you'd want "%" for example 

We already have `[wrap~]` :) Also, % is not a logical operator. This PR is really about audio-rate logic. 

> I have no idea where it would end

This PR ports all the existing control-rate comparison and logical operators to the audio domain, so I think it ends right here.

millerpuckette:
> > well, I'm mostly OK with this, but not so much && and ||, which are integer operations and signals are floating-point.
> 
> Pd only has floats and no true integers. By this logic, we should have no logical operators at all :)
> 

re-reading this, I was mistaking || for | (bitwise operator) - I think you're right that "||" (logical operator) is more logical to include.  I would classify "|", "&", and "%" as very specifically integer (as opposed to logical) operators.

And... although there's a "%" I don't ting there's anything odd about using floats as integers in messages where there are also non-signal constructs like "trigger" or "until".  But in DSP land I think things should be about quantities.  What we're talking about now is using audio signals to carry two different types of thing, numbers and logical "true"/"false" - this is mission creep (which is one of my main fears for Pd's longevity) but in this case there's such a demand for it that it's hard to deny.

Spacechild1:
>  I would classify "|", "&", and "%" as very specifically integer (as opposed to logical) operators.

Yes, I intentionally did not include the bitwise operators. *)

> What we're talking about now is using audio signals to carry two different types of thing, numbers and logical "true"/"false"

I would understand your philosophical concerns if Pd had an actual boolean message type. However, Pd uses floats to represent boolean values (zero = false, non-zero = true) and for all practical purposes it works the same in the control domain and signal domain.

> this is mission creep (which is one of my main fears for Pd's longevity)

I don't think that a few math operators are a danger to Pd's longevity :)

Audio-rate conditional logic is a really common thing and people have been using these operators for decades.  Including them as built-in objects would only recognize existing practice.

---

*) This doesn't mean that audio-rate bitwise operators wouldn't be useful. There's a whole synthesis technique called "bytebeat" that is based on these operators: https://dollchan.net/bytebeat/#4AAAA+kUzNNDSKLGzM68pqQFSZpraJkC+GpBpaAwRAAA. I just considered it out-of-scope for this PR.

porres:
Another selling point is that people already use those in Pd but have to rely on expr~ which is less efficient

Spacechild1:
@millerpuckette Do you still have concerns, or can we merge this? :)

millerpuckette:
Well, I still think it's just bloat.  But I'll bend to the will of the majority on this one, primarily to acknowledge all your many other contrbutions that I _do_ agree with :)

porres:
yay :)

ben-wes:
> it's just bloat

beautiful bloat though. :)

porres:
in https://github.com/pure-data/pure-data/pull/2987 I have created a separate "other-binops-tilde" object, which makes it all more consistent with the current doc structure, anyway, all is properly documented now

PR #2055, opened 2023-07-07
add "channels" method to all table DSP objects

### LINKED ISSUE #2025: change number of channels in table DSP objects
Currently, the number of channels in all the table DSP objects fixed - determined by the creation arguments - and the "set" message can only override the provided tables. There is no way to actually change the number of channels!

`tabsend~` and `tabreceive~` are conceptually similar to `send~` and `receive~` - which have the "channels" method. `tabwrite~` and `tabplay~` are basically in-memory versions of `readsf~` and `writesf~` - which will likely get multi-channel support in the future.

I think all the table DSP objects should support dynamic channel counts, for all the reasons mentioned in https://github.com/pure-data/pure-data/pull/1979.

Here is a simple suggestion: *allow "set" to change the number of channels*. For example, "set foo" -> 1 channel, "set foo bar" -> 2 channels, "set foo bar baz" -> 3 channels, etc. If @millerpuckette agrees in principle, I can make a PR.

Or does anyone have a better idea?

NB: if we change the behavior of the "set" message, we would need to do this *before* releasing Pd 0.54!

---

Side note: IIRC, Miller thinks about adding support for data structure arrays in all DSP table objects. This would enable the usage of multi-dimensional arrays for multi-channel operations. However, in that case there would be no point in restricting the object to a fixed number of channels, and neither should we do for "regular" graphical arrays.

Spacechild1:
For `tabwrite~` and `tabsend~` this will change the number of arrays to write to.

For `tabplay~`, `tabread~`, `tabread4~` and `tabreceive~` this will change the number of arrays to read from and also the number of output channels.

Naturally, the "channels" message causes a DSP graph update, similar to `[send~]` and `[catch~]`.

Closed #2025

---

In practice, the "channels" message would be followed by a "set" message, e.g. `[channels 4, set array1 array2 array3 array4( -> [tabplay~]`.

Alternatively, we could have a message that resizes and sets the arrays in a single operation. I'm not sure. I kind of like the "channels" message because it is already used by `[send~]` and `[catch~]` and in my other PRs (#2049, #2052)

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -93,6 +93,34 @@ static void arrayvec_set(t_arrayvec *v, int argc, t_atom *argv)
         arrayvec_testvec(v);
 }
 
+static void arrayvec_resize(t_arrayvec *v, int newsize)
+{
+    int i, oldsize = v->v_n;
+    void *owner = v->v_vec[0].d_owner;
+    if (newsize < 1)
+        newsize = 1;
+    if (newsize == oldsize)
+        return;
+        /* free old elements */
+    for (i = newsize; i < oldsize; i++)
+        gpointer_unset(&v->v_vec[i].d_gp);
+    v->v_vec = (t_dsparray *)resizebytes(v->v_vec,
+        oldsize * sizeof(*v->v_vec), newsize * sizeof(*v->v_vec));
+    v->v_n = newsize;
+        /* init new elements */
+    for (i = oldsize; i < newsize; i++)
+    {
+        v->v_vec[i].d_symbol = &s_;
+        v->v_vec[i].d_phase = 0x7fffffff;
+        v->v_vec[i].d_owner = owner;
+        gpointer_init(&v->v_vec[i].d_gp);
+    }
+        /* always requires a DSP update because t_dsparray pointers are
+        passed directly to the perform routine. tabplay~, tabread~, etc.
+        also need to update their output signals! */
+    canvas_update_dsp();
+}
+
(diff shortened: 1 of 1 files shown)


umlaeute:
not having looked at the  code, but can we also reverse the `channels` and `set` messages? `[set foo bar pizza benzo, channels 4(`

umlaeute:
>>  not having looked at the code, but can we also reverse the channels and set messages? `[set foo bar pizza benzo, channels 4(`

> You could, but it wouldn't do what you want...

my implicit suggestion was to keep the list of array names, so you can later decide whether you want to use all four arrays or just 2 of them.


> Maybe "channels" should rather take the array names instead of the channel count? 

dunno. i quite like the idea that `channels` has always the same arguments, regardless of object.
(hey, we could have `[send~]` send the various channels of a multichannel signals to different mono `[r~]` instances... :shudder:)

Spacechild1:
> my implicit suggestion was to keep the list of array names, so you can later decide whether you want to use all four arrays or just 2 of them.

Yes, but that would complicate the current implementation: the object maintains a vector of "DSP arrays" and for each channel, the "dsp" routine directly passes a pointer to the corresponding "DSP array" element. If "set" would resize the "DSP array" vector, we might need to update the DSP graph to fixup the pointers, which is what Miller wanted to avoid in the first place.

We *could* avoid passing the "DSP array" directly and instead pass a pointer to the `t_arrayvec` plus the channel index.  `t_arrayvec` would also need to maintain two counts: the allocated number of channels and the number of channels in use. It's certainly possible. I certainly see the appeal of not having to always pass the whole list of array names just for changing the channel count.

Issue #1997, opened 2023-06-17
change mc connection GUI representation


porres:
seems nice to add a new and different representation for mc connections, maybe just making it thicker is enough

umlaeute:
while I was going to suggest the same a few days ago, I am not so sure any more.

should the cord change it's appearance whenever you start the DSP?
and also change it according to some `[channel(` message that somebody sent to some object?

Spacechild1:
> seems nice to add a new and different representation for mc connections, maybe just making it thicker is enough

It would certainly be nice if we could distinguish MC connections visually.

> whenever you start the DSP?
> and also change it according to some [channel( message that somebody sent to some object?

The latter implies the former.

Currently, the core cannot know the *cause* of a DSP graph update or which parts of the graph have changed, so practically we would be forced to redraw all signal connections for all (visible) canvases. This would cause some additional CPU overhead (on the core *and* GUI process) for *every* DSP graph update and we would need to measure and evaluate whether its acceptable.

charlesneimog:
Hi, I was thinking the same, here one simple suggestion...

![multi](https://github.com/pure-data/pure-data/assets/31707161/c0ba9dad-6d97-4b3a-9c61-442310e27595)

In this image, noaudio connections = `1 * x->gl_zoom`, audio connection `2 * x->gl_zoom`, and multichannel `3 * x->gl_zoom`.

It seems that it consists in change the function `canvas_doclick` in `g_editor.c` and `canvas_drawlines` in `g_canvas.c`. Maybe more that I did not see.

I am not one `pd` expert, but I do not know how to get the info about the amount of channels (or if it is multichannel or not) of one object in these functions...

Spacechild1:
> (or if it is multichannel or not

It is easy to query if an object is multi-channel-*capable* by checking for the `CLASS_MULTICHANNEL` flag. However, I guess we only want to show thick signal connections if they actually carry more than 1 channel. There is currently no way to know the number of channels (outside the DSP graph building routine), but we could store that information in `t_outlet` and add a function `obj_getsignaloutletchannels` that we can use for drawing the signal connections.

charlesneimog:
I am yet look in the code of pd, and for fun I implemented using `CLASS_MULTICHANNEL`. It is very small changes in 3 functions, I forgot the `tryconnect` in `g_editor.c`. 

```
noaudio 1
audio 2 
multichannel  2.5 
```
[...]

PR #2057, opened 2023-07-10
Add functions to draw multichannel connections, see #1997

charlesneimog:
As recommended by @Spacechild1, here one first draft. 

There is a need to redraw the connections after the DSP is `on`.

In the video...

https://github.com/pure-data/pure-data/assets/31707161/bf72893b-b272-4567-8d4e-4da2ed60fa4f

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -11,7 +11,7 @@ extern "C" {
 #define PD_MAJOR_VERSION 0
 #define PD_MINOR_VERSION 54
 #define PD_BUGFIX_VERSION 0
-#define PD_TEST_VERSION "test2"
+#define PD_TEST_VERSION ""
 extern int pd_compatibilitylevel;   /* e.g., 43 for pd 0.43 compatibility */
 
 /* old name for "MSW" flag -- we have to take it for the sake of many old
--- a/src/pd.rc
+++ b/src/pd.rc
@@ -9,12 +9,12 @@ BEGIN
     BEGIN
       VALUE "CompanyName", "puredata.info"
       VALUE "FileDescription", "Pure Data Application"
-      VALUE "FileVersion", "0.54-0test2"
+      VALUE "FileVersion", "0.54-0"
       VALUE "InternalName", "pd.exe"
       VALUE "LegalCopyright", "Miller Puckette, et al."
       VALUE "OriginalFilename", "pd.exe"
       VALUE "ProductName", "Pure Data"
-      VALUE "ProductVersion", "0.54-0test2"
+      VALUE "ProductVersion", "0.54-02"
     END
   END
 


src/m_obj.c
```diff
@@ -892,6 +893,29 @@ int obj_nsigoutlets(const t_object *x)
...
+    t_outlet *o2;
+    for (o2 = x->ob_outlet, n = 0; o2 && m--; o2 = o2->o_next);
+    if (o2 && (o2->o_sym == &s_signal))
+    {
+        return o2->o_nchans;
+    }
+}
+
+
+
```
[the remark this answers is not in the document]
Spacechild1: Also, it would make more sense to move the two functions below `obj_sigissignaloutlet`.

porres:
mc connections could be a little bit thicker I guess

src/m_obj.c
```diff
@@ -892,6 +893,29 @@ int obj_nsigoutlets(const t_object *x)
...
+}
+
+int obj_sigoutletgetchans(const t_object *x, int m){
+    int n;
+    t_outlet *o2;
+    for (o2 = x->ob_outlet, n = 0; o2 && m--; o2 = o2->o_next);
+    if (o2 && (o2->o_sym == &s_signal))
+    {
+        return o2->o_nchans;
+    }
```
Spacechild1, line 914: Same as above. Also, what's the return value if the if-statement is false? ;-) Actually, the compiler should give you a warning about missing return value!

I would do the following:
```
    if (o2 && (o2->o_sym == &s_signal))
        return o2->o_nchans;
    else
        return 0;
```
See also https://github.com/pure-data/pure-data/pull/2057/files#diff-5dcf32c92a1ab6eae741228954925258717e876e91c08deff0ce611f94833d14L2400

src/g_editor.c
```diff
@@ -2397,7 +2397,25 @@ static void canvas_doclick(t_canvas *x, int xpos, int ypos, int which,
                 {
                     if (doit)
                     {
+                        //int issignal = obj_issignaloutlet(hitobj, closest);
                         int issignal = obj_issignaloutlet(hitobj, closest);
+                        int nchannels = obj_sigoutletgetchans(hitobj, closest);
```
Spacechild1, line 2402: Actually, `issignal` is redundant because `obj_sigoutletgetchans` should return 0 for non-signal outlets. This means you can simply do the following:
```
int nchannels = obj_sigoutletgetchans(hitobj, closest);
int connection_thickness;
if (nchannels > 1)
    connection_thickness = 2.5 * x->gl_zoom;
else if (nchannels == 1)
    connection_thickness = 2 * x->gl_zoom;
else /* control outlet */
    connection_thickness = x->gl_zoom;
```
Or more tersely:
```
int nchannels = obj_sigoutletgetchans(hitobj, closest);
int connection_thickness =
    (nchannels > 1 ? 2.5 : nchannels == 1 ? 2.0 : 1.0) * x->gl_zoom;
```

Same in `tryconnect`

Spacechild1 (changes requested):

Spacechild1:
Hi, I added a view comments. No need for fix-up comments, just ammend your commit and force-push.

BTW, I think you'd also need to redraw all canvases after a DSP graph update.

Spacechild1:
> mc connections could be a little bit thicker I guess

I agree. 2.5 is *very* subtle. I think it can be at least 3, maybe even 4.

Spacechild1:
BTW, the mc-connection-thickness could be a constant in `g_canvas.h` (probably in the "geometry" section), so you can easily change it in one place. For example:
```
#define MC_CONNECTION_WIDTH 3
```

src/g_canvas.c
```diff
@@ -942,11 +942,27 @@ static void canvas_drawlines(t_canvas *x)
         linetraverser_start(&t, x);
         while ((oc = linetraverser_next(&t)))
         {
+            int issignal = outlet_getsymbol(t.tr_outlet) == &s_signal;
+            int nchannels = obj_sigoutletgetchans(t.tr_ob, t.tr_outno);
+            int connection_thickness;
```
Spacechild1, line 947: `connection_thickness` is a bit of a mouthful. I think in this context `width` would be perfectly clear :)

Spacechild1:
Hint: you don't need to push fixup commits, just ammend your last commit and force-push.

src/g_editor.c
```diff
@@ -2397,7 +2397,9 @@ static void canvas_doclick(t_canvas *x, int xpos, int ypos, int which,
                 {
                     if (doit)
                     {
-                        int issignal = obj_issignaloutlet(hitobj, closest);
+                        int nchannels = obj_sigoutletgetchans(hitobj, closest);
+                        int connection_thickness = (nchannels > 1 ? MC_CONNECTION_WIDTH : nchannels == 1 ? 2.0 : 1.0) * x->gl_zoom;
```
Spacechild1, line 2401: nitpick: use the same indentation as in the other two places to avoid long lines.

Also, you can actually write `2` and `1` instead of `2.0` and `1.0`. We don't need doubles here. My mistake.

src/m_obj.c
```diff
@@ -919,6 +920,35 @@ int obj_issignaloutlet(const t_object *x, int m)
...
+
+    if (o2 && (o2->o_sym == &s_signal)){
+        if (o2->o_nchans == 0){
+            return 1;
+        }
+        return o2->o_nchans;
+    }
+    else{
+        return 0;
+    }
```
Spacechild1, line 949: Again, code style. Opening braces go on their own line. Omit braces for if-statements with a single expressions (also above in `obj_sigoutletsetchans`).

Spacechild1:
```
if (o2->o_nchans == 0){
    return 1;
}
return o2->o_nchans;
```
I think the following is more natural:
```
if (o2->o_nchans == 0)
    return 1;
else
    return o2->o_nchans;
```
Or even:
```
return (o2->o_nchans > 0) ? o2->o_nchans : 1;
```

src/m_obj.c
```diff
@@ -919,6 +920,31 @@ int obj_issignaloutlet(const t_object *x, int m)
...
+}
+
+int obj_sigoutletgetchans(const t_object *x, int m)
+{
+    int n;
+    t_outlet *o2;
+    for (o2 = x->ob_outlet, n = 0; o2 && m--; o2 = o2->o_next);
+
+    if (o2 && (o2->o_sym == &s_signal)){
```
Spacechild1, line 941: Again, open braces go on their own lines. But in this case, no braces :) (Also in `obj_sigoutletssetchans`.)

Sorry, for the nitpicking, but we really try to keep the code style consistent. Soon it will be in your muscle memories :)

src/m_obj.c
```diff
@@ -919,6 +920,31 @@ int obj_issignaloutlet(const t_object *x, int m)
     "MAINSIGNALIN" has been provided, in which case the object won't
     promote scalars correctly.  Nonetheless we provide it so that at least
     such a badly written object won't crash Pd. */
+
+void obj_sigoutletsetchans(const t_object *x, int m, int nchans)
+{
+    int n;
```
Spacechild1, line 926: Oh, and `n` is not used.

charlesneimog: `n` is used here:

``` c
for (o2 = x->ob_outlet, n = 0; o2 && m--; o2 = o2->o_next);
```

You normally use `int n = 0`?

src/m_obj.c
charlesneimog: No problem, I will fix it

src/m_obj.c
Spacechild1: That is just a left-over from `obj_nsigoutlets` or `obj_sigoutletindex`. Here you just assign 0 to `n` and then do nothing with it :) Just remove it.

BTW, I just saw that there is the same issue in `obj_issignaloutlet` above. You can go ahead and remove that one as well.

Spacechild1:
> No need for fix-up comments, just ammend your commit and force-push.

Ah, I only meant your *initial* commit. Pro tip: you can reset your branch to a previous commit - while keeping all your later changes - with `git reset <commit hash>`. Then you can ammend that commit, add all the later changes and force push.

However, don't bother if you're not too comfortable with git. We can also remind Miller to squash-merge :)

Spacechild1:
Oh, btw, I just realized that you are working on your master branch. Typically you would use a dedicated branch for each PR and keep the master branch in sync with upstream. You can easily create a new branch from current master with `git checkout -b <branch_name>`.

BTW, here's a very nice resource for learning git: https://git-scm.com/book/en/v2. The first 3 chapters provide a very good overview of the most important commands.

Spacechild1:
Cool, I think the only thing that's left is redrawing the canvases when the DSP graph is updated.

src/m_obj.c
```diff
@@ -919,6 +920,27 @@ int obj_issignaloutlet(const t_object *x, int m)
     "MAINSIGNALIN" has been provided, in which case the object won't
     promote scalars correctly.  Nonetheless we provide it so that at least
     such a badly written object won't crash Pd. */
+
```
Spacechild1, line 923: This comment belongs to `obj_findsignalscalar` :) Just move the two functions above this comment.

Spacechild1: Or better: after `obj_findsignalscalar` so it matches your order in `s_imp.h`.

charlesneimog:
@Spacechild1, first thanks for the help! I learned important things, and sorry for my noob mistakes. 

I had reset the master to last commit by Miller and now the mc draw things is one separeted branch. I will close this PR to make it in the right way...

Thanks again!!!

Issue #1996, opened 2023-06-16
missing/useful [snake~] methods


porres:
an issue to discuss and propose more [snake~] methods

- [snake~ nchs], get number of channels in a multichannel connection, useful to route signals according to number of channels and also to resize [clone], see https://github.com/pure-data/pure-data/issues/1989
- [snake~ mix], this method could have a 2nd inlet with a toggle and one could use it to mix all channels into one when on, or pass it unchanged otherwise. This would be an alternative to https://github.com/pure-data/pure-data/issues/1994
- [snake~ ltos] "list to sig" or something. Create a multichannel according to a list input. A bit like [sig~] but for lists and multichannels, it would also allow us to create a variable number of channels length.

umlaeute:
what's the advantage of `[snake~ ltos]` over the enhanced version of `[sig~]` as suggested by @Spacechild1 ?

porres:
> what's the advantage of [snake~ ltos] over the enhanced version of [sig~] as suggested by @Spacechild1 ?

being able to create multichannel signals with a different number of channels according to the list input length

Spacechild1:
I think these are all useful, but I would prefer different names:

`[snake~ nchs] -> `[snake~ count]` or `[snake channels]`
`[snake~ mix]` -> `[snake~ sum]`
`[snake~ ltos]` -> `[snake~ fromlist]`

Two more methods:

* `[snake~ split]`: split multichannel signal into two (the split index can be set dynamically)

* `[snake~ merge]` or `[snake~ join]`: merg/join two multi-channel signals together

abreubacelar:
since split feels like the verb from list split, merge could be [snake~ append] and [snake~ prepend] ?

Spacechild1:
> merge could be [snake~ append] and [snake~ prepend] ?

Prepend/append only makes sense for control objects where you have hot and cold inlets. For signal objects you only need a single method.

abreubacelar:
oh, understand, i thought about avoiding crossed connections, but it is silly xD

and what about an object to take specific channels from a signal?

not sure about the verb... get/take/reorder... ?

[snake~ get 1 3 2 4] could take an arbitrary signal with n channels and swap 3 and 2 channels and return just a 4 channel signal from the n channel input, would this be possible with the current objects?

maybe this could even be used for a specific duplication of signals 

[snake~ get 1 2 1 3] could take a 3 channel, and output a 4 channel with the third signal being a copy from the first signal from the input

[...]

PR #2058, opened 2023-07-11
allow to change the number of tables/channels in table DSP objects

Spacechild1:
This is a reboot of https://github.com/pure-data/pure-data/pull/2055. Thanks to @umlaeute for the crucial hint!

---

The "set" message may now resize the DSP array vector. However, it does not require a DSP graph update because we do not pass `t_dsparray` pointers directly to the perform routine; instead we pass the `t_arrayvec` member together with the channel index. The size of the vector is independent from the input/output channel count.

`tabwrite~` and `tabsend~` continue to take the number of channels from their signal inlet. The only difference is that the number of tables is not fixed at creation time anymore.

`tabplay~`, `tabread~`, `tabread4~` and `tabreceive~` now also have a "channels" method that allows to dynamically set the number of output channels (independently from the "set" method).

Tested successfully with all objects.

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -73,14 +73,41 @@ static void arrayvec_testvec(t_arrayvec *v)
     }
 }
 
+    /* This will set the vector to a new list of array names and possibly size it.
+    Note that we do not pass t_dsparray pointers directly to the perform routines;
+    instead we pass a pointer to the t_arrayvec member together with the channel index,
+    so that the actual vector can be reallocated without requiring a DSP graph update.
+    Also, the "set" message does not affect the number of input/output channels!
+    tabwrite~ and tabsend~ simply take the number of channels from their signal inlets.
+    With tabplay~, tabread~, tabread4~ and tabreceive~, the user can dynamically change
+    the number of output channels with the "channels" message. */
 static void arrayvec_set(t_arrayvec *v, int argc, t_atom *argv)
 {
-    int i;
-    for (i = 0; i < v->v_n && i < argc; i++)
+    int i, oldsize = v->v_n;
+    void *owner = v->v_vec[0].d_owner;
+    if (argc < 1)
+        return;
+        /* unset all gpointers */
+    for (i = 0; i < v->v_n; i++)
+        gpointer_unset(&v->v_vec[i].d_gp);
+    if (argc != oldsize)
     {
-        gpointer_unset(&v->v_vec[i].d_gp); /* reset the pointer */
+        v->v_vec = (t_dsparray *)resizebytes(v->v_vec,
+            oldsize * sizeof(*v->v_vec), argc * sizeof(*v->v_vec));
+        v->v_n = argc;
+            /* init new elements */
+        for (i = oldsize; i < v->v_n; i++)
+        {
+            gpointer_init(&v->v_vec[i].d_gp);
+            v->v_vec[i].d_owner = owner;
+            v->v_vec[i].d_phase = 0x7fffffff;
+        }
+    }
+        /* set array names */
+    for (i = 0; i < v->v_n; i++)
+    {
         if (argv[i].a_type != A_SYMBOL)
-            pd_error(v->v_vec[i].d_owner,
+            pd_error(owner,
                 "expected symbolic array name, got number instead"),
                 v->v_vec[i].d_symbol = &s_;
         else
@@ -153,14 +180,18 @@ static void tabwrite_tilde_redraw(t_symbol *arraysym)
 
 static t_int *tabwrite_tilde_perform(t_int *w)
 {
-    t_dsparray *d = (t_dsparray *)(w[1]);
-    t_sample *in = (t_sample *)(w[2]);
-    int n = (int)(w[3]), phase = d->d_phase, endphase;
+    t_arrayvec *v = (t_arrayvec *)(w[1]);
+    int chan = (int)(w[2]);
+    t_sample *in = (t_sample *)(w[3]);
+    int n = (int)(w[4]), phase, endphase;
+    t_dsparray *d;
     t_word *buf;
 
-    if (!dsparray_get_array(d, &endphase, &buf, 0))
+    if (chan >= v->v_n ||
+        !dsparray_get_array((d = &v->v_vec[chan]), &endphase, &buf, 0))
         goto noop;
 
+    phase = d->d_phase;
     if (phase < endphase)
     {
         int nxfer = endphase - phase;
@@ -184,7 +215,7 @@ static t_int *tabwrite_tilde_perform(t_int *w)
     }
     else d->d_phase = 0x7fffffff;
 noop:
-    return (w+4);
+    return (w+5);
 }
 
 static void tabwrite_tilde_set(t_tabwrite_tilde *x, t_symbol *s,
@@ -195,12 +226,11 @@ static void tabwrite_tilde_set(t_tabwrite_tilde *x, t_symbol *s,
 
 static void tabwrite_tilde_dsp(t_tabwrite_tilde *x, t_signal **sp)
 {
-    int i, nchans = (sp[0]->s_nchans < x->x_v.v_n ?
-        sp[0]->s_nchans : x->x_v.v_n);
+    int i, length = sp[0]->s_length, nchans = sp[0]->s_nchans;
     arrayvec_testvec(&x->x_v);
     for (i = 0; i < nchans; i++)
-        dsp_add(tabwrite_tilde_perform, 3, x->x_v.v_vec+i,
-            sp[0]->s_vec + i * sp[0]->s_length, (t_int)sp[0]->s_length);
+        dsp_add(tabwrite_tilde_perform, 4, &x->x_v, (t_int)i,
+            sp[0]->s_vec + i * length, (t_int)length);
 }
 
 static void tabwrite_tilde_start(t_tabwrite_tilde *x, t_floatarg f)
@@ -257,6 +287,7 @@ typedef struct _tabplay_tilde
     t_object x_obj;
     t_outlet *x_bangout;
     int x_limit;
+    int x_nchannels;
     t_clock *x_clock;
     t_arrayvec x_v;
 } t_tabplay_tilde;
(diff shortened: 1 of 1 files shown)


millerpuckette:
I have a question about this one - why should tabread~ and tabread4~ have "channels" messages?  I don't know of any situation where the number of output channels should differ from the number of input channels.

The rest looks good to me.

Spacechild1:
> I have a question about this one - why should tabread~ and tabread4~ have "channels" messages?

Because you apparently suggested it :) See https://github.com/pure-data/pure-data/pull/2055#issuecomment-1628643260

We want to be able to change the number of channels at runtime, but `[set(` does not change the channel count, so I have added a `[channel(` message, similar to other mc objects.

One nice side effect, as noted by @umlaeute, is that we can use the creation arguments to set the max. number of tables and then use `[channels(` to select the first N tables. I would have plenty of use cases for that.

millerpuckette:
I can't see any comment I made on PR 2055...
Anyway, tabread~ has one inlet, which can be multichannel... in which case, creation arguments or a "set" message can set the number of arrays to use.  I imagine also that a "-m" creation argument could _statically_ set the number of output channels.  But why dynamic numbers of channels set by run-time messages?
(The reason adc~ sort of allows that is to keep the "-m number-of-channels starting-channel" format of the creation arguments.  In that case you might want to run-time change the starting channel but not the number of channels.)

Spacechild1:
Anyway, I just realized this branch has conflicts. Will try to resolve them ASAP.

Spacechild1:
When I resolved the merge conflicts I realized that this PR does actually *not* fully preserve backwards compatibility:

Pd 0.55: "set" with a smaller number of arrays preserves the remaining elements (given as creation arguments)

This PR: "set" with a smaller number of arrays resizes the vector, discarding the remaining elements

TBH, I don't really think that anyone is relying on this behavior, so we might get away with it. If not, I'm not really sure what to do... I can think of some solutions, but they all have their problems.

Actually, I now think that this feature needs some more discussion, so I'm not really against postponing it to Pd 0.57. After all, we won't get all the missing MC features into 0.56 anyway (e.g. https://github.com/pure-data/pure-data/issues/1996)

porres:
> I don't really think that anyone is relying on this behavior, so we might get away with it.

me neither :)

Spacechild1:
> TBH, I don't really think that anyone is relying on this behavior, so we might get away with it.

In case we are ok with breaking backwards compatibility in this regard, we actually wouldn't need the "channels" method in the first place. Instead we could just take the channel count directly from the arrayvec and update DSP if the channel count has changed.

If we don't want "set" to update DSP while DSP is on, we could do a similar thing to `[receive~]`: if "set" would resize the arrayvec and DSP is on, we print a warning and only update the output channel count on the next DSP update.

---

Any way, the current state of things has two severe problems:

1. we can't change the channel count at runtime at all
2. multi-channel mode requires the user to provide a list of tables at creation time whereas in single-channel mode we can start with an empty object.

This needs to be fixed in some way or another.

porres:
I've being using mutlichannel for a new way to handle polyphony and i've needed a more flexible way to handle things where hardcoding would be just bad.

In my ELSE library a [nchs~] external has been crucial to check and report the number of channels and adjust things accordingly. I suggested something like that for [snake~]. I am also using something that is more flexible than [sig~], where I give it a list of different sizes and it creates a multichannel signal according to that size and it's dynamic, it changes on runtime. That's also been useful and I also suggested that as a new method for [snake~].

I also have a [sum~] object that can, at runtime, sum and merge multichannels into a single channel, or bypass them unsummed, and a [select~] object can select and pass along signals of different sizes. In all these cases the channel size can change during run time and I play along with it, and these objects emerged from real necessity. 

This is just to say that, in general terms, I believe we often eventually don't want to hardcode things at runtime. Christof gave the examples of ambisonics of unkown order.

Spacechild1:
Here's my longish reply :)

> I wasn't at all thinking about (or wanting to think about) allowing run-time renumbering of channels in Pd - I'm afraid of what feature creep this might bring on. But perhaps it's not as drastic a thing as I imagine.

We actually had this discussion 2-3 years ago when you made the first MC draft. I think you initially wanted to make all channel counts static, but we (successfully) convinced you otherwise. That's why I'm surprised that we are now having the same discussion again.

> My original idea was to have it be posible to have variable-channel-count abstractions by passing a creation argument to the abstraction. This "should" only be necessary for creators of multichannel signals that can't inherit the channel count from their signal inputs: receive~, catch~, delread~, adc~, sig~, readsf~, and perhaps a few others. Hmm, what about line and vline~? The thing grows out of control fast. - so I figured that the only objects that should be snakeableshould be the stateless ones, and the clone~ object is always there for people wanting to, for example, have a bank of filters.

I think you're conflating two questions here:

1. Which objects should support multichannel signals?

2. Should objects be allowed to (actively) change the number of output channels?

These two issues are really orthogonal! This PR is only about 2., so let's focus on that.

Now, the answer to 2. has actually been "yes" since Pd 0.54! The following objects *already* can change the number of output channels *through user interaction*:

- `[adc~]` with the "set" message
- `[catch~]` with the "channels" message
- `[receive~]` with the "set" message (currently still requires manually toggling DSP)
- `[clone -d]` as a side effect of the "resize" message

Also, quite a few externals have gained multichannel support and many of them allow to change the number of output channels. @porres already gave you some examples from the "else" library. Two more examples from myself: `[vstplugin~]` and `[aoo_sink~]`. Changing/setting the number of channels at runtime is something we have come to expect and it works perfectly fine.

> Unfortunately I've lost track of the narrative for a simple reason: since introducing multichannel I've never actually had a use for it myself, and indeed haven't used it except to test the objects. So this means that I can't evaluate whether run-time "channels" messages to everything would have a purpose or not.

Ok, this actually explains a lot :) As someone who actually uses multichannel objects, the need for run-time channel changes is pretty obvious.

> My instinctive reaction is to limit channel count specification to creation arguments. 

And that's exactly the problem. If you limit it to creation arguments, you have to know the number of channels when you create the patch. But that's not always the case! I already gave the example of the ambisonics order.

Generally, we should avoid limiting things to creation arguments whenever possible because certain variables might only be known at runtime (e.g. read from a configuration file). For example, up until very recently, `[poly]` did not allow to set/change the number of voices at runtime, which made it practically useless for me in many situations. (This has been particularly ironic because `[clone]` did have a "resize" method.) Simiarly, I often cannot use the table objects in multichannel signal chains because of the fixed channel count.

There really needs to be *some* way to set/change the number of tables/channels at runtime. The question is how.

millerpuckette:
OK, I'm convinced... and would be OK with throwing this in 0.56, now that it's top of mind (if I put it off for 0.57 I might forget all about it :)

I'm not worried about 0.55 compatibility since the way 0.55 worked seems broken anyhow.  So I _think_ I can merge this as it is presently...?

Spacechild1:
Yay! :)

> I'm not worried about 0.55 compatibility since the way 0.55 worked seems broken anyhow. So I think I can merge this as it is presently...?

Now that you're saying you are willing to change the behavior of the "set" method, I'm wondering if we really need the "channels" method at all. Instead, "set" could directly set the number of output channels, forcing a DSP graph update if the channel number changes. I think this would be more intuitive than having a separate "channels" message.

IIRC I only added the "channels" method so that we could keep the behavior of "set" (which I ironically didn't really accomplish :)

Personally, I tend to removing the "channels" method in favor of empowering the "set" method. If you are not principally against that, I can make a PR. Then we can evaluate both options and make a decision.

---

https://github.com/pure-data/pure-data/pull/2049, on the other hand, should be ready for merging as is.

Spacechild1:
> And... I still don't see why the number of output channels in tabread~/tabread4~ should ever differ from the number
of input channels

Because these two are independent! I think I already pointed this out in another comment. You might either use

a. a single-channel signal to control the offset of all tables
b. a multi-channel signal to control the offsets indepdently

> I'd suggest acting like +~, etc, and cycling through the arrays

That's what's already happening!

So all my previous points still apply.

millerpuckette:
For case 1 - I think that should be a new capability of snake~ in (or perhaps a new verb) - it has nothing to do with the table~ object as such.

millerpuckette:
> I think that is true for objects that have two signal inputs, in which (unfortunately) the channel counts might not match. But it's unnecessary to then oblige every one-input-one-output object to do the same. For tabread~, it's no problem having the same number of inputs as outputs, as in sqrt~, etc.
> 


P.S. of course sqrt~ is different from tabread~ - but so is tabread~ from +~ - so I don't see a good argument based on consistency here.

millerpuckette:
Hmm.. If you make a [tabread~ array1] and then feed it a 3-channel input signal, shouldn't the output be 3 channels (each looking up array1)?

Spacechild1:
> Hmm.. If you make a [tabread~ array1] and then feed it a 3-channel input signal, shouldn't the output be 3 channels (each looking up array1)?

No, it's the other way round. The number of output channels is determined by the number of arrays. The input signal wraps around to accomodate the number of arrays. That's why you can use a singlechannel signal to control all arrays (which for me at least is the common case).

That's how it already works and we can't change it. (And I actually think the current behavior is good and intuitive.)

The only thing I'm asking for is a way to change the number of tables resp. the number of output channels.

Spacechild1:
So back to my original question:

Should we still have a dedicated "channels" method?

Or would be ok for "set" to change the number of channels together with setting the arrays in a single operation?

@umlaeute pointed out *some* advantages of "channels", but IMO having only "set" would be more intuitive and easy to use.

Spacechild1:
> I now think that the number of output channels should be the maximum of the number of channels of the input signal and the number of tables provided (this is sort of what "+~" does), wrapping as necessary.

This sounds good! It would keep the current behavior while still allowing both use cases. Nice!

> If we do that, do we also need an "nchannels" message? I think not - for the same reason that "+~" doesn't need one :)

I already said, the only reason why I added the "channels" message is because you initially didn't want "set" to change the number of output channels.

If you are ok with it, I would get rid of "channels", allow "set" to actually resize the arrayvec (updating DSP if needed) and apply your proposed wrapping behavior. It just needs a slight modification of this PR. I can do that tonight. (Need to head to the playground soon :)

millerpuckette:
OK go for it :)

Spacechild1:
Closing in favor of https://github.com/pure-data/pure-data/pull/2595

PR #2059, opened 2023-07-11
Add functions to draw multichannel connections, see #1997

charlesneimog:
As recommended by @Spacechild1 in #1997 and #2057...

There is a need to redraw the connections after the DSP is turned `on`.

https://github.com/pure-data/pure-data/assets/31707161/bf72893b-b272-4567-8d4e-4da2ed60fa4f


src/m_obj.c
```diff
@@ -919,6 +920,27 @@ int obj_issignaloutlet(const t_object *x, int m)
     "MAINSIGNALIN" has been provided, in which case the object won't
     promote scalars correctly.  Nonetheless we provide it so that at least
     such a badly written object won't crash Pd. */
```
Spacechild1, line 922: this comment belongs to `obj_findsignalscalar`. Just move the two functions above the comment.

Spacechild1:
Don't forget to redraw all canvases after a DSP graph update, probably at the end of `canvas_start_dsp`.

src/m_obj.c
Spacechild1: hmmm... the comment is still at the wrong place.

sebshader:
is it possible to just redraw the lines? redrawing the entire canvas seems like overkill

sebshader:
seems like it would be better if object authors could just set the number of channels for an outlet, then whenever the number of channels changes for that outlet just the connections for that outlet can be redrawn

Spacechild1:
> is it possible to just redraw the lines? redrawing the entire canvas seems like overkill

Sure. That would be even better! I don't immediately see a function for this, but we could add `canvas_updatelines`, based on `canvas_drawlines`, and call it recursively on all (visible) canvases.

> seems like it would be better if object authors could just set the number of channels for an outlet, then whenever the number of channels changes for that outlet just the connections for that outlet can be redrawn (without having to do redrawing everything when the graph gets updated which seems like a lot of extra computer work)

Hmmm. Objects call `signal_setmultiout` with the desired channel count in the "dsp" method. In theory, when we call `obj_sigoutletsetchans` in `ugen_doit`, we could compare it against the previous channel count and send a message to the GUI to change the line width. However, I'm not sure if we should really send graphics updates from within the DSP sorting routine...

> But tbh I haven't familiarized myself w/ the multichannel architecture at all

You should ;-)

sebshader:
It seems like object authors could call a function to set the channel count. That would set the number of channels in t_outlet and redraw the cables if necessary.
then when the dsp graph comes along it can use that channel count (and call ```obj_sigoutletsetchans``` or whatever)

sebshader:
hmm well maybe it should just be based on if it's 'multi-aware' then.. but I'll refrain from further suggestions until I actually look at it

Spacechild1:
> maybe it should just be based on if it's 'multi-aware' then.. 

I don't think that would be particularly useful.

Actually, I have already proposed several solutions for updating the cables based on the actual channel count. Maybe let's just give it a try?

sebshader:
I guess I was trying to find ways to avoid a redraw of every canvas (or even every connection) when the dsp gets turned on. (also, would cables not look multichannel until dsp is turned on? that would be a bit odd imo)

Spacechild1:
> I guess I was trying to find ways to avoid a redraw of every canvas

I think I have proposed a solution in https://github.com/pure-data/pure-data/pull/2059#issuecomment-1631651153

> (also, would cables not look multichannel until dsp is turned on? that would be a bit odd imo)

Signal outlets only get their channel count in the "dsp" method, so naturally they can't appear as multichannel before DSP is turned on.

porres:
Hi, why is this still a draft? Can we try and make it a final proposal for the upcoming 0.55 release?

porres:
Hi @charlesneimog why is this still a draft? Can we try and make it a final proposal for the upcoming 0.56 release?

charlesneimog:
Because there was some bugs with it, I couldn't figure out how to solve then. Also it is weird because it just redraw the line after you turn on the `DSP`, what is weird.

Issue #1870, opened 2023-01-19
allow to [clone] subpatches (not requiring abstractions)


umlaeute:
`[clone]` is great, but has the drawback of requiring a wrapper-abstraction to be cloned.

with the recent talk about multi-channel support for Pd (see pd-dev list), the idea has arisen to have `[clone]` support "atomic" objects (that objectclasses like builtins, or externals...; as opposed to abstractions).

i would like to extend this to suggest support for "subpatches" in `[clone]`.

as "subpatch" in this context is a canvas that is stored as part of the containing patch (that is: a `[pd]`).

subpatches within `[clone]` would use an implicit "dummy-abstraction" that conceptually wraps the subpatch, even though it technically is stored in the patch file that contains the `[clone]` object.
arguments are visible in the subpatches as they are passed to `[clone]`.

### example
consider `[clone pd 10 lop 500]`.
clicking on the `[clone]` object would open up a subpatch `[pd 0 lop 500]`, where you can reference the 3 arguments, with `$1="0"` (that is: the clone-index), `$2="lop"` (which i only put there to make it obvious what the [clone] instance is used for), and `$3="500"` (e.g. the curoff frequency).
all the subpatches share the same $0, but this is distinct from the $0 in the patch that contains the `[clone]` object.

the reason for this is mostly to separate the `[clone pd]` consistently from ordinary `[pd]` subpatches:

we do want *some* way to get the clone index into the subpatch, and the standard way this is handled in `[clone]` is via $1. this however would overwrite any $1 passed to the abstraction containing the `[clone]` object. therefore the other dollargs for the abstraction (including $0) shouldn't propagate to the `[clone pd]` either, as this would be most confusing).

PR #2060, opened 2023-07-11
clone: support anonymous abstractions

Spacechild1:
`[clone [flags] pd ...]` creates an anonymous abstraction. This is particularly useful with the new multichannel functionality, in particular with the "-d" flag. Here's a short demonstration:

https://github.com/pure-data/pure-data/assets/16126632/e3398074-32a7-4457-a7de-4d5d8f42e7ad

You can also pass additional creation arguments, just like with "regular" abstractions.

The only notable difference to "regular" abstraction is that it is not saved as a seperate file, instead it is embedded in the clone object's owning canvas.

A special hack also allows to retext the clone object, e.g. changing the number of instances or creation arguments, without losing the abstraction data.

If an anonymous abstraction is saved, the parent canvas is marked dirty, so that we don't accidentally close the patch and lose our changes.

---

### About the implementation:

`t_canvas` now has a new flag `gl_anonymous` that signifies an anonymous (cloned) abstraction. If such an abstraction is saved, it is not saved to disk, but rather the save binbuf is stored temporarily and copied in `clone_reload`. The binbuf is then used to remake the other instances.

When the clone object itself is saved, it also adds the anonymous abstraction binbuf to the save buffer. The format is very similar to regular subpatches. The main difference is that the first "canvas" message goes to `#A` (= the clone object)` instead of `#N` (= `pd_canvasmaker`). This message temporarily binds the `#N`, `#X` and `#A` symbols to responder objects that just add the message to the save binbuf. The last message is always `#X pop clone;`; it restores the previous symbol bindings and finally creates the actual abstraction instances from the binbuf.

Here's a short example patch that demonstrates the save format:
```
#N canvas 168 377 449 300 12;
#X floatatom 76 94 5 0 0 0 - - - 0;
#X obj 76 125 clone pd 5;
#A canvas 518 258 450 300 12;
#X obj 61 57 inlet;
#X obj 61 94 + 2;
#X obj 62 157 outlet;
#N canvas 104 104 450 300 sub 0;
#X obj 37 37 inlet;
#X obj 37 61 * 2;
#X obj 37 85 outlet;
#X connect 0 0 1 0;
#X connect 1 0 2 0;
#X restore 61 127 pd sub;
#X connect 0 0 1 0;
#X connect 1 0 3 0;
#X connect 3 0 2 0;
#X pop clone;
#X obj 76 154 print;
#X connect 0 0 1 0;
#X connect 1 0 2 0;

```

(The "pop clone" message is a bit arbitrary. I'm open for better suggestions :)

For the sake of simplicity, I based this PR on #2053; the actual relevant commit is c43f68b8a4b54657880670c1b3c87061f9b35c6c.

---

Related to https://github.com/pure-data/pure-data/issues/1870


sebshader:
I wonder if this could be combined with patch-embedded abstractions (another often requested feature)? seems like they would go together

Spacechild1:
> I wonder if this could be combined with patch-embedded abstractions (another often requested feature)?

In what way?

sebshader:
just save the abstraction as embedded, and have clone be able use that rather than having it be specific to clone.

Spacechild1:
> just save the abstraction as 'patch-embedded', and have clone be able use that rather than having it be specific to clone ('clone-embedded').

Patch-embedded abstractions could be a nice feature, but I think that's orthogonal to this PR. If we can embed abstractions in patches, we can also use them in `[clone]`, of course. However, the point of this PR is to avoid all these extra steps, in particular for tiny abstractions such as in the video above. I don't see how embedded abstractions would be an improvement in this regard.

porres:
OH MY GOD THIS IS SOOOO COOOL!!!

Spacechild1:
> Why constrain the feature to clone? An embedded abstraction is what it is really.

Embedded abstractions can use the same saving/loading mechanism. But's more an implementation detail. Apart from that, these two features have little in common. 

> They are really doing the same thing

Not really.

> Also if they are unified clone could take advantage of any features that are added to embedded abstractions.

I guess that "embedded abstractions" would just look like regular abstractions, in the sense that you can create them as objects or pass them to clone; the only difference would be the storage location (disk vs. embedded).

Anyway, if you want "embedded abstractions" or have a specific vision of what they could look like, I would suggest to open a dedicated issue for this. Then we can see if there are overlapping areas and where things could be unified.

Spacechild1:
> Embedded abstractions can use the same saving/loading mechanism.

In the end, the save format is mostly trivial. The only interesting part is the first and last message; everything else just looks like a regular subpatch. Of course, we could try to unify these messages, but it's not something that is visible to end users.

sebshader:
cool, I do think some users tinker with patch files. And if there's functionality to pull embedded abstractions into separate non-embedded abstractions it would be nice to be able to use it for both cases.

Spacechild1:
I have now implemented a general mechanism for embedding patches. Here I use it to implement clone's anonymous abstractions, but it may also be used for "embedded abstractions".

mathiasbredholt:
This is so cool, I hope it will get merged at some point!!

Spacechild1:
Well, it's still a draft, but yes, I would also find this really, really useful.

mathiasbredholt:
What are the missing parts?

Spacechild1:
> What are the missing parts? Let me know if I can be of any help.

I'd need some feedback from @millerpuckette, in particular about the underlying patch embedding mechanism. Personally, I'm not sure about the `#N embed begin` and `#N embed end` messages; it's just what I came up with at that moment.

> Had a few crashes but I managed to get a nice patch going.

As long as you get crashes, this PR is certainly not in a mergable state :) If you manage to get any stacktraces, that would be very helpful.

> I have a bug to report, that whenever I load a patch containing a [clone pd 4], and the embedded patch contains a [float $1], the $1 gets replaced by zero when I load the parent patch.

Thanks, I need to take a look. I won't have time to work on this in the next few weeks, though. Also, I probably won't spend too much time on it unless @millerpuckette indicates interest in this feature.

mathiasbredholt:
> As long as you get crashes, this PR is certainly not in a mergable state :) If you manage to get any stacktraces, that would be very helpful.

Will do! I also will try to and trace the crash.

> Thanks, I need to take a look. I won't have time to work on this in the next few weeks, though. Also, I probably won't spend too much time on it unless @millerpuckette indicates interest in this feature.

I saw @millerpuckette mentioning embedded abstractions on the dev list somewhere in the context of helping beginner users that want to copy the example patches and wonder why Pd cannot find the help dependency patches. So let's hope this feature will get more interest from the community!!

ben-wes:
just leaving a message here that i miss this feature right now. :)

PR #2067, opened 2023-07-27
nsis: installer updates

Lucarda:
This PR (targeted to `develop`) attempts to have a less prone to errors installer and also includes some troubleshooting tools.

- **safer behavior**:

  - the (un)installer refuses to do anything if it finds a currently running Pd(64).

  - check if there's write permission on target dir before attempting to do anything else.

- **troubleshoot tools**:

  - tool to clear Pd(64) preferences. (only runs at `user` scope).

  - tool to set which Pd(64) installation is used to open .pd files (scope can be `user` or `allusers`).

  - also these two tools are included in the uninstaller but here the '.pd' files repair is automatic (you don't have to choose which app).

- **user scope**:

  - admin users can choose to install or run troubleshot tools at `user` or `allusers` scope.

- **alternate install**

  - user can choose to do an alternate install if there is already a default one. it prompts for a short string (like `Pd-0.54-0-(alt)` to be used to name the: install folder, start menu folder,  desktop shortcut and uninstall name.

Tested to work correctly on **WinXp(i386)** and **Win11(amd64)**.

Test artifacts: https://nc.nubegris.com.ar/index.php/s/E7Pyg3M9ZA7Hygo

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -336,7 +336,8 @@ Install git if you want to clone the Pd sources from Github, etc:
 
 and/or the nsis installer tool if you want to build the Pd Windows installer:
 
-    (Currently we only support 32bit installers)
+    # 64 bit
+    pacman -S mingw-w64-x86_64-nsis
 
     # 32 bit
     pacman -S mingw-w64-i686-nsis
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -336,8 +336,7 @@ Install git if you want to clone the Pd sources from Github, etc:
 
 and/or the nsis installer tool if you want to build the Pd Windows installer:
 
-    # 64 bit
-    pacman -S mingw-w64-x86_64-nsis
+    (Currently we only support 32bit installers)
 
     # 32 bit
     pacman -S mingw-w64-i686-nsis
--- a/doc/1.manual/x6.htm
+++ b/doc/1.manual/x6.htm
@@ -9,7 +9,7 @@
 <link rel="stylesheet" type="text/css" href="pdmanual.css" media="screen">
 <link rel="icon" type="image/png" href="favicon.ico">
 </HEAD>
-  
+
 <BODY>
 
 
@@ -26,7 +26,7 @@
 <P>
 <A href="index.htm#s6"> back to table of contents </A>
 <BR><BR>
-</P> 
+</P>
 
 <P>Pd is built on the commandline using traditional Unix-style tools. The source
 distribution comes with two build systems:</P>
@@ -314,15 +314,14 @@ pacman -S mingw-w64-x86_64-toolchain mingw-w64-x86_64-clang \
 pacman -S mingw-w64-i686-toolchain mingw-w64-i686-clang \
           make autoconf automake libtool \
           mingw-w64-i686-gettext</pre>
-		  
+
 <p>Install git if you want to clone the Pd sources from Github, etc:</p>
 
 <pre>pacman -S git</pre>
 
 <p>and/or the nsis installer tool if you want to build the Pd Windows installer:</p>
 
-<pre># 64 bit
-pacman -S mingw-w64-x86_64-nsis
+<pre>(Currently we only support 32bit installers)
 
 # 32 bit
 pacman -S mingw-w64-i686-nsis</pre>
@@ -463,7 +462,7 @@ cd ../fiddle~ &amp;&amp; make pd_nt
 
 <P>
 <BR><BR>
-<A href="index.htm#s6"> back to table of contents </A> 
+<A href="index.htm#s6"> back to table of contents </A>
 <BR><BR>
 </P>
 
--- a/doc/1.manual/x6.htm
+++ b/doc/1.manual/x6.htm
@@ -321,7 +321,8 @@ pacman -S mingw-w64-i686-toolchain mingw-w64-i686-clang \
 
 <p>and/or the nsis installer tool if you want to build the Pd Windows installer:</p>
 
-<pre>(Currently we only support 32bit installers)
+<pre># 64 bit
+pacman -S mingw-w64-x86_64-nsis
 
 # 32 bit
 pacman -S mingw-w64-i686-nsis</pre>
--- a/msw/build-nsi.sh
+++ b/msw/build-nsi.sh
@@ -220,10 +220,15 @@ find "${PDWINDIR}" -type d -printf 'RMDir "$INSTDIR/%P"\n' \
 echo "!insertmacro MUI_PAGE_LICENSE \"${PDWINDIR}/LICENSE.txt\"" \
   > "${LICENSEFILE}"
 
+
+# if we are on mingw/cygwin, we have to unmangle the directories
+WWORKDIR=$(cygpath -m "${WORKDIR}" 2>/dev/null || echo "${WORKDIR}")
+WOUTDIR=$(cygpath -m "${OUTDIR}" 2>/dev/null || echo "${OUTDIR}")
+
 # uninstall/license/... information into pd.nsi script
 cat "${SCRIPTDIR}/pd.nsi" | sed \
-	-e "s|include \"/tmp/|include \"${WORKDIR}/|" \
-	-e "s|OutFile \"/tmp/|OutFile \"${OUTDIR}/|" \
+	-e "s|include \"/tmp/|include \"${WWORKDIR}/|" \
+	-e "s|OutFile \"/tmp/|OutFile \"${WOUTDIR}/|" \
 	> "${NSIFILE}"
 
 # check if we have nsis compiler
--- a/msw/build-nsi.sh
+++ b/msw/build-nsi.sh
@@ -220,15 +220,12 @@ find "${PDWINDIR}" -type d -printf 'RMDir "$INSTDIR/%P"\n' \
 echo "!insertmacro MUI_PAGE_LICENSE \"${PDWINDIR}/LICENSE.txt\"" \
   > "${LICENSEFILE}"
 
-
 # if we are on mingw/cygwin, we have to unmangle the directories
-WWORKDIR=$(cygpath -m "${WORKDIR}" 2>/dev/null || echo "${WORKDIR}")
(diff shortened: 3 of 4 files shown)


umlaeute:
nice.

> troubleshoot tools:

i wonder: could we integrate these tools into the (un)installer ?

IIRC, windows often provides this <kbd>Modify/Uninstall</kbd> button in the Software Center, and I think the "clear properties" might be an excellent use-case.

(I'm not entirely sure whether this would work for file associations as well, without letting the user jump through too many hoops...)

Lucarda:
> i wonder: could we integrate these tools into the (un)installer ?

indeed that is useful. I'll work on it.

Lucarda:
> (I'm not entirely sure whether this would work for file associations as well, without letting the user jump through too many hoops...)

it turned out that it's the simplest and useful one. it had been implemented in the uninstaller along with the 'clear prefs'.

things I'm not totally sure

  - wording
  
  - do we need the tools in the installer? i think they are ok there so people are advertised of them.
  
  
the installers in the link above had been updated.

Lucarda:
yes this is odd. iirc during this days i built on linux without the `/tmp/` prefix and it worked. i'll test this again tomorrow. if there are no problems i'll remove them.

sebshader:
well according to the message at the top of the file, the 'tmp' has to be there for miller's build environment using wine? Just seems like there should be some option or setting, I'm wondering how other people use it.

umlaeute:
> Is there a better way around this? Should that be a separate issue?

how about using the `-o/--outdir`  flag:

```
$ ./build-nsi.sh -h
Helper script to build a proper Windows installer out of a
Pd build tree.

Usage:
./build-nsi.sh [ OPTIONS ] <path_to_pd_build> [<version> [<wish-exec-name> [<arch>]]]

 Options:

  -o,--outdir DIR      Output directory (where the generated installer binary
                       can be found after a succesfull run)

  -h,--help            Print this help

 Arguments:

  <path_to_pd_build>   input directory (containing 'bin/pd.exe')
                       such as created by 'make app'
  <version>            Pd-version to create installer for (e.g. '0.32-8')
                       DEFAULT: autodetect from <path_to_pd_build>
  <wish-exec-name>     name of the Tcl/Tk interpreter executable
                       e.g., 'wish85.exe'
                       DEFAULT: use a "bin/wish*.exe" in <path_to_pd_build>
  <ARCH>               architecture of the Pd executable.
                       valid values are '32' for x86 and '64' for x86_64
                       DEFAULT: autodetect from "bin/pd.exe" in <path_to_pd_build>
```

sebshader:
doesn't work, I think the issue is that the script can't find things in "WORKDIR", and OUTDIR is different. I'm wondering if other people use it on windows without editing pd.nsi.

umlaeute:
@sebshader thanks for the additional info. i think this warrants indeed a separate issue.

umlaeute:
in any case, please check whether [06d6632](https://github.com/pure-data/pure-data/pull/2067/commits/06d66324e5c89e9dd1b4388c10ff8522d5b8acd6) in this PR fixes the issue.

sebshader:
unfortunately not, now it's just: ```!include: could not find: "C:/msys64/tmp/tmp.WvWpL8Vj1h/license_page.nsh"```
actually I had already tried changing the paths in the .nsi to that and it's odd it doesn't work because the file is there.

sebshader:
so trying to use the installer, I'm getting the "Refusing to continue, save and quit pd" message on this branch.
(This is after restarting my computer to make sure no pd is running)

sebshader:
though I thought I copied the same files before, now it isn't working..

sebshader:
I'm pretty unexperienced with windows, but I'm not sure if the 'qprocess' exists..

Lucarda:
### 32-bit NSIS

this should be documented somewhere but the thing is that we have to compile with a 32-bit `nsis`. It has to be done like that because the package that comes with Debian and Fedora are 32bit nsis. Also we need to support 32bit Pd on 32bit machines :)

So get

`pacman -S mingw32/mingw-w64-i686-nsis`

then open the MINGW32 shell and compile.

the `Refusing to continue, save and quit pd` should now work correctly (only if Pd is running).

### About `qprocess`

start a normal `CMD` window. Now you are at the 64bit command prompt.

type `qprocess`: you'll get all running processes.

now switch to the 32-bit command prompt

`C:\WINDOWS\SysWOW64\cmd.exe`

 you'll get 

```
Microsoft Windows [Version 10.0.22000.2176]
(c) Microsoft Corporation. All rights reserved.
```

type `qprocess`. you'll get

```
'qprocess' is not recognized as an internal or external command,
operable program or batch file.
```

so to access the 64bit programs you must do:

`cd %windir%\sysnative` 

`qprocess`


Note that on a 64bit prompt you can't `cd %windir%\sysnative`

sebshader:
> ### About `qprocess`
> start a normal `CMD` window. Now you are at the 64bit command prompt.
> 
> type `qprocess`: you'll get all running processes.

running this gives:
```'qprocess' is not recognized as an internal or external command,
operable program or batch file.
```
this is the ```C:/Windows/System32/cmd.exe```

Lucarda:
`qprocess` works for me on Win10 (enterprise).

Which version do you use?

sebshader:
Edition	Windows 10 Home
Version	22H2
Installed on	‎5/‎20/‎2023
OS build	19045.3208
Experience	Windows Feature Experience Pack 1000.19041.1000.0

Lucarda:
I took from [here](https://ss64.com/nt/query-process.html) that this should work but seems there is missing info

sebshader:
I think we might have to use ```tasklist``` instead

sebshader:
well I am on home addition and have ```tasklist``` but no ```qprocess```..

sebshader:
https://superuser.com/questions/1440493/how-to-get-query-command-on-windows-10

Lucarda:
will work on it. Thanks for the valuable test. have to test if `tasklist` works on xp

umlaeute:
i thought a better way would be to use [DDE](https://en.wikipedia.org/wiki/Dynamic_Data_Exchange) as Pd is already using this on Windows (to open documents with an already running instance).

however, it also seems that the current implementation uses the full path to the `pd-gui.tcl` file into the key, which we do not know during the installer.

https://github.com/pure-data/pure-data/blob/c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c/tcl/pd-gui.tcl#L720-L723


probably a running Pd(-GUI) should *also* register a more generic key which can be checked to see if there is *any* Pd(-GUI) running, and possibly foreground it (or quit it).

also, I do not really know how easy it is to interact with DDE from NSIS (there is a very dated `nsisDDE` plugin at http://wiz0u.free.fr/prog/nsisDDE/, but I think we should probably avoid it)

Lucarda:
@sebshader can you test 95d3fac.

one good thing about `tasklist` is that it works even without `%windir%\sysnative` so the installer can be done with 64bit nsis.

@umlaeute yes, let's try to avoid using nsis plugins.

Lucarda:
I have to revisit this as I suppose it will fail in machines not using english locale. but i think i got a solution.

umlaeute:
@Lucarda should we collect the feedback about the installer here? or on the mailinglist? (or can you setup a poll on your nextcloud instance?)

Lucarda:
> should we collect the feedback about the installer here? or on the mailinglist?

I think the mailinglist is Ok (more people getting the call). 

If we find troubles with `tasklist` reported there, may be, we should resort to https://nsis.sourceforge.io/NsProcess_plugin. or live without detection?.

umlaeute:
Following the call on [the mailinglist](https://lists.puredata.info/pipermail/pd-list/2023-08/132487.html), I'm going to collect the results here:

| OS | cpu | state | tester | notes |
|---|---|---|---|---|
| WinXP (pro) | i386 | :tada: | @Lucarda |
| Win8.1 (pro) | amd64 | :tada: | @Lucarda  |
| Win10 (home) | amd64 | :tada: | @Lucarda  |
| Win10 (enterprise) | amd64 | :tada: | @Lucarda  |
| Win11 (pro) | amd64 | :tada: | @Lucarda  |

ben-wes:
feel free to remove this comment again in order to not bury the summarizing table above ...

since it's not mentioned there: it works perfectly well including the warning for a running Pd instance on:
* Microsoft Windows 11 Business 10.0.22621 Build 22621
* 11th Gen Intel(R) Core(TM) i7-1165G7

(installed at user scope - since no admin rights available here)

ben-wes:
@umlaeute : good point. sorry i didn't clarify. had only tested 64bit. but now i tested both with parallel install and running instances of the same instance, other instance, previous 0.53.2 32bit admin install (installed while i still had those privileges) and no running instance. everything works well here! unfortunately can't test the admin scope anymore though.

Lucarda:
@umlaeute to not get in confusions: both of the installers are 32bit `nsis`.

Lucarda:
@umlaeute warning!

>  the installer-binary itself is not really important

it is. we get 32bit Windows programs from from 32bit `nsis`. see above: we had to do tricks (`cd %windir%\sysnative` to use `qprocess` from a 32bit nsis on 64bit Windows. 

> installer-binary itself is not really important (mostly, because there is only one)

there is no "only one" on `msys2`. in the `docs` we say you can build `nsis` installers with 64bit and 32bit `nsis` packages for `msys2`. I choose to build only on 32bit 'nsis` as this one is the *only* arch for on Linux package managers (and Miller uses that).

umlaeute:
Sure, but that's an implementation detail. Esp with your call for feedback in mind, there is no mention/hint/... of a "64bit installer" (nsis64), but there *is* a "64bit (Pd) installer".

The context where I used "64bit installer" was when talking to a *user* who took the question about "CPU" literally and reported their intel model, while the question was really"which of the two installers from the download folder did you use?"

Lucarda:
tested to work fine on **Win7** (home) amd64

Lucarda:
closing as the commits are already on the `master` branch.

Issue #1915, opened 2023-03-13
allow offset canvas block calculation


Spacechild1:
Pd uses reblocking for its FFT implementation. If you have several large FFTs, this can easily lead to excessive CPU spikes. It would be nice if we could offset the canvas block calculation, so that we can manually "distribute" the CPU load over multiple DSP tickes. This could be done with an additional creation argument for the `[block~]` object resp. an additional argument for the `[set(` method. ~~Naturally, the offset would add some extra delay, so there is a trade-off.~~

FWIW, Max already has such a feature: https://docs.cycling74.com/max8/refpages/pfft~#start-onset

Spacechild1:
Closed via https://github.com/pure-data/pure-data/pull/2072

PR #2072, opened 2023-07-31
block~/switch~: add additional "offset" argument

Spacechild1:
The `offset` argument shifts the time of DSP computation in a reblocked subcanvas by N samples.

This is useful for preventing CPU spikes with many simultaneously reblocked subcanvases, in particular with huge FFTs.

Closes https://github.com/pure-data/pure-data/issues/1915


Spacechild1:
This has apparently been merged via 8da9a9e

PR #2081, opened 2023-08-28
Fall back to system check for endian include

danomatika:
I'm updating libpd for 0.54 and notice an issue with the move to m_private_utils.h.

The existing libpd Makefile builds are broken as either `HAVE_ENDIAN_H` or `HAVE_MACHINE_ENDIAN_H` on non-Windows is *required*, otherwise the build halts on the endian check `#error`. On Windows if MINGW is detected, `sys/param.h` is included which pulls in the endian stuff for you.

Overall, I think this private header is a *good idea* however I don't like that it's focused on auto tools this way. A better approach, IMO is to fall back to a system-based include mechanism which I have put together in this PR.

Why do this when you can simply set a define based on the platform? Well, for instance, the libpd python wrapper is built using setup.py which doesn't seem to have *any* option to apply defines based on platform... they are set for *all*. I can't pre-define them in a source file as I don't have easy control of when m_private_utils.h is included in the build process and/or I'd have to put it before every include to m_pd.h. Not ideal either.

### DIFF
--- a/src/m_private_utils.h
+++ b/src/m_private_utils.h
@@ -82,9 +82,9 @@
 #   include <machine/endian.h>
 #  elif defined(__GNUC__)
 #    include <endian.h>
-#  elif defined(__MINGW32__)
-#    include <sys/param.h>
-#  endif
+#endif
+#if __MINGW32__
+#  include <sys/param.h>
 #endif
 
 /* BSD has deprecated BYTE_ORDER in favour of _BYTE_ORDER
--- a/src/m_private_utils.h
+++ b/src/m_private_utils.h
@@ -72,20 +72,21 @@
 
 // known headers?
 #if defined(HAVE_MACHINE_ENDIAN_H) || defined(HAVE_ENDIAN_H)
-#  if defined(HAVE_MACHINE_ENDIAN_H)
-#    include <machine/endian.h>
-#  elif defined(HAVE_ENDIAN_H)
-#    include <endian.h>
-#  endif
+# if defined(HAVE_MACHINE_ENDIAN_H)
+#  include <machine/endian.h>
+# elif defined(HAVE_ENDIAN_H)
+#  include <endian.h>
+# endif
 #else // try to detect from system
-#  if defined(__APPLE__)
-#   include <machine/endian.h>
-#  elif defined(__GNUC__)
-#    include <endian.h>
-#  endif
+# if defined(__APPLE__)
+#  include <machine/endian.h>
+# elif defined(__GNUC__)
+#  include <endian.h>
+# endif
 #endif
+
 #ifdef __MINGW32__
-#  include <sys/param.h>
+# include <sys/param.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


danomatika:
Also it seems `sys/types.h` is an option for macOS, iOS, Android, & Linux: https://stackoverflow.com/questions/27073837/fatal-error-endian-h-file-not-found

danomatika:
I'm not sure why the mingw Travis build fails.

In any case, there are additional includes for the BSDs: https://stackoverflow.com/a/58418801
~~~
#include <sys/types.h>
#include <sys/endian.h>
~~~

umlaeute:
in general i *really* think that we should strive to get rid of any OS-specific define (which is what i was doing) - it just puts all the cruft into one place.
please let us not get back to special-casing all OSs we ever want to support.

as of libpd/python: the `setup.py` superficially looks very declarative, but of course you have the full power of python.
so it's easy to add OS-specific defines like so:

```python

import sys

my_defines = [
  ('PD', 1),
  ('HAVE_UNISTD_H', 1),
  ('HAVE_LIBDL', 1),
  ('USEAPI_DUMMY', 1),
  ('LIBPD_EXTRA', 1)
]

if sys.name == "darwin":
  my_defines.append(('HAVE_MACHINE_ENDIAN_H', 1)

setup(name='pypdlib',
      version='0.12',
      py_modules = [
        'pylibpd'
      ],
      ext_modules = [
        Extension("_pylibpd",
            define_macros = my_defines,
            # ...
```

danomatika:
Hah, I was afraid this would be your answer. I do appreciate the solution for Python and yes I assumed it was not an active script...

danomatika:
Note: I am putting *required* no-brainer fixes for libpd into develop. When I finalize I can also update the libpd files which are one version behind here.

danomatika:
Ok, python solution is in place: https://github.com/libpd/libpd/blob/develop/python/setup.py

I get the argument of avoiding platform-specifics, however in this case we always include the __MINGW32__ header without a define check. Oh well...

danomatika:
Thinking about this again, I think for my issue it would *nice* to also support `sys/types.h` which is a Posix header pulls in endian.h per-platform. This would simply the makefiles to some degree ala setup.py only needs the single define for this etc.

```C
#ifdef HAVE_SYS_TYPES_H
# include <sys/types.h>
#endif
```

umlaeute:
i'm not at all opposed (that is: sure, go ahead if you find it useful), but from [reading the(?, well some) docs](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_types.h.html) it seems that there is no guarantee that this header pulls in the relevant endian.h. (but i figure you would just additionally include that file).

you probably want to add a comment on why this file gets included (as should have been done with the `sys/param.h` file)

danomatika:
Just adding a note in case anyone runs into this problem: the Commandline Tools must be installed for `machine/endian.h`. I had uninstalled them and Xcode itself does *not* come with this header.

Issue #2071, opened 2023-07-31
t_signal->s_overlap is unset (and thus 0)


umlaeute:
as of Pd-0.54.0, the `t_signal` struct has gained an `s_overlap` member that can be queried for the overlap factor of the patch.
at least, in theory.

in practice, this member is only defined in `m_pd.h` but nowhere used within the Pd code.
it is therefore unset and defaults to `0`.

porres:
Great, we can have an object now that gives us the overlap factor and the actual sample rate huh?

PR #2083, opened 2023-08-29
Set t_signal->s_overlap in signals

umlaeute:
As of Pd-0.54, the `t_signal` struct has an `s_overlap` member to get the overlap factor in the patch.
Unfortunately, so far it is nowhere set (and thus 0).

This PR attempts to fix this.

Like the `s_sr` member, the `s_overlap` accumulates overlap factors (thus nesting two subpatches with `[block~ 256 4]` will report the overlap factor as `16`.
(I don't think that there are any use cases for overlapping already overlapped signals; the current behaviour is merely to get consistent behaviour with the `s_sr` member)

the implementation in the PR is written to ***keep** binary compatibilty* with externals that use `signal_new()`:
- creating a new signal with `signal_new()` will set the initial overlap value to `0` ("unknown").
- the caller of `signal_new()` is responsible for setting the overlap factor on the returned value
- other functions (`signal_newlike()`, `signal_setborrowed()`, `signal_setmultiout()`, `signal_newfromcontext()`) do update the `s_overlap` value (since they have enough information to do so).

i would have  preferred to add an overlap argument to `signal_new()`, but that obviously introduces a binary incompatibility. a quick check on my harddisk has not revealed any externals that call this function, but you never know. an alternative would have been to introduce a `pdsignal_new()` function with the additional argument (and make `signal_new()` a wrapper around this).
because of these concerns i haven't committed directly to the `develop` branch and request a review.

Closes: #2071

### DIFF
--- a/libpd/Makefile
+++ b/libpd/Makefile
@@ -11,7 +11,7 @@ PLATFORM_ARCH ?= $(shell $(CC) -dumpmachine | sed -e 's,-.*,,')
 
 ifeq ($(UNAME), Darwin)  # Mac
   SOLIB_EXT = dylib
-  PLATFORM_CFLAGS = -DHAVE_LIBDL
+  PLATFORM_CFLAGS = -DHAVE_LIBDL -DHAVE_MACHINE_ENDIAN_H
   PLATFORM_LDFLAGS = -dynamiclib -ldl -Wl,-no_compact_unwind
   ifeq ($(FAT_LIB), true)
     # macOS universal "fat" lib compilation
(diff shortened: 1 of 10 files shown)


umlaeute:
i think i'm going to merge this into `develop`, as it doesn't break the API.

if an API-breaking solution (extending `signal_new()` to include an `overlap` argument) is preferred, this can always be added later.

danomatika:
FWIW this change makes sense, ie. 0 as an unknown value etc.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Sep 25, 2023, at 11:00 AM, umläute ***@***.***> wrote:﻿
Merged #2083 into develop.

Issue #635, opened 2019-05-12
feature request - copy to / paste from clipboard


porres:
The idea is to select all of a patch and be able to copy it to the clipboard as a text from the pd file. Conversely, one could copy such a text file and paste it into a Pd patch with a "copy from clipboard" option. 

This would allow you to do things like more easily share patches on the internet as texts, or to copy and paste to different instances/versions of Pd.

mxa:
There is a related comment by @millerpuckette in merge request #1071 https://github.com/pure-data/pure-data/pull/1071#issuecomment-877398575

mxa:
Here is an attempt to implement this with a plugin: https://github.com/eeropic/pd-kopi

porres:
would still like to see this as part of the core. If it works great, please make a PR @eeropic

millerpuckette:
I like Max's idea to have a different "copy" action to spit the copied 
objects as a text patch to the system copy buffer - this solves a 
problem I had - if one copies something large I don't want to make the 
user wait while all the text bubbles up from Pd to the GUI.  (Pd 
probably can't stick things in the copy buffer itself because it's not 
an "app".)  But making it a different "copy" operation 
("copy-to-clipboard" or such) solves that.

(I believe "Paste" doesn't have that problem - the GUI can figure out 
what should be pasted.)

But yes, as Max also mentioned - this could better be done in Pd core, 
not as a GUI plug-in.  So there's no point making a pull request out of 
Max's TCL code.

cheers

Miller

umlaeute:
seems like i am dumb, but i don't fully understand how the problem of "Pd not being an 'app', so it can't stick things directly into the copy buffer itself" is solved with a `copy-to-clipboard` message.

assuming that the `copy-to-clipboard` message is sent from the Pd-GUI to the Pd-core to do the actual work, the core needs to serialize the data and then
- EITHER stuff the serialized data directly into the clipboard by itself
- OR send the serialized data to the GUI which then copies the data to the clipboard (but afaict this is the "wait while all the text bubbles up from Pd to the GUI" solution).

am i missing something?

millerpuckette:
My idea is to have a separate copy operation in which Pd sends a text 
version of the copy buffer up to the GUI which then "copies" it.

I don't know for sure, but I imagine it is hard to get Pd itself to put 
text in the window-system copy buffer.  If in fact it's simple for 
non-window-owning programs to put stuff in copy buffers (on all three 
[...]

PR #2086, opened 2023-09-05
[WIP] Copy-paste patch objects using system clipboard text buffer

eeropic:
Opening wip PR for discussion
Related issue https://github.com/pure-data/pure-data/issues/635

Todo
- [ ] do not use static variables (use gl_privatedata for clipboard length and text)
- [ ] edit commit history so copy/paste functionality are in separate commits, remove unnecessary
- [ ] refactor, DRY, etc..
- [ ] copy: test serialization (escaping special chars in particular) by diffing saved patches and patches serialized to clipboard 
- [ ] memory allocation checks (+error handling)
- [ ] Validate patch data before attempting paste (regex?)
- [x] handle atom width attribute, i.e. `,f 16` in the end of each patch line (atom width in chars)

**Maybe**
- [ ] figure out whether to include #N for new canvas. i.e. on copying, should the copied patch include the canvas in order to be saved as a proper patch, and on pasting, the first #N should be then ignored.
- [ ] Should paste be single action instead of two? Currently there's 2 functional copy buffers (internal and text)

**Testing**
- [x] Test on windows (latest build tested on win10 64bit)
- [x] Test on osx (latest build tested on arm64 m1, osx 13.2.1)
- [x] Test on linux (latest build tested on arm64 debian 11)

**Other**
- [x] escape clipboard string when sending back from tcl to c when pasting
- [x] Text to binbuf chunks - maximum string length is now limited by MAXPDSTRING, which limits the amount of objects you can copy/paste
- [x] split lines to handle longer messages (long arrays for example) through pdsend
- [x] bug: handle double paste on mac, resulting from menu command keybind declaration running the command twice
- [x] handle paste undo

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -38,6 +38,10 @@ struct _instanceeditor
 
 /* positional offset for duplicated items */
 #define PASTE_OFFSET 10
+#define CLIPBOARD_PATCH_TEXT_START 0
+#define CLIPBOARD_PATCH_TEXT_LINE_END 1
+#define CLIPBOARD_PATCH_TEXT_END 2
+#define CLIPBOARD_PATCH_TEXT_LINE_PARTIAL 3
 
 void glist_readfrombinbuf(t_glist *x, const t_binbuf *b, const char *filename,
     int selectem);
@@ -4140,33 +4144,58 @@ static void canvas_paste_from_clipboard_text(t_canvas *x)
     pdgui_vmess("pdtk_get_clipboard_text", "^", x);
 }
 
-void unescape_clipboard_patch(char *str) {
-    char *temp = malloc(strlen(str) + 1);
-    if (!temp) return;
-    int i = 0, j = 0;
-    while (str[i]) {
-        if (str[i] == ';' && str[i+1] == 'n') {
-            temp[j++] = ';';
-            temp[j++] = '\n';
-            i += 2;
-        } else {
-            temp[j++] = str[i++];
-        }
+static t_binbuf *clipboard_patch_bb = NULL;
+
+void canvas_got_clipboard_contents(t_canvas *x, t_floatarg flagf, t_symbol *s) {
+    int flag = flagf;
+    
+    switch (flag) {
+        case CLIPBOARD_PATCH_TEXT_START:
+            if (clipboard_patch_bb) {
+                binbuf_free(clipboard_patch_bb);
+            }
+            clipboard_patch_bb = binbuf_new();
+            break;
+
+        case CLIPBOARD_PATCH_TEXT_LINE_END:
+            if (clipboard_patch_bb) {
+                char* received_data = strdup(s->s_name);
+                t_binbuf *temp_bb = binbuf_new();
+                binbuf_text(temp_bb, received_data, strlen(received_data));
+                binbuf_addsemi(temp_bb);
+                binbuf_add(clipboard_patch_bb, binbuf_getnatom(temp_bb), binbuf_getvec(temp_bb));
+                binbuf_clear(temp_bb);
+                binbuf_free(temp_bb);
+                free(received_data);
+            }
+            break;
+
+        case CLIPBOARD_PATCH_TEXT_LINE_PARTIAL:
+            if (clipboard_patch_bb) {
+                char* received_data = strdup(s->s_name);
+                t_binbuf *temp_bb = binbuf_new();
+                binbuf_text(temp_bb, received_data, strlen(received_data));
+                binbuf_add(clipboard_patch_bb, binbuf_getnatom(temp_bb), binbuf_getvec(temp_bb));
+                binbuf_clear(temp_bb);
+                binbuf_free(temp_bb);
+                free(received_data);
+            }
+            break;
+
+        case CLIPBOARD_PATCH_TEXT_END:
+            if (clipboard_patch_bb) {
+                canvas_dopaste(x, clipboard_patch_bb);
+                binbuf_free(clipboard_patch_bb);
+                clipboard_patch_bb = NULL;
+            }
+            break;
+
+        default:
+            post("Invalid flag received in canvas_got_clipboard_contents.");
+            break;
     }
-    temp[j] = '\0';
-    strcpy(str, temp);
-    free(temp);
 }
 
-void canvas_got_clipboard_contents(t_canvas *x, t_symbol *s) {
-    char* received_data = strdup(s->s_name);
-    unescape_clipboard_patch(received_data);
-    t_binbuf *bb = binbuf_new();
-    binbuf_text(bb, received_data, strlen(received_data));
-    canvas_dopaste(x, bb);
-    binbuf_free(bb);
-    free(received_data);
-}
 
 static void canvas_duplicate(t_canvas *x)
 {
@@ -4962,8 +4991,7 @@ void g_editor_setup(void)
     class_addmethod(canvas_class, (t_method)canvas_motion, gensym("motion"),
         A_FLOAT, A_FLOAT, A_FLOAT, A_NULL);
     class_addmethod(canvas_class, (t_method)canvas_got_clipboard_contents,
-        gensym("got-clipboard-contents"), A_SYMBOL, A_NULL);          
-
+        gensym("got-clipboard-contents"), A_FLOAT, A_SYMBOL, A_NULL);
 /* ------------------------ menu actions ---------------------------- */
     class_addmethod(canvas_class, (t_method)canvas_menuclose,
         gensym("menuclose"), A_DEFFLOAT, 0);
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -4152,31 +4152,20 @@ void canvas_got_clipboard_contents(t_canvas *x, t_floatarg flagf, t_symbol *s) {
             clipboard_patch_bb = binbuf_new();
             break;
 
+        case CLIPBOARD_PATCH_TEXT_LINE_PARTIAL:
         case CLIPBOARD_PATCH_TEXT_LINE_END:
             if (clipboard_patch_bb) {
-                char* received_data = strdup(s->s_name);
                 t_binbuf *temp_bb = binbuf_new();
-                binbuf_text(temp_bb, received_data, strlen(received_data));
-                binbuf_addsemi(temp_bb);
+                binbuf_text(temp_bb, s->s_name, strlen(s->s_name));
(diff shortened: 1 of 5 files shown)


src/g_editor.c
```diff
@@ -3718,6 +3719,37 @@ static void canvas_copy(t_canvas *x)
...
+{
+    t_binbuf *bb = canvas_docopy(x);
+    if (!bb)
+        return;
+    char *buf;
+    int ac;
+    binbuf_gettext(bb, &buf, &ac);
+    char *escaped_buf = escape_clipboard_special_chars(buf);
+    sys_vgui("clipboard clear\n");
+    sys_vgui("clipboard append \"%s\"\n", escaped_buf);
```
umlaeute, line 3747: urgh no.
please use `pdgui_vmess()` instead.

umlaeute (changes requested):
`pdgui_vmess` already takes care of escaping.
just use it.

src/g_editor.c
eeropic: Thanks. Still need to check the other way around, if I can make it work using existing methods when pdsending from tcl

tcl/pd_bindings.tcl
```diff
@@ -54,9 +54,16 @@ proc ::pd_bindings::global_bindings {} {
...
+
+    if {$::windowingsystem eq "aqua"} {
+        set alt "Option"
+    } else {
+        set alt "Alt"
+    }
     bind_capslock all $::modifier-Key a {menu_send %W selectall}
     bind_capslock all $::modifier-Key b {menu_helpbrowser}
     bind_capslock all $::modifier-Key c {menu_send %W copy}
+    bind_capslock all $::modifier-$alt-Key c {menu_send %W copy-to-clipboard-as-text}
```
umlaeute, line 66: the <kbd>Alt</kbd>/<kbd>Option</kbd> is *also* handled in `::pd_bindings::global_bindings`, i think  it should be handled only one (like with  `$::modifier`)

tcl/pd_menus.tcl
```diff
@@ -30,11 +31,14 @@ namespace eval ::pd_menus:: {
...
 proc ::pd_menus::create_menubar {} {
     variable accelerator
+    variable alt
     variable menubar
     if {$::windowingsystem eq "aqua"} {
         set accelerator "Cmd"
+        set alt "Option"
     } else {
         set accelerator "Ctrl"
+        set alt "Alt"
```
umlaeute, line 41: like this.

umlaeute (changes requested):
better but not there yet

eeropic:
Also still need to properly escape brackets, commas, etc when sending back from Tcl. 

Im looking at https://github.com/pure-data/pure-data/blob/c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c/tcl/wheredoesthisgo.tcl#L153 and pd_text.tcl, but not sure if theres already a proc that does this non-destructive (i.e. not removing commas altogether)

src/g_editor.c
```diff
@@ -4123,6 +4135,39 @@ static void canvas_paste(t_canvas *x)
     }
 }
 
+static void canvas_paste_from_clipboard_text(t_canvas *x)
+{
+    pdgui_vmess("pdtk_get_clipboard_text", "^", x);
+}
```
umlaeute, line 4141: i wonder: is there any specific reason for the roundtrip through the core?
(apart from symmetry with the other menu-entries)

eeropic: No - it's leftover from starting work based on existing menu-entries and functions. thanks

src/g_editor.c
```diff
@@ -4135,38 +4139,58 @@ static void canvas_paste(t_canvas *x)
     }
 }
 
-static void canvas_paste_from_clipboard_text(t_canvas *x)
-{
-    pdgui_vmess("pdtk_get_clipboard_text", "^", x);
-}
+static t_binbuf *clipboard_patch_bb = NULL;
```
umlaeute, line 4142: this should go into `(t_glist*)->gl_privatedata`

src/g_editor.c
```diff
@@ -4135,38 +4139,58 @@ static void canvas_paste(t_canvas *x)
...
+                char* received_data = strdup(s->s_name);
+                t_binbuf *temp_bb = binbuf_new();
+                binbuf_text(temp_bb, received_data, strlen(received_data));
+                binbuf_add(clipboard_patch_bb, binbuf_getnatom(temp_bb), binbuf_getvec(temp_bb));
+                binbuf_clear(temp_bb);
+                binbuf_free(temp_bb);
+                free(received_data);
+            }
+            break;
+
```
umlaeute, line 4179: is the `CLIPBOARD_PATCH_TEXT_LINE_PARTIAL` for future use? right now it looks suspiciously identical to `CLIPBOARD_PATCH_TEXT_LINE_END`.

also, i'm not sure why you need to `strdup` the incoming data

src/g_editor.c
```diff
@@ -4962,8 +4986,7 @@ void g_editor_setup(void)
     class_addmethod(canvas_class, (t_method)canvas_motion, gensym("motion"),
         A_FLOAT, A_FLOAT, A_FLOAT, A_NULL);
     class_addmethod(canvas_class, (t_method)canvas_got_clipboard_contents,
-        gensym("got-clipboard-contents"), A_SYMBOL, A_NULL);          
-
+        gensym("got-clipboard-contents"), A_FLOAT, A_SYMBOL, A_NULL);
```
umlaeute, line 4989: if you make this a `A_DEFSYMBOL`, you could get rid of the dummyargument when sending the `START`/`END` commands.

src/g_editor.c
umlaeute: ah, i missed the `binbuf_addsemi()`...

umlaeute:
i think i discovered a bug in `pdgui_vmess()` where `A_COMMA` and a comma.symbol ( `gensym(",")`) are both sent as `,`.

this needs to be fixed before this PR can properly handle appendicles (`, f ...`)

src/g_editor.c
eeropic: I'm doing case fall-through now with conditional addition of semi next

src/g_editor.c
eeropic: Ah nice, didn't know that. DEF as in default? like an empty string passed if not defined?

eeropic: Done,

eeropic:
> i think i discovered a bug in `pdgui_vmess()` where `A_COMMA` and a comma.symbol ( `gensym(",")`) are both sent as `,`.
> 
> this needs to be fixed before this PR can properly handle appendicles (`, f ...`)

Ok, I was wondering about this before.
Now tried printing out on save, and on copy, and can verify.
**Saving:** (printing binbuf before free at canvas_savetofile)
```
#X
text
34
25
look
\,
setting
width
,
f
12
;
```

**Copying (pdgui_vmess()):**
```
#X
text
34
25
look
,
setting
width
,
f
12
;
```

eeropic:
> i think i discovered a bug in `pdgui_vmess()` where `A_COMMA` and a comma.symbol ( `gensym(",")`) are both sent as `,`.
> 
> this needs to be fixed before this PR can properly handle appendicles (`, f ...`)

Even if I have the `, f ...` correctly in text clipboard, it needs to be handled separately (thinking of another flag) as pdsend will split the message / interpret it as 2 messages delimited by comma

src/g_editor.c
```diff
@@ -4143,44 +4144,49 @@ static t_binbuf *clipboard_patch_bb = NULL;
...
             if (clipboard_patch_bb) {
                 binbuf_free(clipboard_patch_bb);
             }
             clipboard_patch_bb = binbuf_new();
             break;
-
         case CLIPBOARD_PATCH_TEXT_LINE_PARTIAL:
         case CLIPBOARD_PATCH_TEXT_LINE_END:
-            if (clipboard_patch_bb) {
+        case CLIPBOARD_PATCH_TEXT_LINE_END_APPENDIX:
```
umlaeute, line 4156: i'm not exactly sure about this.
since a Pd-file can be properly parsed by `binbuf_text()`, even if it contains message delimiting commas (`[foo, bar(`) and gobj attributes (`, f 10`), I do not see why need special handling here.
ideally, the `canvas_got_clipboard_contents()` should get the raw text (as in the clipboard), and that's it (and all the necessary escaping happens in the GUI)

umlaeute: (and sorry for not saying that *before* you started to implement it)

eeropic: Heh no worries. I'm just having trouble transmitting the comma with pdsend, i.e. `something, f32` gets interpreted as 2 messages. So perhaps a special escape sequence for the comma would suffice? So \, gets transmitted as-is, but , gets handled somehow differently (tcl->pd)

umlaeute: that's because a plain comma in a message is a message separator in Pd lingo.
also, a line in a patch file might not be terminated by a semicolon, in which case it is continued on the next line:

```
#X obj 79 136 print hello;
```

is really equivalent to
```
#X
obj
79
136
print
hello;
```

i think this is currently not handled correctly (but haven't actually checked)

umlaeute: i think the line `#X msg 78 108 foo \, bar, f 30;` ought to be transmitted as the string `#X msg 78 108 foo \\, bar\, f 30\;` (note the trailing `\;`)

umlaeute: here's a super stupid (and simple idea):
do not attempt to send the data as a symbol (it's going to pollute the symbol table anyhow), but as a list of numbers (UTF-8-encoded).
This is trivial to reconstruct on the receiving side.

umlaeute: see https://github.com/eeropic-forks/pure-data/pull/1

eeropic: Great stuff!! Just need to implement the chunking, so it doesn't cut off the data in pdsend

umlaeute: Sure. It's just a poc.

I'm pretty sure there is some memory overallocation and what not.
On rethinking, I also would prefer "reset" and "submit" instead of "begin" and "end".

porres:
really hoping this gets done :)

millerpuckette:
I've been thinking about doing this too and am just now looking at this PR.  I would want to figure out what the performance hit is for hitting "copy" on a large object (and having then to dribble it all up to the GUI even if, usually, it's not ever used).  One idea might be to wait and query the user if there would be more than 10K characters or so to send?

porres:
Well, this was already implemented, right? So I guess we can close this one

PR #2114, opened 2023-10-23
document '--enable-debug' configure flag

Lucarda:
document '--enable-debug' configure flag in `INSTALL.txt` and html chapter6

### DIFF
--- a/doc/1.manual/x6.htm
+++ b/doc/1.manual/x6.htm
@@ -118,6 +118,12 @@ midi APIs:            PortMidi</pre>
 # build Pd using a system installed PortAudio
 ./configure --without-local-portaudio</pre>
 
+<p>If you need to run Pd through a debugger (like gdb), you can
+build Pd with debugging symbols using the "--enable-debug" flag.
+
+<pre># build Pd with debugging information
+./configure --enable-debug</pre>
+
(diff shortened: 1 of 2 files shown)


doc/1.manual/x6.htm
```diff
@@ -118,6 +118,25 @@ <h3 id="s6.2">6.2. Autotools Build (recommended)</h3>
 # build Pd using a system installed PortAudio
 ./configure --without-local-portaudio</pre>
 
+<p>The default build will do some moderate optimization.
+This can be tuned with the "--enable-optimizations" flag, which will tell the
+compiler to optimize for speed more aggressively (even at the cost of incorrect
+math operations).
+
+<pre># build with speed optimizations
+./configure --enable-optimizations</pre>
```
umlaeute, line 127: i don't know of any such configure flag.
where did you get it from?

Lucarda:
oops, thanks for checking. While cleaning my branches I came to

https://github.com/pure-data/pure-data/compare/master...Lucarda:pure-data:buildflags/html
        
it seems i did the html based on the txt but this was never implemented.

Lucarda:
how about this on the txt and html:

```
The default build will do some moderate optimization.

If things go wrong and you need to run Pd through a debugger (like gdb), you can
build Pd with debugging symbols using the "--enable-debug" flag.

    # build Pd with debugging information
    ./configure --enable-debug
    
```

?

Issue #2119, opened 2023-10-27
add support for "note off" messages


porres:
[notein]/[noteoff] only deals "note on" messages, which is ok for most cases, but MIDI also has a special "note off" message and this is the only one not supported by Pd (unless you use [midiin]/[midiout]). I'm then proposing adding support for it with a flag in note in/out objects ('-rel' for release maybe). 

The note off message with a note off velocity has been relatively rare for the most part but some new controllers have been supporting it back (there's been some discussion on the pd list about it) and I don't see a good reason not to better support it in Pd. It can play a nice part in designing synths, like being able to set a particular envelope release setting for the note.

It seems that controllers with MPE support are rather likely to also include this. I take this from a MIDI file example that MAX uses in its documentation of MPE objects and also this external that supports note off https://github.com/DanielRudrich/pd_mpe while ignoring many other possible MIDI messages.

Hence, having support for note off can also improve Pd's handling of MPE, allowing users to create abstractions for it without the need of externals like [mpe] referenced above and it could even be an alternate solution for the https://github.com/pure-data/pure-data/issues/2117 issue

porres:
[notein] actually reports "note off" messages, but the velocity is given as '0' regardless of the actual release velocity. This needs to be documented and I'm gonna do it but it narrows down how we could actually implement this.

I'm now thinking of a '-off' flag that gives us an extra outlet/inlet. For [notein] it'd have at least 3 inlets if a channel is set, or 4 if not. 

The extra in/outlet is for release velocity. In case of [notein], a Note Off message with a release velocity of, say, 111, gives us the same value of '0' for the 2nd outlet and comes out as such in the 3rd outlet. A 'note on' message gives us '0' release velocity.

My 1st idea would be to make the object only output release velocity with such a flag, so no new number of in/outlets, so the object without a flag wouldn't respond at all for note off messages and it would be easy to manage things with 2 instances of the object. This is not a pretty idea as someone would obviously always need 2 instances of this object, one just for note on and another for note off, and now it doesnt make much sense because the object without a flag also does output data for note offs anyway. Hence, as I say, this seems to narrow down to the other above solution that I gave.

Spacechild1:
> I'm now thinking of a '-off' flag that gives us an extra outlet/inlet. For [notein] it'd have at least 3 inlets if a channel is set, or 4 if not.

[...]

PR #2124, opened 2023-11-03
add support for MIDI Note Off messages

porres:
closes https://github.com/pure-data/pure-data/issues/2119

as in the above discussion, the strategy was to add a '-off' flag to force the [notein] and [noteout] objects to only receive and output Note Off messages

### DIFF
--- a/src/s_midi.c
+++ b/src/s_midi.c
@@ -293,7 +293,7 @@ typedef struct midiparser
 void inmidi_realtimein(int portno, int cmd);
 void inmidi_byte(int portno, int byte);
 void inmidi_sysex(int portno, int byte);
-void inmidi_noteon(int portno, int channel, int pitch, int velo);
+void inmidi_noteon(int portno, int channel, int pitch, int velo, int flag);
 void inmidi_controlchange(int portno, int channel, int ctlnumber, int value);
 void inmidi_programchange(int portno, int channel, int value);
 void inmidi_pitchbend(int portno, int channel, int value);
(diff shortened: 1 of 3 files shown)


porres:
Ok, I knew I had to also update some libpd stuff, but didn't want to do it right away and test things first. I was able to compile and test, and now I see the libpd test is failng :) 

I'm not comforable in messing with the libpd functions but I can try

I was gonna ask @danomatika to give me some warnings and guidance anyway

danomatika:
libpd relies on the `outmode_noteon` and `inmidi_noteon` functions, so changing their function arguments breaks libpd's current usage.

It essentially
* implements `outmidi_noteon` [here](https://github.com/pure-data/pure-data/blob/master/src/s_libpdmidi.c)
* allows setting a [custom noteon "hook"](https://github.com/pure-data/pure-data/blob/master/src/z_libpd.h#L382) aka user-provided function which is passed messages from libpd's `outmidi_noteon`
* the [`libpd_noteon` function](https://github.com/pure-data/pure-data/blob/master/src/z_libpd.h#L327) passes messages into pd's `inmidi_noteon` function

*All* of these functions rely on the current number and type of arguments, hence the build errors.

---

Without reading in detail, do I assume the MPE protocol requires making a distinction between `noteon vel 0` and `noteoff`? If so, and built-in support is required, then I would recommend an internal noteoff message and move the type filtering logic into the `[noteon]` and/or add a `[noteoff]` object.

danomatika:
Also, `libpd_noteonhook` and `libpd_noteon` have not changed since the project was originally released, circa 2010. Changing them now involves lots of pain to both projects and the various language wrappers which use them. It's not impossible technically but not a likable change for and specific feature addition...

porres:
> If so, and built-in support is required, then I would recommend an internal noteoff message and move the type filtering logic into the [noteon] and/or add a [noteoff] object.

I am doing some logic in both [noteon] and [noteoff] objects, but I guess I know what you mean, adding a new internal message, right?

And with such a new message we can prevent all the pain in the ass and issues with libpd, right? Please shed me some more light, give me some hints and clues, and I'll try to head into the right direction

Issue #2122, opened 2023-11-02
pd_compatibilitylevel not properly exported


umlaeute:
building Pd on systems that only export symbols that are *explicitly* exported (possibly on Windows; but see #569) will miss the `pd_compatibilitylevel` variable.

that's because this variable is declared `extern`, but not `EXTERN`:

https://github.com/pure-data/pure-data/blob/0259d00373b9e6f3bc1d76eebc5b243283ca7063/src/m_pd.h#L15

i'd suggest using the `EXTERN` specifier (and  since `EXTERN` is not yet defined at the current position, move the declaration of the variable down to `sys_getversion()`).

i stumbled upon this when
- building `libpd` with hidden symbols (#569), as is the Default for the Debian *libpd0* package
- trying to load the `sigmund~` external from this libpd

i haven't really checked why Windows people can load `sigmund~` at all (probably data does not need the `__declspec(dllexport)` specifier)

Spacechild1:
> i haven't really checked why Windows people can load sigmund~ at all 

Unlike MSVC, MinGW can automatically import symbols from DLLs - without the need of import libraries and `__declspec(dllimport)`. In the old days, you had to explicitly build with `--enable-auto-import`, but nowadays it is enabled by default.

Also, MinGW exports all symbols by default - unless *at least one* function/data is explicitly exported with `__declspec(dllexport)`. However, it seems that you can force exporting all symbols with `--export-all-symbols` - which is what we're doing in our build system. MSVC *always* hides all symbols by default and only exports those marked with `__declspec(dllexport)`.

> (probably data does not need the __declspec(dllexport) specifier)

AFAIK, functions and data behave the same in this respect.

umlaeute:
thanks for the clarifications.

i think we should strive to get the decorators correct and not rely on some magic linker flags and specific toolchains.

umlaeute:
weirdly enough, while Pd-0.55 has many many `EXTERN` symbols fixed, the `pd_compatibilitylevel` was omitted...

Issue #1018, opened 2020-05-18
[savepanel] doesn't remember last file location


Spacechild1:
unlike `[openpanel]`, `[savepanel]` doesn't remember the last file location, i. e. it will always start in the home directory resp. current working directory which can be quite annoying.

tested on Windows 7 with Pd 0.50-2 and latest master (285c53ff57621801e794d1e0c13425193bde3d61)

umlaeute:
these days you can build this with:

```
[bang(
|        +----------------+
[symbol .]                |
|                         |
[savepanel]               |
|                         |
[t a a]                   |
|     [file splitname]    |
|     [unpack s]          |
|     |                   |
|     +-------------------+
|
```

is this good enough?

Spacechild1:
> is this good enough?

IMO, no. Remembering the last save location is a standard feature on every desktop app. Users shouldn't need to build this themselves.

danomatika:
It should remember. Check the main proc in wheredoesthisgo.tcl and it should set it there when the panel is made.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Dec 14, 2023, at 10:28 AM, Christof Ressi ***@***.***> wrote:﻿

is this good enough?

IMO, no. Remembering the last save location is a standard feature on every desktop app. Users shouldn't need to build this themselves.

Spacechild1:
> It should remember. Check the main proc in wheredoesthisgo.tcl and it should set it there when the panel is made.

With "should" do mean "it already does it" or "this needs to be done"?

At least with Pd 0.54.1, `[savepanel]` does indeed start in the folder of the most recently saved *patch*, but it does not update the path itself (which it really should).

danomatika:
I don't remember exactly. I updated the default path handling to use the Pd do edit when the plugin is enabled. At the time I think I wanted to double check that it worked "as expected" like  stove dialogs but I think I ran out of time.My suggestion is to double check, really. Check what the proc does and where it is called in the tcl. That would be helpful.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Dec 14, 2023, at 10:44 AM, Christof Ressi ***@***.***> wrote:﻿

It should remember. Check the main proc in wheredoesthisgo.tcl and it should set it there when the panel is made.

With "should" do mean "it already does it" or "this needs to be done"?
[...]

PR #2047, opened 2023-07-05
stable development branch

umlaeute:
(just merged, and here it is again)

this is the permanent branch develop that only gets curated, small, no-brainer changes that can be easily merged into master at any time.
(as a general rule-of-thumb we shouldn't include changes to Pd-files in this PR, as it is just too easy to end up with unresolvable file conflicts)

it supersedes https://github.com/pure-data/pure-data/pull/2027

---

### core
- Closes: https://github.com/pure-data/pure-data/issues/2045
- adc~: both arguments to "-m" flag are supposed to be optional
- Closes: https://github.com/pure-data/pure-data/issues/2071 
- Closes: https://github.com/pure-data/pure-data/issues/2101

### GUI
- Closes: https://github.com/pure-data/pure-data/issues/2062
- Closes: https://github.com/pure-data/pure-data/issues/2044

### libpd
- Closes: https://github.com/pure-data/pure-data/issues/2046

PR #2125, opened 2023-11-03
stable development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/2047

---

### Core
- Closes: https://github.com/pure-data/pure-data/issues/2122
- Closes: https://github.com/pure-data/pure-data/issues/2136
- Closes: https://github.com/pure-data/pure-data/issues/2157
- Closes: https://github.com/pure-data/pure-data/issues/2186
- Closes: https://github.com/pure-data/pure-data/issues/2192
- Fix "channels" message for `[send~]` and `[catch~]` (c863bbe)
- Closes: https://github.com/pure-data/pure-data/issues/2199
- Closes: https://github.com/pure-data/pure-data/issues/2187

### GUI
- Closes: https://github.com/pure-data/pure-data/issues/751
- Closes: https://github.com/pure-data/pure-data/issues/1018
- Closes: https://github.com/pure-data/pure-data/issues/2168
- Closes: https://github.com/pure-data/pure-data/issues/2174
- Closes: https://github.com/pure-data/pure-data/issues/2175
- Closes: https://github.com/pure-data/pure-data/issues/2195
- Closes: https://github.com/pure-data/pure-data/issues/2189
- Closes: https://github.com/pure-data/pure-data/issues/2201
- Closes: https://github.com/pure-data/pure-data/issues/2176
- Closes: https://github.com/pure-data/pure-data/issues/2209

#### windows installer
- Closes: https://github.com/pure-data/pure-data/issues/2128

#### linux packages
- Update AppStream metainfo (include issue-tracker, translation page, git-repository and credit the devs)

### DIFF
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -571,13 +571,11 @@ void class_addcreator(t_newmethod newmethod, t_symbol *s,
     va_list ap;
     t_atomtype vec[MAXPDARG+1], *vp = vec, argtype = type1;
     int count = 0;
-    if(!argtype) argtype = A_NULL;
 
     va_start(ap, type1);
     while(argtype != A_NULL && count < MAXPDARG) {
         vec[count++] = argtype;
         argtype = va_arg(ap, t_atomtype);
-        if(!argtype) argtype = A_NULL;
     }
     va_end(ap);
         /* the last argument must be A_NULL */
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -203,7 +203,7 @@ void pdinstance_free(t_pdinstance *x)
     pd_setinstance(x);
     sys_lock();
     pd_globallock();
-    
+
     instanceno = x->pd_instanceno;
     inter = x->pd_inter;
     canvas_suspend_dsp();
@@ -1252,4 +1252,3 @@ int class_getdspflags(const t_class *c)
             (c->c_nopromotesig ? CLASS_NOPROMOTESIG : 0) |
             (c->c_nopromoteleft ? CLASS_NOPROMOTELEFT : 0) );
 }
-
--- a/src/m_class.c
+++ b/src/m_class.c
@@ -142,7 +142,7 @@ static void class_addmethodtolist(t_class *c, t_methodentry **methodlist,
 }
 
 #ifdef PDINSTANCE
-EXTERN void pd_setinstance(t_pdinstance *x)
+void pd_setinstance(t_pdinstance *x)
 {
     pd_this = x;
 }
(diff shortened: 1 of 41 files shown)


src/m_class.c
```diff
@@ -569,28 +569,27 @@ void class_addcreator(t_newmethod newmethod, t_symbol *s,
     t_atomtype type1, ...)
 {
     va_list ap;
-    t_atomtype vec[MAXPDARG+1], *vp = vec;
+    t_atomtype vec[MAXPDARG+1], *vp = vec, argtype = type1;
     int count = 0;
-    *vp = type1;
+    if(!argtype) argtype = A_NULL;
```
Spacechild1, line 574: What is this supposed to do? `A_NULL` is defined as 0, so this line seems entirely redundant.

src/m_class.c
```diff
@@ -569,28 +569,27 @@ void class_addcreator(t_newmethod newmethod, t_symbol *s,
...
-                      MAXPDARG);
-            break;
-        }
-        vp++;
-        count++;
-        *vp = va_arg(ap, t_atomtype);
+    while(argtype != A_NULL && count < MAXPDARG) {
+        vec[count++] = argtype;
+        argtype = va_arg(ap, t_atomtype);
+        if(!argtype) argtype = A_NULL;
```
Spacechild1, line 580: same as above

src/m_class.c
umlaeute: a well, it is not *defined* as `0`. it just happens to be the first enum in the list (thus it is *implicitly* `0`).
the idea of this line (and the one below) is to make this explicit.

i figure any optimizing compiler would detect the same-ness and eliminate the redundant call.

src/m_class.c
umlaeute: same as above  :-)

src/m_class.c
Spacechild1: > it just happens to be the first enum in the list

That means it *is* defined as 0, you don't need to spell it out.

> the idea of this line (and the one below) is to make this explicit.

But why? In the while loop you check for `argtype != A_NULL`, that is all we are interested in. We don't even need to care about the numeric value of `A_NULL`.

> i figure any optimizing compiler would detect the same-ness and eliminate the redundant call.

I'm not worried about the optimizer :) This line just caught me off-guard and I left me scratching my head for a while.

I guess you wanted to say "instead of `A_NULL` you can pass `0`. which are really the same thing". In that case, I would prefer a simple comment :)

umlaeute: indeed, that would be the better solution.
i thought about it, but somehow did not want to change any headers in the commit.

Issue #2136, opened 2023-11-17
`[switch~]` + reblocking out of sync


Spacechild1:
If a subpatch is temporarily switched-off *and* reblocked, inlet/outlet buffering and DSP can easily go out of sync.

Here's a minimal patch that demonstrates the problem: 
[switch-reblock-bug.zip](https://github.com/pure-data/pure-data/files/13397572/switch-reblock-bug.zip)

![grafik](https://github.com/pure-data/pure-data/assets/16126632/3b93d4e6-16ab-4e6d-9cbd-6e714b630609)

Explanation:

If a subpatch is reblocked to a larger blocksize, the inlets/outlets are buffered and DSP is triggered everytime the inlet buffers are ready. A switched-off subpatch jumps over the actual DSP chain, past the block epilog, but it still performs the inlet prolog (which is responsible for filling the inlet buffers) and the outlet epilog (which is responsible for filling the outlets).

When the subpatch is switched on again, the phases of the inlet/outlet buffers and the block prolog/epilog are now very likely out of sync!

A trivial workaround is to put the `switch~` one level above the reblocking (using `block~`), so that the inlet and outlet buffering is switched off together with the DSP.

~~I'm not sure if there is an easy fix for this... In the meantime we could at least add a warning to the help file.~~

Here's a fix: https://github.com/pure-data/pure-data/pull/2137

Spacechild1:
Maybe it would be enough to simply bash the phase when switching a subpatch on again:
```
static void block_float(t_block *x, t_floatarg f)
{
    if (x->x_switched)
    {
        x->x_switchon = (f != 0);
            /* bash phase to keep in sync with inlets and outlets */
        if (x->x_switchon && x->x_reblock)
            x->x_phase = THIS->u_phase & (x->x_period - 1);
    }
}
```

Will give it a try.

PR #2137, opened 2023-11-18
Fix `switch~` + reblocking

Spacechild1:
When a switched-off canvas is reblocked and switched on again, we make sure that the block prolog phase is in sync with the inlet prolog and outlet epilog phase.

Closes https://github.com/pure-data/pure-data/issues/2136

### DIFF
--- a/src/d_ugen.c
+++ b/src/d_ugen.c
@@ -1063,7 +1063,7 @@ void ugen_done_graph(t_dspcontext *dc)
     t_dspcontext *parent_context = dc->dc_parentcontext;
     t_float parent_srate;
     int parent_vecsize, parent_overlap;
-    int period, frequency, phase, calcsize;
+    int period, frequency, calcsize;
     t_float srate;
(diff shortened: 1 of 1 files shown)


umlaeute:
should(n't) this go directly into `develop`?

Spacechild1:
Ok from my side!

Spacechild1:
I think you just accidentally merged into master!

umlaeute:
yikes!

Spacechild1:
(We should really restrict access to `master` to Miller, if that is somehow possible.)

umlaeute:
yes. i thought it was already setup like this.

afaict, the problem is that administrator (like myself) somehow implicitly overrule these restrictions.

I've setup my local clones with extra protection so I don't accidentally push to `master`, but that obviously doesn't work if i just hit the `Rebase & Merge` button in the github webinterface.

umlaeute:
anyhow, i hope i undid my error (hopefully nobody synched `master` in the meantime)

Issue #2142, opened 2023-11-23
CLASS_SIGNALMAININ should ensure that the variable has the correct type


Spacechild1:
The `CLASS_SIGNALMAININ` macro expects a variable of type `t_float`. If you accidentally pass a variable of the wrong type (e.g. `double` in single-precision Pd), the compiler silently accepts it and you get garbage at runtime.

With C11 we could actually check the type/size with `_Static_assert` (https://en.cppreference.com/w/c/language/_Static_assert). I think modern toolchains default to C11< anyway, but we can use conditional compilation to keep C89 compatibility.

Also, we could use `offsetof` (https://en.cppreference.com/w/c/types/offsetof) to calculate the member offset, instead of the terrible `NULL` pointer cast trick. In fact, `offsetof` is part of the C89 standard!

PR #2143, opened 2023-11-23
make CLASS_MAINSIGNALIN more type safe

Spacechild1:
The `CLASS_SIGNALMAININ` macro implicitly expects a variable of type `t_float`. If you accidentally pass a variable of the wrong type, the compiler silently accepts it and you get garbage at runtime.

Here's a real life example: https://lists.puredata.info/pipermail/pd-dev/2023-11/023565.html

This PR adds a `STATIC_ASSERT` macro for cross-platform static assertion support and then uses it in the macro definition of `CLASS_SIGNALMAININ` to make sure that the type has expected size.

Also, I use `offsetof`, instead of the ugly NULL pointer casting trick, to get the member offset.

Closes https://github.com/pure-data/pure-data/issues/2142

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -59,9 +59,16 @@ extern int pd_compatibilitylevel;   /* e.g., 43 for pd 0.43 compatibility */
 #define ATTRIBUTE_FORMAT_PRINTF(a, b)
 #endif
 
-#if !defined(_SIZE_T) && !defined(_SIZE_T_)
-#include <stddef.h>     /* just for size_t -- how lame! */
+#if __STDC_VERSION__ >= 201112L
+#include <assert.h>
+#define STATIC_ASSERT _Static_assert
+#elif __cplusplus >= 201103L
+#define STATIC_ASSERT static_assert
+#else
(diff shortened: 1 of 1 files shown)


Spacechild1:
In fact, the NULL pointer trick of `CLASS_SIGNALMAININ` triggers a warning on clang (at least with C++):
```
/Users/christofressi/Repos/aoo/pd/src/aoo_send~.cpp:1108:5: warning: performing pointer subtraction with a null pointer may have undefined behavior [-Wnull-pointer-subtraction]
    CLASS_MAINSIGNALIN(aoo_send_class, t_aoo_send, x_f);
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/Users/christofressi/Repos/pure-data/src/m_pd.h:529:59: note: expanded from macro 'CLASS_MAINSIGNALIN'
    class_domainsignalin(c, (char *)(&((type *)0)->field) - (char *)0)
                                                          ^ ~~~~~~~~~
```

The use of `offsetof` would fix this.

PR #2148, opened 2023-11-30
document default flag values in [soundfiler] and letting it save files

porres:
closes https://github.com/pure-data/pddp/issues/177

One big issue is that the example used a '/tmp/' directory to write to, but this directory doesn't exist, so noting got saved. I removed this temp file so now it gets saved on the docs folder.

I also documented default flag values for writting

I think it would best to also use [openpanel] and [savepanel] in the examples by the way.


umlaeute:
> I removed this temp file so now it gets saved on the docs folder.

this is *not* the proper way, as you might not have write access to the `docs/` folder.
and if you *do* have write access, then saving files to this folder will pollute it  with temporary files.

the `/tmp`folder might not exist, but at least it conveys the idea of saving a file, without spoiling an existing directory.

please revert this change.

porres:
I do think it is desireable to have the help file allowing us to save files so the user can test things, and I also think it is confusing and might through some people off that this was supposed to get saved somewhere. So let me find a way to work around this instead of just reverting back.

> I think it would best to also use [openpanel] and [savepanel] in the examples by the way.

I guess [savepanel] is the way to go, I was already using them both by the way

PR #2158, opened 2023-12-11
Safe cross-platform `snprintf` and `vsnprintf` alternative

Spacechild1:
Currently, we define `snprintf` as `_snprintf` when building with MSVC. However, `_snprintf` is *not* a drop-in replacement for `snprintf`! Most importantly, it does not null-terminate the output buffer on truncation! Also, it outputs -1 instead of the number of required bytes.

Note that the UCRT (universal C runtime), provided for Windows 10 and above, actually contains a C99-conforming `snprintf` function. When targetting older systems, MSVC would just define `snprintf` as `_snprintf`. See https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/snprintf-snprintf-snprintf-l-snwprintf-snwprintf-l?view=msvc-170.

On MinGW the situation is also a bit complicated, as it traditionally uses the old `msvcrt.dll`, which only has the non-conforming `_snprintf`. Some versions define `snprintf` as an alias for `_snprintf`, but other (more recent?) versions seem to ship their own C99-conforming `snprintf` implementation.

Instead of dealing with all of this incompatibilities, I would suggest to introduce a safe cross-platform alternative: `pd_snprintf` and `pd_vsnprintf`. On non-Windows systems this would just be a macro for `snprintf` resp. `vsnprintf`. On Windows, `pd_vsnprintf` calls `_vsnprintf`; if the function fails (i.e. returns -1), we obtain the required number of bytes with `_vscprintf` and null-terminate the output buffer.

In this PR, I replaced all occurances of `snprintf`/`vsnprintf` with `pd_snprintf`/`pd_vsnprintf`.

---

Currently, this is only meant for internal use, but it may be moved to `m_pd.h` if desired.

### DIFF
--- a/extra/pd~/pd~.c
+++ b/extra/pd~/pd~.c
@@ -34,7 +34,6 @@ typedef int socklen_t;
 
 #ifdef _MSC_VER
 #pragma warning (disable: 4305 4244)
-#define snprintf _snprintf
 #define stat _stat
 #endif
 
@@ -87,6 +86,13 @@ void critical_exit(int z);
 #include "s_stuff.h"
 static t_class *pd_tilde_class;
 #define PDERROR pd_error(x,
+#else
+#ifdef _MSC_VER
+/* Sorry, MSP! */
+#define pd_snprintf _snprintf
+#else
+#define pd_snprintf snprintf
+#endif
 #endif
 
 #if defined(__x86_64__) || defined(_M_X64)
@@ -559,15 +565,15 @@ static void pd_tilde_dostart(t_pd_tilde *x, const char *pddir,
     sprintf(ninsigstr, "%d", ninsig);
     sprintf(noutsigstr, "%d", noutsig);
     sprintf(sampleratestr, "%f", (float)samplerate);
-    snprintf(tmpbuf, MAXPDSTRING, "%s/bin/pd" EXTENT, pddir);
+    pd_snprintf(tmpbuf, MAXPDSTRING, "%s/bin/pd" EXTENT, pddir);
     sys_bashfilename(tmpbuf, pdexecbuf);
     if (stat(pdexecbuf, &statbuf) < 0)
     {
-        snprintf(tmpbuf, MAXPDSTRING, "%s/../../../bin/pd" EXTENT, pddir);
+        pd_snprintf(tmpbuf, MAXPDSTRING, "%s/../../../bin/pd" EXTENT, pddir);
         sys_bashfilename(tmpbuf, pdexecbuf);
         if (stat(pdexecbuf, &statbuf) < 0)
         {
-            snprintf(tmpbuf, MAXPDSTRING, "%s/pd" EXTENT, pddir);
+            pd_snprintf(tmpbuf, MAXPDSTRING, "%s/pd" EXTENT, pddir);
             sys_bashfilename(tmpbuf, pdexecbuf);
             if (stat(pdexecbuf, &statbuf) < 0)
             {
@@ -579,7 +585,7 @@ static void pd_tilde_dostart(t_pd_tilde *x, const char *pddir,
         /* check that the scheduler dynamic linkable exists w either suffix */
     for(dllextent=get_dllextent(); *dllextent; dllextent++)
     {
-      snprintf(tmpbuf, MAXPDSTRING, "%s/pdsched%s", schedlibdir, *dllextent);
+      pd_snprintf(tmpbuf, MAXPDSTRING, "%s/pdsched%s", schedlibdir, *dllextent);
       sys_bashfilename(tmpbuf, schedbuf);
       if (stat(schedbuf, &statbuf) >= 0)
         goto gotone;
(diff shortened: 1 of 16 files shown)


src/m_private_utils.h
```diff
@@ -108,9 +108,15 @@
...
-# define snprintf _snprintf
+    /* safe cross-platform alternatives to snprintf and vsnprintf.
+    NOTE: if the buffer is too small, both functions return a value >= 'size';
+    however, this is not necessarily the number of required bytes! */
+#ifdef _WIN32
+int pd_snprintf(char *buf, size_t size, const char *format, ...);
+int pd_vsnprintf(char *buf, size_t size, const char *format, va_list argptr);
+#else
+# define pd_snprintf snprintf
+# define pd_vsnprintf vsnprintf
```
umlaeute, line 119: hmm, i'm not sure whether `m_private_utils.h` is the right place for this (scratch this: i'm sure it is not the right place), if the symbol must be exported.
`m_imp.h` (or the obvious `s_stuff.h`) would be better choices.
also, it should be marked `EXTERN`.

but if we export the symbol, then it is part of the (semi-private) API, and I think the API should be the same on all platforms. so I would do this instead:

```C
EXTERN int pd_snprintf(char *buf, size_t size, const char *format, ...);
EXTERN int pd_vsnprintf(char *buf, size_t size, const char *format, va_list argptr);
#ifndef _WIN32
# define pd_snprintf snprintf
# define pd_vsnprintf vsnprintf
#endif
```

src/s_print.c
```diff
@@ -14,6 +14,42 @@
...
+    int ret;
+    va_list ap;
+    va_start(ap, format);
+    ret = pd_vsnprintf(buf, size, format, ap);
+    va_end(ap);
+    return ret;
+}
+
+#endif
+
```
umlaeute, line 52: given my comment of having the same API for Windows and non-Windows, we would also need (thin) implementations of `pd_vsnprintf()` and `pd_snprintf()` on systems that do support `snprintf()`.

even if we typically would use `snprintf()` directly (by means of the `#define` in the header)

extra/pd~/pd~.c
```diff
@@ -87,6 +86,19 @@ void critical_exit(int z);
 #include "s_stuff.h"
 static t_class *pd_tilde_class;
 #define PDERROR pd_error(x,
+/* see m_private_utils.h */
+#ifdef _WIN32
+int pd_snprintf(char *buf, size_t size, const char *format, ...);
```
umlaeute, line 91: we must not use explicitly *private* symbols in externals.

umlaeute (changes requested):
while i understand that you put the `pd_snprintf()` into `m_private_utils.h` i strongly believe that this is the wrong place for declaring a symbol that needs to be exported.

src/m_private_utils.h
Spacechild1: > hmm, i'm not sure whether m_private_utils.h is the right place for this (scratch this: i'm sure it is not the right place), if the symbol must be exported.

But the symbol must/should *not* be exported resp. marked `EXTERN`! For now it is meant to be a private helper function. I am not sure about `m_imp.h`, as it seems to be the place for Pd implementation details. Since `pd_snprintf` is a pure utility function, `m_private_utils.h` looks like the right place to me...

Spacechild1: Ah well, I guess, I see your point. Since I also use it in `pd~`, the symbol needs to be public. After all, we might stop exporting all symbols in the future...

umlaeute: yes.

if it wasn't for externals (like `pd~`) then `m_private_utils.h` would of course be fine.

Spacechild1: I think `s_stuff.h` is the better place.

However, we only really need the symbol for Windows, so the following should be ok:
```
#ifdef _WIN32
EXTERN int pd_snprintf(char *buf, size_t size, const char *format, ...);
EXTERN int pd_vsnprintf(char *buf, size_t size, const char *format, va_list argptr);
#else
# define pd_snprintf snprintf
# define pd_vsnprintf vsnprintf
#endif
```

umlaeute: i know that it is only required on Windows, but i don't think this is a reason to have symbols exported on one platform but not on another one.

Spacechild1: Why not? Why should we export a symbol that we don't even need?

If we export `pd_snprintf` on all platforms, I would rather get rid of the macros alltogether. On advantage would be that non-Windows platforms wouldn't need to include `<stdio.h>`, so the behavior would be more consistent. What do you think?

umlaeute:
Yes that's better than having an asymmetric API.

(We could still use a macro in performance-critical code, if we wanted to...)

src/m_private_utils.h
umlaeute: Yes, that's better than having an asymmetric API.

(We could still use a macro in performance-critical code, if we wanted to...)

Spacechild1:
I just figured out how to get the required number of bytes on Windows (`_vscprintf`). Now `pd_snprintf` should really behave the same as `snprintf` :)

umlaeute (approved):

src/s_stuff.h
```diff
@@ -423,3 +423,7 @@ struct _instancestuff
  * 'srclen' can be 0, in which case the 'src' string must be 0-terminated.
  */
 EXTERN char*pdgui_strnescape(char* dst, size_t dstlen, const char*src, size_t srclen);
+
+/* safe cross-platform alternatives to snprintf and vsnprintf. */
+int pd_snprintf(char *buf, size_t size, const char *fmt, ...);
```
umlaeute, line 428: oops i've missed this: the new functions must be declare `EXTERN` if they are to be used in externals (like `[pd~]`)

umlaeute (changes requested):
apart from the missing `EXTERN` in the header, this looks fine.

src/s_stuff.h
Spacechild1: D'oh. Will fix right now.

umlaeute (approved):

extra/pd~/pd~.c
```diff
@@ -87,6 +86,13 @@ void critical_exit(int z);
 #include "s_stuff.h"
 static t_class *pd_tilde_class;
 #define PDERROR pd_error(x,
+#else
+#ifdef _MSC_VER
+/* Sorry, MSP! */
+#define pd_snprintf _snprintf
```
giuliomoro, line 92: why does this require a macro and it cannot use the definition in s_stuff.h and the implementation in s_print.c?

giuliomoro (commented):
I would have preferred to see a dedicated commit with an automated substition, e.g.: `for file in *.c *.h; do echo $file; gsed -i "s/\<\([v]*snprintf\)\>/pd_\1/g" $file; done` in order to minimise the effort of the reviewer.
Left two more comments

Spacechild1:
>  would have preferred to see a dedicated commit with an automated substition, e.g.: for file in *.c *.h; do echo $file; gsed -i "s/\<\([v]*snprintf\)\>/pd_\1/g" $file; done in order to minimise the risk of errors and make reviewing easier.

The substituation has been done automatically (with the help of my IDE), so you just want me to break out the implementation of `pd_snprintf`/`pd_vsnprintf` into a separate commit? Fair enough!

extra/pd~/pd~.c
Spacechild1: because this is the Max part :) the macro is the exact thing that we used to do throughout the Pd codebase.

giuliomoro: Uh makes sense, `pd~` always gets me. At first I thought you were leaving a comment there for MIller. 🤦  But clearly I didn't look at more context.

src/s_print.c
```diff
@@ -14,6 +14,58 @@
...
+}
+
+#else
+
+int pd_vsnprintf(char *buf, size_t size, const char *fmt, va_list argptr)
+{
+    return vsnprintf(buf, size, fmt, argptr);
+}
+
+int pd_snprintf(char *buf, size_t size, const char *fmt, ...)
```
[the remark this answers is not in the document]
Spacechild1: Hah, good catch! The Unix version actually calls `vsnprintf`, but you are completely right that we can just call `pd_vsnprintf` on all platforms.

extra/pd~/pd~.c
Spacechild1: > At first I thought you were leaving a comment there for MIller.

:-D

giuliomoro:
Another thought: it would be nice to be able to detect and prevent the use of plain `snprintf` and `vsnprintf` in the codebase (outside this instance, that is), is this the sort of thing that could be automated in the checks and/or tests?

Spacechild1:
> I think it would help the reviewer to know that it's been done automatically

Well, the PR description says:

> In this PR, I replaced all occurances of snprintf/vsnprintf with pd_snprintf/pd_vsnprintf.

But you are right that the large change set makes it a bit hard to find the actual implementation of these new functions.

Spacechild1:
> Another thought: it would be nice to be able to detect and prevent the use of plain snprintf and vsnprintf in the codebase (outside this instance, that is), is this the sort of thing that could be automated in the checks and/or tests?

Good point! I guess we'd need a linter for that.

PR #2162, opened 2023-12-28
m_pd.h: fix thread-local support for clang on windows

jcelerier:
This fixes the following error when linking against libpd-multi.dll: 

    ld.lld: error: unable to automatically import from pd_this with relocation type IMAGE_REL_AMD64_SECREL in CMakeFiles/pdtest_gui.dir/samples/c/pdtest_gui/pdtest_gui.c.obj

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -969,7 +969,7 @@ EXTERN void pdinstance_free(t_pdinstance *x);
 #endif /* PDINSTANCE */
 
 #if defined(PDTHREADS) && defined(PDINSTANCE)
-#ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


Spacechild1:
Actually, I couldn't reproduce the linker error here on Windows 10 with Msys2/clang64. `pdtest_gui` compiles and links just fine...

Spacechild1:
IMO, https://github.com/pure-data/pure-data/pull/2165 is the actual fix for your problem. Would it be ok if I close this PR?

jcelerier:
absolutely fine with me!

jcelerier:
(although I disagree that clang should behave like GCC, clang does in general a much better work than GCC to match what the native windows toolchain does and GCC should align :))

PR #2172, opened 2024-01-15
Fix typos in German translation, de.po

aaaaalbert:
I've fixed a few small typos.

(If you insist, I can try and go through Weblate instead of this "direct" PR :-)

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -53,7 +53,7 @@ msgid "Undo cut"
 msgstr "Rückgängig: Ausschneiden"
 
 msgid "Undo disconnect"
-msgstr "Rückgängig: Verbindung Trennen"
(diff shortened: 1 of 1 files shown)


porres:
> (If you insist, I can try and go through Weblate instead of this "direct" PR :-)

please do so, it is much more preferable

umlaeute:
it doesn't really matter as long as:
- the translations stay valid and consitent
- we can merge them into the `l10n` branch (because that it where we want to track the translations)

using weblate automatically ensures this, but if you're savvy enough to  not create additional work (like @aaaaalbert obviously is :wink: ), then manual changes are fine as well.

aaaaalbert:
Thanks :smile:

PR #2183, opened 2024-01-28
abort moving a modified abstraction after user refuses to discard changes

### LINKED ISSUE #2176: trying to move a modified abstraction can lock you into an infinite loop
steps to reproduce:

- create a patch containing an abstraction, save the patch.
- modify the abstraction (add, delete or move an object within it). The abstraction is now marked as "dirty", and an asterisk is appended to its window title. Don't save the abstraction.

- in the parent patch, click on the (modified) abstraction, move it, and try to place it somewhere by clicking again;
- Pd then opens a dialog asking _"Discard changes to [the abstraction name].pd? Yes/No"_.
- If you tap "No", you are returned to the parent patch, however the abstraction is still attached to the mouse pointer, so your only choice is to click somewhere to try again to place the abstraction; you are then returned to the previous step, and again and again, and you end up dying from lack of sleep or food.
- If you tap "Yes", then you can at last drop off the abstraction (which was still attached to the pointer), but you lose the information that it was modified.

Ant1r:
fixes #2176

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2920,8 +2920,10 @@ void canvas_mouseup(t_canvas *x,
             t_glist *gl2;
                 /* first though, check we aren't an abstraction with a
                    dirty sub-patch that would be discarded if we edit this. */
-            if (canvas_undo_confirmdiscard(g))
+            if (canvas_undo_confirmdiscard(g)) {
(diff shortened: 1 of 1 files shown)


Ant1r:
closed in favor of #2200.

PR #2200, opened 2024-02-21
ask whether changes should be discarded when typing into a modified abstraction

### LINKED ISSUE #2176: trying to move a modified abstraction can lock you into an infinite loop

Ant1r:
Fixes #2176

Currently, the `"Discard changes?"` dialog is opened each time you "activate" the object, at mouse-up after clicking on it (like after you move it), and there is this #2176 bug.

With this PR, the dialog only opens when you try to type into the object, i.e when there is a real risk that changes will be lost.

This fixes #2176, and, additionally, reduces the number of dialog occurrences to the necessary ones.

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2907,22 +2907,10 @@ void canvas_mouseup(t_canvas *x,
     else if ((x->gl_editor->e_onmotion == MA_MOVE ||
               x->gl_editor->e_onmotion == MA_RESIZE))
     {
-            /* if there's only one text item selected activate the text.
-            LATER consider under sme conditions not activating it, for instance
-            if it appears to have been desired only to move the object.  Maybe
-            shift-click could allow dragging without activating text?  A
-            different solution (only activating if the object wasn't moved
-            (commit f0df4e586) turned out to flout ctrlD+move+retype. */
+            /* if there's only one text item selected activate the text. */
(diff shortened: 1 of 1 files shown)


Ant1r:
@umlaeute I'm not claiming that the situation is ideal after applying this PR, but at least it's much better than the current one.
Do you think it is light enough (aka no-brainer) to be merged into develop (in particular it would be really nice if #2176 could be fixed on the next release)?
Unless @millerpuckette objects?

umlaeute:
looks good. merged.

PR #2206, opened 2024-03-07
readsf~: add third outlet for soundfile info

### LINKED ISSUE #2276: readsf~: allow to enquire the status or "play when ready"
`[print(` currently allows to print to conosle info about the current status. I think it would be useful to be able to read the status programmaticaly, so one knows when enough has been preloaded that it `[play(` can be safely sent. Alternatively/additionally, what about a `[play-when-ready(` message to simplify things.

Spacechild1:
This PR adds a third outlet that will output the soundfile info as a list once the info is available. This happens some time after the user sends the "open" message.

The format of the info message is `list <size> <headersize> <nchannels> <bytespersample> <endianess>`. It is the same as in `[soundfiler]`, except for one additional item (`<size>`).

In addition, you may use the info message to determine when the object is ready for playback, i.e. you don't have to guess the wait time between "open" and "start" anymore!

For example: here's how to open a file and automatically play it once it's ready:

grafik

Pretty straightforward I'd say.

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -81,17 +81,25 @@ const char* soundfile_strerror(int errnum)
     }
 }
 
+static int write_soundfileinfo(const t_soundfile *sf, int argc, t_atom *argv)
+{
+    if (argc < 5)
+        return 0;
+    SETFLOAT((t_atom *)argv, (t_float)sf->sf_samplerate);
+    SETFLOAT((t_atom *)argv+1,
+             (t_float)(sf->sf_headersize < 0 ? 0 : sf->sf_headersize));
+    SETFLOAT((t_atom *)argv+2, (t_float)sf->sf_nchannels);
+    SETFLOAT((t_atom *)argv+3, (t_float)sf->sf_bytespersample);
+    SETSYMBOL((t_atom *)argv+4, gensym((sf->sf_bigendian ? "b" : "l")));
+    return 5;
+}
+
     /** output soundfile format info as a list */
-static void outlet_soundfileinfo(t_outlet *out, t_soundfile *sf)
+static void outlet_soundfileinfo(t_outlet *out, const t_soundfile *sf)
 {
     t_atom info_list[5];
-    SETFLOAT((t_atom *)info_list, (t_float)sf->sf_samplerate);
-    SETFLOAT((t_atom *)info_list+1,
-        (t_float)(sf->sf_headersize < 0 ? 0 : sf->sf_headersize));
-    SETFLOAT((t_atom *)info_list+2, (t_float)sf->sf_nchannels);
-    SETFLOAT((t_atom *)info_list+3, (t_float)sf->sf_bytespersample);
-    SETSYMBOL((t_atom *)info_list+4, gensym((sf->sf_bigendian ? "b" : "l")));
-    outlet_list(out, &s_list, 5, (t_atom *)info_list);
+    write_soundfileinfo(sf, 5, info_list);
+    outlet_list(out, &s_list, 5, info_list);
 }
 
     /* post soundfile error, try to print type name */
@@ -1630,6 +1638,7 @@ typedef struct _readsf
     t_sample *(x_outvec[MAXSFCHANS]); /**< audio vectors */
     int x_vecsize;                    /**< vector size for transfers */
     t_outlet *x_bangout;              /**< bang-on-done outlet */
+    t_outlet *x_infoout;              /**< soundfile info outlet */
     t_soundfile_state x_state;        /**< opened, running, or idle */
     t_float x_insamplerate;           /**< input signal sample rate, if known */
         /* parameters to communicate with subthread */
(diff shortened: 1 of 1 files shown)


porres:
that works for me. Now, could the info on number of channels be used to set number of channels in a single multichannel output? see https://github.com/pure-data/pure-data/issues/1990

Spacechild1:
> Now, could the info on number of channels be used to set number of channels in a single multichannel output? 

Yes! Although I'm not sure it would be a useful thing to do in practice. As we all know, changing the number of output channels may cause a DSP graph update, which is not something you'd want to do during a performance. It probably makes more sense to get the channel count with `[soundfiler]` beforehand to deduce the number of channels you will be working with, but you wouldn't switch between soundfiles with *different* channel counts during runtime (unless you don't care about the possible dropouts).

Here's the multichannel PR, btw: https://github.com/pure-data/pure-data/pull/2052.

porres:
> Here's the multichannel PR, btw: https://github.com/pure-data/pure-data/pull/2052.

didn't know we already had one

millerpuckette:
I think we need some way of flagging non-deterministic actions such as play-when-ready, so that people wanting to make exactly repeatable things that can be controlled with reality check can get warned about this.  For the moment I would welcome the third-outlet feature but would like to wait on the non-deterministic stuff until there's a clear policy in place.  (There's also non-deterministic stuff in other objects already such as getrealtime... I'm trying to keep a lid on that, perhaps by having a "deterministic" flag that disallows them or at least prints something out in red.)

porres:
@Spacechild1 also note this branch has conflicts

Spacechild1:
> I think we need some way of flagging non-deterministic actions such as play-when-ready, so that people wanting to make exactly repeatable things that can be controlled with reality check can get warned about this. 

That's a valid point. Typically, this is done by indicating to non-determistic objects that the application is currently running in "render" or "batch" mode. For example, SuperCollider plugins can check the `mRealtime` member (https://github.com/supercollider/supercollider/blob/6cddadcb8836567f1824c754f09b33974d9f1476/include/plugin_interface/SC_World.h#L91). There are similar mechanisms in all DAW plugin SDKs. 

I think we can do a similar thing in Pd, just by setting some global resp. per-instance variable. In this case, [readsf~] could just read the file and output the info synchronously.

> For the moment I would welcome the third-outlet feature but would like to wait on the non-deterministic stuff until there's a clear policy in place. (There's also non-deterministic stuff in other objects already such as getrealtime... I'm trying to keep a lid on that, perhaps by having a "deterministic" flag that disallows them or at least prints something out in red.)

Keep in mind that the audio output of [readsf~] itself is already non-deterministic because we don't know when the playback actually starts. That's why I don't really think that it's such a problem that the third outlet would be non-deterministic as well. Alternatively, we would need *some* way to tell the user when the sound file info is ready, and I don't see how that could ever be deterministic (at least in realtime mode). Combining these two things just felt natural to me.

millerpuckette:
I would favor something deterministic: simply output the third outlet info (which is clearly useful), but only do it when a "start" message is received.  To make it work in real time, simply wait 1/10 second between the "open" and the "start" message - which I know nobody bothers to do, but I think it's the "correct" thing.

Spacechild1:
> I would favor something deterministic: simply output the third outlet info (which is clearly useful), but only do it when a "start" message is received.

If we want to output the info together with the "start" message, we would have to synchronously access the soundfile, but the soundfile is only read on the helper thread. The whole point of `[readsf~]` is to avoid I/O on the audio thread.

For me the only logical thing is outputting the info once it has been read by the helper thread, which - as a useful sideffect - also indicates that the soundfile is ready to play.

> To make it work in real time, simply wait 1/10 second between the "open" and the "start" message - which I know nobody bothers to do, but I think it's the "correct" thing.

I definitely do this. But that's exactly my issue: I don't want to use arbitrary wait times, I'd like to play the soundfile as soon as it's ready.

millerpuckette:
> If we want to output the info together with the "start" message, we would have to synchronously access the soundfile, but the soundfile is only read on the helper thread. The whole point of `[readsf~]` is to avoid I/O on the audio thread.
> 
I think that, currently,  the subthread isn't synchronized until the next DSP tick, but it could just as well be synchronized at the 'start' message itself, at which point the SF info can (and I think should) be read.

Spacechild1:
> I think that, currently, the subthread isn't synchronized until the next DSP tick, but it could just as well be synchronized at the 'start' message itself, at which point the SF info can (and I think should) be read.

Ah, I see what you mean now. The "start" message would grab the mutex and read the cached soundfile info. If the info is not ready yet, it would wait for the helper thread to deliver the info.

Yes, that could work. However, I would still like to get a notification when the soundfile is ready, so I don't have to use arbitrary delays.

By using the soundfile info as the notification I can kill two birds with one stone. However, we can just as well use your approach and output the info together with the "start" message. In that case, we could maybe just send a bang to indicate that the soundfile is ready? You could easily use `[route]` to separate the notification bang from the info list.

Spacechild1:
Actually, I don't think that "start" should output the info list. There are cases where I want to get the soundfile info before - or even without - playing the soundfile. Tying the info the "start" message would prevent that.

Here's a typical scenario from my own practice: I have a playlist and I'm using two `[readsf~]` instances. When I'm playing a file in one instance, I'm already opening the next one in the other instance. I would like to see the info of the upcoming file, but I don't want to play it yet.

I see at least two solutions:
1. keep the current behavior of this PR, i.e. automatically send the info when it's ready
2. add a "info" method that outputs the info (waiting for the helper thread if necessary). In addition, automatically send a bang when the soundfile is ready.

umlaeute:
if you need the soundfile info prior to playing, why don't you use `[soundfiler]`? (it's not threaded, but then at least you know when you will get the info)

Spacechild1:
> if you need the soundfile info prior to playing, why don't you use [soundfiler]?

Because I don't want to do file I/O on the audio thread during a performance.

Spacechild1:
I just came back from a theater project where I would have needed this feature: the performers have tablets with a "soundboard" UI; when they push a button, a message is sent to Pd where the corresponding soundfile is played with `[readsf~]`. I want to play the soundfile as soon as its ready, but not too soon because I don't want to risk blocking Pd. I had no other choice than take a conservative guess and delay the "play" message by a fixed number of milliseconds. This is not good.

IMO we really need a way of knowing when the soundfile is ready to play.

---

I actually agree with @millerpuckette that the soundfile info should be requested by the user and not be sent automatically as part of the "ready" notification. This way you can have it both ways: you can get the info deterministically, but with the "ready" notification you can still do it without blocking Pd.

The soundfile info can be requested by sending an "info" message. If the info is not available yet, Pd waits for the worker thread, just like it would for the "start" message.

Here are some open questions:

- which outlet should output the soundfile info?
- should the info be sent as a plain Pd list (e.g. "list 48000 ...") or as a typed message (e.g. "info 48000 ...")?

@millerpuckette Would you be interested in including these two features in Pd 0.57? If yes, I'm happy to rework this PR.

Issue #2194, opened 2024-02-13
Patch appears outside the screen


Ars-subtilior:
I have created a patch, which I have opened several times.  Today, when I tried to open it, it appears in the taskbar, but will not open and the preview in the taskbar is empty.  I have tried shutting down Pd and restarting it, I have restarted my PC (running Windows 11) but the problem persists.  I had noticed this with some older patches I was not worried about losing after getting a new PC - my old one ran Windows 10.  Is this a known problem?  Is this a Windows 11 problem?  Is there a solution?

ben-wes:
this sounds familiar and is most likely due to a canvas position outside your screen area (maybe the patch was saved on a multi-monitor setup before?).

maybe check if you can recover the window via right-click on the taskbar and maximizing it (i currently don't remember what other options there are and am not on windows)? i remember that this was an issue in the past and thought it was fixed. are you working with the current Pd version?

Spacechild1:
This can happen when you save a patch that is on a second screen and then open it again with a single screen. The patch would still have the original coordinates and therefore will not be visible. That's obviously a bug. One workaround is to select the patch window in the taskbar and then bring it back to the main display with Windows+Left/Right/Up/Down.

ben-wes:
related: #1805

Ars-subtilior:
Maximising the window from the task bar and resaving it has fixed the problem.  I have not been using two monitors, so I am not sure what the original cause was, but if it happens again, I have a solution now.  Thanks for the help.

ben-wes:
> Workaround available.

good to see that it worked for you this way! i agree with @Spacechild1's opinion though: it's a bug and other users will experience this again. maybe not a critical issue - but would be good nonetheless to have a better solution for this (like limiting the offset in a way that at least parts of the window stay visible for interaction ... or even aligning windows at the top left corner if they are otherwise overflowing the screen area on any side. not sure what behaviours are common for other apps. but it's obviously a common problem).

@Spacechild1: would you suggest to keep this open or wait for similar reports?

Spacechild1:
Let's keep it open for reference. It is important to point out that @Ars-subtilior did *not* use a second screen, so it is exactly not the same issue as in #1805 (which appears to have been fixed), but they are probably related.

ben-wes:
[...]

PR #2210, opened 2024-03-13
Windows: restore canvas inside screen bounds

ben-wes:
Changes canvas positioning for Windows users once more according to discussion in https://github.com/pure-data/pure-data/issues/2194.

This replaces the current logic where positions get wrapped with a more predictable solution that keeps the canvas inside screen bounds (these bounds also apply for multi-screen setups and can have a negative origin)

This PR is to be merged into develop branch.

- Closes: https://github.com/pure-data/pure-data/issues/2194

### DIFF
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -52,24 +52,25 @@ if {[tk windowingsystem] eq "win32" || \
     # also check for Tk Cocoa backend on macOS which is only stable in 8.5.13+;
     # newer versions of Tk can handle multiple monitors so allow negative pos
     proc pdtk_canvas_wrap_window {x y w h} {
-        foreach {width height} [wm maxsize .] {break}
+        foreach {width height} [wm maxsize .] break
 
-        if {$w > $width} {
-            set w $width
-            set x 0
-        }
-        if {$h > $height} {
-            # 30 for window framing
-            set h [expr $height - $::menubarsize]
-            set y $::menubarsize
-        }
-
+        # get virtual root coordinates for minimum position
         set xmin [winfo vrootx .]
         set ymin [winfo vrooty .]
-        set x [expr ($x - $xmin) % $width + $xmin]
-        set y [expr ($y - $ymin) % $height + $ymin]
 
-        return [list ${x} ${y} ${w} ${h}]
+        # clip window size to screen size
+        set w [expr {min($w, $width)}]
+        set h [expr {min($h, $height - $::menubarsize)}]
+
+        # get max position
+        set xmax [expr {$xmin + $width - $w}]
+        set ymax [expr {$ymin + $height - $h}]
+
+        # clip given position
+        set x [expr {max(min($x, $xmax), $xmin)}]
+        set y [expr {max(min($y, $ymax), $ymin + $::menubarsize)}]
+
+        return [list $x $y $w $h]
     }
 } {
     proc pdtk_canvas_wrap_window {x y w h} {


Lucarda:
@Spacechild1 and @umlaeute i had tested this on Win11 with "single" and "dual" display settings and it all went good.

Spacechild1:
@Lucarda I'll trust you on that :)

ben-wes:
> @Lucarda I'll trust you on that :)

just don't trust me! :) 

(and on a more serious note: thanks! i'm also quite confident that this will improve things after testing with different setups myself)

umlaeute (changes requested):
in https://github.com/pure-data/pure-data/issues/2194#issuecomment-1944226690 you say that
>  removed `$::menubarsize` here since this function is only for windowingsystem "win32" and the value in that case is set to 0 in 


however, this assumption is not true.
the full condition for applying the code  is  https://github.com/pure-data/pure-data/blob/d153c362707b72872c9e9a62a9670a101329c29d/tcl/pdtk_canvas.tcl#L46-L50 which obviously includes macOS (at least for some Tcl/Tk versions), and i'm sure that on macOS the `$::menubarsize` is non-zero.

i think this needs testing on macOS before merging it.

ben-wes:
@umlaeute ah ... damn. the condition is quite obviously not just checking win32. not sure why i missed that. sorry and thanks for checking! i tested on sonoma now and it looks all good. i'd prefer to have a less intrusive change then though and consider `$::menubarsize` just in case.

will test more with an adapted version and update this PR (is that how it works or will we close this one and i'd create another one?)

Spacechild1:
> or will we close this one and i'd create another one?)

No! Just apply the changes in another commit. Or ammend the last commit and force push.

ben-wes:
> Or amend the last commit and force push.

Ok, thanks. Did that now and it should consider the `$::menubarsize` - the problem for me is that it looks good on Win11 and macOS sonoma. But I can't really tell for other setups where this change might be relevant.

ben-wes:
i tested a bit more with a forced `menubarsize` on win11 and sonoma now (where i applied this new logic as well without the conditions just for testing):
* it behaves as expected on win11, respecting an imaginary menubar
  * the menubar is always expected at the top of the whole screen area though - so for screen setups that have a negative origin, the menubar would be expected at `y: -1080` for example. not sure if this is ideal - but the current solution also only takes this variant into account.
  * additionally, the current solution only compensates for a menubar if the restored canvas height is bigger than the available screen height. this means that i should still be able to cover the menubar when `h == height`
* for sonoma, this test doesn't really make sense since it seems like `winfo vrooty` doesn't give negative values there ... at least, i can't properly restore windows at a negative position. this is not a regression though and also my tested scenario with this logic on sonoma doesn't apply in the "real world"

in short: i think, this change makes sense to avoid the disappearing patches for windows users. menubars are considered if present now and the behaviour is more transparent than it was before.

no problem at all though if this PR won't make it and if someone with actual tcl experience can solve this in a better way!

ben-wes:
@umlaeute : this is obviously still not critical at all - but the requested changes should be included (considering the menubar). i rebased to `develop` now and imho, this will potentially avoid confusion (caused by invisible patches) for some Windows users.

umlaeute:
doh, i think i missed the point of the ping: while I approved the changes, i forgot to merge into `develop`

ben-wes:
no problem. this confirmed my point of low priority. :)
anyway: thanks!

Issue #2075, opened 2023-08-04
API function to dynamically get an API function pointer by name


Spacechild1:
When Pd adds a new API function, any external that uses that function would refuse to load in older Pd versions, as they do not export the necessary symbol.

As a workaround, you can load the API function explicitly with `dlsym` resp. `GetProcAddress`. Here is an example with recently added `signal_setmultiout` function:

```
typedef void (*signal_setmultiout_fn)(t_signal **, int);
signal_setmultiout_fn my_signal_setmultiout;

// in setup function:
#ifdef _WIN32
my_signal_setmultiout = (signal_setmultiout_fn)GetProcAddress(GetModuleHandle(NULL), "signal_setmultiout");
#else
my_signal_setmultiout = (signal_setmultiout_fn)dlsym(dlopen(NULL, RTLD_NOW), "signal_setmultiout");
#endif

// in DSP method:
if (my_signal_setmultiout) // Pd has multichannel support
    ...
else // no multichannel support
    ...
```

---

Instead of expecting developers to figure this out on their own, we could add a function like `sys_getfunbyname` as a thin wrapper around `dlsym` resp. `GetProcAddress`:

```
my_signal_setmultiout = (signal_setmultiout_fn)sys_getfunbyname("signal_setmultiout");
```

The declaration in of `sys_getfunbyname` in `m_pd.h` should have a small description and probably also a short example.

umlaeute:
thanks! this has been on my todo list of a while (and then i forgot about it!)

Spacechild1:
Interestingly, REAPER seems to use a similar approach for its entire extension API:

https://github.com/justinfrankel/reaper-sdk/blob/main/sdk/reaper_plugin_functions.h

The plugin entry function (`REAPERAPI_LoadAPI`) receives an `getAPI` function which returns a function pointer from a C string. `reaper_plugin_functions.h` declares a list of API function pointers; one file implements the entry function (by defining `REAPERAPI_IMPLEMENT`) which actually initializes all the pointers. If an API function is not present in the host program, the function pointer will be `NULL`.

This is a clever way of writing a plugin API that does not require the plugin to actually link back to the host, while keeping the illusion of simple function calls. The more orthodox approach would be to just pass a function pointer table (e.g. `InterfaceTable` in SuperCollider) or some kind of host object (e.g. `IHostApplication` in the VST3 SDK).

[...]

PR #2211, opened 2024-03-13
add sys_getfunbyname() API function

Spacechild1:
This allows get a Pd API function pointer by name. It returns NULL if the function does not exist.
For example, this allows to use recently introduced API functions while providing a fallback for older Pd versions.

Closes https://github.com/pure-data/pure-data/issues/2075


umlaeute:
just for the record: I've included this PR in the following meta-PR (it is included verbatim, so once the meta-PR is merged, this one will (hopefully) be closed automatically):

PR #2212, opened 2024-03-14
simplify compile-time and runtime version-checks

umlaeute:
this PR implements a `PD_VERSION` macro that calculates a single number from the major/minor/bugfix version, as well as a `PD_VERSION_CODE` which does this for the current values of PD_MAJOR_VERSION and friends.

This makes it simple to do compile-time checks for the required Pd-version:

```C
#if PD_VERSION_CODE < PD_VERSION(0,56,0)
# error this code requires at least Pd-0.56.0
#endif
```

additionally, this PR extends the `sys_getversion()` function to return the runtime Pd version code at.
this makes it simple to do runtime checks for the required Pd-version

```C
if (sys_getversion(0,0,0) > PD_VERSION(0, 99, 0) {
    post("Pd is still around...");
}
```

the macro follows best practice from the linux kernel (where it is called `KERNEL_VERSION` resp `KERNEL_VERSION_CODE`; see your favorite `/usr/include/linux/version.h`).

this is somewhat related to @Spacechild1's #2211 , as both PRs allow external writers to better provide backward and forward compatibility.

of course, these macros can only be used with newer Pd headers, so they should have been introduced about 2 decades ago.

### DIFF
--- a/src/m_glob.c
+++ b/src/m_glob.c
@@ -198,7 +198,7 @@ void glob_init(void)
 
     /* function to return version number at run time.  Any of the
     calling pointers may be zero in case you don't need all of them. */
-void sys_getversion(int *major, int *minor, int *bugfix)
+unsigned int sys_getversion(int *major, int *minor, int *bugfix)
 {
     if (major)
         *major = PD_MAJOR_VERSION;
@@ -206,6 +206,8 @@ void sys_getversion(int *major, int *minor, int *bugfix)
         *minor = PD_MINOR_VERSION;
     if (bugfix)
         *bugfix = PD_BUGFIX_VERSION;
+
+    return PD_VERSION_CODE;
 }
 
 unsigned int sys_getfloatsize()
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -911,8 +911,8 @@ static inline int PD_BIGORSMALL(t_float f)  /* exponent outside (-512,512) */
 #endif
 #endif /* _MSC_VER */
 
-    /* get version number at run time */
-EXTERN void sys_getversion(int *major, int *minor, int *bugfix);
+    /* get major/minor/bugifx version numbers and version code at run time */
+EXTERN unsigned int sys_getversion(int *major, int *minor, int *bugfix);
 
     /* get floatsize at run time */
 EXTERN unsigned int sys_getfloatsize(void);


umlaeute:
just for the record: I've included this PR in the following meta-PR (it is included verbatim, so once the meta-PR is merged, this one will (hopefully) be closed automatically):

src/m_glob.c
```diff
@@ -198,14 +198,16 @@ void glob_init(void)
 
     /* function to return version number at run time.  Any of the
     calling pointers may be zero in case you don't need all of them. */
-void sys_getversion(int *major, int *minor, int *bugfix)
+unsigned int sys_getversion(int *major, int *minor, int *bugfix)
```
Spacechild1, line 201: I don't think that this works. Older Pd versions don't know that they are supposed to return a value, in which case RAX will just contain garbage.

umlaeute: my assumption was, that nobody will use this for the next 5 years, and then it's going to be ok :-)
this probably should be mentioned in the comments.
(i would prefer having as few functions for such meta information as possible)

Spacechild1: > my assumption was, that nobody will use this for the next 5 years, and then it's going to be ok :-)

I wouldn't count on this... Personally, I would like to use this function right away :)

> (i would prefer having as few functions for such meta information as possible)

I totally understand. However, we usually have a very strict stance on ABI compat, so I wouldn't really make an exception here.

On a second thought, adding a new function wouldn't be great either, because the external wouldn't load in older Pd versions, which would defeat the very purpose. An inline function would work though:

```
static inline int sys_getversioncode(void)
{
    int major, minor, bugfix;
    sys_getversion(&major, &minor, &bugfix);
    return PD_VERSION(major, minor, bugfix);
}
```

umlaeute: yes, the point is that a new function wouldn't be usable by old Pds anyhow, so it would require you to use `sys_getfunbyname()` or similar - which requires a recent enough Pd as well...

> we usually have a very strict stance on ABI compat,

well, we are ABI forward compatible at least (or backward, or whatever - i always confuse them; in any case, an external build against Pd-0.54-1 could not use the return value of `sys_getversion()` (as there is none), but would be usable with Pd-0.56)

> An inline function would work though:

if we go for such hacks, i would rather have the following (totally untested):

```C
unsigned int sys_getversion(int *major, int *minor, int *bugfix);

/* ABI compat hack: do not use _sys_getversioncode() directly! */
static inline unsigned int _sys_getversioncode(int*major, int*minor, int*bugfix)
{
    int a, b, c;
    sys_getversion(major, minor, bugfix);
    sys_getversion(&a, &b, &c);
    return PD_VERSION(a, b, c);
}
#define sys_getversion _sys_getversioncode
```

after a *fair amount* of time, we could then remove the compat hack.

Spacechild1: > yes, the point is that a new function wouldn't be usable by old Pds anyhow, so it would require you to use sys_getfunbyname() or similar - which requires a recent enough Pd as well...

Keep in mind that a runtime version does not necessarily involve a new API function. For example, if I have a multi-channel signal object with only signal inputs, there is no need to call `signal_setmultiout`; I only need to do the runtime check to ensure that `t_signal` has the `s_nchans` member. There might be other examples.

> well, we are ABI forward compatible at least

I am not 100% sure on this either. With the System V and Microsoft x64 ABI, the caller should save the RAX register because the function would use that register to store the return value. However, if the function returns void, the compiler may omit this. Now, if such code would be loaded by a binary where the function suddenly *does* return something, it could inadvertently overwrite the caller's RAX register. Now, I *think* that the caller really must save the RAX register in any case, so it wouldn't cause a problem. A quick check with godbolt seems to confirm this: https://godbolt.org/. But I'm not 100% sure. I am even less sure about ARM (unfortunately, I can't read ARM assembly).

In general, I don't really feel comfortable messing with function signatures unless we are 100% what we are doing.

So I would rather suggest to have an inline function `sys_getversioncode`, as outlined in my previous post. Actually, we may already export a "real" API function internally - for forward compatibility -, but declare an inline function publically to prevent linker errors with older Pd versions:
```
#if PDINTERNAL
EXTERN int sys_getversioncode(void);
#else
/* for externals */
static inline int sys_getversioncode(void)
{
    int major, minor, bugfix;
    sys_getversion(&major, &minor, &bugfix);
    return PD_VERSION(major, minor, bugfix);
}
#endif
```
After some time we can just remove the `#ifdef` together with the second class, so that externals would always call the "real" API function.

Spacechild1: Actually, if we decide to have a dedicated API function (inline or not), we might go ahead and cater to the most typical use case:

```
static inline int sys_checkminversion(int major, int minor, int bugfix)
{
    int a, b, c;
    sys_getversion(&a, &b, &c);
    return PD_VERSION(major, minor, bugfix) >= PD_VERSION(a, b, c);
}
```

Does not look too bad to me.

Note that we already have inline functions in `m_pd.h` (`PD_BADFLOAT` and `PD_BIGORSMALL`) and these are *a lot* messier.

umlaeute: i don't really like the `sys_checkminversion()`, as this is just one (although typical) usecase.

```C
if (sys_getversioncode(0,0,0) >= PD_VERSION(0, 54, 3)) { ... }
```

is obviously longer than:

```C
if (sys_checkminversion(0,54,3)) { ... }
```

but I think both are readable and the first one is more versatile.

Spacechild1: I'm fine with both. 

Note that `sys_getversioncode` wouldn't take arguments:
```
if (sys_getversioncode() >= PD_VERSION(0, 54, 3)) { ... }
```
I also find it quite readable. I'd just add a comment with a little example.

umlaeute: i did a bit of duckduckgo, and found https://stackoverflow.com/questions/15626579 which is exactly about whether changing the return type of a function from `void` to `int` would break ABI.

while the answers (and comments) are not very detailed nor authoritative, they tend to agree that such as change will not break any ABI.

umlaeute: another related question (which a windows expert like @Spacechild1 can hopefully answer easily):
m_pd.h says that
> Microsoft compiler can't handle "inline" function/macros

with the implication that one cannot use `static inline` on `_MSC_VER`. is this actually (stlil) true?
i guess that for the case of `sys_checkminversion` (or my suggested forward-compatible hack), we can just omit the `inline` - there's not much performance critical code in here anyhow.

Spacechild1: > Microsoft compiler can't handle "inline" function/macros

This is pretty out of date. Visual Studio 2015 already supports most of C99 (including `inline`).

> there's not much performance critical code in here anyhow.

`inline` is not really about performance, it is only a hint to the compiler that is largely unnecessary for modern compilers. (You can *force* inlining, but it requires compiler-specific directives, e.g. `inline __attribute__((always_inline))` for GCC.)

In C++, `inline` actually disables ODR, so that the function definition can be placed into header files without causes linker error about duplicate symbols. C does not really have this, but you can simulate it with `static` resp. `static inline`. (IIRC, the `static inline` just helps to prevent warnings about unused static functions).

`static inline` is a lazy way to define (pseudo)inline functions that kind of works for both C(99) and C++.

This is what I tend to do in my own C/C++ APIs:
```
#if defined(__cplusplus)
# define MY_INLINE inline
#else
# if (__STDC_VERSION__ >= 199901L)
#  define MY_INLINE static inline
# else
#  define MY_INLINE static
# endif
#endif
```

Such an `INLINE` macro could be applied to `PD_BADFLOAT` and `PD_BIGORSMALL`, removing quite a bit of code duplication.

---

Side note: `m_pd.h` contains quite a few relics from the distant past. For example, I'm pretty sure `EXTERN_STRUCT` hasn't been necessary for the last 20 years or so. In fact, I couldn't even find a source that it has ever been a thing.

Spacechild1: Since you've asked @millerpuckette to merge the PR, I'm wondering what your opinion is regarding the inline `sys_getversioncode`? Or do you still prefer to add a return value to `sys_getversion`?

umlaeute: i still prefer to return a value with `sys_getversion()`.

as a sidenote, i have now also added (and used) `PD_INLINE`

PR #2228, opened 2024-03-20
print: add '-l' flag to specify the log level

Spacechild1:
This would finally allow to print error or debug messages on the patch level. Long overdue IMO.

`[print -l 0]` -> critical
`[print -l 1]` -> error
`[print -l 2]` -> normal (default)
`[print -l 3]` -> debug
`[print -l 4]` -> verbose

### DIFF
--- a/src/x_interface.c
+++ b/src/x_interface.c
@@ -19,30 +19,55 @@ extern void startlogpost(const void *object, const int level, const char *fmt, .
 
   /* avoid prefixing with "verbose(PRINT_LOGLEVEL): "
   when printing to stderr or via printhook. */
-#define print_startlogpost(object, fmt, ...) do{ \
+#define print_startpost(object, fmt, ...) do { \
     if (STUFF->st_printhook || sys_printtostderr) \
         startpost(fmt, __VA_ARGS__); \
-    else startlogpost(object, PRINT_LOGLEVEL, fmt, __VA_ARGS__); \
+    else startlogpost(object, x->x_level, fmt, __VA_ARGS__); \
 } while(0)
 
 typedef struct _print
 {
     t_object x_obj;
     t_symbol *x_sym;
+    int x_level;
 } t_print;
 
 static void *print_new(t_symbol *sel, int argc, t_atom *argv)
 {
     t_print *x = (t_print *)pd_new(print_class);
-    if (argc == 0)
+    int level = PRINT_LOGLEVEL;
+    int suppress = 0;
+    while (argc > 0 && argv->a_type == A_SYMBOL)
+    {
+        const char *c = atom_getsymbol(argv)->s_name;
+        if (!strcmp(c, "-n"))
+        {
+            suppress = 1;
+            argc--; argv++;
+        }
+        else if (!strcmp(c, "-l"))
+        {
+            if (argc >= 1 && argv[1].a_type == A_FLOAT)
+            {
+                level = atom_getfloat(argv + 1);
+                argc -= 2; argv += 2;
+            }
+            else
+            {
+                pd_error(x, "missing argument for '-l' flag");
+                argc--; argv++;
+                break;
+            }
+        }
+        else break;
+    }
+    x->x_level = level;
+    if (suppress)
+        x->x_sym = &s_;
+    else if (argc == 0)
         x->x_sym = gensym("print");
     else if (argc == 1 && argv->a_type == A_SYMBOL)
-    {
-        t_symbol *s = atom_getsymbolarg(0, argc, argv);
-        if (!strcmp(s->s_name, "-n"))
-            x->x_sym = &s_;
-        else x->x_sym = s;
-    }
+        x->x_sym = atom_getsymbol(argv);
     else
     {
         int bufsize;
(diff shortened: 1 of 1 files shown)


umlaeute:
indeed, long overdue.

two remarks:
- how about symbolic loglevels?
  i always tend to forget which loglevel is what, so i usually prefer `[print -l CRITICAL]`
- the current implementation links the printout to the object (which is great), but it links to the `[print]` object (which is suboptimal).
  when i encounter a problem within an abstraction i probably want to direct the use to the abstraction itself, rather to the messenger object `[print]`.
  iemguts uses a complicated depth parameter for this, but i *think* we could just get away with attaching the parent object when using the `-l` flag. if `[print]` lives on a top-level patch, the message would instead be bound to the `[print]` object itself.

Spacechild1:
> how about symbolic loglevels?

I already thought about it. I guess we can support both. Will add it.

> when i encounter a problem within an abstraction i probably want to direct the user to the abstraction itself, rather to the messenger object [print].

Good point! I think both use cases are valid.  Personally, I wouldn't tie it to the `-l` flag, but rather introduce a separate option.

> iemguts uses a complicated depth parameter for this
> but i think we could just get away with attaching the parent object

I think a depth option would be better:
`-d 0` = this object (default)
`-d 1` = this canvas
`-d 2` = parent canvas
etc.

As you said, if I want to print an error in the context of an abstraction, it would be great if the user could be pointed to the abstraction and not some subpatch within that abstraction.

---

On the other hand, I'm wondering if we shouldn't just introduce a dedicated object for this kind of advanced logging:

`logpost <level> <depth> [<label>...]`

Spacechild1:
I just went ahead and implemented the `-d` flag. Works fine in my tests.

umlaeute:
if we (plan to) introduce a separate object `[logpost]`, then *this* PR is not really needed.

i'm not totally opposed to the `-d` flag, but it complicates things further.
otoh, i think most use-cases can be covered rather simply with the proposed depth of `1` (the parent).

Spacechild1:
> otoh, i think most use-cases can be covered rather simply with the proposed depth of 1 (the parent).

Are you sure? This assumes that all the print objects are typically located in the abstraction's root canvas, which is definitely not case.

umlaeute:
Well, if you want the user to end up within your abstraction (being "directed to the `[print]` object"), just put it within a (possibly trivial) subpatch 😀.

Probably it is as easy as using a *default* depth of 1, but allow the user to override it to 0, of they don't like it.


Ah, and if you need to resort to send/receive within your abstraction to get to the parent canvas, you probably should rethink your patch design 😉

umlaeute:
> Again, I don't think it makes sense for the loglevel option to automatically mess with the target object.

but it already does (by setting the target to `[print]` itself).
if you want to keep them orthogonal, set the target object to `NULL` unless the user specified a depth.

Spacechild1:
> but it already does (by setting the target to [print] itself).

No, the default log level is "normal" and the default depth is 0. This is perfectly fine. The `-l` and `-d` flags would allow to override both independently.

> if you want to keep them orthogonal, set the target object to NULL unless the user specified a depth.

Orthogonal does not mean that there can't be default values.

umlaeute:
Right.

That's why I was proposing a default value of 1 (as i think it makes more sense in the general case).

But we obviously don't agree (and I just want to have the last word)

Spacechild1:
> That's why I was proposing a default value of 1 (as i think it makes more sense in the general case).

I am confused. We can't change the default value. Or do you still think it's a good idea to change the default value based on some unrelated flag?

Anyway, this PR should cover all use cases, as it allows to set the log level and target depth independently. If anyone would prefer a different interface, I am open to suggestions.

umlaeute:
> We can't change the default value.

I was not aware that `[print]` currently attaches to an object (the `[print]` object itself). So no change of a "default value".

However, i just checked the code (being afk) and it seems it does so.
Probably few people are aware of that (but my reference is myself).

I still don't see a big problem with changing the default depth iff a loglevel is given (as a loglevel slightly shorts the context)

Ant1r:
I'm personally fine with @Spacechild1 's proposition.
If log level (`-l`) isn't specified, it should default to normal, the current behavior.
If the depth (`-d`) isn't specified, then it should default to 0 (the `[print]` object).
They should be independent each other.

Spacechild1:
> I was not aware that [print] currently attaches to an object (the [print] object itself).

Now that explains the confusion! :-D FWIW, this has been introduced in Pd 0.52. Personally, I find it quite useful for debugging.

Ant1r:
> this has been introduced in Pd 0.52

I think I have some responsibility here ;-) And I'm finding it useful too. It will be even more useful with this PR!

millerpuckette:
The whole logpost thing has caused me a fair bit of misery - I often can't figure out how to get debugging messages to appear on the Pd window.  (Do I need to change a "GUI preference"?)  I'd rather not deepen the confusion by adding
more logpostiness if it can be helped.  OTOH, being able to post errors and show the owning abstraction would be real useful - and this would extend to pd_error() calls in general.  So perhaps this should be a 'declare' thing that could reroute calls to pd_error() to the owning abstraction, and perhaps this should also be true of "print".

Spacechild1:
> The whole logpost thing has caused me a fair bit of misery - I often can't figure out how to get debugging messages to appear on the Pd window. (Do I need to change a "GUI preference"?) I'd rather not deepen the confusion by adding
more logpostiness if it can be helped. 

I'm not sure which problems you've had, but the `-l` flag simply specifies the log level. This works just fine and, as @umlaeute agreed, is long overdue.

> OTOH, being able to post errors and show the owning abstraction would be real useful - and this would extend to pd_error() calls in general. So perhaps this should be a 'declare' thing that could reroute calls to pd_error() to the owning abstraction, and perhaps this should also be true of "print".

You have a point that there might be other (and maybe better) solutions than manually providing a "depth" value. This definitely needs more discussion.

---

So here's a suggestion: what if I rip out the `-d` (= depth) flag and only keep the `-l` (= log level) flag?

It would be really nice if we could finally print our own error messages or debug messages on the patch level.

Issue #2241, opened 2024-04-01
class dispatcher missing prototypes? ex. clicking on pd subpatch no longer calls canvas_click


danomatika:
This looks like a bug introduced when merging #688 which added Emscripten support.

@millerpuckette  has noticed a problem with the class dispatcher which indicates it may not cover all required function prototypes. @Spacechild1 maybe we actually do need the full 36?

I am personally not surprised as those working with the original PR likely ran the Emscripten version as a runtime via libpd and not with normal desktop-style patch creation with the GUI.

I am copying in Miller's email here in the case that anyone building the current master wonders what is going on:

---

> It looks like the code now in m_class_dispatcher* does "objectmaker" calls OK but omits some cases (and includes nonexistent ones) for plain messages. All plain messages are called with at least one pointer-sized argument that is the object to pass the message to, and no top of that may have up to five integer or floating-point arguments.  So for instance,
> 
>     case 0x00 : (*(t_vfun00)(m->me_fun))(); break;
> and
> 
>     case 0x01 : (*(t_vfun01)(m->me_fun))(ad[0]); break;
> 
> aren't needed but there should be a
> 
>     case 0x15: [...]
> 
> etc.  The number of cases is still 21.  I think they are precisely the cases currently listed, each incremented by 16 with one leading integer argument.
> 
> (symptom: clicking on a "pd" doesn't open it since canvas_click has 5 floating-point args :)

danomatika:
Also, I see the `m_class_dispatcher.c` file ... but how do I run it? I assume I compile it then run it in the `src` directory or...? Would be nice to have some further documentation about this so we could, possibly, integrate these steps into the auto tools makefiles etc (if this makes sense).

Spacechild1:
@millerpuckette already found the issue:

> All plain messages are called with at least one pointer-sized argument that is the object to pass the message to, and no top of that may have up to five integer or floating-point arguments. 

The solution is simple: just invoke `m_class_dispatcher.c` with 6 maxargs. This would still create 2 unnecessary cases (`0x00` and `x06`), but we don't need to care.

> I assume I compile it then run it in the src directory or...?

Yeah, I guess so. It's just a single source file with no dependencies. I can generate new `m_class_dispatcher` headers and make a PR.

> Would be nice to have some further documentation about this so we could, possibly, integrate these steps into the autotools makefiles etc (if this makes sense).

[...]

PR #2243, opened 2024-04-02
update m_class_dispatcher headers to support up to 6 arguments

Spacechild1:
We must support up to `MAXPDARGS+1` arguments because the message receiver itself is passed as an additional `t_int` argument.

Following @umlaeute's suggestion, I have modified `m_dispatch_gen.c` (former `m_class_dispatcher.c`) to generate a single header (`m_dispatch.h`) that contains the function prototypes and a static function `mess_dispatch()`.

Closes https://github.com/pure-data/pure-data/issues/2241


Spacechild1:
@millerpuckette Should be ready for merging.

umlaeute:
thanks for updating.

i think the generator (+ source) should be renamed, e.g. to `generate-m_class_dispatcher`.

this is obviously minor, but i think it makes it obvious that the generator source code is not meant to be compiled *into* Pd (unlike the generated code)

umlaeute:
also i wonder whether we could merge the two header files into one (that declares a single static function which is then called from `m_class`)

danomatika:
> thanks for updating.
> 
> i think the generator (+ source) should be renamed, e.g. to `generate-m_class_dispatcher`.
> 
> this is obviously minor, but i think it makes it obvious that the generator source code is not meant to be compiled _into_ Pd (unlike the generated code)

I agree. It was a bit confusing to me as the .c file has the same name as the headers.

Note: We could merge into the `develop` branch. I brought it up to date the other day.

danomatika:
Maybe something simpler: "m_class_dispatch_generator.c" This can follow the other .c files with `main` functions: "u_pdreceive.c", "u_pdsend.c", and "s_watchdog.c"

Spacechild1:
> this is obviously minor, but i think it makes it obvious that the generator source code is not meant to be compiled into Pd (unlike the generated code)

+1 This would be much cleaner! I can give it a shot.

> Maybe something simpler: "m_class_dispatch_generator.c"

I like this.

umlaeute:
> > Maybe something simpler: "m_class_dispatch_generator.c"
> 
> I like this.


actually i tried to avoid the `m_` prefix, picking a totally *different* prefix.
(my trivial pick on building Pd would be "compile all `?_*.c` files")

i general, i think there's a distinction from e.g. `"u_pdsend.c`: when cross-compiling (something everybody should do at least once per day), you want `u_pdsend.c` to compile for the destination architecture, whereas the generator (if you need it) needs to be built for the architecture running the build.

but, ymmv...

umlaeute:
if you say so :-)
i would have though that the boilerplate of the header files would be sufficient:
```C
/* IMPORTANT: EDIT m_class_dispatcher.c INSTEAD OF THIS FILE */
```

but anyhow, i'll try to remain silent on this topic from now on :speak_no_evil:

danomatika:
> if you say so :-) i would have though that the boilerplate of the header files would be sufficient:
> 
> ```c
> /* IMPORTANT: EDIT m_class_dispatcher.c INSTEAD OF THIS FILE */
> ```
> 
> but anyhow, i'll try to remain silent on this topic from now on 🙊

True, that is how I figured it out *but* it saved me a couple seconds finding the file in the dir list. Actually, to better match the other main file naming it might be something like "m_gendispatch.c" or "u_gendisptach.c".

Spacechild1:
Done! I took the liberty and went with `m_dispatch.h` and `m_dispatch_gen.c` :)

danomatika:
> Done! I took the liberty and went with `m_dispatch.h` and `m_dispatch_gen.c` :)

Thanks for thinking about it. My stand is to at least do a round of back and forth on things like this still have to make sense 10+ years later... ;)

Spacechild1:
@millerpuckette *now* it should be ready to merge :)

src/m_dispatch_gen.c
```diff
@@ -0,0 +1,151 @@
...
+    if (MAXARGS >= MAXMAXARGS)
+    {
+        fprintf(stderr, "error: too many arguments for today\n");
+        return 1;
+    }
+    o = fopen("m_dispatch.h", "wb");
+    if (!o)
+        return 1;
+
+    fprintf(o, "/* IMPORTANT: EDIT m_dispatch_gen.c INSTEAD OF THIS FILE */\n\n");
```
umlaeute, line 53: even better:
```C
fprintf(o, "/* IMPORTANT: EDIT %s INSTEAD OF THIS FILE */\n\n", __FILE__);
```

danomatika:
Small niggle form clang:
~~~
m_class_dispatcher.c:49:14: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  fprintf(o, warning);
             ^~~~~~~
m_class_dispatcher.c:49:14: note: treat the string as an argument to avoid this
  fprintf(o, warning);
             ^
             "%s",
m_class_dispatcher.c:91:14: warning: format string is not a string literal (potentially insecure) [-Wformat-security]
  fprintf(o, warning);
             ^~~~~~~
m_class_dispatcher.c:91:14: note: treat the string as an argument to avoid this
  fprintf(o, warning);
             ^
             "%s",
~~~

danomatika:
Apps, ignore my last post. You already fixed this in your refactor.

src/m_dispatch_gen.c
```diff
@@ -0,0 +1,151 @@
...
+    ./m_dispatch_gen[.exe] <maxargs>
+
+NB: the <maxargs> argument should be at least MAXPDARGS+1 because the
+receiver itself is passed as an additional t_int argument!
+*/
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#define MAXMAXARGS 16
```
umlaeute, line 31: this seems to be wrong:

using `./m_dispatch_gen 11`, i get:
```C
typedef void (*t_vfun110)(t_int i0, t_floatarg d0, t_floatarg d1, t_floatarg d2, t_floatarg d3, t_floatarg d4, t_floatarg d5, t_floatarg d6, t_floatarg d7, t_floatarg d8, t_floatarg d9);
typedef void (*t_vfun110)(t_int i0, t_int i1, t_int i2, t_int i3, t_int i4, t_int i5, t_int i6, t_int i7, t_int i8, t_int i9, t_int i10);
```

so either the MAXMAXARGS must be set to 10, or the typedef naming algorithm must be fixed (e.g. `t_vfun_i%d_f%d`)

src/m_dispatch_gen.c
Spacechild1: :)

src/m_dispatch_gen.c
umlaeute: yikes, @millerpuckette caught me by surprise.

Spacechild1: you're of course right. maybe in 50 years we will support 11 type-checked arguments :) feel free to push a fix to develop.

Spacechild1: or just limit `MAXMAXARGS` to 10 :)

Spacechild1: actually, I'm wondering why we need to manually provide the number of arguments in the first place. I guess `m_dispatch_gen.c` could just include `m_pd.h` and directly use `MAXPDARG`  🤷 

but then again, I don't really care too much because we probably won't have to touch it again in the foreseeable future.

danomatika: This could also be done with a shell script, but why change a working system...

danomatika:
I can confirm this fixes clicking on `[pd ]` sub patches (and I assume other things).

PR #2252, opened 2024-04-09
update to libpd 0.14.1

danomatika:
This PR updates libpd to the current 14.1 version, with small edits for the upcoming 0.15 version. The important changes pertain to additions in libpd 0.13 including:

* C: renamed ringbuffer to z_ringbuffer to match integration in upstream
     pure-data repo, this may require updating existing projects
* C: print and midi hooks finally per-instance (!)
* C: added double precision support: libpd_read_array_double(),
     libpd_write_array_double(), libpd_double(), libpd_add_double(),
     libpd_set_double(), t_libpd_doublehook, libpd_set_doublehook(),
     libpd_get_double() (IOhannes Zmoelnig)
* C: added libpd_set_instancedata() & libpd_get_instancedata() for optional
     per-instance user data storage
* C: added t_libpd_freehook type
* C: fixed libpdinstance_new() memory issue
* C: fixed ringbuffer mem barrier on Windows (Ali Somay)
* C: call socket_init() during libpd_init() for Windows (reported by Ali Somay)


danomatika:
This builds fine in the libpd repo with the current `pure-data` master branch, however I am seeing linking errors when building in in `pure-data/libpd`:
~~~
ld: Undefined symbols:
  _s_float, referenced from:
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
      ...
  _s_list, referenced from:
      _libpd_finish_list in z_libpd.o
      _libpd_list in z_libpd.o
      _pique_new in pique.o
      _pique_list in pique.o
      _sigmund_new in sigmund~.o
      _sigmund_new in sigmund~.o
  _s_signal, referenced from:
      _bob_new in bob~.o
      _bonk_new in bonk~.o
      _loop_new in loop~.o
      _pd_tilde_new in pd~.o
      _pd_tilde_new in pd~.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make: *** [libpd.dylib] Error 1
~~~

umlaeute:
> ```
> ld: Undefined symbols:
>   _s_float, referenced from:
> ```


is `PDINSTANCE` (un)defined consistently when building `src/` vs `extra/`?

(i think the error only occurs if `src/m_class.c` is built with multi-instance support (`PDINSTANCE=1`) and some other file (e.g. `extra/sigmund~/sigmund~.c`) is built without (`PDINSTANCE` is undefined).

millerpuckette:
Yes, there's no loger an _s_float (because it would have to be 
instance-local and I don't know how

to make an "extern t_symbol *) appear as an instance-local variable.)  
Perhaps there could be a

macro to supply that?

cheers

Miller

Spacechild1:
> Perhaps there could be a macro to supply that?

There already *is* such a macro :)
```
#ifdef PDINSTANCE
...
#define s_float     (pd_this->pd_s_float)
...
```
So if you get a linker error about the `_s_float` symbol, it means that `PDINSTANCE` hasn't been defined for that source file.

millerpuckette:
(duh) - it must have been me that wrote that macro :)

danomatika:
Is `PDINSTANCE` now required, even for non-multi instance builds? Currently, libpd allows for building for "single" or "multi" instances.

danomatika:
Doh, my bad. The linker error comes from the Makefile not cleaning the .o files in `/extra`. I had built desktop pd first, then later tried building libpd which sets `PDINSTANCE` but without rebuilding the extras.

danomatika:
I'll fix the Makefile to handle this. I think someone mentioned this issue last year but I forgot about it.

danomatika:
I updated the makefile and fixed a few problems with it, for instance building some extra files not needed by libpd and a couple missing ones such as the libpd midi hooks.

millerpuckette:
OK, merged it.  Meanwhile, I've hit an ouch - when I use this inside a large Ableton project on MACOS, sometimes file descriptors exceed 1024 and select() fails.  My workaround is to define DARWIN_UNLIMITED_SELECT and to override FD_SETSIZE (defined in sys/ytpes.h) to 10240 (the current MACOS per-process fd limit).  I saw this on the web as a workaround, but it seems like a fragile hack - I'd like to have a better way to do this.

millerpuckette:
like this:
PLATFORM_CFLAGS = -DHAVE_LIBDL -DHAVE_MACHINE_ENDIAN_H \
      -D_DARWIN_UNLIMITED_SELECT -DFD_SETSIZE=10240

Spacechild1:
>  My workaround is to define DARWIN_UNLIMITED_SELECT and to override FD_SETSIZE (defined in sys/ytpes.h) to 10240

The proper fix, of course, would be to use `poll` instead. Generally, `select` has so many problems that it shouldn't be used in new code at all.

danomatika:
Ah thanks for the note about FD_SIZE. I'll add that as well here and to the libpd repo Makefile.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Apr 15, 2024, at 1:05 PM, Christof Ressi ***@***.***> wrote:﻿

My workaround is to define DARWIN_UNLIMITED_SELECT and to override FD_SETSIZE (defined in sys/ytpes.h) to 10240

The proper fix, of course, would be to use poll instead. Generally, select has so many problems that it shouldn't be used in new code at all.

danomatika:
Ok, I've set this now in the relevant makefiles in the `develop` branch.

PR #2260, opened 2024-04-15
PD_BADFLOAT: let zero count as a valid float

myQwil:
I recently came across some strange behavior with sliders. They don't allow being set to zero from their inlet:

https://github.com/pure-data/pure-data/assets/3401272/9fe3eda6-cd2f-41a5-95ec-f8dc29468ac4

It turns out that `PD_BADFLOAT` currently makes no distinction between zero and denormalized numbers.

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -847,7 +847,7 @@ static inline int PD_BADFLOAT(t_float f)  /* malformed float */
     t_bigorsmall32 pun;
     pun.f = f;
     pun.ui &= 0x7f800000;
-    return((pun.ui == 0) | (pun.ui == 0x7f800000));
+    return (f != 0) & ((pun.ui == 0) | (pun.ui == 0x7f800000));
 }
 
 static inline int PD_BIGORSMALL(t_float f)  /* exponent outside (-64,64) */
@@ -870,7 +870,7 @@ static inline int PD_BADFLOAT(t_float f)  /* malformed double */
     t_bigorsmall64 pun;
     pun.f = f;
     pun.ui[1] &= 0x7ff00000;
-    return((pun.ui[1] == 0) | (pun.ui[1] == 0x7ff00000));
+    return (f != 0) & ((pun.ui[1] == 0) | (pun.ui[1] == 0x7ff00000));
 }
 
 static inline int PD_BIGORSMALL(t_float f)  /* exponent outside (-512,512) */


src/m_pd.h
```diff
@@ -844,6 +844,7 @@ typedef union
 
 static inline int PD_BADFLOAT(t_float f)  /* malformed float */
 {
+    if (f == 0) return 0;
```
Spacechild1, line 847: small nitpick: Pd is still nominally a C89 codebase, so all declarations need to be at the top of the block. Just move the statement down one line and you're good. Same for the double precision version below.

Spacechild1:
@umlaeute lgtm, but can you have a quick look as well?

src/m_pd.h
```diff
@@ -868,6 +869,7 @@ typedef union
 static inline int PD_BADFLOAT(t_float f)  /* malformed double */
 {
     t_bigorsmall64 pun;
+    if (f == 0) return 0;
```
umlaeute, line 872: hmm, this turns the highly performance critical `PD_BADFLOAT` into code with branches, whereas previously it was branchless...

a good optimizing compiler might of course generate branchless code from that, but it would be good if that was the case even for simple compilers...

sebshader: that was my first reaction as well but at least it isn't used in very critical areas now afaict

umlaeute: indeed, after checking  it seems that in Pd-vanilla it's currently only used in slider/vumeter/`[oscparse]`.

i see that cyclone uses it a bit in its dsp perform routines (but it does so after calling to things like `acosf()` so i think performance is not a concern here :-)).

it would still be great if we had it non-branching.

sebshader: maybe I'm missing some ieee thing but wouldn't it just work to take out the ```(pun.ui == 0)``` in the return?

umlaeute: i *think* that is testing for denormals, where the exponent is 0 but the mantissa is non-null, which should be caught.

checking the `pun.ui` value *before* masking should do the trick though.
something like the following untested code:

```C
    t_bigorsmall32 pun;
    unsigned int nonzero;
    pun.f = f;
    nonzero = pun.ui;
    pun.ui &= 0x7f800000;
    return nonzero && ((pun.ui == 0) | (pun.ui == 0x7f800000));
```

sebshader: maybe something like
```
int tmp = pun.ui;
pun.ui &= 0x7f800000;
return (tmp <  0x7f800000) | (pun.ui == 0x7f800000);
```

sebshader: maybe something like
```
tmp = pun.ui
pun.ui &= 0xff800000;
return (tmp <  0x00800000) | (pun.ui >= 0x7f800000);
```

sebshader: ah I always feel like there's some other clever bit puzzle but can't quite figure out how to keep zero out of them

Spacechild1: Why not simply:
```
t_bigorsmall32 pun;
pun.f = f;
pun.ui &= 0x7f800000;
return (f != 0) & ((pun.ui == 0) | (pun.ui == 0x7f800000));
```

myQwil: Haha, yes, I was just thinking this. This approach will translate way more easily into the 64-bit version.

Ant1r:
**WARNING**: on Linux 32bit (i686), PD_BADFLOAT currently triggers for the value zero (0.0).
Example: if you send `[0.0(` to a slider, this value is silently ignored and the slider outputs its last stored value.
I know, there are only a few 32bit installations left these days, but I guess it deserves to be addressed anyway...

myQwil:
Can confirm that the latest version satisfies all the test cases that I can think of:

https://github.com/pure-data/pure-data/assets/3401272/fc796f7f-bc9f-4f23-ba24-1c244acd2e65

millerpuckette:
I think the name PD_BADFLOAT could be considered in error - we could alias it to a better name.  It's correct to use it as in (if (PD_BADFLOAT(f)) f=0; but not correct that it was necessarily bad.

umlaeute:
@millerpuckette probably, but:

with this PR accepted
- `if(PD_BADFLOAT(f)) f=0;` is still valid code and behaves the same as before
- other uses of `PD_BADFLOAT()` which take the name as correct are now also possible

a quick survey of externals on my harddisk and on github shows, that the `PD_BADFLOAT()` macro is
- not in widespread use
- mostly used as `if(PD_BADFLOAT(f)) f=0;` (with the notable exception of `g_slider.c` and `g_vu.c`.

so i do not see any issue by fixing the functionality according to the pre-existing name.

---

i also think that ff70aacf954594cc91dbbd2f81e8e91534a809f0 is not the correct fix, as now a slider will reset to 0 whenever it receives a really bad float, like a `NaN` or an `Inf`.
if you insist on keeping the current behaviour of `PD_BADFLOAT`, then I think `g_slider.c` should use something like this instead:

```C
     if (f != 0 && PD_BADFLOAT(f))
        return;
```

porres:
done with https://github.com/pure-data/pure-data/commit/f8a791dc695abd9b6442b1f91e9d0027f2981eb3

myQwil:
Don't forget, there's another 64-bit equivalent function

porres:
> Don't forget, there's another 64-bit equivalent function

ping @millerpuckette

porres:
>  there's another 64-bit equivalent function

done with https://github.com/pure-data/pure-data/commit/12203aef6e34f21e540cffc46c1a06fece8639a0

Issue #295, opened 2018-01-24
support for 64bit soundfiles (and/or improve  soundfile error messages)


umlaeute:
recently i created a soundfile with floating-point samples, using something like:

~~~python
import numpy
import scipy.io.wavfile
scipy.io.wavfile.write("test.wav", 44100, numpy.array([1., 2., 3.]))
~~~

now opening this in Pd yields:
~~~
tried .../test.wav and succeeded
soundfiler_read: .../test.wav: unknown or bad header format
~~~

so i checked the file:

~~~sh
$ sndfile-info test.wav 
========================================
File : test.wav
Length : 82
RIFF : 74
WAVE
fmt  : 18
  Format        : 0x3 => WAVE_FORMAT_IEEE_FLOAT
  Channels      : 1
  Sample Rate   : 44100
  Block Align   : 8
  Bit Width     : 64
  Bytes/sec     : 352800
fact : 4
  frames  : 3
data : 24
End
~~~

it took a while (and firing up a printf debugger) to figure out the root of the problem: the created file has a bit-width for `64` (rather than the supported `16`, `24` (for PCM) or `32` (for IEEE_FLOAT).

it would be very nice if Pd could support reading 64bit IEEE_FLOAT wavs.

it would also be nice if Pd would yield a more specific error message, e.g. "unsupported sample width (64)" or similar

danomatika:
I believe this is resolved with the sound file objects update and support for CAF files.

Spacechild1:
Can we close this?

umlaeute:
unfortunately not.

i still get:
> tried /tmp/test.wav and succeeded
> soundfiler read wave: test.wav: unsupported sample format

umlaeute:
sidnote: opening a 64bit CAF file (created with `sndfile-convert test.wav test.caf`) doesn't work either.

in case someone cannot run the above python code or the conversion themselves, i'm attaching my test-patch and two test soundfiles: [doublesound.zip](https://github.com/pure-data/pure-data/files/14955906/doublesound.zip)

danomatika:
I can look at it. For single-precision Pd, I'd assume it would cast to float samples? For double-precision it should "just work." IMO I agree it should either play or complain. In any case, fix will be in s_audio_caf.c header check. I can look later.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Apr 12, 2024, at 10:07 AM, umläute ***@***.***> wrote:﻿
[...]

PR #2261, opened 2024-04-16
Add 64 bit float sample support for WAV, AIFF, CAF

danomatika:
This PR is a response to #295 and adds support for reading and writing 64 bit float samples for the WAV, AIFF, and CAF sound file types.

It also fixes a small bug in the writesf~ where passing `-little` argument for AIFF with `-bytes 4` would not force big endian for floating point samples. The AIFF sound file format spec only allows little endian for integer samples.

The readsf~, sound filer, and writesf~ help files have been updated to refer to supporting "8-byte floating point samples".

Note: If you want to generate test files:
* Don't use `sox` for 64-bit CAF. it has a bug which results in *extremely* loud samples on playback. If you have access to a mac, use the built-in `afconvert` command ala `afconvert infile.wav -b F64 outfile.caf`.
* `sox` will convert to 64-bit float AIFF *only* if you specify the `aifc` file extension, not `aiff`. AIFC is an extended AIFF with additional formatting chunks, although most everyone just uses ".aiff" for it.


src/d_soundfile.c
```diff
@@ -19,7 +19,9 @@ objects use Posix-like threads. */
 #include <stdio.h>
 #include <pthread.h>
 
-/* Supported sample formats: LPCM (16 or 24 bit int) & 32 bit float */
+/* Supported sample formats: LPCM (16 or 24 bit int) & 32 or 64 bit float */
+
+#define VALID_BYTESPERSAMPLE(b) (b == 2 || b == 3 || b == 4 || b == 8)
```
umlaeute, line 24: prefer putting parens around the macro vars:

```
#define VALID_BYTESPERSAMPLE(b) ((b) == 2 || (b) == 3 || (b) == 4 || (b) == 8)
```

src/d_soundfile.c
```diff
@@ -495,6 +502,34 @@ static void soundfile_xferin_sample(const t_soundfile *sf, int nvecs,
...
+                    *fp = (t_sample)alias.d;
+                }
+            }
+            else
+            {
+                for (j = 0, sp2 = sp, fp = vecs[i] + framesread;
+                    j < nframes; j++, sp2 += sf->sf_bytesperframe, fp++)
+                {
+                    alias.ui = (((uint64_t)sp2[7] << 56) | ((uint64_t)sp2[6] << 48) |
+                                ((uint64_t)sp2[5] << 40) | ((uint64_t)sp2[4] << 32) |
```
umlaeute, line 526: there's quite a lot of code-duplication here, probably use a macro here and for the reverse operation (also for single)

```C
#define BYTES2DOUBLEUINT(bytes, b0, b1, b2, b3, b4, b5, b6, b7) \
    (((uint64_t)((bytes)[(b0)]) << 56) | ((uint64_t)((bytes)[(b1)]) << 48) | \
     ((uint64_t)((bytes)[(b2)]) << 40) | ((uint64_t)((bytes)[(b3)]) << 32) | \
     ((uint64_t)((bytes)[(b4)]) << 24) | ((uint64_t)((bytes)[(b5)]) << 16) | \
     ((uint64_t)((bytes)[(b6)]) <<  8) | ((uint64_t)((bytes)[(b7)]) <<  0)

            if (sf->sf_bigendian)
            {
                for (j = 0, sp2 = sp, fp = vecs[i] + framesread;
                    j < nframes; j++, sp2 += sf->sf_bytesperframe, fp++)
                {
                    alias.ui = BYTES2DOUBLEUINT(sp2, 0, 1, 2, 3, 4, 5, 6, 7);
                    *fp = (t_sample)alias.d;
                }
            }
            else
            {
                for (j = 0, sp2 = sp, fp = vecs[i] + framesread;
                    j < nframes; j++, sp2 += sf->sf_bytesperframe, fp++)
                {
                    alias.ui = BYTES2DOUBLEUINT(sp2, 7, 6, 5, 4, 3, 2, 1);
```

umlaeute (commented):
lgtm.

src/d_soundfile.c
danomatika: Out of curiosity, why?

src/d_soundfile.c
danomatika: The same is true for the other bytespersamples...

umlaeute:
Just general style advice for macros (where precedence often plays weird tricks. I haven't been able to come up with a convincing example though...)

danomatika:
I'll try to finish this up tonight.

src/d_soundfile.c
danomatika: done in latest commit

src/d_soundfile.c
umlaeute: (moving this post to the right thread, and hopefully fixing the auto-correction nonsense)

Yes (that's why I wrote "also for single").

Of course the amount of duplicate code is linearly proportional to the bytes/sample, so this is more useful for double precision than e.g. for 16bit PCM.

The idea is mostly to reduce copy/paste errors that are hard to spot on the repetitive sequences

danomatika:
Sorry all, I am too busy working on a project for an exhibition which opens next weekend. If anyone wants to jump in and implement the last suggestions, feel free. This PR is pretty much ready, otherwise.

umlaeute:
i'll give it a shot (probably tomorrow)

porres (commented):
For the help files, shouldn't we say that 8 bytes is really pertinent only if you are using Pd in double precision?

umlaeute:
Dunno.

Are 16bit sound files only pertinent when using Pd-anywhere?

I was encountering #295 on a single precision Pd, but most likely I'm impertinent...

porres:
for instance, how are you recording 64bit wave files with writesf~ if you are using Pd with single precision?

or saving them from arrays with [soundfiler]?

I mean, you can, but you're gonna lose precision, right? Same as if you play 64 bit files on single precision, huh?

porres:
for instance, how are you recording 64bit wave files with writesf~ if you are using Pd with single precision?

or saving them from arrays with [soundfiler]?

I mean, you can, but you're gonna lose precision, right? Same as if you play 64 bit files on single precision, huh?

danomatika:
I added a sub patch about this to the writesf~ help file.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Apr 29, 2024, at 7:54 AM, porres ***@***.***> wrote:﻿
for instance, how are you recording 64bit wave files with writesf~ if you are using Pd with single precision?
or saving them from arrays with [soundfiler]?
I mean, you can, but you're gonna lose precision, right? Same as if you play 64 bit files on single precision, huh?

porres:
nice, missed that one. I think we can also bring it up in [soundfiler], and also, if I'm playing a 64 bit sample with [readsf~] in single precision, the output values will get truncated to 32 bits, right?

umlaeute:
i think i would prefer a more general description, along the lines of
> Pd has a fixed sample precision.
> If you are reading soundfiles with a greater precision than the one of Pd, the extra bits are discarded.
> If you are writing soundfiles with a lower precision, information may be lost. if you are writing soundfiles with a greater precision, redundant data is recorded and you are wasting bytes.

phrased like this, it works for any precision Pd is running at, and for any value of "bytes" you might want to chose.

and by all means: improve the wording...

umlaeute:
> the output values will get truncated to 32 bits, right?

I'm not sure what exactly this means (with floating point numbers), but anyhow: yes. If Pd uses 32bit floats for `t_sample` (aka: single precision Pd), then the 64bits of the double values in the soundfile have to be squeezed into Pd's 32bit representation.
This is done using a static **cast** (`float out = (float)((double)in)`).

What exactly happens, is defined in the C standard, as mentioned in [this stackoverflow answer](https://stackoverflow.com/a/64828602/1169096):

> C 2018 6.3.1.5 discusses conversions between real floating types and says:
>
>>    … If the value being converted is in the range of values that can be represented but cannot be represented exactly, the result is either the nearest higher or nearest lower representable value, chosen in an implementation-defined manner.


i think the Pd help files shouldn't get into the details of "implementation-defined behavior of the chosen compiler".

porres:
It is sufficient to say 8-byte works 'best' in double precision Pd without getting into much details. I will propose edits in the documentation branch now

danomatika:
I stuck with the "8 byte" nomenclature since that's the value you give to writesf~ and what the main help patch refers to.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Apr 29, 2024, at 6:43 PM, porres ***@***.***> wrote:﻿
It is sufficient to say 8-byte works 'best' in double precision Pd without getting into much details. I will propose edits in the documentation branch now

PR #2262, opened 2024-04-17
Clarify a comment in G05.execution.order.pd

maftkd:
Previously the comment made a distinction between left and right branches of the patch, but that was kind of confusing because the left and right halves of the patch sort of criss-cross.

I modified the comment to be more explicit, describing how the toggle would alternate between the "right" version with correct sort order using subpatches, and the "wrong" version without.


maftkd:
Deleting - instead going to bring up an Issue to discuss since I'm actually not 100% the best way to improve the comment in G05

PR #2272, opened 2024-04-25
fix incorrect mutex being destroyed for s_inter_free()

Samson026:
Changes destroyed mutex to correct instance.

Current implementation can lead to a crash when running multiple instances of pure data as INTER points to the main instance, not the instance being freed.


Spacechild1:
@Samson026 Looks good! That was really an oversight. Can you rebase the branch on `develop`? Then I can merge it right into `develop`.

PR #2281, opened 2024-04-29
Documentation updates

porres:
This is the permanent PR for documentation, maintained/curated by me.

As a general rule-of-thumb this should **only** include changes to the documentation (help-patches, HTML); other changes should either go into the `develop` branch or in specific feature-branches

closes https://github.com/pure-data/pddp/issues/178


porres:
Thanks for the merge

> For the record, this branch was perfectly synced to master before this PR. I'm having trouble rebasing this branch to master if it changes, let's see if it still happens now.

Ok, now that it is merged, this branch is 8 commits behind and one of these commits interests me, which is the release notes and I wanted to add changes to [oscparse]. But I can't sync or update the branch to current master. Rebasing is not even an option, but it wasn't working either when I tried, because of the merging restriction. Now I tried "Update from master" 

Screen Shot 2024-05-06 at 13 49 07

Not surprisingly, it doesn't work and doesn't let me because of the protection against merging commits.

Screen Shot 2024-05-06 at 13 47 41

I'm still not sure why this branch can't be updated to current master or take merging commits, unlike other branches. I found a hack I mentioned before to make things in sync by including the commits from but it doesn't look good at al... I wouldn't like to do that, seems very wrong.

Well, please, @umlaeute / @Spacechild1 help me with this. Keeping things in sync is also quite useful to document new features that were just added (like the new right outlet to [oscparse]). Can we consider allowing the branch to easily keep in sync? Get rebased and updated? What is the concern? What caution should be taken if merging commits are allowed?

thanks

PR #2288, opened 2024-05-06
bytestruct: parse lists of bytes into messages according to a struct definition (and vice versa)

umlaeute:
(reboot of #1360, rebased on current `master` (3810461b9faf5c0d2559bcd0bb6e059543e5aa71), moved the source of the PR from the main Pd repository to my personal fork)

---

Pd has a couple of objects to convert between binary data and Pd messages (`list {to,form}symbol`, `osc{parse,format}`, `fudi{parse,format}`,...)

this PR adds another - more generic - one: `bytestruct`.
it allows one to specify a memory layout of a [POD structure](https://en.wikipedia.org/wiki/Passive_data_structure) via a format-string, and it will turn lists of bytes into an atom-list according to this definition.

It is closely modelled after the [Python `struct` standard library](https://docs.python.org/3/library/struct.html)

as with `text`, `array`, `list` & `file`, the `bytestruct` uses verbs to distinguish between different modes of operation:

- `[bytestruct pack]` turns a general atom-list into a bytelist
- `[bytestruct unpack]` turns a bytelist back into a general atom-list
- `[bytestruct size]` returns the number of bytes required to represent the structure

the format specification is rather flexible and can express, integers (8bit, 16bit, 32bit, 64bit; all signed and unsigned), floats (16bit(!), 32bit, 64bit), strings (wchar[], char[], pascal strings), bools, bytes and padding bytes (that are present in the binary representation but are ignored otherwise).

### limitations
the struct format string, describes a fixed-sized struct. while this is quite flexible, it does not lend itself to variable size-data (zero-terminated strings, run-length encoded data).

### examples
the help-patch comes with some examples:
- creating/parsing OSC-bundles
- reading a WAV file info (using the `fmt` chunk)
- reading loops from a WAV-file (using the `smpl` chunk)

### DIFF
--- a/doc/5.reference/bytestruct-help.pd
+++ b/doc/5.reference/bytestruct-help.pd
@@ -1,0 +1,646 @@
+#N canvas 196 105 753 818 12;
+#X text 459 69 details:;
+#X text 457 52 click for;
+#X text 38 703 see also:;
+#X text 405 712 updated for Pd version 0.52;
+#X obj 28 96 bytestruct pack;
+#X obj 28 126 bytestruct unpack;
+#X obj 28 166 bytestruct size;
+#X obj 123 702 oscformat;
+#X obj 123 732 oscparse;
+#X obj 215 702 fudiformat;
+#X obj 215 732 fudiparse;
+#X obj 304 716 file;
+#N canvas 458 97 1119 822 pack 0;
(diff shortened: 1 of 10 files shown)


umlaeute:
closing this for now.
"bytestruct" is available as an external via deken.

Issue #2294, opened 2024-05-07
starting a jack-less Pd with jack permanently yields errors


umlaeute:
i usually (build and) run Pd with JACK, so my .pdsettings contains:
```
audioapi: 5
```

today i accidentally built a Pd that lacks JACK (by simply omitting the `--enable-jack` flag to configure).

When i now start Pd, i get an endless stream of "Unknown API" messages:

```
$ ./bin/pd -stderr
Pd-0.54.1 ("") compiled 10:43:45 May  7 2024
float precision = 32 bits
verbose(4): input channels = 2, output channels = 2
port 46747
TCL_LIBRARY="<<SRCDIR>>/pd/_build/lib/tcl/library" TK_LIBRARY="<<SRCDIR>>/pd/_build/lib/tk/library"   wish "<<SRCDIR>>/pd/_build/tcl/pd-gui.tcl" 46747
Waiting for connection request... 
verbose(4): running at normal (non-real-time) priority.

... connected
verbose(4): priority 92 scheduling enabled.

opened alsa MIDI client 128 in:2 out:3
verbose(4): input channels = 2, output channels = 2
unknown audio API specified
unknown API
verbose(4): running at normal (non-real-time) priority.

<<SRCDIR>>/pd/_build/bin/pd-watchdog
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
unknown API
```

Issue #2274, opened 2024-04-26
core-dump on Raspberry pi - only with Gui


cooksoundscapes:
Hello everyone, I'm facing an issue when running Pd on my raspberry pi 3 B+;

When started in a graphical environment it immediately fails, but runs fine if started with `-nogui` flag;

I'm using Pd Vanilla 0.54 on Arch Linux ARM, in a minimal openbox session;

By using `coredumpctl`  and `gdb` I was able to trace the issue up to the following point:
```
Reading symbols from /usr/bin/pd...
(No debugging symbols found in /usr/bin/pd)
[New LWP 1100]
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
Core was generated by `pd'.
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00525f98 in ?? ()
(gdb) bt
#0  0x00525f98 in ?? ()
#1  0x00528330 in ?? ()
#2  0x00528eac in sys_startgui ()
#3  0x0052f374 in sys_main ()
#4  0xb6a1d9dc in __libc_start_call_main (main=0xb6a1d9dc <__libc_start_call_main+152>, main@entry=0x491ec8 <main>, argc=-1229410728, 
    argc@entry=1, argv=0x491ec8 <main>, argv@entry=0xbeb69a44) at ../sysdeps/nptl/libc_start_call_main.h:58
#5  0xb6a1daf0 in __libc_start_main_impl (main=0x491ec8 <main>, argc=1, argv=0xbeb69a44, init=<optimized out>, fini=<optimized out>, 
    rtld_fini=<optimized out>, stack_end=<optimized out>) at libc-start.c:392
#6  0x00491f14 in _start ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

Unfortunately I cannot pinpoint when it first started, since it's been a while since I tried that;
All help is highly appreciated!

umlaeute:
it seems you are using the `puredata` package from Raspbian (and/or Debian).

in order to make it easier to track down the source of the problem, please follow the steps outlined at https://wiki.debian.org/HowToGetABacktrace to create a backtrace with more info.

umlaeute:
@cooksoundscapes if you cannot provide more information, i'm afraid we will have to close this issue.
if you have trouble creating a meaningful backtrace, do not hesitate to ask.

cooksoundscapes:
[...]

PR #2295, opened 2024-05-07
stable development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/2285

---

### Core

- Closes: #2274
- Initialize some more variables that might otherwise be used uninitialized
- Include headers (in d_fft_fftw.c) for function prototypes.
- Add `const` qualifier to `pd_extraflags` (as it proxies another `const char*` value)
- Closes: #2294


danomatika:
Thanks for this. Always good to fix these kinds of warnings.

PR #2298, opened 2024-05-08
fix About doc version info

danomatika:
This PR contains a simple fix to filter out the new PD_VERSION_CODE define when pulling version info from m_pd.h.


danomatika:
This was only tested on macOS, so maybe double-check the `grep -Fv` flags are universal to other versions of grep.

umlaeute (commented):
personally, i find the new filter still too lax.
it matches any line that happens to contain `PD_MAJOR_VERSION` but not `PD_VERSION_CODE`, e.g. any potential comment.

i've addressed this in my alternative implementation at #2299 

also, i don't fully see the reason to use `-F` here (but not in the positive filter above).

apart from this, i looks good.

danomatika:
Closed in favor of #2299.

PR #2299, opened 2024-05-08
stable development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/2295

---

### Build system

- Fix calculation of `PD_VERSION` (and prefer to use the autotools `$(VERSION)` instead)


umlaeute:
i notice that @danomatika already submitted a fix for the version parsing as well in #2298.

the differences are:
- this PR also takes the `src/makefile.{gnu,mac,mingw}` build-systems into account
- the (default) autotools based `doc/Makefile.am` now uses the `$(VERSION)` variable that derives it's value from the `AC_INIT()` call in `configure.ac` - no need to parse at all!

PR #2303, opened 2024-05-13
new API functions for thread-safe messaging

Spacechild1:
Currently, there is no obvious and safe way for posting a message from a helper thread to a Pd object.

People sometimes try to execute code between a pair of  `sys_lock` + `sys_unlock`, but this can actually deadlock when you join the thread. The correct solution would be to implement your own thread-safe queue and poll it regularly with a clock; this is neither obvious nor trivial.

Therefore I propose the following API:

* `EXTERN void pd_queue_mess(struct _pdinstance *instance, t_pd *obj, void *data, t_messfn fn)`

  This sends a message to a Pd object from another (helper) thread. `fn`will be called on the scheduler thread with `obj` and `data`.  If the message has been canceled, the `obj` argument is `NULL`; the function will still be called because it is reponsible for freeing the message data.

* `EXTERN void pd_queue_cancel(t_pd *obj)`

  Cancel all pending messages for the given object. This would be typically called in the object destructor AFTER joining the helper thread.

As you can see, the implementation is very simple.

---

This PR is somewhat related to https://github.com/pure-data/pure-data/pull/1357. Ideally we would have both. #1357 should really be preferred for any kind of asynchronous command; this API would be more useful in situations where an object receives a continuous stream of data from another thread (e.g. a websocket server).

### DIFF
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -236,6 +236,9 @@ typedef struct _atom
     union word a_w;
 } t_atom;
 
+EXTERN_STRUCT _pdinstance;
+#define t_pdinstance struct _pdinstance
+
 EXTERN_STRUCT _class;
 #define t_class struct _class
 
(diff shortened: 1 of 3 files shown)


Spacechild1:
See https://github.com/pure-data/pure-data/discussions/2424 for an example use case.

umlaeute:
\o/

PR #2305, opened 2024-05-14
Update portmidi scripts

danomatika:
This PR updates the portmidi update.sh mechanism:

* move from old Sourceforge SVN to active Github git repo: [https://github.com/PortMidi/portmidi.git](https://github.com/PortMidi/portmidi.git)
* update mac MIDI rate limit patch (will need to be tested... try [this macOS arm64 build for now](http://docs.danomatika.com/pdbuilds/0.55/Pd-0.55.0-pm-no-ratelimit.zip))
* remove mac unicode name fixes, they are now included in PM sources
* update script to match portaudio/update.sh changes: remove bashisms, no longer overwrite by default (requires -f)

As with portaudio, this should make it easier to keep portmidi up to date.

Note: I haven't yet updated the portmidi sources yet. I wanted to start with this first as it's late in the release process.

~UPDATE: this PR + PortMidi v2.0.6 are in the `update/portmidi-v206` branch -> PR #2688~

UPDATE 2: this PR + PortMidi v2.0.8 are in the `update/portmidi-v208` branch -> PR #2930


portmidi/update.sh
```diff
@@ -1,62 +1,119 @@
...
 
-# checkout source
+# clone source
 echo "==== Downloading portmidi $VERSION"
-if [ -d $SRC ] ; then
-    rm -rf $SRC
+if [ -d "${SRC}" ] ; then
+    rm -rf "${SRC}"
+fi
+git clone https://github.com/PortMidi/portmidi.git "${SRC}"
```
sebshader, line 99: would it make sense for this to be --depth 1? (I guess it would be combined w/ 'checkout' in that case..)

danomatika: Possibly, but how do you then checkout an older tag if you only do a shallow clone?

danomatika:
The LIMIT_RATE handling should be fine for macOS. Should we consider updating the portmidi sources now or wait until after the 0.55 release?

portmidi/update.sh
sebshader: hmm I guess you can't..

millerpuckette:
Let's hold off for the next one - I need to update port* earlier in the 
release cycle (and need to learn to remember to do that)

danomatika:
Ok, we leave it as is for now. We can easily update from git now with the script changes. Same for portaudio.

danomatika:
Maybe revisit this for 0.57?

danomatika:
Updating this again.

danomatika:
Ok this should be working now. There were a few files dropped upstream and the Makefile.am is changed to reflect that. Compiles for me on macOS.

Also, I made a separate branch which updates the PortMidi sources to v2.0.6: `update/portmidi-v206`

danomatika:
@millerpuckette Now updated for PortAudio v2.0.8.

danomatika:
There is now an `update/portmidi-v208` branch from this branch which has the Portmidi v2.0.8 tag sources. Compiels for me on macOS 15.

Spacechild1:
@danomatika Since you are at it: can you also add `-DLIMIT_RATE=0` in `makefile.mac`? Apparently we are currently only defining it in the autotools build system.

danomatika:
> @danomatika Since you are at it: can you also add `-DLIMIT_RATE=0` in `makefile.mac`? Apparently we are currently only defining it in the autotools build system.

Got it in #2930. I wrote with Miller and he can just take that PR, so I will close this one to avoid more CI builds / two discussion threads.

PR #2308, opened 2024-05-16
Documentation updates

porres:
This is the permanent PR for documentation, maintained/curated by me.

As a general rule-of-thumb this should only include changes to the documentation (help-patches, HTML); other changes should either go into the develop branch or in specific feature-branches

--------------------

There are some commits that do update the manual and take care of listed tasks in the issu below

closes https://github.com/pure-data/pddp/issues/197
closes https://github.com/pure-data/pddp/issues/207
closes https://github.com/pure-data/pddp/issues/210
closes https://github.com/pure-data/pddp/issues/213


porres:
@millerpuckette this next manual revision just got ready, maybe you can have a look and merge it too. I severely improved the part that compares Pd to MAX and me and @ben-wes also took care of many more issues

krugre999 (approved):

umlaeute:
e41e12891ac28985297fc2b791d66e34c1a2d809 broken the 'install' target of the Makefile-based build-systems (`src/makefile.gnu`, `src/makefile.mac`, `src/makefile.mingw`).

this was not caught by the github actions as these only do a simple `make` (and no `make install`).

i'm going to fix the actions to also run `make install`, but you should fix the PR nevertheless.

porres:
> you should fix the PR nevertheless.

sure, but not sure exctly what to do. I had built myself befor4 commiting and it all went fine

umlaeute:
> I had built myself before commiting and it all went fine

Pd comes with multiple build-systems.
1. the *primary* autotools-based build-system (where you run `./autogen.sh && ./configure && make && make install`; on macOS and Windows the `make install` is typically replaced with a `make app`)
2. a secondary, stripped down pure-make buildsystem that works without `./configure`. it comes with separate makefiles for the major OSs), and needs to be invoked specifically. e.g. (for your convenience i'm using the macOS variant, but others are similar):
    ```sh
    make -C src -f makefile.mac
    make -C src -f makefile.mac install
    ```

(note: the `make install` invocations - both for the autotools and pure-make buildsystem  will install stuff into `/usr/local/` which is probably not what you want; for testing you would typically call tell it to install into some temporary directory, e.g. `make -C src -f makefile.mac install DESTDIR=$(pwd)/_install`)


so:
- just building with a *single* buildsystem might not be enough
- just *building* might not be enough as well, there's an `install` step that must succeed as well

you typically do not have to worry about those things, unless you start to "re-organize" folders.

umlaeute:
btw, i think re-organizing folders at this stage of the release is a not a good idea.

umlaeute:
I've updated the CI-configuration to now also run `make install` for the makefile-based buildsystems.
as you can see with https://github.com/pure-data/pure-data/actions/runs/9168337187/, it now fails.

umlaeute (changes requested):
i think that re-organizing folders at this stage of the 0.55 release without a compelling reason (that is: fixing bugs) should not be done.

porres:
seems the error comes from renamng 1.introduction.txt to 'about.txt' and placing it somewhere else. I reverted and it now passes the checks that were failing, but there is still a fail chcek that I am not sure if it is relevant or if it is something else I need to fix.

I don't get how renaming and relocating this txt file ruins everything cause I did change it accordingly in the makefiles... could anyone help. Can you tell me @umlaeute what was wrong with the way things were? Check the diff and what I changed in https://github.com/pure-data/pure-data/pull/2308/commits/8b8b668f9bc4f3fbc1424934ee54ab141e1f4b77

porres:
hmmm, I see I may have missed something

umlaeute:
> but there is still a fail chcek that I am not sure if it is relevant 

it *is* relevant as it breaks the default build of Pd).

for now, i think the correct way is to undo the entire folder re-organisation (back to 666ecfae9ecdcc40ccd4f4b0234a9b52ac42dc01).

umlaeute:
> hmmm, I see I may have missed something

yes. `grep about.txt` is your friend :-)


> I don't get how renaming and relocating this txt file ruins everything 

because you try to install the text *file* as a *directory*.

millerpuckette:
HI all -

A reorganization migth be a good thing early (not this late) in a 
release cycle.  OTOH, Im not sure it's a good idea to rename the 
numbered files "1.x" etc... - they're intended as a sequence and it 
would help someone trying to understand the file layout for the first 
time to see them in order.

cheers

Miller

porres:
>  Im not sure it's a good idea to rename the numbered files "1.x" etc.

Hi Miller, for now I renamed this file back to what it was (1.introduction.txt), but the thing is that this file is actually the text we get when requesting the "About Pd" menu entry, and opening it is not really meaningful as the txt gives us an unexpanded path file. 

For instance, opening it in Pd 0.54-1 gives me

"Pd is copyrighted, but is free for you to use for any reasonable purpose.
See the file:
     PD_BASEDIR/LICENSE.txt"

so 'PD_BASEDIR' is not replaced.

Also, another reasoning is that this is not an actual introduction to the manual, which has a 'chapter 1' for a real introduction.

I thought this file could be relocated out of the manual folder and just reamed to 'About.Pd', but then it ruined things and I just reverted back, anyway...

I can leave it as is, but for next time I would like to discuss a proper renaming and relocation of this. In fact, I was actually wondering if this .txt should be included in the doc folder at all, since it's just used for the 'About Pd' window.

cheers

porres:
>  Im not sure it's a good idea to rename the numbered files "1.x" etc.

Hi @millerpuckette , for now I renamed this file back to what it was (1.introduction.txt), but the thing is that this file is actually the text we get when requesting the "About Pd" menu entry, and opening it is not really meaningful as the txt gives us an unexpanded path file. 

For instance, opening it in Pd 0.54-1 gives me

"Pd is copyrighted, but is free for you to use for any reasonable purpose.
See the file:
     PD_BASEDIR/LICENSE.txt"

so 'PD_BASEDIR' is not replaced.

Also, another reasoning is that this is not an actual introduction to the manual, which has a 'chapter 1' for a real introduction.

I thought this file could be relocated out of the manual folder and just reamed to 'About.Pd', but then it ruined things and I just reverted back, anyway...

I can leave it as is, but for next time I would like to discuss a proper renaming and relocation of this. In fact, I was actually wondering if this .txt should be included in the doc folder at all, since it's just used for the 'About Pd' window.

cheers

millerpuckette:
Yep - I'm thinking about just filling in the URLs and the version number 
by hand from now on - the process of text-substituting into the intro is 
simply too fragile (as we've seen recently :)

cheers

M

porres:
>  I'm thinking about just filling in the URLs and the version number by hand from now on

The problem here @millerpuckette is that it should be expanded and replaced according to each OS, so it's not just some URL, but like, it should have read

_Pd is copyrighted, but is free for you to use for any reasonable purpose.
See the file:
     /Applications/Pd-0.54-1.app/Contents/Resources/LICENSE.txt_

instead of:
 _Pd is copyrighted, but is free for you to use for any reasonable purpose.
See the file:
PD_BASEDIR/LICENSE.txt"_

Anyway, just leaving as is for now and keeping it as it originally was, but I think it would at least make sense to rename this from '1.introduction' to '1.about'. I can open a new issue for that in pddp later and we take it from there.

For now, the main changes in this PR refer to issues that this PR closes. I hope you can revise it soon. Let me know if I'm getting too crazy, but you should at least cherry pick some fixes for Data Structures examples and other help files.

Thanks

porres:
> But that's filled in at run-time by Pd (I think - that's how I used to do it anyway).

Well, it does and works fine when I choose the "About Pd" menu entry, but this .txt file has this issue and I don't see much point in carrying it if it is just a copy of what we get from "About Pd" anyway. On the other hand, I guess it doesn't hurt to carry it in the doc folder if it gets fixed. 

Anyway, I wonder if you are ok in naming this as 'about.pd' and having it outside the manual folder. In my test, when checking the browser window, it did look nice

Screen Shot 2024-05-21 at 05 23 48

You can see how 'about.pd' actually sticks out and has a more descriptive name I think

umlaeute:
i fail to see *any* value in moving the chapters into a subdirectory.
unless you are determined to break existing patches.

@millerpuckette said:
> I'm not sure it's a good idea to rename the numbered files "1.x" etc

you replied to that with reverting the renaming of `1.introduction.txt` to `about.pd.txt`.
but you still have dropped the `x1.html` in favor of `resources/chapter1.html`

from my point of view, the main "features" of this change are:
1. we now have an almost empty `1.manual` folder (as shown by your screenshot). this just adds another layer of indirection.
2. all external links are broken (and some internal links as well, e.g. `doc/2.control.examples/02.editing.pd`


@millerpuckette 's also said:
> A reorganization migth be a good thing early (not this late) in a release cycle.

which i think is in line with my
> i think that re-organizing folders at this stage of the 0.55 release without a compelling reason (that is: fixing bugs) should not be done.

so please revert 5fd3870bb265b15bdc5a9f31918ce50fed26a35e and leave the old file structure intact.

porres:
> i fail to see any value in moving the chapters into a subdirectory

Well, I do, as it was in 0.54 and still is in 0.55 the folder is very polluted. 

Here's 0.54-1

Screen Shot 2024-05-21 at 16 58 07


And now current 0.55 test 2

Screen Shot 2024-05-21 at 00 07 45

To mention the current state, we now have 2 new subfolders that just add a new layer of noise when you open things via the browser, that don't really reveal any new files to the user. At least a subfolder is needed though, as having all files into a single folder with way over 100 pictures and chapters and cc and everything is just chaos.

The 1.introduction.txt file as I said doesn't belong at all inside 'manual', it's just a file needed on the tcl/tk side to open 'About Pd'. I can argue it shouldn't even be available inside the 'doc' folder to begin with.

I see a lot of value from a user perspective that when you check the browser you see a single index.htm file that unequivocally shows you this is all you need to check out.

Files named 'x1', 'x2', etc are not meaningful as they do not tell you what they are at all, from a user perspective, that's just noise and cofusion, for no good reason, for any really.

About '2', I see I missed fixing internal links in the docs, and I also understand that I am taking lots of risks in missing things at this point where the final release is about to come out. So yeah, I can revert and we can avoid mistakes and discuss things. 

As for external links, if I get it, you mean speculating that maybe someone out there has linked to a very specific section of the manual, which in any way probably is already dead and broken because new sections have been added, new chapters even, and if we take this seriously into consideration, I guess we basically can't update the manual any more and many revisions would need to be reverted.

porres:
> it should be expanded and replaced according to each OS

For the record, looking closely, PD_BASEDIR should not be expanded at all, we can't and this is only replaced on the tcl tk side by pd_menucommands.tcl, so this .txt could and should just be part of the tcl folder. I don't think we need to worry on carrying it in 'doc' or even try to expand 'PD_BASEDIR' so it's meaningful. Anyway, not to bother and do anything right now at this point, but I wanted to make things clear.

umlaeute (approved):

porres:
I guess I can open a new tickect and discussion in pddp and we take it from there

> i agree that that moving both img and css into a singe resource folder is nice.

In https://github.com/pure-data/pure-data/pull/2323 I am removing the 'css' folder we added, I don't like that and this is what triggered me to change things. I'm gonna do more minor changes to the manual over there that are not controversial, then merge it here so Miller can take it and we can finalize things for this release.

Controversy and headaches aside, I hope you all appreciate the countless changes we've made :) I'm very excited and happy with the jump in content we're providing here. 

cheers

umlaeute:
@millerpuckette
>  I'm thinking about just filling in the URLs and the version number
by hand from now on - the process of text-substituting into the intro is
simply too fragile (as we've seen recently :)

well, it showed us that the *version extraction from a header-file* is a bit fragile.
however, at least with the autotools build-system we already have the Pd-version (without any *test* suffix) readily available in the makefile via the `VERSION` variable.
the actual text-substitution (replacing `PD_VERSION`) is not fragile.

we still have the fragile version extraction code in `src/makefile.*`. i *think* that builds using `src/makefile.*` are not widely distributed, so we probably could just skip parsing of `m_pd.h` altogether (and use an empty PD_VERSION when building via the pure-make buildsystems).

so i think we should just ditch parsing `m_pd.h` altogether (falling back to no version; and always lacking the `test`-suffix), but keep replacing the version in the files.

alternatively, we could dynamically replace `PD_VERSION` at runtime by Tcl (though then i start to agree with @porres that `1.introduction.txt` is too much of a template to be actually displayed in the `doc/1.manual/` directory.

porres:
Ok, non controversial changes have been made, I think this is safe for @millerpuckette to merge. 

I opened a new ticket to discuss the "1.introduction.txt" file at https://github.com/pure-data/pddp/issues/215 and I have many ideas.

---------------

As for other major structural overhaul, renaming chapters and stuff, I will open yet a new issue. Actually, I forgot to mention, but I do still have big plans for the manual, like new chapters, changing order of chapters and stuff. Even though we made over 100 tasks, made countless edits and added a great deal of new stuff, pictures and subsections, I left the original skeleton and structure intact, so it's all still the same. When I open this new ticket I will reason and suggest my plans to actually overhaul things. The thing is that I reached the limit of what can be done without changing some things and I wanted to reorganize tihngs to add more stuff. And when and if doing so, renaming or changing files would be a lesser issue

porres:
by the way, earlier today me and @ben-wes figured out the issues around renaming and relocating 1.introduction.txt, see https://github.com/pure-data/pure-data/pull/2322

I will reboot that PR and link to the first tasks in https://github.com/pure-data/pddp/issues/215

porres:
by the way, I hope @Spacechild1 is proud cause I squashed so damn many commits...

umlaeute:
> wow, where did this conflicts came from?

i guess from 25a46cc18d8b05ec51102f6dcddafd497adac072

porres:
Ok. By the way, this already happened soon ago, but I was able to solve the conflicts and miller did "re-merge". Not sure if I squashed things back then.

And I see this branch has been merged, but a few of the last commits didn't make it. I also saw that I couldn't solve them because they were "too complex" see below.

Screen Shot 2024-05-22 at 12 50 26

And I thought it'd be easier to just reapply the changes, so I reverted the last 3 commits. I now have a simpler conflict to resolve, but I just can't do it anymore, huh?

Screen Shot 2024-05-22 at 12 57 55

Has something changed for it to get protected from me solving conflicts? :)

I also see that the last commit here being applied was "partially merged", because I was squashing commits and I actually turned 3 commits on fixing data structures examples into a single one, and the one fixing 17.partialtracer didn't make it.

I was actually wondering if squashing commits isn't really making things more problematic. So... ok, yes. Thanks for the answer :) damn, I was trying to make @Spacechild1  proud  :( hahaha

I can also revert this last commit and see if no conflicts exist anymore... but I need help and guidance on how to, once again, just sync this branch with master and solve all conflicts.

I've also been long asking about the merge restrictions on this branch that eventually just makes things impossible and I've requesting if we can simply remove some of the restrictions here. And I wonder if there are even more restrictions, as I can't even see what are the conflicts now. Anyway, merging and syncing to master has always been impossible. 

So, what's the next step?

porres:
Ok, I wasn't able to fix the conflict in 'x2.htm' so I cheated and made a new commit just including the x2.htm file from master. I guess @millerpuckette can try and re-merge now? At least the partialtracer issue can be solved and closed. Then we can sync to master please and I can reboot and restart to apply the last changes that didn't make it for the next round (it was merging 'css' and 'img' into a single 'resources' folder).

porres:
Ok, I tried a new commit to correct a number in the manual and it doesn't let me... it shows a conflict... it'd be really simple to resolve, but it is now protected :) I need this protection to be lifted :) and hopefully also sync up

porres:
For the record, when rebasing, I am allowed to fix conflicts.

Screen Shot 2024-05-22 at 15 41 10

But I can't rebase... it'll complain about the merge limits. Well, @umlaeute, I am sorry, but I asked this in several occasions where we had problems and they keep showing up and I get no answers. It was much easier when I just ran my documentation PRs from my repository. I just can't work with this branch anymore unless you just let me rebase, fix conflicts and stuff.

porres:
I'm closing this in favor of https://github.com/pure-data/pure-data/pull/2326 if we solve things here I can target this branch instead of master and open a new PR

PR #2323, opened 2024-05-21
manual updates

porres:
- [x] - merge 'css' and 'img' into a new 'resources' folder
- [x] - add navigation bar and logo


umlaeute (commented):
i agree that `img/` is a good place for `favicon.ico`, and that having a `css` folder with a single `pdmanual.css` is a bit of an overkill.

otoh, we now have the `pdmanual.css` as *noise* in the toplevel directory. would it be a bad idea to just stuff it into the `img/` folder as well?
for obvious reasons it is not technically an image, but well...

porres:
> otoh, we now have the pdmanual.css file as noise in the toplevel directory

I know, I could jam it into 'img' or even rename it for that matter, but at least it doesn't show up in the browser one way or another

ben-wes:
the `img` folder was only added with the recent changes for 0.55, so i also think this name is still "flexible". not completely sure what the aim is right now - but if we're looking for a folder to include images (including favicon) and css, why not just go with `resources` or `assets`?

porres:
> if we're looking for a folder to include images (including favicon) and css, why not just go with resources?

yeah, a single 'resources' folder with images, icon and css looks good. Images are numbered so they are nicely grouped anyway. I will go for that

PR #2328, opened 2024-05-25
Fix scheduler when audio device numbers are out of range

Spacechild1:
portaudio: warn if audio device numbers are out of range

portaudio/jack: return failure if there is no input and output device, so that the scheduler is set to `SCHED_AUDIO_NONE`.


umlaeute:
should this go into `develop`?

PR #2306, opened 2024-05-15
stable development branch

umlaeute:
(just merged, and here it is again)

this is the permanent branch `develop` that only gets curated, small, no-brainer changes that can be easily merged into master at any time.
(as a general rule-of-thumb we shouldn't include changes to `.pd`-files (including help patches!) in this PR, as it is just too easy to end up with unresolvable file conflicts. use the `documentation` branch for updating the documentation).

it supersedes https://github.com/pure-data/pure-data/pull/2299

---

### core

- rename `PD_CODE_VERSION` back to `PD_VERSION_CODE` (the version-extraction code has been fixed!)

### vendored libraries

- update portaudio updater script (indentation,...)

### docs

- small clarification in `writesf~-help.pd`

PR #2331, opened 2024-06-02
stable development branch

umlaeute:

(as a general rule-of-thumb we shouldn't include changes to `.pd`-files (including help patches!) in this PR, as it is just too easy to end up with unresolvable file conflicts. use the `documentation` branch for updating the documentation).

it supersedes https://github.com/pure-data/pure-data/pull/2306

---

### core

- disable dynamic guiprefs domains on macOS (Fixes the extraordinarily long startup time on macOS 14)

### DIFF
--- a/tcl/pd_guiprefs.tcl
+++ b/tcl/pd_guiprefs.tcl
@@ -68,12 +68,13 @@ proc ::pd_guiprefs::init {} {
     switch -- $::platform {
         "Darwin" {
             set backend "plist"
-            # on macOS the domain should be the same as the bundle ID
-            catch {
-                set ::pd_guiprefs::domain [exec defaults read
-                                           [file join $::sys_guidir .. .. Info]
-                                           CFBundleIdentifier]
-            }
+            ## on macOS the domain should be the same as the bundle ID
+            ## (disabled for now, as it slows down startup by several seconds)
+            #catch {
+            #    set ::pd_guiprefs::domain [exec defaults read
+            #                               [file join $::sys_guidir .. .. Info]
+            #                               CFBundleIdentifier]
+            #}
         }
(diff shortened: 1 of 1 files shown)


tcl/pd_guiprefs.tcl
```diff
@@ -68,12 +68,13 @@ proc ::pd_guiprefs::init {} {
     switch -- $::platform {
         "Darwin" {
             set backend "plist"
-            # on macOS the domain should be the same as the bundle ID
-            catch {
-                set ::pd_guiprefs::domain [exec defaults read
-                                           [file join $::sys_guidir .. .. Info]
-                                           CFBundleIdentifier]
-            }
+            ## on macOS the domain should be the same as the bundle ID
```
danomatika, line 71: Should we bump the plugin version to 0.2 with this? It could;d *potentially* break things if a fork uses a different domain and relied on this "feature."

umlaeute: No. 
This "feature" has never been part of a stable release. 

Also, it was meant as a way to ensure that the bundle id and the preferences domain are the same (by using a single source of truth). Now we are back to manually enforcing this.

danomatika: I agree. It's not really needed.

umlaeute: just to clarify...

> Also, it was meant as a way to ensure that the bundle id and the preferences domain are the same (by using a single source of truth). Now we are back to manually enforcing this.

what i meant is: reading the bundleID and using it as the preferences domain, doesn't really change the current behaviour, where we just short-circuit the readout and replace it with a hardcoded value (which the same as the bundle ID)

PR #2332, opened 2024-06-02
Update gui dialog headers & controls for case consistency

danomatika:
This PR updates the default dialog section headers (and a new few controls) to prefer upper case. Some recent editions make the dialogs a bit inconsistent, so this attempts to rectify this, ie. "audio system" in the dialog when the other headers are "Input Devices", etc. I used the following convention:

Sections
* initial upper case every word for short names, ie. "Audio System"
* initial upper case only on first word for long, description names, ie. "Pd search path for objects, help,"...

Options
* Initial upper case for control options, ie. "Use tabs"

I separated changes to the "default" strings in the Tcl from updates to the actual translation files. IMO the defaults used in the TCL should match those in en.po as much as possible. I left the tab naming alone however.

NOTE: I screwed up and accidentally pushed the commits to master when I thought I was on the feature branch. My mistake. I had to make reverts then un-revert the changes on this branch. I could squash this later, etc.


danomatika:
*Audio Before*
Screenshot 2024-06-02 at 2 22 07 PM

*Audio After*
Screenshot 2024-06-02 at 2 23 10 PM

*Misc Before*
Screenshot 2024-06-02 at 2 22 17 PM

*Misc Afte
Screenshot 2024-06-02 at 2 23 14 PM
r*

umlaeute:
Note that we have an ["English" translation](https://hosted.weblate.org/projects/pure-data/pure-data/en/) of the default strings, which could handle this as well.

That would be a less invasive change

danomatika:
Ja... but then the default EN in the TCL/code diverges from the translations. In my own projects, I prefer to keep the normative EN in sync, although I suppose we could drop a lot of it for token names when adding new stuff aka "audio.input.devices" but this can also be limiting.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Jun 2, 2024, at 7:00 PM, umläute ***@***.***> wrote:﻿
Note that we have an "English" translation of the default strings, which could handle this as well.
That would be a less invasive change

danomatika:
Also to note: I did go through the diffs in the `po` directory and make corrections. I didn't just run make and commit. ;)

danomatika:
Another small point: I think it's helpful for the translators to also have an indication how the string is used. For example, if everything is lower-case, it's likely the translations will similarly be lower-cased as I doubt every translator re-builds Pd to check the location of the strings.

umlaeute:
sure. i've uploaded a lot of screenshots to weblate, so translators can see the strings in use.
but of course weblate (and poedit) like to complain if a translatable starts with capitalization and the translation does not (or vice versa), so i guess that people tend to mimick the original casing.

i'm not opposed to the change: consistency is always good, even if we went for "token names".

i hadn't checked the `po` files but saw that they were edited and assumed that you indeed fixed the translatables. it's good to hear that this is actually the case :rocket:

danomatika:
> sure. i've uploaded a lot of screenshots to weblate, so translators can see the strings in use. but of course weblate (and poedit) like to complain if a translatable starts with capitalization and the translation does not (or vice versa), so i guess that people tend to mimick the original casing.

AH ok, I haven;t checked that out yet, but it sounds pretty good then.

> i'm not opposed to the change: consistency is always good, even if we went for "token names".

👍 

> i hadn't checked the `po` files but saw that they were edited and assumed that you indeed fixed the translatables. it's good to hear that this is actually the case 🚀

Yeah, I wouldn't do this without covering everything. I also noticed that the TR and HY files aren't in the Makefile.am, but I assume that they are simply that out of date (another issue).

umlaeute:
> I also noticed that the TR and HY files aren't in the Makefile.am, but I assume that they are simply that out of date (another issue).

they are not exactly out of date, but
- the turkish translation contains exactly one (1) translation: `OK`  -> `Tamam`
- the armenian translation is a bit better, as it contains six (6) translations (of similar triviality)

so i drew an (arbitrary) line and did not include translations that have translated no more than 1% of the strings.

once there are more strings translated, they will of course be added.

danomatika:
Should we put this in `develop` or sync with translation branch?

umlaeute:
dunno about `develop` (your pick).

i think there's no urge to merge this into the translation branch. (weblate is based on `master` btw)

danomatika:
Ok, merged to `develop`.

porres:
> Ok, merged to develop

so we can close this?

umlaeute:
> so we can close this?

it will be closed automatically, once `devel` is merged.

PR #2340, opened 2024-06-11
Documentation updates

porres:
my usual doc updates
------------
- closes https://github.com/pure-data/pddp/issues/219
- closes https://github.com/pure-data/pddp/issues/220
- closes https://github.com/pure-data/pddp/issues/226
- closes https://github.com/pure-data/pddp/issues/229
- closes https://github.com/pure-data/pddp/issues/228 
- closes https://github.com/pure-data/pddp/issues/231
- closes https://github.com/pure-data/pddp/issues/234
- closes https://github.com/pure-data/pddp/issues/237
- closes https://github.com/pure-data/pddp/issues/233
- closes https://github.com/pure-data/pddp/issues/232
- closes https://github.com/pure-data/pddp/issues/235
- closes https://github.com/pure-data/pddp/issues/238
- closes https://github.com/pure-data/pddp/issues/236
- closes https://github.com/pure-data/pddp/issues/240
- closes https://github.com/pure-data/pddp/issues/223
- closes https://github.com/pure-data/pddp/issues/244


porres:
Please, I need help from @umlaeute or @Spacechild1 as I can't see what the compilation error is actually about, check the minor change in https://github.com/pure-data/pure-data/pull/2340/commits/624a1c637a249fbe06c108c154a0c043f7a1bb03 that triggers this error, it doesn't make sense to me and I would just like to fix this comment as "patch number 68" is just bogus information.

porres:
I will fix and squash and force push when this is worked out

umlaeute:
the problem here is ,that `qlist.txt` (and - btw - `qlist2.txt`), uses DOS line endings (CRLF) rather than the native line endings.

see https://github.com/pure-data/pure-data/actions/runs/9945600225/job/27474219481?pr=2340

the extra `CR` 

now, this is not your fault, as the file has always had CRLF line endings. the tests just fail, because you touched these lines.
(an intentional side-effect of the tests is, that file formatting inconsistencies get fixed when the files are updated)

we have a couple of options
- fix the line-endings: this is *probably* the preferred option (but might have undesired side-effects on Windows, so needs a bit of testing)
- specify in `.gitattributes` that the files are supposed to have CRLF line endings and yield some magic that makes the CI tests respect these settings (preferred, if the files really have to have CRLF endings; probably hard to to)
- white-list these files in the CI tests
- ignore the CI error

umlaeute:
> fix the line-endings: this is _probably_ the preferred option (but might have undesired side-effects on Windows, so needs a bit of testing)

i just did a quick test on a Win10 VM, and `[qlist]` (on Windows) can read files with both CRLF and LF endings.
so I think the best way forward is to just convert the files.
i would typically use the `dos2unix` cmdline tool (available in `brew`), but your favorite editor might offer this as well...

porres:
I am failing and have no real idea on how to convert this

porres:
ok, all good now I guess, thanks

porres:
@millerpuckette this is ready for review and merge

porres (commented):
I should also mention some other edits that I forgot while pushing this commit.

in some examples I standardized inputs to 'ms' and 'cents', some examples had other less common inputs like 'tenths' of seconds or 'tenths' of tones. I imagine this was done before for better click and drag experience into a good range. For that, in some cases, I added a slider.

I will write more if I remember something else I missed

porres:
I should also mention some other edits that I forgot while pushing this commit for the big review on audio examples.

In some examples I standardized inputs to 'ms' and 'cents', some examples had other less common inputs like 'tenths' of seconds or 'tenths' of tones. I imagine this was done before for better click and drag experience into a good range. For that, in some cases, I added a slider.

I will write more if I remember something else I missed

PR #2344, opened 2024-06-12
Fix read at index -1 in [sigmund~]

timothyschoen:
If bestbin comes back at index 0, it will try to read the adjacent bin at index bestbin-1, which is out of bounds.

Instead, if bestbin is 0, we can just sample bin 0 twice. This solves the memory error, and should lead to well defined behaviour if bestbin equals 0.

### DIFF
--- a/extra/sigmund~/sigmund~.c
+++ b/extra/sigmund~/sigmund~.c
@@ -488,9 +488,9 @@ static void sigmund_getpitch(int npeak, t_peak *peakv, t_float *freqp,
 
         /* first guess by parabolic peak fitting */
     fbestbin = bestbin + (ppt[bestbin+1].p_weight
-        - ppt[bestbin-1].p_weight) /
+        - ppt[bestbin == 0 ? 0 : bestbin-1].p_weight) /
             (ppt[bestbin+1].p_weight +  ppt[bestbin].p_weight +
-                ppt[bestbin-1].p_weight);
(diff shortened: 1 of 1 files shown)


extra/sigmund~/sigmund~.c
```diff
@@ -488,9 +488,9 @@ static void sigmund_getpitch(int npeak, t_peak *peakv, t_float *freqp,
 
         /* first guess by parabolic peak fitting */
     fbestbin = bestbin + (ppt[bestbin+1].p_weight
-        - ppt[bestbin-1].p_weight) /
+        - ppt[bestbin == 0 ? 0 : bestbin-1].p_weight) /
```
umlaeute, line 491: shouldn't that be `bestbin < 1` (to catch all kinds of invalid values)?

umlaeute: and do the clamping only once (so we don't have to repeat the code in https://github.com/pure-data/pure-data/pull/2344/files#diff-c5ff8229d6a761a660b6ac33ae780bd8c92f30d9e4fa2616a0653d88e845458bR493)

millerpuckette: So perhaps it makes sense simply to toss any peak at bin 0... or perhaps it's worth writing some sort of special case code for it?

porres:
so, this was fixed in https://github.com/pure-data/pure-data/commit/c09505611f33f8abfb425759217d29483241a41f

right?

can we close this?

PR #2348, opened 2024-06-14
cross-platform atomics macros

Spacechild1:
This PR adds cross-platform macros for atomic integer operations.

We try to use C11 atomics if available; otherwise we emulate them with Win32 resp. GCC/Clang intrinsics.

These new macros are now used in `z_ringbuffer` and `s_audio_paring`.

I have added these to `m_private_utils.h` because they might be useful for other things. (In fact, I am currently working on PRs that would make use of them.)

---

I'm targeting the `develop` branch. If you think this would require Miller's attention and I should target `master`, just tell me.

### DIFF
--- a/src/m_private_utils.h
+++ b/src/m_private_utils.h
@@ -18,6 +18,63 @@
 # include "config.h"
 #endif
 
+/* ------------------------------ atomics ----------------------------------- */
+
+#ifdef _MSC_VER
+/* NB: Visual Studio only added C11 atomics support in version 17.5 */
+#if _MSC_VER >= 1935
+#define HAVE_C11_ATOMICS
+#endif
+#elif __STDC_VERSION__ >= 201112L
+#define HAVE_C11_ATOMICS
+#endif
+
+#ifdef HAVE_C11_ATOMICS
+/* use C11 stdatomic if available. */
+#include <stdatomic.h>
+#define atomic_int _Atomic int
+#define atomic_int_load atomic_load
+#define atomic_int_store atomic_store
+#define atomic_int_fetch_add atomic_fetch_add
+#define atomic_int_fetch_sub atomic_fetch_sub
+#define atomic_int_exchange atomic_exchange
+#define atomic_int_compare_exchange atomic_compare_exchange_strong
+#else
+/* emulate C11 atomics with platform specific intrinsics */
+#ifdef _MSC_VER
+/* Win32 instrinsics */
+#include <windows.h>
+#define atomic_int volatile int
+#define atomic_int_load(ptr) _InterlockedOr((volatile long *)(ptr), 0)
+#define atomic_int_store(ptr, value) _InterlockedExchange((volatile long *)(ptr), value)
+#define atomic_int_fetch_add(ptr, value) _InterlockedExchangeAdd((volatile long *)(ptr), value)
+#define atomic_int_fetch_sub(ptr, value) _InterlockedExchangeAdd((volatile long *)(ptr), -(value))
+#define atomic_int_exchange(ptr, value) _InterlockedExchange((volatile long *)(ptr), value)
+static int atomic_int_compare_exchange(volatile int *ptr, int *expected, int desired)
+{
+    long old = _InterlockedCompareExchange((volatile long *)ptr, *expected, desired);
+    if (old == *expected)
+        return 1;
+    else
+    {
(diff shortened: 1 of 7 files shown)


danomatika:
Actually, I'd rather keep the macros in `z_ringerbuffer.c` so it's self-contained and 1. I can back port this to libpd *now* without having to wait for the next Pd core release tag and 2. it can be dropped into another C/C++ project  without dependencies.

Spacechild1:
> Actually, I'd rather keep the macros in z_ringerbuffer.c so it's self-contained 

As I said, I would need these in other PRs.

> and 1. I can backport this to libpd now without having to wait for the next Pd core release tag

I don't think these changes are urgent. Also, I'm pretty sure Pd 0.55.1 will come pretty soon, given that I've accidentally broken the "quit" message :)

> 2. it can be dropped into another C/C++ project without dependencies.

`m_private_utils.h` is actually pretty much solved contained as well, so it's just a matter of copying a second header.

Alternatively, we could move the atomic macros to a dedicated header, but that can always be done later, probably when we add more synchronization-related stuff.

danomatika:
Why not have the macros in `z_ringbuffer.h`?

Spacechild1:
> Why not have the macros in z_ringbuffer.h?

Because they are not specific to libpd's ringbuffer.

Spacechild1:
I have now also applied the new macros to `s_audio_paring` (which didn't have any atomics resp. memory barriers at all and thus wasn't guaranteed to work on non-x86 platforms).

umlaeute (commented):
i don't really have much of an opinion here (though i tend to agree with @Spacechild1 that re-usable functionality should not be buried within some very specific header like `z_ringbuffer.h`)

the only thing i wonder is whether this should target 0.55-1 or 0.56

danomatika:
At this point, probably 0.56.enohp ym morf tnes-----------Dan Wilcoxdanomatika.comrobotcowboy.comOn Jun 17, 2024, at 8:32 AM, umläute ***@***.***> wrote:﻿
@umlaeute commented on this pull request.

i don't really mind one way or another.
the only thing i wonder is whether this should target 0.55-1 or 0.56

millerpuckette:
> the only thing i wonder is whether this should target 0.55-1 or 0.56

I think this is more a bug-fix than a new feature... so I'm leaning toward merging this for 0.55-1.  "What could possibly go wrong?"(TM)

Spacechild1:
> so I'm leaning toward merging this for 0.55-1

Fine with me :)

Spacechild1:
@millerpuckette Just a gentle reminder that you've been considering this PR for Pd 0.55-1 :)

millerpuckette:
yep, thanks :)

porres:
so, Isee it's been merged but it's still in here :) can we close this?

danomatika:
Ok, I will see if/how I integrate this with libpd. I can probably just include the header in the copy of the ring buffer included with the libpd repo. Ideally, we can remove the core libpd C stuff for the repo itself since it's included with pure-data proper, but I'm not really ready for that yet.

Spacechild1:
> Ideally, we can remove the core libpd C stuff for the repo itself since it's included with pure-data proper,

That would be great. I've always been wondering why libpd still contains the z_* sources although they have been included in pure-data. I guess it allows for faster release cycles on your end because you don't have to wait for the next Pd release. But does this matter in practice? Alternatively, libpd could have its own fork of pure-data so can you can work independently on the z_* sources without actually duplicating the files.

danomatika:
> > Ideally, we can remove the core libpd C stuff for the repo itself since it's included with pure-data proper,
> 
> That would be great. I've always been wondering why libpd still contains the z_* sources although they have been included in pure-data. I guess it allows for faster release cycles on your end because you don't have to always wait for the next Pd release. But does this matter in practice? Alternatively, libpd could have its own fork of pure-data so can you can work independently on the z_* sources without actually duplicating the files.

It was mainly because things weren't complete, as far as I was concerned when the z_* sources were added here. Now that all messaging and MIDI events are multi-instance compatible, I would say the libpd C sources are pretty stable.

The only other major development I can think of would be generalizing the GUI messaging through libpd for clients, but that can be done via separate branches, both here and in libpd. Ultimately, I'd like to somehow split up the language wrappers into their own separate repos... a discussion best made elsewhere. l)

Issue #2373, opened 2024-07-02
do we really need the same menu for all windows?


umlaeute:
PdWindow (the console) and patch-windows share the same menu.

since many entries for a patch window (e.g. `Put->Object`) do not make sense for the PdWindow, we go through all kinds of activating/deactivating code, to make sure that only those items are available that we actually want.

i wonder, why we cannot just have different menus for these two types of windows.

(afaik, the only other window that has a menu is `deken`: here it appears to work without any problems)

porres:
Sounds good

umlaeute:
the big question is: why do we do it like we do in the first place?

reduzent:
> the big question is: why do we do it like we do in the first place?

Can't answer that. I think less is more. I don't need to access Preferences and Media Settings from every patch window. OTOH, the Put menu is totally obsolete in Pd console (and also 'Pd window' and 'Parent Window' from the Window menu).

umlaeute:
while being there, how about adding a new `Tools` (or `Utilities`) menu, that could hold things that do not fit anywhere else.
e.g.
- `Message...`  (to send a message to Pd; currently in the <kbd>File</kbd> menu)
- `Find Externals`  (deken; currently in the <kbd>Help</kbd> menu)

I never liked deken to be in the `Help` menu - it's really not about "helping".
- see https://github.com/pure-data/pure-data/issues/537

and while playing with settable shortcuts (#2353), i stumbled across the `Message...`, which I seldom use but which has definitely nothing to do with anything "File"-based.

porres:
I never use 'message' either and it really doesn't feel like something useful... for instance, you can't set a receive address with "$0", only global ones. I wonder if we should even keep it and I agree it's jammed in somewhere it shouldn't.

'Find Externals', seems better placed at 'find' :) and, well, yeah, I agree it is not 'help'

umlaeute:
For me "find" doesn't seem to be the right place (apart from the obvious).
I wouldn't expect "find" (which is just your ordinary "find" as in MS word) to do a web search...

ben-wes:
> I wouldn't expect "find" (which is just your ordinary "find" as in MS word) to do a web search...

+ 1

... i know this has been discussed before - but adding my thoughts here anyway: to me, the phrasing of the entry itself feels a bit off. the intended action is not to _find_ an external, but to _install_ it (at least in the typical case).

[...]

PR #2383, opened 2024-07-11
Cleanup menus

umlaeute:
rather than using a single menubar for all windows and doing heavy reconfiguration whenever we are switching between a patch-window and the main PdWindow, this PR uses two separate menubars (that partly share the various pull-down menus).

Closes: https://github.com/pure-data/pure-data/issues/2373

For PatchWindows, nothing much has changed (except for a new `Tools` menu, for stuff that doesn't belong anywhere).

| Menu | PatchWindow | PdWindow |
|---|---|---|
|  `File`    | reorganized (*) | removed `Save` and `Print` |
|  `Edit`    | reorganized (**) | removed: the patching stuff, Undo/Redo, Edit Mode |
|  `Put`     | no changes | completely removed |
|  `Find`    | no changes | (same as PatchWindow) |
|  `Media`   | no changes | (same as PatchWindow) |
|  `Window`  | no changes | (same as PatchWindow) |
|  `Tools`   | new menu (***) | (same as PatchWindow) |
|  `Help`    | no changes | (same as PatchWindow) |

#### * reorganized `File` menu

i've changed the order of entries so now the macOS and X11/Win32 menus are practically the same (except for the `Quit` menu.

File

the obvious differences are:
- `Recent Files` is now in a separate cascade menu
- `Close` is much closer to `Save` rather than `Quit`

i guess the `Close` move might be disputed, as it is now easy to accidentally Close a file when you are trying to open  (or Save) instead.
however, I did check with other applications, and it seems that many multi-windowed applications put it there.
it is also the place where it is currently found on macOS, and people there seem to be able to cope with it.

#### ** reorganized `Edit` menu

For PatchWindows I tried to better organize the entries into categories:

Edit

(I guess one could move the Zoom/Font section below the intelligent patching stuff)

For PdWindow must stuff is gone:
Edit

I left in the `Cut` and `Paste` entries for consistency (even though they are useless and thus disabled)

#### *** Tools

as discussed in https://github.com/pure-data/pure-data/issues/2373#issuecomment-2209268225 i've created a new `Tools` menu for things like "Send a message to Pd" or "Install externals".

### further ideas
as mentioned at https://github.com/pure-data/pure-data/issues/2373#issuecomment-2203126946, one could probably do away the `File->Preferences` for PatchWindows (but it's a cascading menu anyhow, so doesn't really add much mental load).
I thought about getting rid of the `Audio/MIDI` settings for PatchWindows, but since we do need to have the `DSP on` and `DSP off` available somewhere, we would end up with a 2 element menu, so I thought it's better to leave the settings & testpatch in the menu


umlaeute:
the current state of the PR is: testing what breaks if we do it like this.

umlaeute:
- [ ] opening the `font` or the `find` dialog, now gives me an error (on Windows and Linux):
     ```
    (Tcl) UNHANDLED ERROR: can't read "::menubar": no such variable
        while executing
    "$mytoplevel configure -menu $::menubar"
        (procedure "::pd_menus::menubar_for_dialog" line 6)
        invoked from within
    "::pd_menus::menubar_for_dialog .font"
        (procedure "create_dialog" line 3)
        invoked from within
    "create_dialog $gfxstub"
        (procedure "pdtk_canvas_dofont" line 16)
        invoked from within
    "pdtk_canvas_dofont {.gfxstub57dbcda0} 12 "
        ("uplevel" body line 1)
        invoked from within
    "uplevel #0 $docmds"
   ```

umlaeute:
@millerpuckette if possible, I would like this to be merged early in the release cycle as well.

to recap:
- from a user perspective:
  - it removes unused entries (rather than disabling them), which makes the menu easier to navigate
  - it makes the menus (more) uniform across platforms (when there's no obvious need to diversion, like OS UI guidelines)
- from a dev perspective
  - simplifies the code by removing platform specific handling
  - pushes the handling of different menus down to Tcl/Tk, and (most likely) down to the window manager (rather than re-configuring menus with Tcl/Tk callbacks whenever you switch windows, each window now has a fixed menu)

the only issue i know of so far is that the keyboard accelerators for *recent files* (<kbd>1</kbd>, <kbd>2</kbd>, ...) no longer seem to work on Linux (Xfce4; i haven't tested other window-managers , like the more popular gnome or kde. MSW seems to have no problem, and on macOS there is no change to previous versions). but to be honest, i never use accelerators for accessing the recent files anyhow (as the actual file that is opened changes all the time anyhow) so this seems very minor to me. if this is a show stopper, i would probably revert the folding of the recent files into a submenu (restoring the original - platform specific - behavior)

umlaeute:
@millerpuckette or would be great if this cleanup could make it into 0.56

it reduces per-platform code (and eventually I would like to use this as the basis for some more elaborate menu handling; but for now, this cleanup would already be a nice under-the-hood improvement) 

of course I'm open for suggestions and critique

PR #853, opened 2019-12-30
Indexing table with signal inlet in expr~

nivlekp:
This PR implemented a feature which we can index a table with a signal inlet in the expr~ object.
For example [expr~ array1[$v1]] is now possible.
Here is a simple test patch.
[sig_set_ind.pd.txt](https://github.com/pure-data/pure-data/files/4009029/sig_set_ind.pd.txt)
This is a potential solution to the issue #719

PR #2401, opened 2024-08-20
allow reading array indexes with signals in [expr~]

porres:
This is a remake of https://github.com/pure-data/pure-data/pull/853 but on top of the new code changes in [expr] (see https://github.com/pure-data/pure-data/commit/cad91dc791c7b17972d4d508f6bf6a93f1ff0613 )

The thing is that we can previously access array values with float inlet inputs for indexes in [expr], as in [expr arrayname[$f1]]. It's also possible to use signal inputs in [fexpr~], as in [fexpr~ arrayname[$x1]], but we couldn't do it with [expr~].

This PR allows something like [expr~ arrayname[$v1]].

One problem that I see now is that currently - as of https://github.com/pure-data/pure-data/commit/cad91dc791c7b17972d4d508f6bf6a93f1ff0613 - a signal input for index makes it interpolate! This is funny because using signals as inputs is still not possible, but changes were made so that using signals would interpolate, see https://github.com/pure-data/pure-data/blob/master/src/x_vexp_if.c#L1087

But then, [fexrp~] doesn't interpolate. So before merging this and allowing signal inlets for indexes, we must get things straight and think carefully about what to do...

I think [expr~] should not interpolate and be consistent with how [expr] and [fexpr~] already work. Another option is that all ([expr], [expr~] and [fexpr~]) do interpolate and we have a compatibility flag to not interpolate with [expr] and [fexpr~]. Another option is to have a new table reading function that we can choose if we want interpolation or not... we could even choose and set what kind of interpolation with a flag. For instance '0' is no interpolation, '1' is linear interpolation and maybe something like '2' could be cubic (lagrange) interpolation.

### DIFF
--- a/src/x_vexp.c
+++ b/src/x_vexp.c
@@ -1623,8 +1623,11 @@ eval_var(struct expr *expr, struct ex_ex *eptr, struct ex_ex *optr, int idx)
 
         }
 
+/* (nivlekp) commenting this for now;
+* Why are we assuming the type of optr?
+* This does not work well with table
         optr->ex_type = ET_INT;
-        optr->ex_int = 0;
+        optr->ex_int = 0; */
         if (!novar)
                 (void)max_ex_var(expr, (t_symbol *)var, optr, idx);
         return (++eptr);
--- a/src/x_vexp_if.c
+++ b/src/x_vexp_if.c
@@ -1042,6 +1042,9 @@ max_ex_tab(struct expr *expr, t_symbol *s, struct ex_ex *arg, int interpol,
         int size;
         long indx;
         t_word *wvec;
+        t_float *op;    // output pointer
+        t_float *ap;    // arg pointer
+        int j;
 
         if (!s || !(garray = (t_garray *)pd_findbyclass(s, garray_class)) ||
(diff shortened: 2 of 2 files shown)


porres:
On a second thought, maybe one could choose what kind of interpolation is desired by using either a float input or signal input. And if you must use a signal input and do not want interpolation even though the input has non integer values (like if you're using the output of a [phasor~] or something), you can just use 'int' and stuff to round/truncate the signal input.

I don't like it much because of the inconsistency... I'd rather if things could be different as I discussed in the previous comment, but It's still 'acceptable' I guess....

The problem that would still remain is the inconsistency to [fexpr~] and signal input, which does not interpolate, and maybe is something that nobody knew until now :) one way or another, [fexpr~] should behave like [expr~] with signal input for indexes.

porres:
Something else occurred to me, which would be the ability to use signal input as indexes for the 'store' function in arrays, as in [expr~ arrayname[$v1] = 1], and also with [fexpr~]. This is another issue and request, but I wanted to make it clear that this PR does not enable that!

porres:
for the record, signal support will be something for the next [expr] version, we're focusing on the symbol support and other bugs right now.

PR #2405, opened 2024-08-22
allow concurrent calls for pdgui_vmess()

### LINKED ISSUE #2404: opening files via [readsf~]  can crash the GUI.
with a patch under heavy GUI-load (loads of moving bangs and GOPs and whatnot), i sometimes experience a weird lockup (eventually making the GUI unresponsive).

when the error occurs, the GUI displays the following:
```
tried /home/umlaeute/_____/________/__________/______/_______/49.wav and succeeded
tried /home/umlaeute/_____/________/__________/______/_______/40.wav and succeeded
tried /home/umlaeute/_____/________/__________/______/_______/17.wav and succeeded
(Tcl) INVALID COMMAND NAME: invalid command name ".x5564f5fbe700.c:coordsow::logpost"
    while executing
".x5564f5fbe700.c:coordsow::logpost obj:(nil) 4igur {0x5564f60bd020BASE} 334  {tried /home/umlaeute/_____/________/__________/______/_______/52.w..."
    ("uplevel" body line 147)
    invoked from within
"uplevel #0 $docmds"tried /home/umlaeute/_____/________/__________/______/_______/10.wav and succeeded
tried /home/umlaeute/_____/________/__________/______/_______/15.wav and succeeded
tried /home/umlaeute/_____/________/__________/______/_______/24.wav and succeeded
tried /home/umlaeute/_____/________/__________/______/_______/08.wav and succeeded
```

turning on debugging for the core<->gui communication i see:

```
>> .x5564f5208ab0.c delete {0x5564f62ae790IN0} ;
>> ::pdwindow::logpost obj:(nil) 4 {tried /home/umlaeute/_____/________/__________/______/_______/17.wav and succeeded
} ;
>> pdtk_text_set .x5564f5208ab0.c {0x5564f61f2df0LABEL} {⮊} ;
>> .x5564f5208ab0.c coords {0x5564f61f2df0LABEL} 330 485 ;
[...]
>> .x5564f5208ab0.c coords {0x5564f5fbdd50LABEL} 430 585 ;
>> .x5564f5208ab0.c itemconfigure {0x5564f5fbdd50LABEL} -font {{DejaVu Sans Mono} -22 {bold} } -fill #404040 ;
>> pdtk_text_set .x5564f5208ab0.c {0x5564f5fbdd50LABEL} {⮊} ;
>> .x5564f5208ab0.c coords {0x5564f5fbdd50LABEL} 430 585 ;
>> .x5564f5208ab0.c delete {0x5564f5fbdd50OUT0} ;
>> .x5564f5208ab0.c delete {0x5564f5fbdd50IN0} ;
>> .x5564f5fbe700.c coords {0x5564f60bd020RECT} 334 51 382 99 ;
>> .x5564f5fbe700.c itemconfigure {0x5564f60bd020RECT} -fill #e0e0e0 -outline #e0e0e0 ;
>> .x5564f5fbe700.c :coords obj:(nil) coords {0x5564f60bd020BASE} 3344 {tried /home/umlaeute/_____/________/__________/______/_______/52.wav and succeeded
} ;
  {tried /home/umlaeute/_____/________/__________/______/_______/52.wav and succeeded
} ;
51 382 99 ;
>> .x5564f5fbe700.c itemconfigure {0x5564f60bd020BASE} -width 1 -outline #e0e0e0 ;
>> .x5564f5fbe700.c coords {0x5564f60bd020LABEL} 348 75 ;
>> .x5564f5fbe700.c itemconfigure {0x5564f60bd020LABEL} -font {{DejaVu Sans Mono} -30 {bold} } -fill #e0e0e0 ;
>> pdtk_text_set .x5564f5fbe700.c {0x5564f60bd020LABEL} {} ;
```

so obviously some messages for the GUI get interleaved before sending them over.

this is with Pd-0.55-0

umlaeute:
This introduces a new (per-instance) lock, to ensure that messages sent to the GUI from different threads do not produce garbage.

this only protects the new-style `pdgui_vmess()` calls (so users of `sys_vgui()` and friends should switch ;-))

- Closes: https://github.com/pure-data/pure-data/issues/2404 

see also https://github.com/Spacechild1/pure-data/commit/3be5d4ac3a4b763a056c0f498889c25781757c32#commitcomment-145643167 for a bit of discussion

### DIFF
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -160,6 +160,7 @@ struct _instanceinter
 #endif
 #if PDTHREADS
     pthread_mutex_t i_mutex;
+    pthread_mutex_t i_guimutex;
     pthread_mutex_t i_messqueue_mutex;
     t_messqueue *i_messqueue_head;
     t_messqueue *i_messqueue_tail;
@@ -854,7 +855,7 @@ void sys_vgui(const char *fmt, ...)
     if(msglen < 0)
     {
         fprintf(stderr,
-            "sys_vgui: pd_snprintf() failed with error code %d\n", errno);
+            "Pd: buffer space wasn't sufficient for long GUI string\n");
         return;
     }
     if (msglen >= INTER->i_guisize - INTER->i_guihead)
@@ -932,9 +933,11 @@ static const char**namelist2strings(t_namelist *nl, unsigned int *N) {
     }
     return result;
 }
-
+void pdgui_lock(void);
+void pdgui_unlock(void);
 static int sys_flushtogui(void)
 {
+    pdgui_lock();
     int writesize = INTER->i_guihead - INTER->i_guitail,
         nwrote = 0;
     if (writesize > 0)
@@ -942,7 +945,6 @@ static int sys_flushtogui(void)
             INTER->i_guisock,
             INTER->i_guibuf + INTER->i_guitail,
             writesize, 0);
-
 #if 0
     if (writesize)
         fprintf(stderr, "wrote %d of %d\n", nwrote, writesize);
@@ -953,8 +955,10 @@ static int sys_flushtogui(void)
         perror("pd-to-gui socket");
         sys_bail(1);
     }
-    else if (!nwrote)
+    else if (!nwrote) {
+        pdgui_unlock();
         return (0);
+    }
     else if (nwrote >= INTER->i_guihead - INTER->i_guitail)
         INTER->i_guihead = INTER->i_guitail = 0;
     else if (nwrote)
@@ -969,6 +973,7 @@ static int sys_flushtogui(void)
             INTER->i_guitail = 0;
         }
     }
+    pdgui_unlock();
     return (1);
 }
 
@@ -1925,6 +1930,7 @@ void s_inter_newpdinstance(void)
     pthread_mutex_init(&INTER->i_mutex, NULL);
     pthread_mutex_init(&INTER->i_messqueue_mutex, NULL);
     pd_this->pd_islocked = 0;
+    pthread_mutex_init(&INTER->i_guimutex, NULL);
 #endif
 #ifdef _WIN32
     INTER->i_freq = 0;
@@ -1946,6 +1952,7 @@ void s_inter_free(t_instanceinter *inter)
     }
 #if PDTHREADS
     pthread_mutex_destroy(&inter->i_mutex);
+    pthread_mutex_destroy(&inter->i_guimutex);
         /* flush message queue */
     while (inter->i_messqueue_head)
     {
@@ -2095,6 +2102,14 @@ void messqueue_dispatch(void)
     }
 }
 
+void pdgui_lock(void)
+{
+    pthread_mutex_lock(&INTER->i_guimutex);
+}
+void pdgui_unlock(void)
+{
+    pthread_mutex_unlock(&INTER->i_guimutex);
+}
 #else /* PDTHREADS */
 
 #ifdef TEST_LOCKING /* run standalone Pd with this to find deadlocks */
@@ -2126,5 +2141,8 @@ void pd_queue_mess(struct _pdinstance instance, t_pd *obj, void *data, t_messfn 
 void pd_queue_cancel(t_pd *obj) {}
 
 void messqueue_dispatch(void) {}
+
+void pdgui_lock(void) {}
+void pdgui_unlock(void) {}
 
 #endif /* PDTHREADS */
--- a/src/s_inter.c
+++ b/src/s_inter.c
@@ -990,7 +990,7 @@ static int sys_flushqueue(void)
     {
         if (INTER->i_bytessincelastping >= GUI_BYTESPERPING)
         {
-            pdgui_vmess("pdtk_ping", "");
+            pdgui_vmess("pdtk_ping", 0);
             INTER->i_bytessincelastping = 0;
             INTER->i_waitingforping = 1;
             return (1);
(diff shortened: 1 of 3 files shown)


umlaeute:
So far, i haven't thought about PDINSTANCE support - does the mutex has to be per-instance?

Spacechild1:
I would really make this per instance, like in https://github.com/Spacechild1/pure-data/commit/3be5d4ac3a4b763a056c0f498889c25781757c32.

umlaeute:
ok, i'm now using a per-instance lock.

Spacechild1 (approved):

millerpuckette:
Well, this just pushes the problem deeper... calls to pd_vmess() can't be re-ordered at will (e.g., the main thread might delete a screen object that a sub-thread tries to move or change).

Calls to post() and its friends can be happily re-ordered.  So perhaps the post() mechanism is the place to do this.

Or... simply refuse all GUI actions from non-main thread and leave it to anyone unfortunate enough to want to make printout in a sub-thread to bubble the error up to the main thread where it can be sent to post() etc.  I can volunteer to try to do that for sfread~/sfwrite~ which need more work anyway - the whole open_via_path() mechanism is problematic.

For now, perhaps it will work to suppress printout when in a sub-thread, and if verbose is set, try pre-opening soundfiles in the main thread (with printout).  the pre-opening might cause a gap in the sound, but that will go away when the user wisely turns verboseness back off :)

umlaeute:
well my plan is, that `pdgui_vmess()` calls are practically self-contained and rather stateless.

in reality it is not going to be fully stateless (e.g. a window needs to be created before a canvas can be created before we can draw anything to it), but we can teach the GUI to not spit in our face if you update an object that has already been destroyed.

but yes, in theory there needs to be synchronization on the GUI side as well.

---

anyhow, I do not like the idea of different code-paths for verbose-mode vs normal-mode, as this makes issues even harder to track down (after all, we are in threading land!): i'd like to avoid situations like "oh, you should have told me you were running Pd as user *idle*, because we are special-casing the cosine function in this case."

Spacechild1:
> Or... simply refuse all GUI actions from non-main thread and leave it to anyone unfortunate enough to want to make printout in a sub-thread to bubble the error up to the main thread where it can be sent to post()

How would you do that? You would need to maintain your own thread-safe queue. It's *possible* (I'm currently doing this in `[vstplugin~]`), but it's pretty cumbersome for such a basic task.

https://github.com/pure-data/pure-data/pull/2303 would make this a little bit easier, but that PR is really meant for more elaborate tasks than simply logging a message.

From my side it would also be ok to just protect the logging calls, but I don't think it hurts to go ahead and protect `pdgui_vmess()`.

porres:
just checking, has this fix been superseded by something else?

millerpuckette:
So I _think_ the only real reason for doing this is so that printout (post(), etc) can work from other threads than the main one.  And yes, simply putting a lock around sys_vgui(0 (as called in s_inter_gui.c) would take care of this.  Other calls to pdgui_vmess() would cause trouble (and I don't think the GUI can't ever be made stateless without causing some major code bloat).  So right now I'm thinking perhaps this is OK but it would be even better if you snooped the message and only allowed printout if we aren't on the main thread.  This check should be done inside s_inter.c so that the thread ID of the main thread can be stored in the t_instanceinter structure which is opaque outside that file.

Spacechild1:
> This check should be done inside s_inter.c so that the thread ID of the main thread can be stored in the t_instanceinter structure which is opaque outside that file.

We careful with that! This might work for the Pd application, where we know on which thread the scheduler runs. But even there we might run into troubles, e.g. if the user switches between polling and callback mode!

It gets even more troublesome with libpd because we don't really know which thread the scheduler is running on. In fact, libpd clients may even call `libpd_process` on different threads for each tick, e.g. when using a thread pool to run multiple patches in parallel.

So I would strongly advice against working with thread IDs.

Generally, I see two solutions:
1. allow to call `post` and friends from different threads and protect the GUI queue with a lock
2. tell users that they must use the new `pd_queue_mess` function if they want to log from helper threads.

Personally, I would opt for 1.

Side note: in my experimental multi-threading fork (https://github.com/Spacechild1/pure-data/tree/multi-threading) I *have* to protect the GUI queue because perform routines can run in parallel. In that case, there is really no other option and it works fine. That's also what Supernova (= multithreaded version of the scsynth) has to do so that the user can call `Print` in the process function.

millerpuckette:
I'm worried that there would then be four separate mutexes in Pd... is there any way to avoid deadlocks in which the owner of lock a tries to lock b and the owner of lock b is trying to lock a?  It seems to me that this new lock wouldn't cause any new problems since nobody will ever call a function within sys_gui that tries to open any of the other locks.  But I don't know how we set about guaranteeing that we'll stay out of trouble with all these locks.

millerpuckette:
(OTOH it's a crash bug currently so I guess _something_ has to be done)

Spacechild1:
> I'm worried that there would then be four separate mutexes in Pd... is there any way to avoid deadlocks in which the owner of lock a tries to lock b and the owner of lock b is trying to lock a? It seems to me that this new lock wouldn't cause any new problems since nobody will ever call a function within sys_gui that tries to open any of the other locks. But I don't know how we set about guaranteeing that we'll stay out of trouble with all these locks.

Spacechild1:
> I'm worried that there would then be four separate mutexes in Pd... 

The number of mutexes doesn't really matter. The important thing is what you do in the critical sections. The real problem are coarse mutexes like `sys_lock()/sys_unlock()` with huge critical sections that may run arbitrary code. It's always better to have several fine-grained mutexes that only protect specific resources.

> is there any way to avoid deadlocks in which the owner of lock a tries to lock b and the owner of lock b is trying to lock a? It seems to me that this new lock wouldn't cause any new problems since nobody will ever call a function within sys_gui that tries to open any of the other locks. 

Exactly. As long as you only hold the lock while pushing/popping messages from the queue, there is no risk of deadlocking at all. In fact, message passing via concurrent queues is pretty much the safest way of parallel programming there is. (It's also the idea behind https://github.com/pure-data/pure-data/pull/1357)

millerpuckette:
OK, I'll cede to your deeper understanding of the issues, even if I still think letting GUI calls get out of order can potentially hang the GUI... different threads taling to the same TCL tag could trip this.

Issue #2406, opened 2024-08-23
expr's string-functions cause FTBFS on VisualStudio


umlaeute:
Since cad91dc791c7b17972d4d508f6bf6a93f1ff0613 `expr` supports a number of string functions.

Unfortunately, this makes the [Appveyor CI](https://ci.appveyor.com/project/umlaeute/pure-data) (where be build using VisualStudio) fail,due to missing symbols:

```
  Creating library ../bin/pd.lib and object ../bin/pd.exp
x_vexp_fun.obj : error LNK2019: unresolved external symbol _snprintf referenced in function _ex_exp
x_vexp_fun.obj : error LNK2019: unresolved external symbol _strcasecmp referenced in function _ex_exp
x_vexp_fun.obj : error LNK2019: unresolved external symbol _strncasecmp referenced in function _ex_exp
../bin/pd.dll : fatal error LNK1120: 3 unresolved externals
```

(this is just here to track the issue in case anybody is wondering why the CI is currently failing)

PR #2407, opened 2024-08-23
expr: MSVC compat for snprintf/strcasecmp/strcasecmp

umlaeute:
Since https://github.com/pure-data/pure-data/commit/cad91dc791c7b17972d4d508f6bf6a93f1ff0613 expr supports a number of string functions.

Unfortunately, some of these functions are named differently on Visual Studio.
This PR does the olde define magic to use usable versions of `snprintf()`/`strcasecmp()`/`strcasecmp()`

Closes: https://github.com/pure-data/pure-data/issues/2406

### DIFF
--- a/src/m_private_utils.h
+++ b/src/m_private_utils.h
@@ -18,6 +18,63 @@
 # include "config.h"
 #endif
 
+/* ------------------------------ atomics ----------------------------------- */
+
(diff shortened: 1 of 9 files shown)


umlaeute:
ah no, it works for `snprintf()` but not yet for `strcasecmp()`.

PR #2419, opened 2024-09-01
doc updates

porres:
Final updates for the final 0.55-1 release


Spacechild1:
> ALSO NOTE: As of Pd 0.55-1 \, [readsf~] can files in folders added via [declare]!

First off, there is a word missing ) Second, I would rather say:

> ALSO NOTE: As of Pd 0.55-1 \, [readsf~]'s "open" method searches in the canvas environment, just like [soundfiler] and similar objects. This means you can now use paths relative to parent canvases or folders added via [declare].

This would make it clear that it's not only about folders added via `[declare]` and parent canvases are also considered. A very typical situation is that you have a `[readsf~]` in an abstraction and want to open a file relative to the *calling* canvas. This is now possible.

porres:
> We should make it clear that it's not only about folders added via [declare], but that parent canvases are also considered. A very typical situation is that you have a [readsf~] in an abstraction and want to open a file relative to the calling canvas. This is now possible.

never knew this was possible :)

Spacechild1:
> ALSO NOTE: As of Pd 0.55-1 , [readsf~]'s "open" method searches in the canvas environment, just like [soundfiler] and similar objects. This means you can now use paths relative to parent canvases or folders added via [declare].

D'oh! Actually, this only works if you explicitly do `[declare -lib .]` in the parent canvas. Here's a possible rewording:

> ALSO NOTE: As of Pd 0.55-1 , [readsf~]'s "open" method searches in the canvas environment, just like [soundfiler] and similar objects. This means you can now use paths relative to folders added via [declare] - in this canvas or in parent canvases.

What do you think? Sorry for the confusion!

porres:
> D'oh! Actually, this only works if you explicitly do [declare -lib .] in the parent canvas. Here's a possible rewording:

yeah, as I suspected, problem is that I did a flawed test and it seemed to work, to my surprise, when I knew for long that this wasn't possible.

I guess I need to go back to my previous description. And the way [declare] works for abstractions is already defined in its help file...

Spacechild1:
> that I did a flawed test

me too...

> I guess I need to go back to my previous description

Yes. Sorry again.

Issue #2422, opened 2024-09-04
allow hiding the name of an array


umlaeute:
putting an array into a graph, will show the name of the array at the top of the graph.
putting multiple arrays into a single graph, will stack the names.

this is mostly nice, but somewhat annoying when using graphs in a GOP-abstraction.
E.g. building an oscilloscope object, you would typically use a $0-qualified named in a GOP-abstraction.
The actual name is just an implementation detail, but it will be shown to the user:

![array-gop](https://github.com/user-attachments/assets/f92ad355-8b81-4e8f-8cb7-346c7fb6aeeb)

i'd like to avoid this.

therefore i propose a a new message to hide the name of the array:

```
[labelvis 0(
|
[s array]
```

umlaeute:
i just noticed that there has been a flag to hide the name since the beginning of the git-history: https://github.com/pure-data/pure-data/blob/79305da10834f92e45912cfdf4e84040af839282/src/g_array.c#L120

and it is actually used if saved correctly in the patch.

so the only thing i had to do in f9149da was to add a message to set this flag...

Issue #2421, opened 2024-09-04
array  ticks draw outside of gop


umlaeute:
if we set starting point for the `xticks` resp `yticks` of an array outside the value range for the the given axis, the ticks are continued outside of the array.

Consider an array with a xrange of 0..100 and a y-range of -1..+1.

Letting the ticks start at the origin, everything looks nice:
![normal](https://github.com/user-attachments/assets/38e5d331-b175-48c4-ae87-70be23adba0a)

we can shift the ticks, by using an offset:
![within](https://github.com/user-attachments/assets/5a54d735-1b73-4b96-87fa-fafb8cd7d7c2)

however, if we shift beyond the allowed range for a given axis, the ticks are extended in the exceeded direction.

E.g. with a tick-offset outside the upper-limits:
![outofbounds+](https://github.com/user-attachments/assets/f5fec8c5-5a5f-458a-8bde-aec193a11364)

Resp. with a tick-offset outside the lower-limits:
![outofbounds-](https://github.com/user-attachments/assets/7117cb45-659c-4024-ac81-ed8775acba14)

is this intentional? (i cannot think of a good use-case)

Issue #2036, opened 2023-06-29
on-canvas arrays cannot have names starting with a single dash


umlaeute:
steps to reproduce:
1. create a new patch
2. Put a new array (<kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>a</kbd>
3. in the dialog, set the the *Name* of the array to `-dash` (or some other name that starts with a single dash)
4. Click on <kbd>OK</kbd> to create the array
5. notice that the array is labeled `--dash` (with **two** dashes)

The array can only be accessed with the double-dash name (e.g. `[tabread --dash]`)

6. Right-click the array to open its *Properties*
7. Notice that the name shows as `-dash` (single dash)

umlaeute:
it seems that the GUI and the core do not use the same escaping scheme:
- the GUI uses the escaping code for gatom dialogs, assuming that an empty label is transmitted as `-`, and thus a single leading `-` is stripped from the name
- the core doesn't know about this (since there's no point in having arrays without a name)

Issue #1838, opened 2022-11-24
allow patch to save array's color/width and include it in properties


porres:
At least these 2 could be saved with the patch and we could also have this setting under its preferences

porres:
moreover, perhaps, setting its visibility and mouse interaction as well (why not?)

porres:
@Spacechild1 any reason why this would be hard to do (asking cause you added these)

porres:
Another idea, allow us to send a message to arrays to make it save contents or not. This is the only thing that it can't do yet via messages

PR #2427, opened 2024-09-05
improve garrays: add message to hide name, save properties

umlaeute:
- add a `visname` message to *garray* to hide the name
   Closes: https://github.com/pure-data/pure-data/issues/2422
  (this is automatically persisted, as the `hidename` flag has been there forever... Just without an option to set it)
- add a `keep` message to *garray* to set the "saveit" flag
  Closes: https://github.com/pure-data/pure-data/issues/2428
- store the following properties when saving an array:
   - `color`
   - `width` (linewidth)
  (they are only stored if compatibility is >0.52 where these messages were introduced)
  Closes: https://github.com/pure-data/pure-data/issues/1838 (well partly, there's no corresponding properties dialog entry)
  
  i think these properties are meant to be persistent (you typically either need them for an array or you don't, but not both)
  the `vis` and `edit` properties are kept as runtime configuration and are not saved (as i think they are typically changed at runtime, and also can cause much confusion when being saved in the "wrong" state)
  th e`visname` property is stored in the patch (as was always the case)
- changing the `style` now no longer resets the line width (it just makes sure, that the minimum width (which is different for the points case) is not exceeded)
- store the following properties when saving a graph (and they are non-empty):
   - `xticks`, `yticks`
   - `xlabels`, `ylabels`
- confine `[xy]ticks` to the graph size
  Closes: https://github.com/pure-data/pure-data/issues/2421
- force a redraw when adding more garrays to a graph
   Closes: https://github.com/pure-data/pure-data/issues/1486
- force a redraw when deleting garrays from graph
   Closes: https://github.com/pure-data/pure-data/issues/2464
- force a redraw when renaming garrays
   Closes: https://github.com/pure-data/pure-data/issues/2465
- switch to edit-mode when creating graphs (and as a consequence garrays), and follow the mouse-pointer for positioning
  Closes: https://github.com/pure-data/pure-data/issues/1088
- Fix undo for adding garrays
  Closes: https://github.com/pure-data/pure-data/issues/1546
- Fix updating of arrays on rename (and have [tab...~] mark arrays as used in DSP)
  Closes: https://github.com/pure-data/pure-data/issues/2533
- Closes: https://github.com/pure-data/pure-data/issues/2036

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -57,6 +57,7 @@ static int dsparray_get_array(t_dsparray *d, int *npoints, t_word **vec,
         else
         {
             gpointer_setarray(&d->d_gp, garray_getarray(a), *vec);
+            garray_usedindsp(a);
             return 1;
         }
     }
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -9,6 +9,7 @@
 #include "g_canvas.h"
 
     /* common struct for reading or writing to an array at DSP time. */
+#define MAX_PHASE 0x7fffffff
 typedef struct _dsparray
 {
     t_symbol *d_symbol;
@@ -94,7 +95,7 @@ static void arrayvec_set(t_arrayvec *v, int argc, t_atom *argv)
                 v->v_vec[i].d_symbol = &s_;
         else
         {
-            v->v_vec[i].d_phase = 0x7fffffff;
+            v->v_vec[i].d_phase = MAX_PHASE;
             v->v_vec[i].d_symbol = argv[i].a_w.w_symbol;
         }
     }
@@ -119,7 +120,7 @@ static void arrayvec_init(t_arrayvec *v, void *x, int rawargc, t_atom *rawargv)
     for (i = 0; i < v->v_n; i++)
     {
         v->v_vec[i].d_owner = x;
-        v->v_vec[i].d_phase = 0x7fffffff;
+        v->v_vec[i].d_phase = MAX_PHASE;
         gpointer_init(&v->v_vec[i].d_gp);
     }
     arrayvec_set(v, argc, argv);
@@ -187,11 +188,11 @@ static t_int *tabwrite_tilde_perform(t_int *w)
         if (phase >= endphase)
         {
             tabwrite_tilde_redraw(d->d_symbol);
-            phase = 0x7fffffff;
+            phase = MAX_PHASE;
         }
         d->d_phase = phase;
     }
-    else d->d_phase = 0x7fffffff;
+    else d->d_phase = MAX_PHASE;
 noop:
     return (w+4);
 }
@@ -228,10 +229,10 @@ static void tabwrite_tilde_stop(t_tabwrite_tilde *x)
 {
     int i;
     for (i = 0; i < x->x_v.v_n; i++)
-        if (x->x_v.v_vec[i].d_phase != 0x7fffffff)
+        if (x->x_v.v_vec[i].d_phase != MAX_PHASE)
     {
         tabwrite_tilde_redraw(x->x_v.v_vec[i].d_symbol);
-        x->x_v.v_vec[i].d_phase = 0x7fffffff;
+        x->x_v.v_vec[i].d_phase = MAX_PHASE;
     }
 }
 
@@ -307,10 +308,10 @@ static t_int *tabplay_tilde_perform(t_int *w)
     if (phase >= endphase)
     {
         int i, playing = 0;
-        d->d_phase = 0x7fffffff;
+        d->d_phase = MAX_PHASE;
             /* set the clock when all channels have run out */
         for (i = 0; i < x->x_v.v_n; i++)
-            if (x->x_v.v_vec[i].d_phase < 0x7fffffff)
+            if (x->x_v.v_vec[i].d_phase < MAX_PHASE)
                 playing = 1;
         if (!playing)
             clock_delay(x->x_clock, 0);
@@ -349,7 +350,7 @@ static void tabplay_tilde_list(t_tabplay_tilde *x, t_symbol *s,
     int i;
     if (start < 0) start = 0;
     if (length <= 0)
-        x->x_limit = 0x7fffffff;
+        x->x_limit = MAX_PHASE;
     else
         x->x_limit = (int)(start + length);
     for (i = 0; i < x->x_v.v_n; i++)
@@ -360,7 +361,7 @@ static void tabplay_tilde_stop(t_tabplay_tilde *x)
 {
     int i;
     for (i = 0; i < x->x_v.v_n; i++)
-        x->x_v.v_vec[i].d_phase = 0x7fffffff;
+        x->x_v.v_vec[i].d_phase = MAX_PHASE;
 }
 
 static void tabplay_tilde_tick(t_tabplay_tilde *x)
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -64,6 +64,33 @@ static int dsparray_get_array(t_dsparray *d, int *npoints, t_word **vec,
     return 0;
 }
 
+static int dsparray_reset_array(t_dsparray *d) {
+    int size0 = 0, size1 = 0;
+    int phase = d->d_phase;
+    t_word *vec0 = 0, *vec1 = 0;
+    t_garray *a = 0;
+    if (d->d_symbol == &s_)
+        return 0;
+    a = (t_garray *)pd_findbyclass(d->d_symbol, garray_class);
+    if(!a)
+        return 0;
+    if (garray_getfloatwords(a, &size1, &vec1) && dsparray_get_array(d, &size0, &vec0, 0)) {
+        if(vec0 == vec1)
+                /* no change */
+            return 0;
+    }
+
+    gpointer_unset(&d->d_gp);
+#if 0
+        /*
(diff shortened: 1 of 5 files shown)


umlaeute:
> - `xticks`, `yticks`
> - `xlabels`, `ylabels`

i wonder whether these should be stored as a *graph* property, rather than an *array* property.
(conceptually, they belong to the former)

millerpuckette:
They should, yep.

(I haven't looked but I suppose this PR makes them save with the patcher 
now - I should have done that years ago :)

cheers

M

porres:
> (conceptually, they belong to the former)

but it's quite specific to array visualization, while 'graph' for me is a more general idea. So I say it should be in array properties

I suppose this also closes https://github.com/pure-data/pure-data/issues/1838

porres:
If I may, I'd like to also point to relevant issues that we can relate to these changes/fixes and close them too

https://github.com/pure-data/pure-data/issues/1486

https://github.com/pure-data/pure-data/issues/2428

https://github.com/pure-data/pure-data/issues/1088

umlaeute:
> but it's [note: ticks and labels] quite specific to array visualization, while 'graph' for me is a more general idea. So I say it should be in array properties

but it's not specific for a single array.
if a single graph holds multiple arrays, you can only set one set of ticks/labels. so they (currently) are a property of the graph (which they really are).
this could be changed (but i think it is out-of-scope for this PR)

porres:
> but it's not specific for a single array.

yeah, that's right!

porres:
Hi, let me please inquire this PR to also consider and include a minor change to help with the documentation

I've been hoping for a while to get help changing the source code just so we can have a help file dedicated to arrays, see https://github.com/pure-data/pddp/issues/24

What I want to highlight is that I really hope we could have a better help file for arrays and not need to call 2 examples from the control tutorial series instead. I think it should be the other way around. We could have a single file to introduce arrays in the tutorial and tell people to look into the help file for details.

Ideally, the [table] object should also call a separate help file, just like [array].

umlaeute:
> Hi, let me please inquire this PR to also consider and include a minor change to help with the documentation

this is definitely not a *minor* change, so I don't think i will tackle this.

umlaeute:
undo/redo:
- [x] add graph
- [x] delete graph
- [x] add array-in-new graph
- [x] add array-in-last-graph
       - undo works, but redo does not redraw (and therefore hides label)
- [x] delete array from graph

porres:
Hi @umlaeute about https://github.com/pure-data/pure-data/issues/1486 I also mentioned that if you delete the array with a "clear" message sent to the graph, the array goes away but its name keeps being shown. If you move the graph around, it disappears. Have you fixed this as well? Should I open a new issue for that?

Also, when renaming an array with the 'rename' message to an array, the new name doesn't show right away. You also have to select the graph and move it to force an update and show the new name. This is not mentioned anywhere. Is it enough to brin this up here or should I open new issues?

porres:
here are some other array related stuff that maybe could be considered here

- https://github.com/pure-data/pure-data/issues/2079
- https://github.com/pure-data/pure-data/issues/2109
- https://github.com/pure-data/pure-data/issues/901

porres:
> if you delete the array with a "clear" message sent to the graph, the array goes away but its name keeps being shown.

geez, I just copied the text without really thinking, and now that I am trying to test this again I don't even know how one would actually do that :) 

How can someone send a 'clear' message to a graph? Can anyone make sense of what I meant?  otherwise we should just ignore it :/

porres:
Ok, I think I know what I mean now.

We are able to delete an array in the properties, but when I do that, the array name keeps being shown! Not sure if there is a message we can send to the array that would make it delete itself.

porres:
> We are able to delete an array in the properties, but when I do that, the array name keeps being shown! Not sure if there is a message we can send to the array that would make it delete itself.

Geez, this is actually more convoluted. In the properties, if you delete the array, you remain with the canvas properties. In this case, the array is deleted but the name keeps shown. If you cancel the canvas properties, the name remains, but disappears when you select it and move it around. On the other hand, if you hit 'ok' on the canvas properties, the array name does in fact disappear.

The best way to manage this, I think, is also deleting the name right away when the array is deleted. It does seem weird anyway that when the array is delete the name is still there. Then it doesn't matter if we hit ok or cancel in the canvas properties next.

umlaeute:
@millerpuckette seeing that your recent commits (22f4801a0cae1d26297359d6da446bb8a9759ea5, b4ad52871dc5375584d31bec90cd46980f3dde9e) improve some editing issues with GUIs, i'd like to gently ping for this PR.

from your comment:
> I should have done that years ago

... i figure that you are OK with this PR in principle.

umlaeute:
@millerpuckette any chance for this to make it into 0.56?

umlaeute:
@millerpuckette ping

porres:
So, I'm testing this, also to document it, and I see that edit/vis is not saved with the array, to my surprise. My intuition would expect them to be saved as well.

> the vis and edit properties are kept as runtime configuration and are not saved (as i think they are typically changed at runtime, and also can cause much confusion when being saved in the "wrong" state)

I've been using 'edit 0' consistently the idea is that it's always non editable, cause the data is "read/display only". I'd like to require it to be able to be saved to the patch as well, cause it'd make it easier for me not needing to always initialize the patch for this.

As for hiding the array, I haven't really been using this, so I'm not sure. I don't have the same strong opinion, but I don't think it'd hurt to let the patch save your configuration.

cheers

umlaeute:
> > the vis and edit properties are kept as runtime configuration and are not saved (as i think they are typically changed at runtime, and also can cause much confusion when being saved in the "wrong" state)
> 
> I've been using 'edit 0' consistently the idea is that it's always non editable, cause the data is "read/display only".

sure.
the reasoning was (and is) that arrays *typically* are read/writeable, so that being no longer able to "draw" them will breed confusion.
if there's a actual code that turns editing off, i figure that it's easier to understand why the garray is not editable.
and typically you will want to use this in abstractions/subpatches where you have additional code anyway, so just set the *editable* state while being there.

"edit" and "vis" are different from other settings that only change the visual appearance: the former have the potential at making the object appear to be "broken", the latter don't.

i'm aware that this has a number of unverified assumptions (what's "typical", really?), so my argument is rather weak.
if others disagree and want to create a PR, I don't think i'll oppose :sunglasses:

Spacechild1:
> the reasoning was (and is) that arrays typically are read/writeable

I wouldn't necessarily agree with that. A very typical use case is to display the wave form of an audio file, in which case the array should be read-only.

> so that being no longer able to "draw" them will breed confusion.

I agree, but the actual reason is that we cannot set the "edit" mode in the array properties. Once it's in the properties dialog, we can also save it.

Similarly, it would be great if the color and line width could be set in the array properties. I understand that this was likely out of scope for this PR, but it would be great if you could still consider this, since you're much more competent in Tcl than me :) It would make things *much* more consistent.

---

Another thing: I just noticed that the "visname" message is not reflected in the canvas properties dialog!
1. send `[visname 0(`
2. right click -> properties -> "hide object name and arguments" is not ticked 

Even worse: the "visname" state is saved with the array. When I save the patch and reopen it, the name is hidden, but it's still not reflected in the canvas properties. This is *really* confusing!

Actually, shouldn't "visname" be a *canvas* message instead? After all, it is really part of the canvas properties, not the array properties. The message would also be useful for GOP abstractions.

porres:
> I agree, but the actual reason is that we cannot set the "edit" mode in the array properties. Once it's in the properties dialog, we can also save it.
> 
> Similarly, it would be great if the color and line width could be set in the array properties. I understand that this was likely out of scope for this PR, but it would be great if you could still consider this, since you're much more competent in Tcl than me :) It would make things much more consistent.

yeah, I agree and had opened this yesterday :)

https://github.com/pure-data/pure-data/issues/2618

porres:
> i'm aware that this has a number of unverified assumptions (what's "typical", really?), so my argument is rather weak. If others disagree and want to create a PR, I don't think i'll oppose 😎

I guess me and @Spacechild1 already consider at least 'edit 0' should be saved? Well, maybe we can open a new issue.

> Another thing: I just noticed that the "visname" message is not reflected in the canvas properties dialog!

That didn't confuse me as that is for the subpatch name, not the array name, which is actually the name we need to change things like the array zie, by the way, see https://github.com/pure-data/pure-data/pull/627#issuecomment-2985059730 which is kinda related to this.

Also, the way the names are displayed are different, for a GOP soboatch the name is displayed inside the rectangle, whereas the array name shows on top of it. So I think this should be a setting of the array properties, not the canvas properties

Spacechild1:
> I guess me and @Spacechild1 already consider at least 'edit 0' should be saved? Well, maybe we can open a new issue.

This is what I said:

> I agree, but the actual reason is that we cannot set the "edit" mode in the array properties. Once it's in the properties dialog, we can also save it.

In other words: let's please only save properties that are also available in the array properties dialog. At the moment, this does not include "edit".

> That didn't confuse me as that is for the subpatch name, not the array name, which is actually the name we need to change things 

D'oh, you are of course right! I don't know why I thought that "hide object name" would hide/show the array name...

---

Anyway, I'm still not quite happy that we are currently saving several properties (visname, color, width) that are not available in the array properties dialog. May I gently nudge @umlaeute to activate his Tcl super powers and implement https://github.com/pure-data/pure-data/issues/2618?

PR #2435, opened 2024-09-18
fix libpd ios build by wrapping dlopen call with HAVE_LIBDL

danomatika:
This fixes an oversight when a new function was added to s_loader.c: iOS doesn't support dlopen, so we need to wrap it with `HAVE_LIBDL`.

### DIFF
--- a/doc/5.reference/list-help.pd
+++ b/doc/5.reference/list-help.pd
@@ -661,7 +661,7 @@
 #X obj 182 527 float;
 #X obj 143 527 text;
 #X obj 230 527 symbol;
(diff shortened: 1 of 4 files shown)


danomatika:
@millerpuckette If this is approved by the others, can you merge? It's a block for making a libpd release as, currently, building on iOS is broken.

umlaeute (approved):
lgtm.

umlaeute:
merged into the `develop` branch.

danomatika:
Aw, `master` would be better. ;)

Issue #2392, opened 2024-07-27
JACK autoconnect only connects input (but not output)


umlaeute:
i'm runnig Pd with JACK on top of the pipewire audio daemon.

when I start Pd, it only connects a device to Pd's input, but Pd's output is not connected anywhere, so I get no sound out by default:

![qjackctl](https://github.com/user-attachments/assets/57a90828-dc51-4991-9eb7-5b329da21585)

the available inputs are:
- **`Tiger Lake-LP Smart Sound Technology Audio Controller Digital Microphone`** (capture)
- `Tiger Lake-LP Smart Sound Technology Audio Controller HDMI / DisplayPort 1 Output` (monitor)
- `Tiger Lake-LP Smart Sound Technology Audio Controller HDMI / DisplayPort 2 Output` (monitor)
- `Tiger Lake-LP Smart Sound Technology Audio Controller HDMI / DisplayPort 3 Output` (monitor)
- `Tiger Lake-LP Smart Sound Technology Audio Controller Headphones Stereo Microphone` (capture)
- `Tiger Lake-LP Smart Sound Technology Audio Controller Speaker` (monitor)

the available outputs are:
- `Tiger Lake-LP Smart Sound Technology Audio Controller HDMI / DisplayPort 1 Output` (playback)
- `Tiger Lake-LP Smart Sound Technology Audio Controller HDMI / DisplayPort 2 Output` (playback)
- `Tiger Lake-LP Smart Sound Technology Audio Controller HDMI / DisplayPort 3 Output` (playback)
- **`Tiger Lake-LP Smart Sound Technology Audio Controller Speaker`** (playback)

(I've marked the *primary* devices in bold)

afaict, Pd assumes that the primary input device has the same name as the primary output device, and only tries to connect one.
https://github.com/pure-data/pure-data/blob/4cdf007e337b3e5f1b9c59aba42d85066b6009c3/src/s_audio_jack.c#L546-L547

jamshark70:
I just built from latest sources (cloned just now, with `--enable-jack`) and find that the behavior is unchanged. Upon launching Pd (`./bin/pd`) and enabling DSP:

- "Alder Lake PCH-P High Definition Audio Controller Digital Microphone" is connected to Pure Data's 2 inputs.
- Pd's 2 outputs are *not* connected to "Alder Lake PCH-P High Definition Audio Controller Speaker + Headphones"

[...]

PR #2437, opened 2024-09-19
JACK: autoconnections

umlaeute:
previously, Pd would (internally) get a list of JACK-clients, and then try to connect both input and output to the very first available client (prioritizing any client that happens to be called `alsa_pcm`).

This worked nicely in the olden days.

Modern systems (e.g. using PipeWire as a JACK replacement daemon), will use more verbose naming schemes, which might (and actually *will*, at least on my systems) result in one soundcard registering multiple names for both input and output, and the primary input client-name can be different from the primary output client-name:

qjackctl

This PR improves the logic on how JACK clients are selected:
- connect Pd's `[adc~]` to the first JACK client that has audio *source* ports
- connect Pd's `[dac~]` to the first JACK client that has audio *sink* ports (even if different from the above)
- for both directions, prefer clients that have *physical* ports over ones that don't.

it also only considers clients that have ports of the `JACK_DEFAULT_AUDIO_TYPE` (which is what we are using to register the Pd JACK client as well)

- Closes: https://github.com/pure-data/pure-data/issues/2392

### DIFF
--- a/src/s_audio_jack.c
+++ b/src/s_audio_jack.c
@@ -88,6 +88,8 @@ static volatile int jack_dio_error;
 static volatile int jack_didshutdown;
 static t_audiocallback jack_callback;
 static int jack_should_autoconnect = 1;
+static int jack_defaultsource = 0;
+static int jack_defaultsink = 0;
 static int jack_blocksize = 0; /* should this be PERTHREAD? */
 #ifdef THREADSIGNAL
 t_semaphore *jack_sem;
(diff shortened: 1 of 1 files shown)


reduzent:
I took it for given that with new names for soundcards it's impossible for Pd to know what port to conncect to automatically. Currently,  I start Pd with a cumbersome script that connects Pd's output to the port I want.

Picking the one having physical ports makes totally sense.

umlaeute:
of course we still do not know, which device is the "first" one (w or w/o physical ports)  - that's up to the JACK daemon.
in practice i found, that my patch always did what i wanted/expected.

of course, if the setup is so complicated that Pd cannot possibly figure out what to do, there's always `-nojackconnect` to prevent autoconnecting altogether.

PR #2439, opened 2024-09-21
add 'clear' message to [vcf~]

### LINKED ISSUE #2438: add 'clear' message to [vcf~]
Seems like this is the only filter without a clear and/or set message in Pd...

so... for the sake of consistency...

no382001:
- closes #2438

### DIFF
--- a/src/d_osc.c
+++ b/src/d_osc.c
@@ -342,12 +342,20 @@ static void sigvcf_dsp(t_sigvcf *x, t_signal **sp)
                 &x->x_cspace, (t_int)sp[0]->s_n);
 }
 
+static void sigvcf_clear(t_sigvcf *x)
+{
+    x->x_cspace.c_re = 0;
+    x->x_cspace.c_im = 0;
+}
+
 static
 void sigvcf_setup(void)
 {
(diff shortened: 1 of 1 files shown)


porres:
hi, I don;t think we need a 'set' message, right? that's for lower level filters.

would you considering just implementing a 'clear' message?

no382001:
sure, i just added that bc of the issue's wording.

umlaeute (changes requested):
the `clear` message for other filters only clear the hidden internal state, not the parameters set from outside:
- https://github.com/pure-data/pure-data/blob/177350fc4999b74ea28a12ba2981baa6ae04c6f0/src/d_filter.c#L122-L125
- https://github.com/pure-data/pure-data/blob/177350fc4999b74ea28a12ba2981baa6ae04c6f0/src/d_filter.c#L165-L168
- https://github.com/pure-data/pure-data/blob/177350fc4999b74ea28a12ba2981baa6ae04c6f0/src/d_filter.c#L320-L323
- ...

so i think we shoud do the same here (which basically boils down to leaving the `x_f` ontouched.

umlaeute (approved):

Issue #2431, opened 2024-09-10
allow the var() function to also store a value


porres:
we can store values in variables as in [expr x = $f1]

But something like [expr var($s1) = $f2] doesn't work, the object doesn't create and we get a 'expr: syntax error'

no382001:
Hi, can i work on this?

porres:
Sure, but now this is already in the radar of Shahrokh, you can make a PR and propose a fix yourself here

PR #2442, opened 2024-09-23
allow a function to store a value

no382001:
related to #2431

`we can store values in variables as in [expr x = $f1].
But something like [expr var($s1) = $f2] doesn't work, the object doesn't create and we get a 'expr: syntax error'`

since (all of the?) functions in `expr` resolve to a concrete value, to solve this problem, i'd propose another function that returns only a variable reference to the value

so in the case of `[expr function($s1) = $f2] ` where `function` is only syntactically correct in its current position, if a variable matches `$s1` it would internally resolve to: first `[expr a = 2.5]` where `a` is the variable and `2.5` is the value of `$f2`, and then `a` would be bound to that variable.

the function implementation is identical to `ex_var` with the exception of variable binding, of course.

i'm open to suggestions.

### DIFF
--- a/src/x_vexp.c
+++ b/src/x_vexp.c
@@ -834,12 +834,22 @@ ex_checklval(struct expr *e, struct ex_ex *eptr)
         extmp = eptr->ex_end;
         while (eptr->ex_type && eptr != extmp) {
                 if (eptr->ex_type == ET_OP && eptr->ex_op == OP_STORE) {
+
                         if (eptr[1].ex_type != ET_VAR &&
                             eptr[1].ex_type != ET_SI &&
                             eptr[1].ex_type != ET_TBL) {
-                                post("expr: '%s' - Bad left value: ", e->exp_string);
-                               return (1);
-                         }
+
+                                if (eptr[1].ex_type == ET_FUNC) {
+                                        t_ex_func *f = (t_ex_func *) eptr[1].ex_ptr;
+                                        if (strcmp(f->f_name, "ref")) {
+                                                post("expr: '%s' - only ref() is permitted on lhs", e->exp_string);
+                                                return (1);
+                                        }
+                                } else {
+                                        post("expr: '%s' - Bad left value: ", e->exp_string);
+                                        return (1);
+                                }
+                        }
                 }
                 eptr++;
         }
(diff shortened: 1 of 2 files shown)


porres:
> the function implementation is identical to ex_var with the exception of variable binding, of course.

sorry, I didn't get the difference

no382001:
the internal logic is the same, it just does not resolve the variable by `max_ex_var`
https://github.com/pure-data/pure-data/blob/1ed152ab5a0dfae13d701a82c53035b8ca484a2b/src/x_vexp_fun.c#L1533C9-L1543C20
https://github.com/no382001/pure-data/blob/1ed152ab5a0dfae13d701a82c53035b8ca484a2b/src/x_vexp_fun.c#L1504C9-L1510C32

no382001:
the internal logic is the same, it just does not resolve the variable by `max_ex_var`
https://github.com/pure-data/pure-data/blob/1ed152ab5a0dfae13d701a82c53035b8ca484a2b/src/x_vexp_fun.c#L1533-L1542

```c
        switch(argv->ex_type) {
                case ET_SYM:
                        if (argv->ex_flags & EX_F_TSYM)
                                optr->ex_ptr = gensym(argv->ex_ptr);
                        else
                                optr->ex_ptr = argv->ex_ptr;
                        return;
```

porres:
Well, sorry, I'm really confused. Are you saying you did create a new function named function? 

Or is this just a PR that closes https://github.com/pure-data/pure-data/issues/2431 ? Meaning, if the PR is taken, can [expr var(a) = 2.5] work and set [v a] to '2.5' when taking a bang?

umlaeute:
i'm not sure whether i groked `function()` either.

to assign values to symbols defined at run-time, how about using a `value = assign(sym, value)` function?
as in:

```
[foo 10(
|
[expr assign($s1, $f2)]
|
|10\
```

umlaeute:
> i'm not sure whether i groked `function()` either.

after reading the code, i see that the problem is actually just one of bad wording in the description.
the newly introduced function is `ref()` (**not** ~`function()`~), so the syntax would be:
```
[expr ref($s1) = $f2]
```

i still think an `assign(name, value)` function is easier to understand.

porres:
Assign makes more sense, or var could just take one or two arguments. If two, the second is assigned to the first

no382001:
sorry for not being clear, i wasnt decided on the name thats why i wrote 'function', but your 'assign' actually makes more sense! and is considerably less messy

umlaeute:
to be honest, i do not really see how `var($s2, $f1)` is really readable.
but anyhow, Shahrokh is the one to decide.

Spacechild1:
What about `set(name, value)`? Should be pretty self-explaining.

no382001:
wouldn't preventing it from being callable before a `=` make the syntax less intuitive? if you can already assign a value to a variable like this: `[expr x = 1]`, then calling `set(name, value)` to reassign a value might feel less intuitive.

umlaeute:
depends.
`name = value` is certainly intuitive.
however, i find `ref(name) = value` *less* intuitive than `set(name, value)`.

(i also think that `=` assignments shouldn't have gone into `expr` in the first place; there's a reason that s-expressions don't use it)

sebshader:
> What about `set(name, value)`? Should be pretty self-explaining. I'd find this a bit more concise than `assign`, but I'd also be ok with the latter.

there's no 'upvote' on github so.. +1 vote for set
though I have to say also, the string processing and accessing variables makes sense to be part of [expr] but imo setting variables seems somewhat out of scope

PR #2443, opened 2024-09-24
add support for 8-bit wav audios in [soundfiler]

### LINKED ISSUE #2393: add support for 8-bit wav audios in [soundfiler]
I was trying to open some 8-bit audio and just realized [soundfiler] can't load or save 1 byte formats.

I know it's a cheap and probably undesired format, but it just posed as an issue to me and it feels like an unnecessary restriction.

no382001:
closes #2393

- add support for 8-bit pcm audio for `[soundfiler]`

note: `[write]` still defaults to 2 bytes when banging `[write ~/test.wav sample]` so to test writing, use `[write -bytes 1 ~/test.wav sample]`

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -23,7 +23,7 @@ objects use Posix-like threads. */
 
 /* Supported sample formats: LPCM (16 or 24 bit int) & 32 or 64 bit float */
 
-#define VALID_BYTESPERSAMPLE(b) ((b) == 2 || (b) == 3 || (b) == 4 || (b) == 8)
+#define VALID_BYTESPERSAMPLE(b) ((b) == 1 || (b) == 2 || (b) == 3 || (b) == 4 || (b) == 8)
 
 #define MAXSFCHANS 64
 
@@ -445,7 +445,15 @@ static void soundfile_xferin_sample(const t_soundfile *sf, int nvecs,
     t_sample *fp;
     for (i = 0, sp = buf; i < nchannels; i++, sp += sf->sf_bytespersample)
     {
-        if (sf->sf_bytespersample == 2)
+        if (sf->sf_bytespersample == 1)
+        {
+            for (j = 0, sp2 = sp, fp = vecs[i] + framesread;
+                j < nframes; j++, sp2 += sf->sf_bytesperframe, fp++)
+            {
+                *fp = ((t_sample)(*sp2) - 128) / 128.0;
+            }
+        }
+        else if (sf->sf_bytespersample == 2)
         {
             if (sf->sf_bigendian)
             {
@@ -545,7 +553,15 @@ static void soundfile_xferin_words(const t_soundfile *sf, int nvecs,
     size_t j;
     for (i = 0, sp = buf; i < nchannels; i++, sp += sf->sf_bytespersample)
     {
-        if (sf->sf_bytespersample == 2)
+        if (sf->sf_bytespersample == 1)
+        {
+            for (j = 0, sp2 = sp, wp = vecs[i] + framesread;
+                j < nframes; j++, sp2 += sf->sf_bytesperframe, wp++)
+            {
+                wp->w_float = ((t_sample)(*sp2) - 128) / 128.0;
+            }
+        }
+        else if (sf->sf_bytespersample == 2)
         {
             if (sf->sf_bigendian)
             {
@@ -709,7 +725,7 @@ static int soundfiler_parsewriteargs(void *obj, int *p_argc, t_atom **p_argv,
         else if (!strcmp(flag, "bytes"))
         {
             if (argc < 2 || argv[1].a_type != A_FLOAT ||
-                ((bytespersample = argv[1].a_w.w_float) < 2) ||
+                ((bytespersample = argv[1].a_w.w_float) < 1) ||
                     !VALID_BYTESPERSAMPLE(bytespersample))
                         return -1;
             argc -= 2; argv += 2;
@@ -859,7 +875,17 @@ static void soundfile_xferout_sample(const t_soundfile *sf,
     for (i = 0, sp = buf; i < sf->sf_nchannels; i++,
         sp += sf->sf_bytespersample)
     {
-        if (sf->sf_bytespersample == 2)
+        if (sf->sf_bytespersample == 1){
+            for (j = 0, sp2 = sp, fp = vecs[i] + onsetframes;
+                j < nframes; j++, sp2 += sf->sf_bytesperframe, fp++)
+            {
+                int sample = (int)((*fp * 128.0) + 128);
+                if (sample < 0) sample = 0;
+                if (sample > 255) sample = 255;
+                *sp2 = (unsigned char)sample;
+            }
+        } 
+        else if (sf->sf_bytespersample == 2)
         {
             t_sample ff = normalfactor * 32768.;
             if (sf->sf_bigendian)
@@ -995,7 +1021,18 @@ static void soundfile_xferout_words(const t_soundfile *sf, t_word **vecs,
     for (i = 0, sp = buf; i < sf->sf_nchannels;
          i++, sp += sf->sf_bytespersample)
     {
-        if (sf->sf_bytespersample == 2)
+        
+        if (sf->sf_bytespersample == 1){
+            for (j = 0, sp2 = sp, wp = vecs[i] + onsetframes;
+                j < nframes; j++, sp2 += sf->sf_bytesperframe, wp++)
+            {
+                int sample = (int)((wp->w_float + 1.0) * 127.5);
+                if (sample < 0) sample = 0;
+                if (sample > 255) sample = 255;
+                *sp2 = (unsigned char)sample;
+            }
+        }
+        else if (sf->sf_bytespersample == 2)
         {
             t_sample ff = normalfactor * 32768.;
             if (sf->sf_bigendian)
--- a/src/d_soundfile_wave.c
+++ b/src/d_soundfile_wave.c
@@ -286,7 +286,7 @@ static int wave_readheader(t_soundfile *sf)
             bytespersample = swap2(format->fc_bitspersample, swap) / 8;
             switch (bytespersample)
             {
-                case 2: case 3: break;
+                case 1: case 2: case 3: break;
                 case 4: case 8:
                     if (formattag == WAVE_FORMAT_FLOAT) /* 32 bit int? */
                         break;


porres:
hi, thanks for this, but can you please leave the help file out of the PR? I usually handle those separately

porres:
@danomatika can you revise this please?

porres:
Hi @no382001 I forgot about [readsf~] and [writesf~] and have a new issue for them, see https://github.com/pure-data/pure-data/issues/2459

Do you think you can also take care of 8-bit support for both of them?

danomatika:
I would not consider this complete. It simply adds read support to WAVE but not write support. Also, IMO the other file types which would support 8 bit samples should also be updated similarly.

porres:
> It simply adds read support to WAVE but not write support.

Do you mean writing with [soundfiler] is not supported? Because the PR says it does

"note: [write] still defaults to 2 bytes when banging [write ~/test.wav sample] so to test writing, use [write -bytes 1 ~/test.wav sample]"

src/d_soundfile.c
```diff
@@ -859,7 +875,17 @@ static void soundfile_xferout_sample(const t_soundfile *sf,
     for (i = 0, sp = buf; i < sf->sf_nchannels; i++,
         sp += sf->sf_bytespersample)
     {
-        if (sf->sf_bytespersample == 2)
+        if (sf->sf_bytespersample == 1){
```
danomatika, line 878: Please match formatting: '{' below not on the same line.

danomatika:
> > It simply adds read support to WAVE but not write support.
> 
> Do you mean writing with [soundfiler] is not supported? Because the PR says it does
> 
> "note: [write] still defaults to 2 bytes when banging [write ~/test.wav sample] so to test writing, use [write -bytes 1 ~/test.wav sample]"

I haven't tried it, I just noticed there weren't any changes to the wave output section, however, you are write, I suppose it's not needed there if you force the output width. I'd need to test but cannot get to this until November.

porres:
> the other file types which would support 8 bit samples should also be updated similarly

For the sake of clarity can we list them all? I am assuming you mean all the supported files, right? Namely: wave, aiff, caf & next formats.

danomatika:
> > the other file types which would support 8 bit samples should also be updated similarly
> 
> For the sake of clarity can we list them all? I am assuming you mean all the supported files, right? Namely: wave, aiff, caf & next formats.

Yup. I can see it now: "I thought Pd can read 8 bit files, why doesn't my 8 bit AIFF work? etc..."

porres:
For the record, .aiff allows signed and unsigned 8-bit PCM, .wave unsigned only.

I also see in https://pubs.opengroup.org/external/auformat.html that the next format does support 8-bit linear PCM 

.caf does not support 8-bit files, couldn't find it in https://developer.apple.com/library/archive/documentation/MusicAudio/Reference/CAFSpec/CAF_spec/CAF_spec.html

porres:
@no382001 we're starting another release cycle and it'd be good to have something good to go around now so it could be merged early in the game. Are you able and willing to add more formats and work on other objects? thanks

umlaeute:
sidenote: i think "we are starting another release cycle" is probably raising too high expectations.

my understanding (which might be totally wrong) is, that miller does not expect any more bugfix releases for 0.55, so all the things that go into the `master` branch now, are already targeting the next major release (0.56).
technically this starts a new release cycle.

it doesn't say anything about *when* the release will actually be done.
esp. miller's note on ["[potentially buggy changes are] done _early_ in a release cycle"](https://github.com/pure-data/pure-data/commit/b2ee7fafcd98cad3638f8b04f5fcfca13d68d7ab) indicates (to me) that there's still plenty of time until the next release (so that there's enough time to discover these bugs).

of course i do not know anything about the release plans.

danomatika:
I would really appreciate not pushing feature changes like this without testing. We have had way too many instances in the last releases of adding too much to bugfixe `-1`, `-2` releases. The best result is that people have time other than test and the PR is ready to be merged when 0.56 starts to come together. It takes longer but it is much more sustainable, IMO, for all involved.

Spacechild1:
The thing is, people typically don't go ahead and test random PRs. Of course, PRs should be tested by the author (and ideally by other people) before merging. But it is also good for PRs to get merged early because quite a few people use the current master branch, e.g. for development, and so it is much more likely to catch bugs before the release.

What is not sustainable IMO is merging all PRs in a tight window of 2-3 weeks. I think *that* is a major reason for many bug fixes.

danomatika:
> What is not sustainable IMO is merging all PRs in a tight window of 2-3 weeks. I think _that_ is a major reason for many bug fixes.

I suppose this is what I am responding to: "quick, hurry up and finish it." Sustainable development can't always require "quick, hurry up" but part of the issue falls on our side as well: we can also prepare things early enough instead of waiting until an arbitrary couple of weeks. I am guilty of this as well as I don't have infinite time.

I think Alex's message would have been worked better as: "I see Miller is fixing a couple of things in readsf~." Maybe it's a good time to revisit this PR and fleshing it out for testing, etc."

danomatika:
> @no382001 we're starting another release cycle and it'd be good to have something good to go around now so it could be merged early in the game. Are you able and willing to add more formats and work on other objects? thanks

Actually I am making a lot of noise without reading in detail, as usual. "merged early in the game" is a good idea, assuming there is more than a couple weeks before 0.56 ;)

porres:
Hi, I saw Miller's remark on https://github.com/pure-data/pure-data/commit/b2ee7fafcd98cad3638f8b04f5fcfca13d68d7ab  that changes to these objects should be done early on the release cycle. Here's a full citation "_Improvements" to readsf~/writesf~ usually cause unexpected problems so this is done _early_ in a release cycle :)_"  

So I thought it'd be good timing opportunity to focus on this. And I wanna check if this is being considered for more work because if that's not happening I could try and adventure myself into this. I wanna take things to the next level with how I can and collaborate.

danomatika:
I would do it, I just didn't have the time before and probably don't have the time this week, however I will spend an hour on it tonight. My original comment was mainly that it's easy to *quickly* add support but that support should really work across all the formats with appropriate error message if the format doesn't support it. This PR is definitely useful as a start, I can try fleshing it out.

no382001:
@porres yes, i plan on finishing it soon. the PR as of now does not contain `aiff`, `caf` & `next` as noted by @danomatika earlier, i've tested the `wave` format, im just very lost on `aiff`, the 8bit support of this format afaik is handled in a quite liberal way in most audio players, especially the s8 part. conversion with `ffmpeg` and `sox` for example results in totally different headers.

danomatika:
> For the record, .aiff allows signed and unsigned 8-bit PCM, .wave unsigned only.
> 
> I also see in https://pubs.opengroup.org/external/auformat.html that the next format does support 8-bit linear PCM
> 
> .caf does not support 8-bit files, couldn't find it in https://developer.apple.com/library/archive/documentation/MusicAudio/Reference/CAFSpec/CAF_spec/CAF_spec.html

It's not as simple as this. Some formats support both unsigned and signed int values with the single byte, so the current sample conversion doesn't work in all cases. I am trying with test patches and a bunch of audio files in various sample rates and formats leftover from the sound file overhaul.

danomatika:
FYI: I am putting updates and tests with this into a feature branch: `feature/sf-8bit`

danomatika:
Ok, after looking through the file format spec again, the following seems to be the case for 8 bit samples:
* WAVE: 8 bit unsigned int, 16 & 24 bit signed int
* AIFF: 8, 16, 24 bit signed int
* NEXT: 8, 16, 24 signed int
* CAFF: 16, 24 signed int

This means there needs other be a way to check for signedness when converting 8 bit samples. I suppose I can add a new function call to the api, possible with an enum to allow for future expansion.

porres:
what's different than what I saw from checking the specs is that I mentioned .aiff also allows unsigned 8-bit PCM.

from page 2 [here](https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/AIFF/Docs/AIFF-1.3.pdf) I see

Data types
A C-like language will be used to describe data structures in this document. The data types used
are listed below:
char: 8 bits, signed. A char can contain more than just ASCII characters. It
can contain any number from -128 to 127 (inclusive).
unsigned char: 8 bits, unsigned. Contains any number from zero to 255 (inclusive).

I may be missing something, of course

danomatika:
> what's different than what I saw from checking the specs is that I mentioned .aiff also allows unsigned 8-bit PCM.
> 
> from page 2 [here](https://www.mmsp.ece.mcgill.ca/Documents/AudioFormats/AIFF/Docs/AIFF-1.3.pdf) I see
> 
> _Data types A C-like language will be used to describe data structures in this document. The data types used are listed below: char: 8 bits, signed. A char can contain more than just ASCII characters. It can contain any number from -128 to 127 (inclusive). unsigned char: 8 bits, unsigned. Contains any number from zero to 255 (inclusive)._
> 
> I may be missing something, of course

I saw that too, it's just a reference to the "types" which are used for the fields descrbing the chunks. GO to the "Sound Data Chunk" page and check the last paragraph:

> Sample Points
>
> Each sample point in a sample frame is a linear, 2's complement value. The sample points are
from 1 to 32 bits wide, as determined by the sampleSize parameter in the Common Chunk.
Sample points are stored in an integral number of contiguous bytes. One to 8 bit wide sample
points are stored in one byte, 9 to 16 bit wide sample points are stored in two bytes, 17 to 24 bit
wide sample points are stored in 3 bytes, and 25 to 32 bit wide samples are stored in 4 bytes.
When the width of a sample point is less than a multiple of 8 bits, the sample point data is left

"2's complement" is signed integer

danomatika:
Reading 8 bit WAVE, AIFF, and NEXT now working. Writing *should* work but the `-bytes` argument is being ignored in some places so output ends up being 16 bit. Tracking that down now.

danomatika:
Writing in the `feature/sf-8bit` branch should work now and an error message will be printed if trying to write an unsupported format, ie. `[open -bytes 1 test.caf<`

Someone please double check the sample conversions in the *xferin* and *xferout* functions. I am also not 100% sure about using the function for checking the signdnyss, but that seemed the easiest way and doesn't require adding another bool to the `t_soundfile` struct which is only used for a particular end case.

@umlaeute @Spacechild1 I notice that the [soundfiler] `-normalize` write option assumes 16 bit samples, see [1](https://github.com/pure-data/pure-data/blob/feature/sf-8bit/src/d_soundfile.c#L1685) and [2](https://github.com/pure-data/pure-data/blob/feature/sf-8bit/src/d_soundfile.c#L1719) Should this be updated to handle other sample ranges? I think it's been this way for a while...

danomatika:
I opened a new PR for the feature branch: #2498

porres:
@no382001  I guess you can close this, right?

Cause now we have https://github.com/pure-data/pure-data/pull/2498 that uses your first commit and expands it. Can you also test the PR so we have more certain things are alright? It's ready for me...

no382001:
@porres sure you can close this

PR #2445, opened 2024-09-25
fix [sigmund~ -nharmonics] reporting false warning

### LINKED ISSUE #2440: sigmund~ with -nharmonics flag reports error
It took me a while to find out, when using the -nharmonics flag in sigmund~ it needs two numbers, e,g. [sigmund~ -nharmonics 6 0]. This worked in Pd 0.54 and in 0.55-0, but in Pd 0.55-1 the console reports an error:
sigmund~: : unknown flag or argument missing

I am not sure if this is a bug or just an incorrect warning, because sending messages to nharmonics (inside sigmund~) seems to work, nevertheless.
Besides that, it would be good if this special situation is documented in the help file, I think.

no382001:
closes #2440

### DIFF
--- a/extra/sigmund~/sigmund~.c
+++ b/extra/sigmund~/sigmund~.c
@@ -1367,11 +1367,16 @@ static void *sigmund_new(t_symbol *s, int argc, t_atom *argv)
             sigmund_minpower(x, atom_getfloatarg(1, argc, argv));
             argc -= 2; argv += 2;
         }
-        else if (!strcmp(firstarg->s_name, "-nharmonics") && argc > 2)
+        else if (!strcmp(firstarg->s_name, "-nharmonics") && argc > 1)
         {
-            sigmund_nharmonics(x, atom_getfloatarg(1, argc, argv),
-                atom_getfloatarg(2, argc, argv));
-            argc -= 2; argv += 2;
+            if (argc <= 2){
+                pd_error(x, "sigmund~: -nharmonics: takes 2 arguments");
+                argc -= 2; argv += 2;
+            } else {
+                sigmund_nharmonics(x, atom_getfloatarg(1, argc, argv),
+                    atom_getfloatarg(2, argc, argv));
+                argc -= 3; argv += 3;
+            }
         }
         else if (!strcmp(firstarg->s_name, "-amppowerlaw") && argc > 1)
         {


millerpuckette:
I don't think this will work - what if the "-nharmonics" is followed by a single number and then another flag?  At the moment, I think -nharmonics as a flag requires two arguments, while the run-time message nharmonics allows the second one to default to zero.  This probably should be documented but isn't strictly an error, just an inconsistent feature :)

porres:
Documentation is actually wrong as it only mentions a single argument. It is described as

"nharmonics (default 6): harmonic number at which pitch detection rolls off."

So, is that correct and what is the 2nd argument about?

porres:
@millerpuckette 

Documentation is actually wrong as it only mentions a single argument. It is described as

"nharmonics (default 6): harmonic number at which pitch detection rolls off."

So, is that correct and what is the 2nd argument about?

porres:
> At the moment, I think -nharmonics as a flag requires two arguments

@millerpuckette 

hmmm, but as it is, if I guve it 2 arguments, we get an error!

I am getting at the moment "**_sigmund~: : unknown flag or argument missing_**" when doing something like [sigmund~ -nharmonics 6 0], so we have a bug in the code while checking for the correct number of arguments

porres:
a screenshot 

Screenshot 2024-10-11 at 10 08 21

porres:
> what if the "-nharmonics" is followed by a single number and then another flag?

well, but as you mentioned, it is supposed to take two arguments, right? So if you give it one argument, we get an error, which already happens in Pd (tested with 0.55-1). in this PR we also get an error but it now warns you that you should have 2 arguments. And this PR then does not print any error when giving 2 arguments, which currently happens.

So, as I see it, this PR really does fix the issue @millerpuckette 

Or do you want the flag to be able to take either 1 or 2 arguments?

porres:
as for the 2nd argument, I see it is "octave bias", from -100 to 100, but I don't understand what it is supposed to do, we need more information!

millerpuckette:
OK, I think this is fixed now in a slightly more general way.  "-nharmonics" as a flag takes two numberic arguments, whereas for the "nharmonics" run-time message the second arg is optional.

porres:
cool, just tested it, closing this on then.

@millerpuckette please help me make sense of how the 2nd argument works so I can document it, see https://github.com/pure-data/pddp/issues/254

Issue #2462, opened 2024-10-19
Add -alsaname cmdline flag - so Alsa midi ports can be unique and predictable.


tstex:
On Linux, Alsa midi client numbers are never guaranteed to be the same for each Pd session. Pd might use 129 today, and 130 tomorrow. (This is the same for all Alsa midi clients, and is not limited to Pd). This becomes a problem for systems (such as Jack and Carla) that attempt to reliably make connections between different midi clients.  

Fortunately a2jmidid provides a "-u" flag which forces a2jmidid to use port names that do not include the number. From the a2jmidid man page:

       "ALSA does not guarantee client names to by unique. I.e. it is possible to have two apps that create two  clients  with
       same  ALSA  client  name.  JACK however requires port names to be unique. To ensure this uniqueness, a2jmidid will add
       the unique numeric ALSA client ID to the JACK port name.  However this behaviour  is  known  to  be  problematic  when
       restoring  connections  using  simplistic tools like aj-snapshot and jack_connect.  In order to make them work, the -u
       option can be used. This option will cause a2jmidid to omit the numeric ALSA Client ID from JACK port names.  In  this
       mode, ALSA client name uniqueness must be guaranteed externally."

With this flag enabled in Linux, the problem then arises that every pd session is hard coded to use the name "Pure Data" for its port name, and so multiple Pd sessions appear as a single session in Jack, Carla etc.  This makes patching impossible. 

There needs to be a "-alsaname" cmdline flag in Pd - similar to the "-jackname" flag - so that Alsa midi ports can be both unique, and not attached to an indeterminate Alsa client number.

tstex:
Fix in this Pull Request: https://github.com/pure-data/pure-data/pull/2463

tstex:
New pull request here: https://github.com/pure-data/pure-data/pull/2466

PR #2463, opened 2024-10-19
Add -alsaname command line flag to allow Pd to use unique Alsa midi port names

tstex:
Fix for: https://github.com/pure-data/pure-data/issues/2462

### DIFF
--- a/src/m_private_utils.h
+++ b/src/m_private_utils.h
@@ -13,11 +13,6 @@
 # error m_private_utils.h is a PRIVATE header. do *not* use it in your externals
 #endif
 
-#ifdef HAVE_CONFIG_H
-/* autotools might put all the HAVE_... defines into "config.h" */
-# include "config.h"
-#endif
-
 /* ------------------------------ atomics ----------------------------------- */
 
 #ifdef _MSC_VER


umlaeute:
i wonder whether we shouldn't just re-use the `-jackname` for this.

obviously the name of the flag is sub-optimal (there's no JACK), but maybe the solution is to just rename the flag to `-devicename` (keeping `-jackname` as an alias).

tstex:
I agree that is a good idea. I believe I tried that initially, but ran into problems. Let me try again tonight after work to verify.

reduzent:
The flag `-jackname` does not only set the jack client name, but it implicitly switches the audio backend to JACK. I think a new flag like `-devicename` that only sets the name (for whatever audio or MIDI backend) would make a lot of sense and would allow a better behavior without breaking backwards compatibility.

tstex:
That kind of makes -jackname redundant in some situations yes? How do we resolve situations where the user uses both flags - but with different values?

I'll look at re-using the jackname value shortly. Are there any other use cases for the more generic -devicename on other platforms etc?

As an aside, I don't think jackname should set or switch the backend, but backwards compatibility seems to be highly prized.

reduzent:
> As an aside, I don't think jackname should set or switch the backend, but backwards compatibility seems to be highly prized.

I absolutely agree on both points.

umlaeute:
> Hmm. Ok. I remember why I didn't re-use -jackname now. Linkage hell. Makefile changes: (USEAPI_JACK in USEAPI_ALSA code). Much harder (impossible?) to do cleanly from a code point of view.

Nah, much simpler. 
`-devicename` sets an API independent name. As such, it *must not* be depending on any preprocessor define like `USEAPI_JACK` or the like. 
It will simply set a global symbol, that can then be used by any API that requires it (Jack, alsa-midi, futuresound,...)

> As an aside, I don't think jackname should set or switch the backend, but backwards compatibility seems to be highly prized.

I agree with the former.
Personally I could happily  live with `-jackname` not setting the backend. 
Iirc, the current behaviour is not documented (at least I hope so. There's a tendency to document every voidness, turning arbitrary (and not well-thought-of) decisions into canonical and expected behaviour), so we could just regard it as a bug. 
It's easy (and backward compatible) enough to just add an additional "-jack" flag. 
I *would* keep the compost flag name ("-jackname") around, so that startup scripts don't start to fail.

reduzent:
I fully agree. I also rather would have `-jackname` not implicitly set the back-end. What I meant to say: I fear backwards compatibility concerns (I don't agree with them).

src/s_midi_alsa.c
```diff
@@ -103,9 +104,13 @@ void sys_alsa_do_open_midi(int nmidiin, int *midiinvec,
         if (port < 0) goto error;
     }
 
+    if (!alsa_client_name || !strlen(alsa_client_name))
+        alsa_set_client_name("Pure Data");
+    post("Alsa MIDI client name:%s", alsa_client_name);
```
umlaeute, line 109: this debugging output shouldn't be here

umlaeute (changes requested):
instead of writing to an API-specific variable, i would instead use a global (but private) `sys_devicename[MAXPDSTRING]`  that get's initialized to `"Pure Data"` (or `"pure_data"`, but I'd actually prefer the former even if the rename breaks my qjackctl patchbays).

than JACK, or ALSA or whoever wants can use this name (as is, without having to check for emptiness).

umlaeute:
why has this PR been closed?

tstex:
I cleaned up my git history - which inadvertently closed the PR. I will
make a new one. I'm still working on it. There will be another one
involving autoconf probably. CFLAGS includes both -o0 and -o2 at present,
so the things I'm trying to look at in GDB have been optimized away. It's
been quite the side track. :D

But the pd code is written. I just need to debug this last thing....

PR #2466, opened 2024-10-24
Add '-devicename' flag for the alsa midi device (and JACK)

tstex:
This time the devicename is accessible from alsa midi, but also other files - like jack audio. If -jackname is provided - then that will be used for jack as previously. If -jackname is not used, then devicename, or a default name ("Pure Data") will be used. Removed the silent setting of the Jack back end that was hidden in the -jackname arg check. Use the -jack flag instead.


src/s_stuff.h
```diff
@@ -47,6 +47,10 @@ EXTERN int sys_noloadbang;
 EXTERN int sys_havetkproc(void);    /* TK is up; we can post to Pd window */
 EXTERN int sys_havegui(void);       /* also have font metrics and can draw */
 extern const char *sys_guicmd;
+static const char * sys_default_device_name = "Pure Data";
+static char * sys_devicename = NULL;
```
umlaeute, line 51: please don't declare `static` variables in a (quasi-)public header file.

umlaeute (changes requested):
sorry for being obtuse, but why do you keep the JACK name separate?
IMHO, this needlessly complicates the code (as opposed to just using a single `sys_devicename` for everything).
(your solution currently allows the user to specify a different name for the JACK client and the ALSA seq client; but I cannot think of any use valid case for this).

or put otherwise, what is wrong with just using:

```diff
diff --git a/src/s_main.c b/src/s_main.c
index 455d8736..8d311eca 100644
--- a/src/s_main.c
+++ b/src/s_main.c
@@ -84,6 +84,8 @@ int sys_nmidiin = -1;
 int sys_midiindevlist[MAXMIDIINDEV] = {1};
 int sys_midioutdevlist[MAXMIDIOUTDEV] = {1};
 
+char sys_devicename[MAXPDSTRING] = "Pure Data";
+
 #if __APPLE__
 char sys_font[100] = "Menlo"; /* hack until DVSM bug is fixed on macOS 10.15+ */
 char sys_fontweight[10] = "normal";
@@ -814,6 +816,13 @@ int sys_argparse(int argc, const char **argv)
             as.a_noutdev = as.a_nchoutdev =  as.a_nindev = as.a_nchindev = 0;
             argc--; argv++;
         }
+        else if (!strcmp(*argv, "-devicename"))
+        {
+            if (argc < 2)
+                goto usage;
+            pd_snprintf(sys_devicename, MAXPDSTRING-1, argv[1]);
+            argc -= 2; argv +=2;
+        }
 #ifdef USEAPI_OSS
         else if (!strcmp(*argv, "-oss"))
         {
diff --git a/src/s_audio_jack.c b/src/s_audio_jack.c
index 6d5bfccd..351fa358 100644
--- a/src/s_audio_jack.c
+++ b/src/s_audio_jack.c
@@ -55,6 +55,7 @@ static char *jack_outbuf;
 static sys_ringbuf jack_outring;
 static char *jack_inbuf;
 static sys_ringbuf jack_inring;
+extern char sys_devicename[MAXPDSTRING];
 
     /* callback routine for non-callback client... throw samples into
         and read them out of a FIFO.  Since we don't know at compile time
@@ -365,9 +366,7 @@ int jack_open_audio(int inchans, int outchans, t_audiocallback callback)
         jack_client_open() will start uone up by default.  It's not clear
         whether or not this is desirable; see long Pd list thread started by
         yvan volochine, June 2013) */
-    if (!desired_client_name || !strlen(desired_client_name))
-        jack_client_name("pure_data");
-    jack_client = jack_client_open (desired_client_name, JackNoStartServer,
+    jack_client = jack_client_open (sys_devicename, JackNoStartServer,
       &status, NULL);
     if (status & JackFailure) {
         pd_error(0, "JACK: couldn't connect to server, is JACK running?");
diff --git a/src/s_midi_alsa.c b/src/s_midi_alsa.c
index 7a011164..d1cd050d 100644
--- a/src/s_midi_alsa.c
+++ b/src/s_midi_alsa.c
@@ -41,6 +41,7 @@ static int alsa_midioutfd[MAXMIDIOUTDEV];
 static snd_seq_t *midi_handle = NULL;
 
 static snd_midi_event_t *midiev;
+extern char sys_devicename[MAXPDSTRING];
 
 void sys_alsa_do_open_midi(int nmidiin, int *midiinvec,
     int nmidiout, int *midioutvec)
@@ -105,7 +106,7 @@ void sys_alsa_do_open_midi(int nmidiin, int *midiinvec,
 
     snd_seq_client_info_malloc(&alsainfo);
     snd_seq_get_client_info(midi_handle, alsainfo);
-    snd_seq_client_info_set_name(alsainfo,"Pure Data");
+    snd_seq_client_info_set_name(alsainfo, sys_devicename);
     client = snd_seq_client_info_get_client(alsainfo);
     snd_seq_set_client_info(midi_handle, alsainfo);
     snd_seq_client_info_free(alsainfo);
```

(for brevity's sake, i left out the cleanup of now-unused code)


(**NOTE**: I "request changes" for the "`static` in public header", not necessarily for this.)

umlaeute:
> why do you keep the JACK name separate?

ah, I see that the `desired_client_name` is also used for keeping the client name stable when re-opening JACK (if the primary name was already taken).

e.g.
1. start Pd with JACK - it will register itself as `pure_data`)
2. start another instance of Pd with JACK - it wlil register itself as something like `pure_data-200`
3. quit the 1st instance
4. in the 2nd instance, change the audio backend to *ALSA* and then back to *JACK* (so it has to create a new client)
5. notice that the JACK-name is still `pure_data-200` (even though the original `pure_data` name is now available)


in this case, i think a better variable name would be `actual_client_name` or somesuch.
I also wonder whether it is such a good idea in the first place, to try keeping the (uniquified) client name stable.
I don't think any other application does this (I just checked with Ardour - which I consider canonical in all things JACK - and 
it always tries to re-claim the original name):

1. start Ardour with JACK - it will register itself as `Ardour`)
2. start another instance of Ardour with JACK - it wlil register itself as something like `Ardour-123`
3. quit the 1st instance
4. in the 2nd instance, change the audio backend to *ALSA* and then back to *JACK* (so it has to create a new client)
5. notice that the JACK-name is now `Ardour` (as this name is no longer locked by another process)

tstex:
Heya,

>  why do you keep the JACK name separate?

Well I thought we'd discussed wanting that backwards compatibility in the
previous PR chat? (wish that thread still existed. Sigh.)

But yes, jackname has it's own odd behavior. One of them seemed to be
making devicename NULL.

It also occured to me that these changes might be leaking. We put another
copy of devicename on the heap every time we call get_device_name(). But I
think having that value on the heap was needed for accessibility across
files? The heap usage part of this change is based off the way jackname is
implemented, although it has admittedly drifted. Alsa midi required a const
char*, so things changed for that

I will do some more testing with multiple instances tonight, but given the
difficulty I had getting this -devicename alternative to the original
-alsaname PR working, I'm becoming rather fond of any "ain't broke"
status... 😏

Cheers

umlaeute:
> Heya,
> why do you keep the JACK name separate?
> Well I thought we'd discussed wanting that backwards compatibility in the previous PR chat? (wish that thread still existed. Sigh.)

where has it gone to? i still see https://github.com/pure-data/pure-data/pull/2463


in any case: what **I** meant with "backward compatibility" is simply: If I start Pd with `-jackname foobar`, Pd should show up in JACK as `foobar`. period.
now this can easily be obtained if `-devicename` sets the JACK client name **and** `-jackname` is a simple alias for `-devicename`.
So adjust my example code to:
```C
        else if ((!strcmp(*argv, "-devicename") || !strcmp(*argv, "-jackname")))
        {
            if (argc < 2)
                goto usage;
            pd_snprintf(sys_devicename, MAXPDSTRING-1, argv[1]);
            argc -= 2; argv +=2;
        }
```

or if you would like to keep the automatic enabling of the JACK audio backend for *full* compatibility, do something like:
```C
        else if (!strcmp(*argv, "-devicename"))
        {
            if (argc < 2)
                goto usage;
            pd_snprintf(sys_devicename, MAXPDSTRING-1, argv[1]);
            argc -= 2; argv +=2;
        }
        else if (!strcmp(*argv, "-jackname"))
        {
            if (argc < 2)
                goto usage;
            pd_snprintf(sys_devicename, MAXPDSTRING-1, argv[1]);
#if HAVE_JACK
            as.a_api = API_JACK;
#endif
            argc -= 2; argv +=2;
        }
```

(though I think the consensus was, that we do not need such a level of backwards compatibility)


> But yes, jackname has it's own odd behavior. One of them seemed to be making devicename NULL.

i'm not entirely sure what you are talking about.
Yes, the JACK client name is initially NULLed https://github.com/pure-data/pure-data/blob/c0f1664bbf600c9e7d27d2b5defa806f7bd1474a/src/s_audio_jack.c#L44 and if it is NULL (or empty) a default value is used https://github.com/pure-data/pure-data/blob/c0f1664bbf600c9e7d27d2b5defa806f7bd1474a/src/s_audio_jack.c#L368-L369

there's nothing wrong with this, but otoh, there's also nothing wrong with making sure in the first place, that the devicename is valid.

> It also occured to me that these changes might be leaking. We put another copy of devicename on the heap every time we call get_device_name(). 

where do you see a leak?
the original solution freed the data before re-allocating it. my suggestion would just use pre-allocated data.

> But I think having that value on the heap was needed for accessibility across files? 

no. why do you believe this?

> Alsa midi required a const char*, so things changed for that 

how so?
a `const` argument is really justa promise of the called function that it will not change the argument.
you can pass a `char*` directly as a `const char*`. it's just the other way that is forbidden (passing a `const char*` as a `char*` - the non-const argument indicates that the data might change, and the `const` specifier explicitly forbids this).

and of course, in C you can pass an array as a pointer (e.g. in `char foo[100]`, the `foo` is really just a pointer to a pre-allocated memory holding 100 `char`s)

tstex:
Ok. That is a much more elegant solution. It's always nice to add/improve functionality and yet remove code. :D

I did not see your line below until after I submitted my most recent update. 
> else if ((!strcmp(*argv, "-devicename") || !strcmp(*argv, "-jackname")))

The jack code is in an ifdef for jack, so perhaps it's best to leave it as I submitted it?

I tested by running a bunch of pd instances with different combinations of both flags. The jackname does have to be unique, and there is code to increment a number in the name if it is not. This does not effect the -devicename value. Users will need to make that unique on their own.

umlaeute (approved):

src/s_audio_jack.c
```diff
@@ -376,10 +373,10 @@ int jack_open_audio(int inchans, int outchans, t_audiocallback callback)
         /* jack spits out enough messages already, do not warn */
         STUFF->st_inchannels = STUFF->st_outchannels = 0;
         return 1;
-    }
+    } 
     if (status & JackNameNotUnique)
-        jack_client_name(jack_get_client_name(jack_client));
-    logpost(NULL, PD_VERBOSE, "JACK: registered as '%s'", desired_client_name);
+        pd_snprintf(sys_devicename, MAXPDSTRING-1, jack_get_client_name(jack_client));
```
umlaeute, line 378: hmm, this will overwrite the `sys_devicename` if the name was already taken by JACK.
the consequence is, that ALSA MIDI might or might not use the unique-ified name.

e.g.
1. start `pd -devicename Foo -jack -alsamidi -mididev 1`
   - the JACK client name will be "Foo"
   - the ALSA MIDI client name will be "Foo"
2. start another instance of `pd -devicename Foo -jack -alsamidi -mididev 1`
   - the JACK client name will be "Foo-123"
   - the ALSA MIDI client name will be "Foo"
2. start another instance, this time with `pd -devicename Foo -jack -ossmidi`
   - the JACK client name will be "Foo-124"
   - now select "ALSA-MIDI"  via the Media-menu (and make sure to have some MIDI channels)
   - the ALSA MIDI client name will now be ***"Foo-124"***

personally I find it irritating if the ALSA MIDI name is different depending on *when* i switch to ALSA MIDI.
as mentioned in https://github.com/pure-data/pure-data/pull/2466#issuecomment-2434738595, I think we should not try to store a uniquified JACK client name, and instead always try to use the non-unique name.

tstex: Fair enough. Pushed another change.

tstex: One other thing I'm not completely comfortable with - is that if the user uses both` -jackname` and `-devicename`, only the value set by `-jackname` will be used for sys-devicename. Since we are presumably going to deprecate -jackname, I'm not sure that this is something we need to code a fix for, but presumably the fix would be to change the order in which the args are parsed in main. What do you think?

umlaeute: ??? i cannot reproduce this.

it's the last flag that wins, so if you specify `-jackname Foo -devicename Bar` then the name will be **Bar**.
otoh, if you specify `-devicename Bar -jackname Foo`, the name will be **Foo**.

This is consistent with specifying `-jackname Foo -jackname Bar` (**Bar**) resp `-devicename Bar -devicename Foo` (**Foo**)

umlaeute: just to clarify: we loop over the user-provided arguments and check each one if it matches a given string, and then we do something with this.
I *think* you think that instead we are looping over the list of potential arguments, and check whether the currently selected one is present in the list of user-provided arguments. but this is not the case

tstex:
You are correct. My bad

Issue #2471, opened 2024-10-30
[expr] strpbrk() bug/crash


porres:
I used to get an arguably improper 'symbol float' output for something like
[expr strpbrk("bang", "l")]. I opened an issue at https://github.com/pure-data/pure-data/issues/2399 and I closed it because I don't get this anymore, but an error instead. 

"expr:'strpbrk("bang" \, "lx")': bang: unrecognized result 0 "

I also don't like the output of an error instead of something like an empty symbol instead.

The thing now is that Pd is also crashing with these errors!

<img width="490" alt="Screenshot 2024-10-30 at 13 30 04" src="https://github.com/user-attachments/assets/3e9243e2-589e-458f-89b3-a4107d25118a">

I am attaching this patch. Steps to reproduce, click first on "symbol hello", then click on "symbol goodbye" and Pd crashes.

[test.pd.zip](https://github.com/user-attachments/files/17576111/test.pd.zip)

porres:
@umlaeute maybe you could try and fix this in the develop branch?

PR #2472, opened 2024-10-30
fix crash in expr

porres:
- Closes https://github.com/pure-data/pure-data/issues/2471


porres:
@millerpuckette I got this file from Shahrokh, it was supposed to be the final and correct code for the 0.55-1 release, somehow we ended up with a different one. Besides the fact that it fixes a Pd crash, I wanted to clear this out, so you'd consider this a no brainer to merge right away

PR #2481, opened 2024-11-14
API to allow external objects (libraries, abstractions) to load an accompanying gui plugins

umlaeute:
This adds a global message "plugin-load" to the `pd` receiver, that is forwarded to the GUI to load a plugin script.

E.g. this
```
[pd plugin-load helloworld-plugin /path/to/helloworld(
```
will load the */path/to/helloworld/helloworld-plugin.tcl*.

- the message is modelled after the `open` message (and saves the user from joining a path with a filename)
- the API omits the filename extension (to support multiple GUI backends, when we get there in 2300)
- the duplicate check is somewhat relaxed to only match exact filenames (so if both `[foo/guiabs]` ans `[bar/guiabs]` come with their `guiabs-plugin.tcl`, both can be loaded)

imo, this obsoletes https://github.com/pure-data/pure-data/pull/1766

### DIFF
--- a/src/m_glob.c
+++ b/src/m_glob.c
@@ -112,6 +112,11 @@ void max_default(t_pd *x, t_symbol *s, int argc, t_atom *argv)
     endpost();
 }
 
+/* loading/interacting with GUI plugins */
+void glob_pluginload(t_pd *dummy, t_symbol *plugin, t_symbol *path)
+{
+    pdgui_vmess("load_plugin", "ss", plugin->s_name, path->s_name);
+}
 void glob_plugindispatch(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
 {
     pdgui_vmess("pdtk_plugin_dispatch", "a", argc, argv);
@@ -195,6 +200,8 @@ void glob_init(void)
         gensym("compatibility"), A_FLOAT, 0);
     class_addmethod(glob_pdobject, (t_method)glob_plugindispatch,
         gensym("plugin-dispatch"), A_GIMME, 0);
+    class_addmethod(glob_pdobject, (t_method)glob_pluginload,
+        gensym("plugin-load"), A_SYMBOL, A_DEFSYM, 0);
     class_addmethod(glob_pdobject, (t_method)glob_helpintro,
         gensym("help-intro"), A_GIMME, 0);
     class_addmethod(glob_pdobject, (t_method)glob_fastforward,
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -739,14 +739,38 @@ proc check_for_running_instances { } {
 # ------------------------------------------------------------------------------
 # load plugins on startup
 
-proc load_plugin_script {filename} {
+proc load_plugin_script {filename {duplicate basename} {warn 1}} {
+    # read $filename and evaluate it (for loading plugins)
+    # $duplicate: specifies how to check for duplicates; possible values are
+    # - ignore: do not do any duplicate checking (might load scripts multiple times)
+    # - basename: only check the base filename (without path) [ DEFAULT ]
+    # - filename: check the full $filename
+    # $warn: whether an attempt to load a script multiple times leads to a warning or not
+    # - 1 : warn if duplicate [ DEFAULT ]
+    # - 0 : silently ignore duplicates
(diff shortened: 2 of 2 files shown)


umlaeute:
for completeness sake, here's how you would use this an external.

assuming we have an objectclass "helloworld" and an accompanying "helloworld-plugin.tcl". both live in the same directory.

To load the GUI-plugin, modify the `helloworld_setup()` callback as follows:

```C
void helloworld_setup(void) {
    t_atom ap[2];
     helloworld_class = class_new(gensym("helloworld"), ...);
    /* class setup skipped */

    SETSYMBOL(ap+0, gensym("helloworld-plugin"));
    SETSYMBOL(ap+1, gensym(class_gethelpdir(helloworld_class)));
    pd_typedmess(gensym("pd")->s_thing, gensym("plugin-load"), 2, ap);
```

umlaeute:
similarily, if your "helloworld" object is an abstraction, you would load the accompanying GUI-plugin like so:

```
[loadbang]
|
[file patchpath]
|
[; pd plugin-load helloworld-plugin $1(
```

Spacechild1:
Looks good, but for externals I would really add a proper C API function. As I've written on the Pd list, I find it strange to have functionality that is *only* accessible through a Pd message. I don't know of any other example.

Not only do I find it awkward to assemble a Pd message just to make an API call, I see two general issues with such an approach:
1. it's discoverable via the public headers
2. there's no possibility for error handling

Speaking of error handling: the C API function could check whether an appropriate file exists for the current GUI backend and return failure otherwise.

umlaeute:
> Speaking of error handling: the C API function could check whether an appropriate file exists for the current GUI backend and return failure otherwise.

in which case the core needs to be full aware of the GUI backend, which i would like to avoid.

also, the mere existence of a file is not a good indicator of whether the plugin is usable.
if your library needs to know whether its plugin was loaded successfully, then i think the only way is to send a message from the GUI back to the core.

> Ah well, we obviously can't do that if the GUI runs on a different machine...

in which case many things start falling apart anyhow (for starters you would need to put your GUI-plugin in exactly the place where your library expects it to be)

chr15m (approved):
LGTM on the abstraction loader side (though I defer to others more qualified for a proper analysis).

umlaeute:
> Doesn't this conflict with your request above?

not sure i can follow (probably just lack of :coffee:): which request do you mean?

or put otherwise:
@Spacechild1, are you saying that 
```C
pdgui_vmess("load_plugin", "ss", "helloworld-plugin", class_gethelpdir(helloworld_class));
```
is not enough, but instead we need the following?
```C
pdgui_loadplugin("helloworld-plugin", class_gethelpdir(helloworld_class));
```

chr15m:
Sorry, I was not very clear. I was basically asking the same thing.

millerpuckette:
I'm not clear on what will happen if the GUI side gets rewritten (most likely not in a scripting language I think).  For instance, what if there's a GTK or QT front end?

umlaeute:
@millerpuckette well the idea is that whatever toolkit is used to re-implement the GUI, it will also (need to) have some replacement for loading GUI-plugins.
this could mean one of the following:
- the new GUI is able to interpret the existing (TclTk) GUI plugin
- an external (that requires a GUI-plugin) needs a separate implementation for its GUI part
- the external can (or: must) live without the GUI plugin

I think that the proposed solution is an improvement for all these possibilities, when compared with what we have now (that the external embeds the functionality of the GUI-plugin as direct Tcl/Tk commands in the  C-code)
- a generic plugin (written in Tcl/Tk) could have specific code for the idiosyncracies of the various GUI backends (given that it knows more about the GUI than the Pd-core process) - in my experience it's much easier to write such special casing in a single language (whatever it is), rather than having to juggle with a long stick through the peephole of the core/gui connection.
- if the new GUI doesn't allow for executing legacy plugins, you need a way to somehow provide the functionality in the new GUI (and being able to implement it in a native format, seems like a good idea)
- if the GUI plugin is somewhat optional, i think it's good to have a single point of failure (when trying to load the plugin)

speaking of *optional* plugins, i wonder whether we should add a flag to the `load_plugin` proc, that will reduce any loading errors to mere warnings. (but it's probably over-engineering)

PR #2496, opened 2024-12-11
clip_perf8()

giuliomoro:
clip~ was missing a perf8 callback and that can give a significant performance boost on embedded platforms. I tested different implementations on a 1GHz Cortex-A8; CPU usage is running 64 objects in parallel. 

```
for (; n; n -= 8, in += 8, out += 8)
    {
#if 0
            // gcc 1.4%
            // clang 1.6%
            t_sample f[8];
            for(int i = 8; i--; )
                    f[i] = in[i] < x->x_lo ? x->x_lo : in[i];
            for(int i = 8; i--; )
                    out[i] = f[i] > x->x_hi ? x->x_hi : f[i];
#endif
#if 0
            // gcc: 8.3%
            // clang: 1.9%
            t_sample f[8];
            f[0] = in[0] < x->x_lo ? x->x_lo : (in[0]);
            f[1] = in[1] < x->x_lo ? x->x_lo : (in[1]);
            f[2] = in[2] < x->x_lo ? x->x_lo : (in[2]);
            f[3] = in[3] < x->x_lo ? x->x_lo : (in[3]);
            f[4] = in[4] < x->x_lo ? x->x_lo : (in[4]);
            f[5] = in[5] < x->x_lo ? x->x_lo : (in[5]);
            f[6] = in[6] < x->x_lo ? x->x_lo : (in[6]);
            f[7] = in[7] < x->x_lo ? x->x_lo : (in[7]);

            out[0] = f[0] < x->x_lo ? x->x_lo : (f[0]);
            out[1] = f[1] < x->x_lo ? x->x_lo : (f[1]);
            out[2] = f[2] < x->x_lo ? x->x_lo : (f[2]);
            out[3] = f[3] < x->x_lo ? x->x_lo : (f[3]);
            out[4] = f[4] < x->x_lo ? x->x_lo : (f[4]);
            out[5] = f[5] < x->x_lo ? x->x_lo : (f[5]);
            out[6] = f[6] < x->x_lo ? x->x_lo : (f[6]);
            out[7] = f[7] < x->x_lo ? x->x_lo : (f[7]);
#endif
#if 0
            // gcc: 8.3%
            // clang: 1.8%
            t_sample f0, f1, f2, f3, f4, f5, f6, f7;
            f0 = in[0] < x->x_lo ? x->x_lo : (in[0]);
            f1 = in[1] < x->x_lo ? x->x_lo : (in[1]);
            f2 = in[2] < x->x_lo ? x->x_lo : (in[2]);
            f3 = in[3] < x->x_lo ? x->x_lo : (in[3]);
            f4 = in[4] < x->x_lo ? x->x_lo : (in[4]);
            f5 = in[5] < x->x_lo ? x->x_lo : (in[5]);
            f6 = in[6] < x->x_lo ? x->x_lo : (in[6]);
            f7 = in[7] < x->x_lo ? x->x_lo : (in[7]);

            out[0] = f0 < x->x_lo ? x->x_lo : (f0);
            out[1] = f1 < x->x_lo ? x->x_lo : (f1);
            out[2] = f2 < x->x_lo ? x->x_lo : (f2);
            out[3] = f3 < x->x_lo ? x->x_lo : (f3);
            out[4] = f4 < x->x_lo ? x->x_lo : (f4);
            out[5] = f5 < x->x_lo ? x->x_lo : (f5);
            out[6] = f6 < x->x_lo ? x->x_lo : (f6);
            out[7] = f7 < x->x_lo ? x->x_lo : (f7);
#endif
#if 0
            // gcc: 12.4%
            // clang: 7.5%
            for(int i = 8; i--; )
                    out[i] = in[i] < x->x_lo ? x->x_lo : (in[i] > x->x_hi ? x->x_hi : in[i]);
#endif
#if 0
            // gcc: 10.6%
            // clang: 7.4%
            out[0] = in[0] < x->x_lo ? x->x_lo : (in[0] > x->x_hi ? x->x_hi : in[0]);
            out[1] = in[1] < x->x_lo ? x->x_lo : (in[1] > x->x_hi ? x->x_hi : in[1]);
            out[2] = in[2] < x->x_lo ? x->x_lo : (in[2] > x->x_hi ? x->x_hi : in[2]);
            out[3] = in[3] < x->x_lo ? x->x_lo : (in[3] > x->x_hi ? x->x_hi : in[3]);
            out[4] = in[4] < x->x_lo ? x->x_lo : (in[4] > x->x_hi ? x->x_hi : in[4]);
            out[5] = in[5] < x->x_lo ? x->x_lo : (in[5] > x->x_hi ? x->x_hi : in[5]);
            out[6] = in[6] < x->x_lo ? x->x_lo : (in[6] > x->x_hi ? x->x_hi : in[6]);
            out[7] = in[7] < x->x_lo ? x->x_lo : (in[7] > x->x_hi ? x->x_hi : in[7]);
#endif
    }
```

I eventually picked the first implementation for my PR because it is the only one that works well with both clang and gcc.

### DIFF
--- a/src/d_math.c
+++ b/src/d_math.c
@@ -51,10 +51,29 @@ static t_int *clip_perform(t_int *w)
     return (w+5);
 }
 
+static t_int *clip_perf8(t_int *w)
+{
+    t_clip *x = (t_clip *)(w[1]);
+    t_sample *in = (t_sample *)(w[2]);
+    t_sample *out = (t_sample *)(w[3]);
+    int n = (int)(w[4]);
+    for (; n; n -= 8, in += 8, out += 8)
+    {
+        // this partial unrolling gives best results on cortex-a8 with both clang and gcc
+        t_sample f[8];
+        for(int i = 8; i--; )
+            f[i] = in[i] < x->x_lo ? x->x_lo : in[i];
+        for(int i = 8; i--; )
+            out[i] = f[i] > x->x_hi ? x->x_hi : f[i];
+    }
+    return (w+5);
+}
+
 static void clip_dsp(t_clip *x, t_signal **sp)
 {
     signal_setmultiout(&sp[1], sp[0]->s_nchans);
-    dsp_add(clip_perform, 4, x, sp[0]->s_vec, sp[1]->s_vec, SIGTOTAL(sp[0]));
+    dsp_add((SIGTOTAL(sp[0]) & 7) ? clip_perform : clip_perf8,
+        4, x, sp[0]->s_vec, sp[1]->s_vec, SIGTOTAL(sp[0]));
 }
 
 static void clip_setup(void)


Spacechild1:
There are few problems with your code snippets.

First, in Nr.2, Nr.3 and Nr.5 you are accidentally using the same ternary clause twice...

Second, none of your snippest enable full auto-vectorization! To explain why, let's take a step back and look what the `perform8` methods actually try to achieve.

Not only do we manually unroll the loops, we also save the input samples into local variables. Why? Because the compiler must assume that `in` and `out` may alias each other, i.e. point to the same memory location. Let's consider the last example in your snippet:
```
for (; n; n -= 8, in += 8, out += 8)
    out[0] = in[0] < x->x_lo ? x->x_lo : (in[0] > x->x_hi ? x->x_hi : in[0]);
    out[1] = in[1] < x->x_lo ? x->x_lo : (in[1] > x->x_hi ? x->x_hi : in[1]);
    ...
```
Everytime we write into `out`, the compiler must load `in` from memory in the next loop iteration. This is because every item of `out` can potentially alias any item of `in`. (The two arrays may overlap at arbitrary points.) As a consequence, we have an implicit depency between loop iterations that prevents auto-vectorization.

Aliasing is not restricted to `in` and `out`, though! The compiler has to assume that *any* pointers *of the same type* can alias each other. In our case, `out` may also alias `x->x_lo` or `x->x_hi`! As I showed in https://github.com/pure-data/pure-data/pull/2496#issuecomment-2536440229, this can fully prevent auto-vectorization.

See also https://en.wikipedia.org/wiki/Aliasing_(computing) and https://en.wikipedia.org/wiki/Restrict

Actually, in the case of Pd, `in` and `out` *can* -- and often do! -- point to the same signal buffer -- to facilitate buffer reuse --, but they only alias within a single loop iteration. By saving all input samples to local variables before accessing any output variables, we effectively tell the compiler that it can run all iterations in parallel.

Similarily, we save members like `x_lo` to the stack to tell the compiler that it can keep them in registers.

That's where auto-vectorization kicks in! Instead of saving the input to individual registers, the compiler can pack 4 or 8 floats into a single SIMD register and use appropriate SIMD instructions.

Side note: modern optimizing compilers even try to auto-vectorize code that is not unrolled by checking for overlapping ranges. That's why you also see sections with vector instructions in "normal" `perform` methods. Why do we care about manual unrolling then? Because the extend of auto-vectorization varies from compiler to compiler. Also, manual loop unrolling enables even more aggressive automatic loop unrolling and auto-vectorization by the compiler. For example, GCC even tries to unroll up to 32 iterations!

---

When you try these kind of low level optimizations, it is crucial to always look at the generated assembly. You don't have to understand every instruction, but at least you can recognize when two versions of a function produces the same assembly. You can also see to which extend the compiler is able to vectorize the code.






Since the compiler must assume that `in` and `out` can alias each other, it has to load `in[i]` from memory

giuliomoro:
Thanks for the insights. I would normally look at the assembly when attempting these things, but in this case I just went for the performance test, as that's more accurate; factors such as out-of-order execution (or lack thereof), proper interleaving of load/store operations, pipeline stalls may make it complicated to infer performance from looking at assembly without specific knowledge of the target's FPU.
In this case, by separating out the max and min operations I was actually trying to achieve performance parity with the use of `[min~]` and `[max~]` in series, which was much cheaper than `[clip~]` in my tests, which in turn prompted this exercise.

FWIW, your `clip_perform8` implementation in #2497 is only marginally (7%) faster than mine on the same hardware.

> Actually, in the case of Pd, in and out can -- and often do! -- point to the same signal buffer (to facilitate buffer reuse), but they only alias within a single loop iteration. By saving all input samples to local variables before accessing any output samples, we effectively tell the compiler that it can run all iterations in parallel.

Is that true per-class or can it be true or false for a given class depending on the DSP graph? I.e.: could we safely use `restrict` in the dsp callback to encourage compiler optimisation in at least some cases?

Spacechild1:
>  I would normally look at the assembly when attempting these things, but in this case I just went for the performance test, as that's more accurate

Just to be clear: these two should complement each other. Of course, you can't fully determine which function is faster just from looking at the assembly. (Although in some cases it will be obvious and spare you some benchmarks.)

More importantly, as I said, the assembly output shows you which functions produce (near) identical assembly! In you case you could have reduced your benchmark suite down to only two functions. Also, you can check if the compiler really does what you intended. You might be happy with the benchmarks, but the assembly would show you a missed opportunity for optimization, just like in your case with the `x_lo` and `x_hi` members. (As I've shown in the Godbolt link, Clang refused to auto-vectorize it!)

Finally, as Pd runs on many different platforms and is built with different compilers, we can't really optimize for specific target platforms. (For that we would really need to write platform-specific code - which we rather try to avoid.) In the case of `perform8` functions, we are just looking for code that the major compilers are able to properly auto-vectorize.

(If Pd was written in C++, we could use a SIMD library for the `perform` methods, so we would rely less on (current) compiler behavior.)

> Is that true per-class or can it be true or false for a given class depending on the DSP graph?

It entirely depends on the graph.

Our case is interesting because the buffer aliasing still allows in-place operation for most algorithms. (There are some externals that *must* buffer the input to get correct resuls, e.g. `[zexy/demux~]`.) However, the `restrict` keyword really tells that pointers do *not* alias - which in our case would be a lie. In theory, this would be undefined behavior as per the standard. In practice, it could still work but optimizing compilers have become so smart and unpredictable that I'd rather not bet on it :)

Spacechild1:
Do you think we can close this in favor of https://github.com/pure-data/pure-data/pull/2497?

giuliomoro:
tes

PR #2498, opened 2024-12-11
soundfile 8 bit sample support

danomatika:
This PR is an extension to #2443 which adds support for 8 bit soundfile samples:
* read and write 8 bit WAVE, AIFF, and NEXT files
* CAF files are 16 bit and above, so print an error when trying to write a header with 1 byte samples
* also fixed the NEXT format not printing an error when trying to write a header with 4 byte samples

Please double check the sample conversions in the `d_soundfile.c` xferin and xferout functions. I am also not 100% sure about using the new `t_soundfile_signedness` function for checking the 1 byte signedness, but that seemed the easiest way and doesn't require adding another bool to the t_soundfile struct which is only used for a particular end case.


danomatika:
Help files updated, please test.

porres:
Hi, here's my test with [soundfiler]! I can't find an app to play or investigate .snd files, other than Pd. But the finder shows that the aif and wav files are indeed 8 bit. Reloading them into Pd with [soundfiler] shows that there is just 1 byte.

So, saving from [soundfiler] to 8 bits seems to work fine, but there's a problem. It saves it with half the amplitude values, so the sound is twice as quiet. Reloading the saved file into arrays makes that clear, see picture below.

Screen Shot 2024-12-11 at 21 00 44

.caf prints an error and doesn't save 1 byte sound files, but it prints the error twice.

[soundfiler] write caf: file: unsupported sample format

should be just once, of course.

porres:
My test with [readsf~] and [writesf~]

[readsf~] seems simple to test, it does load 1-byte sample and plays just fine.

Saving files with [writesf~] seems also just fine and I don't see the issue I saw with [soundfiler] where the saved file would have half og the amplitude.

Trying to open a 1-byte .caf file with [writesf~] gives us an error, but not related to [writesf~], we get instead an error from "soundfiler", as in 

[soundfiler] write caf: ./foo.caf: unsupported sample format]

After that, when trying to "start" the recording, I get a proper error from [writesf~]

[writesf~] caf: ./foo.caf: unsupported sample format

Anyway, I'd say where almost there and pretty close :) 

Thanks Dan!

porres:
> also fixed the NEXT format not printing an error when trying to write a header with 4 byte samples

I didn't get an error, it did work just fine to save and load with [soundfiler]. I do get the error when trying to save 8 bytes! 

And like the 1-byte .caf error, this error comes out twice

[soundfiler] write next: file: unsupported sample format
[soundfiler] write next: file: unsupported sample format

porres:
> Trying to open a 1-byte .caf file with [writesf~] gives us an error, but not related to [writesf~], we get instead an error from "soundfiler", as in
> 
> [soundfiler] write caf: ./foo.caf: unsupported sample format]
> 
> After that, when trying to "start" the recording, I get a proper error from [writesf~]
> 
> [writesf~] caf: ./foo.caf: unsupported sample format

same situation happens in [writesf~] when trying to open 8-byte .snd files, it prints an error as "soundfiler"

danomatika:
The `[soundfiler]` is printed because there is shared code invoked for a background thread without a way to tell which type of object is calling it.

danomatika:
If anyone else wants to fix the last bits, feel free. I spent more time on it than I intended. ;)

danomatika:
> The `[soundfiler]` is printed because there is shared code invoked from a background thread without a way to tell which type of object is calling it.

No, I am wrong in this. It seems that the writesf~ background thread creates the header, checks afterward and sets an error which should be picked up by the main thread for posting, however that error value is never printed.

danomatika:
Ah nevermind: the error print did work, I just forgot it is not triggered until your start recording. It prints something useful now at least since the errno is now set. Also, half sized samples should be fixed.

porres (commented):
all good by me

porres:
I see this now has conflicts because of the help files, which is wise to leave out as this is always potentially likely to happen and it's a pain to deal with. 

My suggestion for now is to just conform with current master and as this is on my radar I can make the necessary changes later

porres:
I was able to resolve the doc issues, but I guess that worst case scenario miller could have cherry picked and left the doc commit out

millerpuckette:
> > Last time I tried to cherry-pick among a bunch of commits I ended u pbreaking thigfs so badly that I had to go to a backup :)
> 
> I had solved the issues, but they appeared back and now I just solved them again :) no need to cherry pick.
> 
> > This seems like a lot of trouble - is it really ever useful
> > to read/write 8-bit soundfiles?
> 
> I actually need this for an 8-bit audio project. Anyway, seems like an odd and arbitrary choice to not support it. I know it's not that common though but what makes Pd great and awesome is being able to do all sorts of things and not just sticking to most common scenarios :) and any software that deals with audio can actually record and open 8-bit files.
> 
> Now, what exactly is a lot of trouble? Cherry picking commits or something related to the PR itself?
> 
> For the record, I've been testing this and it's all good.

I think we simply have to stop tryint to every soundfile type in the world... when willl it ever end?  FLAC? MP3?  If someone has a bunch of MP3 files I'd simply tell them to convert to a format Pd knows about - and I don't see any fundamental difference between that and 8-bit files.

And does "any music software" include csound?  I'm pretty sure it doesn't support 8-bit either.

I was happy to add "caf" support since that allowed using soundfiles that couldn't be converted to the previously existing formats.  But 8-bit is taking things in the opposite direction.

Spacechild1:
> I think we simply have to stop tryint to every soundfile type in the world... 

I would agree with that. However, this PR doesn't really add a new soundfile type, it just implements the missing 8-bit support for WAV files. which does feel like a somewhat arbitrary omission to me. Personally, I don't have a use case, but some people apparently do.

I just wanted to give my perspective. I'm not reallying pushing for it.

> when willl it ever end?

I guess it would end right here, because there wouldn't be any WAV sample types left :)

>  FLAC? MP3? If someone has a bunch of MP3 files I'd simply tell them to convert to a format Pd knows about

Just a side note: It would be cool if there was a plugin API for soundfile backends, so that things like FLAC and MP3 support could be added as externals. I think @umlaeute and @danomatika have already been floating this idea in the past. It might be time to start working on it ;-)

porres:
> And does "any music software" include csound? I'm pretty sure it doesn't support 8-bit either.

Not a csound user, but I'm googling it and have found entries for  8-bit on the documentation.

I see these options for https://www.csounds.com/manual/html/GEN01.html (which is like [soundfiler] I believe).

1 - 8-bit signed character    4 - 16-bit short integers 
2 - 8-bit A-law bytes         5 - 32-bit long integers 
3 - 8-bit U-law bytes         6 - 32-bit floats

more impressively, searching for '8-bit' in the command line flags make me believe it supports running the whole environment with 8-bit quality? https://www.csounds.com/manual/html/CommandFlags.html

-8, --format=uchar
Use 8-bit unsigned character audio samples.


> I think we simply have to stop tryint to every soundfile type in the world... when willl it ever end? FLAC? MP3?

I say it definitely stops here :) I see the reasoning in not supporting new file formats (flac/mp3/ogg/whatever), but this is just a missing quality supported.

Again, 8-bit music (aka 'chiptune') is a thing. I just think it makes sense to also support this to enable people to experiment with poor quality audio for distorted wavetable synthesis, making some chiptune in Pd and the like.

Spacechild1:
@danomatika Wow, I didn't know about `soundfile_addtype`!

Are there any roadblocks for calling this at runtime, e.g. in an external? In other words: are there more things we need to do other than exporting it in `m_pd.h` (and raising the value of `SFMAXTYPES`)?

Side note: instead of imposing a static limit of soundfile types with `SFMAXTYPES`, we could use a linked list instead.

danomatika:
> @danomatika Wow, I didn't know about `soundfile_addtype`!
> 
> Are there any roadblocks for calling this at runtime, e.g. in an external? In other words: are there more things we need to do other than exporting it in `m_pd.h` (and raising the value of `SFMAXTYPES`)?
> 
> Side note: instead of imposing a static limit of soundfile types with `SFMAXTYPES`, we could use a linked list instead.

I didn't design it for runtime use. It may work but would need to be thread safe. There is a limit as that was the easiest way to make the soundfile type list while avoiding dynamic memory. I did *not* want to overly complicate what was there, it was the first step: formalize the interface and existing sound file types through modularization, then add CAF support.

Note: I did ask for feedback on this internal API back then, so it's funny to me that you only see it now. The pattern is repeated at the bottom of each d_sound_type.c file. ;)

Spacechild1:
> It may work but would need to be thread safe.

Good point! Since we only want to add backends - and never remove them - this could even be done in a lockfree way.

> Note: I did ask for feedback on this internal API back then, so it's funny to me that you only see it now. 

I may just have forgotten about it :) Anyway, I will keep an eye on it. One thing I should probably do is try to implement a MP3 or Opus backend and see if there any stumbling blocks in the API.

PR #2502, opened 2024-12-20
WIP: setting up minimal color config vars.

chr15m:


sebshader:
there's some discussion in the beginning of https://github.com/pure-data/pure-data/pull/196 that would probably still be relevant; @umlaeute suggested to use a function on the tcl side rather than just an array

sebshader:
anyways I still don't really understand the decision to do this incrementally or partly. It will be more work in the end if full customization is wanted by users (and I think it is..)

chr15m:
Smaller patches are always easier for a maintainer to reason about and merge than larger patches that touch a lot of the codebase. 

Maybe I am missing something but one issue I find with pdnext is the user experience of configuration feels very complicated. Some (most?) users just want a simple toggle to enable dark mode or maybe two or three values to change. The pdnext way seems to require the configuration of a lot of vars by hand to achieve this - correct me if I'm wrong. Maybe there is a way to have both.

I agree a proc is better. 🤔

Thanks for your comments, and for pdnext which has been very helpful as a reference.

sebshader:
Although it's slightly off-topic I guess I'll continue as it relates to the features and implementation of this feature:
There are 2 separate things here:
1. creating a particular color theme
2. setting a particular color theme

Both are the same in pd-next: the user is responsible for assigning values to the colors array or overriding the mentioned proc itself.
However, I created a tcl plugin to manage and choose different color themes which is very easy to use, and comes with many different themes. Users simply open a preference menu and choose the theme from a GUI selector, and can then apply and/or save the selected theme:
https://github.com/sebshader/color-themes-plugin

I think this design is preferable, though of course certain tcl-plugins are present in PD 'vanilla'. Some users might not need the extra preference or might want to manage the color-theme array or proc themselves

chr15m:
@sebshader I finally got this working by checking out pdnext, compiling, installing the plugin. It's super cool. :+1:

Issue #2506, opened 2025-01-02
add support for the (') flag in [makefilename]


porres:
I've been testing and reading the documentation of printf for [makefilename] and I was able to find 3 flags that the documentation wasn't telling us about. One of which is the apostrophe (') flag, which doesn't work with [makefilename]. BTW, see https://github.com/pure-data/pddp/issues/263

Sp, I'm also working on externals and I see that the apostrophe flag won't just simply work out of the box.

First, the apostrophe (') flag in printf-style formatting in C is used to include a locale-specific thousands separator (such as a comma , or a space) in the output of numbers. It is useful when formatting large integers or floating-point numbers for readability. For example, the integer 1234567 becomes: 1,234,567, and a floating point would be 1,234,567.00000

The ' flag works only if the program sets a locale that supports it. By default, the C locale does not enable thousands separators, so you need to use a locale like en_US.UTF-8.

In my external I was able to enable it by doing

#include <locale.h>

and then at the setup function doing:

setlocale(LC_NUMERIC, "en_US.UTF-8");

I haven't tested altering the code of [makefilename] for this. I can try and make a PR.

porres:
AI was helping me with this and I don't really know how setting the locale might screw up or not really work well with Pd. The main devs should know better and could hopefully make it clear if this is harmless or of any caveats.

porres:
as the discussion in https://github.com/pure-data/pure-data/pull/2507 shows, this is not worth it

PR #2486, opened 2024-11-26
Documentation updates

porres:
Here we go again, more documentation updates...

- closes https://github.com/pure-data/pure-data/issues/2489
- closes https://github.com/pure-data/pddp/issues/262
- closes https://github.com/pure-data/pddp/issues/263
- closes https://github.com/pure-data/pddp/issues/264
- closes https://github.com/pure-data/pddp/issues/265
- closes https://github.com/pure-data/pddp/issues/268
- closes https://github.com/pure-data/pddp/issues/266
- closes https://github.com/pure-data/pddp/issues/270
- closes https://github.com/pure-data/pddp/issues/271
- closes https://github.com/pure-data/pddp/issues/274
- closes https://github.com/pure-data/pddp/issues/275
- closes https://github.com/pure-data/pddp/issues/276

PR #2507, opened 2025-01-02
add support for apostrophe flag in [makefilename]

porres:
closes https://github.com/pure-data/pure-data/issues/2506

I tested and it works, see screenshot

Screenshot 2025-01-02 at 03 21 26

but I'm not 100% sure I'm not messing with things I shouldn't be messing by setting this locale thing and if I am doing it in the best way. Please let me know

ping @millerpuckette and @umlaeute

### DIFF
--- a/src/x_connective.c
+++ b/src/x_connective.c
@@ -9,6 +9,7 @@
 #include <string.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <locale.h>
 #ifdef _WIN32
 # include <malloc.h> /* MSVC or mingw on windows */
 #elif defined(__linux__) || defined(__APPLE__) || defined(HAVE_ALLOCA_H)
@@ -1345,6 +1346,7 @@ static void *makefilename_new(t_symbol *s)
     x->x_format = s;
     x->x_accept = NONE;
     makefilename_scanformat(x);
+    setlocale(LC_NUMERIC, "en_US.UTF-8");
     return (x);
 }
 


umlaeute (commented):
calling `setlocale` can have all kinds of unexpected side-effects (esp. with externals that use 3rd party libraries).

not sure if this is worth the benefit of supporting `'`

Spacechild1:
`setlocale` is evil because it modifies global state. A hard no from my side, unfortunately.

porres:
This is what my intuition was telling me, and AI also mentioned some caveats, but I don't really know much, so I proposed. And now I hope to learn a bit more from you and try to grasp what is the potential damage, especially now that I am considering supporting this in an external. How can it ruin other externals for instance, and which "kind" of external exactly?

Spacechild1:
`setlocale` globally changes the behavior of *all* `printf` calls. Do you feel comfortable doing that? As for the exact implications, I'm not 100% sure, that's why I would be very wary of doing it. In particular for such a tiny obscure feature. Also note that Pd can be used as a library and in that case setting the locale would be even worse.

Fun fact: `libpd_init` explicitly calls `setlocale(LC_NUMERIC, "C");` -- unless `LIBPD_NO_NUMERIC` is defined -- to get the default C locale, probably for the case where some other library has (accidentally) changed the locale. None of these shenanigans would be necessary with other programming languages, it's really a defect of the C language.

---

Side note: the Pd repository currently has 461 open issues and 109 open PRs. I have the feeling that sometimes your are requesting/implementing things just for the sake of it ("consistency"), but this takes away mental capacity -- including your own! --- that could be better spent elsewhere. Check your priorities! Do you *really* think we need to support an obscure `printf` specifier that most people have never heard of? If it would be trivial to implement - no problem, we can merge it to develop. But this is obviously not the case here.

porres:
> Do you feel comfortable doing that? 

Well, I was clear that I suspected maybe this shouldn't be messed with, so it's not like I'm insisting on it :) I'm still curious about the implications as I am still thinking wether to support this on externals, but I guess that's a separate concern I can bring to the pd-dev list.

> sometimes your are requesting/implementing things just for the sake of it ("consistency")

Not sure what you mean about consistency, but this (and maybe other stuff) do come from me checking things throughly via a deep documentation revision. This one is motivated by scrutinizing the features and trying to finally get right the help file of [makefilename].

In my last update in https://github.com/pure-data/pure-data/pull/2486/commits/c725ae282e4ae43fc33933a584ce6fa29bdf7e31 I had made it clear that the apostrophe flag isn't supported, so we can leave it at that. I just wanted to be sure we're not gonna support it and be done with it. I also thought that if we were not to support it that just mentioning it is not supported is enough about it.

Spacechild1:
> I am still thinking wether to support this on externals

Please do not mess with the globale locale! If you need better formatting options, try C++.

> I also thought that if we were not to support it that just mentioning it is not supported is enough about it.

+1

porres:
> Not sure what you mean about consistency, but this (and maybe other stuff) do come from me checking things throughly via a deep documentation revision.

Well, I just opened another issue about [makefilename] not handling lenght modifiers, and as I've been told about not bothering with some details that aren't important, I'd like to add on what I already said.

You know, I can see that point. On the other hand, I see this as part of my documentation work which requires a deep check up revision in some cases. At least as part of this task I think it is a valid effort, but sometimes it involves some nitpicking. Even so, I think it is for the best and makes things more solid in the long run and as many of little details are worked on.

As for priorities, I'd say the main priorities and issues of the documentation revision have been worked on and I may stumble upon lesser important issues by now, which is somewhat natural.

By the way, this is also motivated by my work in Cyclone and I'm finally deeply revising [sprintf] after 9 years of work on this project. It's got just 11 issues opened and this is one of them I've been ignoring for like 5 years for being a low priority indeed, but I got to it.

I guess I wanna say I agree and that I will try not to bother with silly and low priority things, but I hope you can see why I can't eventually help it.

Issue #2508, opened 2025-01-05
length modifiers 'l'/'L' are not working in [makefilename]


porres:
As I mention in https://github.com/pure-data/pddp/issues/264  the 'h' modifier does work, but we don't seem to handle the 'l'/'L' modifiers. And I guess all we need are these ones.

By the way, I am also revising and checking cyclone/sprintf and how this works in MAX. Cyclone's sprintf has been handling 'h'/'l' modifiers correctly.

While we're at it, I also see in https://github.com/pure-data/pure-data/pull/2491/commits/c9f6a63da9e6d09ff71afe955b716a301f917cef a mentioning that "we still do not handle length-modifiers properly".

porres:
I am documenting the behaviour of [makefilename] in both single and double precision Pd and I have realized that '%ld' won't really work in Pd double precision as it can only represent digits up to 2ˆ53. Also, only Pd in double precision can handle all possible values of '%d'.

I have documented 'h' and 'hh' but I also made a remark that this shouldn't be bothered with in the context of Pd.

That is to say that maybe we shouldn't bother in implementing 'l'/'L' modifiers as it'd only work in Pd64 if it had a special INT data type. 

What do you say @umlaeute, since you were making the remark in https://github.com/pure-data/pure-data/commit/c9f6a63da9e6d09ff71afe955b716a301f917cef that we don't handle length modifiers properly?

Right now in my last documentation revision I am only saying that 'l'/'L' isn't supported "yet". Maybe we could say it is intentionally not support it and if we do support it we need to document that Pd is still limited to take care of it.

porres:
I get what happens now. The code check for some characters for flags and width/precision fields ("-.#0123456789"), but it does not check for all possibilities (such as the "+" or "space" flags) and there is no logic to handle when the some invalid characters is not found.

So this lets the modifiers go through and they kinda do work, but we have a fixed size regardless of the length modifier. For instance, currently we just have "INT" type, but the https://github.com/pure-data/pure-data/commit/c9f6a63da9e6d09ff71afe955b716a301f917cef commit also adds the "UINT" type (which fixes an existing bug).

The 'h'/'hh' modifiers seems to work because it shortens the length and this is taken care of by sprintf itself internally. The 'l'/'ll' and 'L' go through but don't really work as the size is fixed.

Therefore, ideally, we would need to check for these modifiers and change the size properly and accordingly. Something like:

*typ = INT;  // Default for %d, %i
[...]

PR #2510, opened 2025-01-07
improvements to [makefilename]

porres:
1st commit is a simple addition to include '%F'.

I wasn't planing on making a 2nd commit, but here it is and it: 
- closes https://github.com/pure-data/pure-data/issues/2508

If it is too intrusive for the 'development' branch let me know

Of course I will add these to the docs after it all gets merged

### DIFF
--- a/src/x_connective.c
+++ b/src/x_connective.c
@@ -1331,6 +1331,7 @@ static void makefilename_scanformat(t_makefilename *x)
     const char *str;
     t_printtype typ;
     if (!x->x_format) return;
+    x->x_long = 0;
     str = x->x_format->s_name;
     str = _formatscan(x, str, &typ);
     x->x_accept = typ;
@@ -1353,7 +1354,6 @@ static void *makefilename_new(t_symbol *s)
     outlet_new(&x->x_obj, &s_symbol);
     x->x_format = s;
     x->x_accept = NONE;
-    x->x_long = 0;
     makefilename_scanformat(x);
     return (x);
 }
--- a/src/x_connective.c
+++ b/src/x_connective.c
@@ -1279,10 +1279,11 @@ typedef struct _makefilename
 {
     t_object x_obj;
     t_symbol *x_format;
+    t_int x_long;
     t_printtype x_accept;
 } t_makefilename;
 
-static const char* _formatscan(const char*str, t_printtype*typ) {
+static const char* _formatscan(t_makefilename *x, const char*str, t_printtype*typ) {
     int infmt=0;
     for (; *str; str++) {
         if (!infmt && *str=='%') {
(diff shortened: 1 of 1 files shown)


porres:
Hi @umlaeute for the next develop merging, would you consider at least including the commit https://github.com/pure-data/pure-data/pull/2510/commits/886804537a4107eb8fb55a608f0a582bafed450b which also adds support to '%F'?

Again, I think I may have stretched things a bit on the 2nd commit for a develop branch merge, what do you say?

src/x_connective.c
```diff
@@ -1365,10 +1370,12 @@ static void makefilename_float(t_makefilename *x, t_floatarg f)
         sprintf(buf, "%s",  x->x_format->s_name);
         break;
     case INT:
-        sprintf(buf, x->x_format->s_name, (int)f);
+        sprintf(buf, x->x_format->s_name,
+            x->x_long ? (long)f : (int)f);
```
umlaeute, line 1374: this may not work, as the ternary operator can only return a single type, so it promotes the `(int)f` to a `long` before passing it to `sprintf()`.

umlaeute (changes requested):
make sure to pass the correct types to variadic functions.

umlaeute:
> commit [8868045](https://github.com/pure-data/pure-data/commit/886804537a4107eb8fb55a608f0a582bafed450b) which also adds support to '%F'?

why?
i'm not opposed per se, but i fail to see why anyone would need  `%F`

porres:
> i'm not opposed per se, but i fail to see why anyone would need %F

The difference is that 'inf' and 'nan' are printed in capital letters. The same happens in this case for '%A', '%E' and '%G'.

Well, it's simply because it is a valid specifier, and we already have '%E', and '%G', and now '%A' (not to mention %X), and arguably I would also not see why the capital letter ones would be 'needed'. So it's just for the sake of completeness and consistency. I can't see why it should be specifically the only one to be left out. Specially now that we are also adding the '%a'/'%A' specifiers, which I also consider not to be something people were really needing and expecting.

src/x_connective.c
porres: An 'if' statement should solve this the, huh?

porres: > An 'if' statement should solve this then, huh?

@umlaeute did you see https://github.com/pure-data/pure-data/pull/2510/commits/81d6cdbb2f18a0f2837e04d3e8f0fbe6ac4df375 ? I tried using an if statement

src/x_connective.c
```diff
@@ -1279,10 +1279,11 @@ typedef struct _makefilename
 {
     t_object x_obj;
     t_symbol *x_format;
+    t_int x_long;
```
umlaeute, line 1282: please do not use `t_int` outside of the process functions.
use a simple `int` instead.

porres: done with https://github.com/pure-data/pure-data/pull/2510/commits/fc2b457aac7daf62ea39cc932be1eb246a6f7208

porres:
> sending [1e10( to [makefilename %d] will output -2147483648

I actually get '2147483647', which is the same output as before by the way.

> sending [set %d, 1e10( to [makefilename %ld] will output 10000000000 instead of -2147483648 💣

Good Catch! I see x->x_long needs to be zeroed when using the set message... oops. Just made a fix for it and will updload right next!

Issue #2446, opened 2024-09-27
Toggling 'use callbacks'  in the jack menu hangs Pd when JACK is provided by PipeWire


reduzent:
How to reproduce:
- Make sure that PipeWire's JACK shim is running (as opposed to the original JACK server)
- Run Pd with `-jack`
- Open menu `Menu` -> `jack`
- Toggle `Use callbacks`
- Click `Apply`

Pd freezes and the watchdog starts printing `watchdog: signaling pd...`

Tested:
- on Ubuntu 24.04 (comes with PipeWire as default backend)
- with Pd 0.55.1

reduzent:
@umlaeute I recently read that you're using JACK as provided by Pipewire. Do you experience the same issue? I haven't had the opportunity to test on another system than mine.

reduzent:
It seems the PR #2512 fixes it

umlaeute:
i just checked with my jack replacement daemon (aka: pipewire) and toggling the "use callbacks" option does indeed not freeze. (i get a short drop out, but i guess that is expected :-))

i've also tested with a proper `jackd2` daemon (but only with the `dummy` backend), and no freezes either.

so I guess this can be closed for good.

reduzent:
Works for me, too.

PR #2512, opened 2025-01-12
Try to open audio device synchronously

Spacechild1:
### Problem

Since Pd 0.55, `sys_reopen_audio()` does not immediately (re)open the audio device -- which would deadlock in the callback scheduler --, instead it posts a message to the scheduler and the device is only (re)opened at the beginning of the next scheduler tick.

Unfortunately, this means that for the remaining tick Pd objects still see the old samplerate, which breaks patterns like these:

grafik

This happens, for example, when you start Pd with the Jack backend and with a samplerate other than 48000 (= the default samplerate). It also happens when you change the samplerate in the audio settings.

---

### Solution

If we are in the main thread, we can safely close and (re)open the audio device *synchronously*. This makes sure that objects immediately see the actual samplerate. For example, the sample rate might have been changed by the user in the audio settings, or the actual rate might differ from the initial (default) rate.

If we are in the audio thread, however, we can't do this because it would deadlock; instead we send a request to main thread. However, we cannot do this immediately because we don't want the main thread to interrupt us halfway and accidentally enter the polling scheduler, just because we temporarily turned off DSP.
Instead we capture all requests to open and close the device in `sched_request`. Only at the end of the audio callback we send the most recent request to the main thread, which in turn will either close or (re)open the device. This allows users to
toggle the DSP state multiple times within a single scheduler tick.
See `m_callbackscheduler()` and `sched_audio_callbackfn()`.

When we start audio with callbacks we need to break from the polling scheduler. This is done by setting `sys_quit` to `SYS_QUIT_RESTART`.
*NOTE*: in this case it is **NOT** possible to further (re)open or close the audio device in that same scheduler tick because it would deadlock. For now, we just print an error message, but LATER we might want to fix this. In practice, there is hardly a reason why a user would want to start audio and then immediately stop or (re)open it. Note, however, that it's always ok to toggle DSP states while audio is already running (which has its use cases).

I also make sure that `sys_do_reopen_audio()`, `sys_do_close_audio()`, etc. are always called with Pd locked. (Previously, only `sys_setchsr` would lock Pd when updating DSP.) These functions contain various calls to `pd_error()` and `logpost()` which are not thread-safe.

---

I did a few regression tests and the following scenarios all still work:

- [x] Portaudio/Jack: start/stop DSP with polling and callback scheduler

- [x] Portaudio/Jack: switch between polling and callback scheduler while DSP is on.

- [x] Portaudio/Jack: toggle DSP off and on in the same clock tick, e.g. `[1, 0, 1( -> [;pd dsp $1(`, with polling and callback scheduler

- [x] Portaudio: automatically reopen audio device when disconnected (polling and callback scheduler)

---

I also took the chance to fix two other issues I've encountered on the way:

1. Jack: don't automatically start audio if already open, e.g. with `[loadbang] -> [;pd dsp 1(` in a patch opened on the command line.

2. Jack: don't update DSP in `jack_srate()` when called from within `jack_open_audio()`

### DIFF
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -28,7 +28,9 @@
 #define SYS_QUIT_QUIT 1
 #define SYS_QUIT_REOPEN 2
 #define SYS_QUIT_CLOSE 3
+#define SYS_QUIT_RESTART 4
 static int sys_quit;
+static pthread_t mainthread_id;
 static pthread_cond_t sched_cond;
 static pthread_mutex_t sched_mutex;
 static int sched_useaudio = SCHED_AUDIO_NONE;
@@ -226,6 +228,7 @@ void sched_init(void)
 {
     pthread_mutex_init(&sched_mutex, 0);
     pthread_cond_init(&sched_cond, 0);
+    mainthread_id = pthread_self();
 }
 
 void sched_term(void)
@@ -234,10 +237,15 @@ void sched_term(void)
     pthread_cond_destroy(&sched_cond);
 }
 
+int sys_ismainthread(void)
+{
+    return pthread_equal(pthread_self(), mainthread_id);
+}
+
 void dsp_tick(void);
 
     /* ask the scheduler to quit; this is thread-safe, so it
-       can be safely called from within the audio callback. */
+    can be safely called from within the audio callback. */
 void sys_exit(int status)
 {
     pthread_mutex_lock(&sched_mutex);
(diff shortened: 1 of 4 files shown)


Spacechild1:
@reduzent @danomatika @umlaeute Can you give this a try and see if you notice any weirdness?

Spacechild1:
I fear there might be issues with libpd...

`libpd_init` calls `sched_set_using_audio(SCHED_AUDIO_CALLBACK);`.

If I would send a "dsp 1" message from the same thread where I called `libpd_init()`, `sys_reopen_audio()` would mistakenly post an error message because it thinks we are reopening the callback scheduler from the main thread.

One easy workaround would be to call `sched_set_using_audio` with `SCHED_AUDIO_NONE` instead (or just omit the call alltogether). What would be the implications?

---

Ideally, libpd should not use any audio backend functionality at all. There is no reason why it should call functions like `sys_reopen_audio()` in the first place.

Here's an idea: AFAICT, `USEAPI_DUMMY` is only defined if there is no audio backend. In this case we can do the following:
```
void sys_reopen_audio(void)
{
#if !USEAPI_DUMMY
// current code
}

void sys_close_audio(void)
{
#if !USEAPI_DUMMY
// current code
}
```
This should be ok because there is no audio device that could be opened or closed in the first place.

libpd already defines `USEAPI_DUMMY=1`, so this would solve any potential issues.

@umlaeute @danomatika What do you think?

reduzent:
Re weirdness: When toggling 'use callbacks' all existing JACK connections disappear. I _believe_ this is a new behavior, but I cannot really test since I'm on a Pipewire system since a few months and toggling 'use callbacks' would hang Pd when using JACK from Pipewire. 

With your PR, I can toggle 'use callback' again. In case it gets merged, this issue can be closed:

https://github.com/pure-data/pure-data/issues/2446

Spacechild1:
> Re weirdness: When toggling 'use callbacks' all existing JACK connections disappear.

That's expected. The Jack backend uses different audio callback functions for audio with and without callbacks, so we necessarily have to reopen the stream. In general, you can't expect to keep the current audio stream when you change audio settings.

reduzent:
> > Re weirdness: When toggling 'use callbacks' all existing JACK connections disappear.
> 
> That's expected. The Jack backend uses different audio callback functions for audio with and without callbacks, so we necessarily have to reopen the stream.

Thanks for the explanation. It seems to work fine for me.

Spacechild1:
I just went ahead and made `sys_reopen_audio()` and `sys_close_audio()` no-ops if `USEAPI_DUMMY` is defined.

@danomatika If you have time, can you test on libpd? I'm pretty sure this should avoid any unwanted side effects, but you never know :)

danomatika:
You can try running the libpd test project which is included in pure-data now. I didn't write that part of the wrapper but I believe those functions are called in order to setup internal buffers etc. The test project would blow up if something is missing.

I need to setup a libpd branch here to use with the original repo, but have not done so yet.

Spacechild1:
`test_libpd` seems to work fine with this PR!

Jonathhhan:
@Spacechild1 do you know, if my issue: https://github.com/cuinjune/Ofelia/issues/102 could be potentially solved with your PR?

Issue #2505, opened 2025-01-01
soundfiler drops initial entry from ascii file


kaimikael:
pd version 0.55-1
macOS 15.2/Intel

In the shown patch (.pd files seem not to be attachable here) I attempt to read numbers from a text file (also included) with soundfiler.  I have truncated the actual data to just two numbers to demonstrate that the first value (1800) gets read as 0.  In the full data file, some values in the middle series also get misread and replaced by random values, but this varies between runs and isn’t reliably repeatable.
I am also confused by the graph shown, which seems to run upside-down—does it use a top-left origin?  (Why?).

[trunc.txt](https://github.com/user-attachments/files/18286415/trunc.txt)
<img width="1054" alt="Screenshot 2025-01-01 at 20 25 10" src="https://github.com/user-attachments/assets/e94203d2-7cc2-4411-b285-7864ea90521c" />

ben-wes:
> the first value (1800) gets read as 0

i assume that the parsing fails because there's a BOM in your text file, which Pd doesn't handle (maybe someone else can correct me on this if i'm wrong) - so it can't interpret that as a numeric value and replaces it with a zero (as mentioned in the help file: "non-numeric fields are replaced by zero").

for comparison (i saved the second one without BOM):
```
xxd trunc.txt               
00000000: efbb bf31 3830 3020 3138 3031 0a         ...1800 1801.

xxd trunc_test.txt
00000000: 3138 3030 2031 3830 31                   1800 1801
```
* file is [here](https://github.com/user-attachments/files/18286543/trunc_test.txt)
* more info: https://de.wikipedia.org/wiki/Byte_Order_Mark

> it use a top-left origin? (Why?).

Pd's canvases have top-left origins, yes. garrays compensate for that with their default range, which goes `from 1 to -1`. so you could do the same here and set your Y range from 3000 to 0. again, someone else might explain the details here better - but i hope it helps!

kaimikael:
Oh shoot, I didn’t even think of checking for a BOM, but that makes sense.  Thank you!

porres:
I have no idea what you're talking about :) but you can propose changes ;)

umlaeute:
> at the same time, a BOM is not ASCII

that would have been my answer.

kaimikael:
[...]

PR #2516, opened 2025-01-15
soundfiler: report parse errors when loading ascii file

ben-wes:
adds warning for unparseable elements when loading ASCII data.

suggestion came up in #2505 - more specifically https://github.com/pure-data/pure-data/issues/2505#issuecomment-2590440250

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -1146,7 +1146,7 @@ static int soundfiler_readascii(t_soundfiler *x, const char *filename,
     t_asciiargs *a)
 {
     t_binbuf *b = binbuf_new();
-    int i, j, vecsize;
+    int i, j, vecsize, parse_errors = 0;
     ssize_t framesinfile, nframes = a->aa_nframes;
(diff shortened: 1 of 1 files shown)


ben-wes:
@umlaeute thanks for the previous comment on the original commit (for some reason, i could see that only in my mail and not on github). indentation is fixed.

what i'm not sure about here is the redundant comparison `ap->a_type != A_FLOAT` and then again in `atom_getfloat`. let me know if i should remove that (by omitting `atom_getfloat` - although that doesn't seem like the most elegant option neither).

umlaeute:
the only reason to worry about the redundant comparison is performance.

and if anybody is worrying about the performance penalty if a single comparison, they probably should not use text-files for reading numeric data in the first place.

umlaeute:
and about:

> thanks for the previous comment on the original commit (for some reason, i could see that only in my mail and not on github).


github has a tendency to hide comments in ways that are beyond me. I find that they are often hidden behind miniscule user-icons that happen to be drop-down menus.

Issue #672, opened 2019-07-03
allow for shebang as the first-line of a patch


umlaeute:
on unix-system, executable scripts can specify the interpreter via a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) in the first line.

If Pd would sport such a feature, this would allow to create "Pd-applications" that can be launched from the cmdline
- without wrapper scripts
- without file-associations (that's so early 80s!)
  - you could even have a patch without the `.pd` extension that would automatically launch Pd...

the shebang is simply a first-line of a script that starts with `#!`.

The only thing that would need to be implemented on the Pd-side is to ignore a line in the patch
- if it starts with `#!`
- and is the very first line in the patch-file

danomatika:
Good idea.

umlaeute:
if we are special-handling the beginning of files, we might as well handle BOMs gracefully (https://github.com/pure-data/pure-data/issues/2505).

PR #2517, opened 2025-01-15
shebang support

umlaeute:
this PR adds basic support for Pd-files that contain a shebang.

Closes: https://github.com/pure-data/pure-data/issues/672

### what's a shebang

a [shebang](https://en.wikipedia.org/wiki/Shebang_(Unix)) is a special character sequence at the beginning of an executable file, that tells the program loader which interpreter to use to execute the file.

a `.pd` file with a shebang could be used as a standalone script.

e.g. consider a Pd file `hello.pd`:

```pd
#!/usr/bin/env -S pd -nogui -open
#N canvas 2655 476 280 264 12;
#X msg 142 143 world;
#X obj 142 168 print hello;
#X obj 72 91 loadbang;
#X obj 72 123 t b b;
#X msg 72 148 \; pd quit;
#X connect 0 0 1 0;
#X connect 2 0 3 0;
#X connect 3 0 4 0;
#X connect 3 1 0 0;
```

when made executable (`chmod +x hello.pd`) you could simply run:

```sh
$ ./hello.pd
hello: world
$
```

## about the PR

to support shebangs, Pd simply has to ignore the first line of a patch file if it starts with `#!`.

unfortunately:
- Pd uses a unified method for reading text-files into binbufs, and then evaluates the binbuf
- a shebang line must not be semicolon-terminated (it is simply terminated with `\n`)

this is a bit problematic as:
- we do not want to ignore the shebang when opening general file with `binbuf_read()` (e.g. with the `[text]` object)
- since Pd-patches use semicolon as line terminator, the CR terminating the shebang-line is simply swalled by `binbuf_read()`

the solution is to turn the `binbuf_read()`'s `crflag` into a general flag, so we can tell `binbuf_read()` to explicitly ignore the shebang if so desired (e.g. in the context of `binbuf_evalfile()`)

I have checked all the pd-externals packaged in Debian, and none use `crflags` with any value other than `0` or `1`.
Since the new flag for ignoring the shebang is `| 2`, I do not expect any problems.

### persistence

as of now, saving a patch-file within Pd will clear any shebang it might have had.

### DIFF
--- a/src/m_binbuf.c
+++ b/src/m_binbuf.c
@@ -781,10 +781,10 @@ broken:
 
 int binbuf_read(t_binbuf *b, const char *filename, const char *dirname, int crflag)
 {
-    long length;
+    long length, length0;
     int fd;
     int readret;
-    char *buf;
+    char *buf, *buf0;
     char namebuf[MAXPDSTRING];
 
     if (*dirname)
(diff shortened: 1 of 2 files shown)


umlaeute:
Ah well... This pr depends on a commit in current `develop`.
But you are probably right and I should just rebase including the required commits...

ben-wes:
just noting https://github.com/porres/pure-data/pull/86 here. it will revert the soundfiler comment on BOMs for ASCII loading, which becomes obsolete with c714ff6.

umlaeute:
https://github.com/pure-data/pure-data/commit/c714ff6de22ef3d254dd3d85ee07d64b49beeb68 is also in `develop` as 43b4e8291a7d8f981a876a2526e532e49c0daeac (together with your 589028bfcdfe74d7243b850f2ac4f0ee88bb0f42, this is the reason why the `develop` branch is marked as closing  #2505)

ben-wes:
> this is the reason why the `develop` branch is marked as closing #2505

got it. since my previous changes to the help file were already merged on @porres' side, i mainly wanted to comment here that there's another PR considering your additional stuff here.

(might have caused more confusion by trying to prevent it)

PR #2522, opened 2025-01-20
Support right clicks on macOS with Wish9.x

### LINKED ISSUE #2521: Wish 9.0.1 UI not getting right clicks
this likely solvable 3rd-button problem occurred when testing the changes solving #2503 (#2520).

i can't see right clicks working - not with an actual mouse and not with 2-finger-tap, whereas ctrl+click works.

system:
* macOS 15.3
* arm64

ben-wes:
closes #2521


tcl/pd_bindings.tcl
```diff
@@ -293,6 +293,9 @@ proc ::pd_bindings::patch_bindings {mytoplevel} {
     # "right clicks" are defined differently on each platform
     switch -- $::windowingsystem {
         "aqua" {
+            # TclTk-9.x
+            bind $tkcanvas <ButtonPress-3>    "pdtk_canvas_rightclick %W %x %y %b"
```
ben-wes, line 297: not sure if a `foreach` would be preferable here instead?

```tcl
foreach btn {2 3} {
    bind $tkcanvas <ButtonPress-$btn> "pdtk_canvas_rightclick %W %x %y %b"
}
```

umlaeute: this obviously binds both button-2 and button-3 to the right-click.
is this desired? (i dunno, just asking)

ben-wes: yep, this is certainly not the cleanest option. the problem is that `Button-3` and `Button-2` were obviously swapped on macOS with Tcl/Tk 9.0. With a mouse connected, `Button-2` is now the middle button and `Button-3` the right button (consistent with other systems), while it was the opposite before.

so with this update, right click is now triggered with both, middle and right button.

while i first thought that this was acceptable since there's no function for the middle button, i now force pushed a change querying the tcl version.

ben-wes:
now thinking that the declaration of `$rightbtn` is concise, but rather unnecessary - maybe we should just go with this simple and slightly more verbose version?:

```tcl
if {$::tcl_version < 9} {
    bind $tkcanvas <ButtonPress-2> "pdtk_canvas_rightclick %W %x %y %b"
} else {
    bind $tkcanvas <ButtonPress-3> "pdtk_canvas_rightclick %W %x %y %b"
}
```

PR #2531, opened 2025-02-03
Check for multiple dollar signs in binbuf_restore

myQwil:
In the event that the first dollar sign we come across isn't related to a dollarsym atom, we need to keep looking. Otherwise, there will be inconsistencies between a patch's state upon saving, and its state after closing and re-opening.

Example: in the message "$$1", the 1st $ is actually a dollarsym's prefix, so we need to check the 2nd $ as well.

Here is a video showing a before and after of this PR:

https://github.com/user-attachments/assets/41db09b5-5ce1-4f67-a648-7bdc5dc0cc4c

### DIFF
--- a/src/m_binbuf.c
+++ b/src/m_binbuf.c
@@ -20,6 +20,15 @@
 
 #include "m_private_utils.h"
 
+    /* returns the start of a valid dollar or dollsym */
+static const char *str_dollar(const char *s)
+{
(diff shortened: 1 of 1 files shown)


millerpuckette:
This makes my head hurt...  but OTOH if you wanted to generate a symbol "$1" this would be one way to do it :)

PR #2535, opened 2025-03-02
add new API function `pd_findbyclassname`

Spacechild1:
The standard method for data sharing in Pd is to bind objects to symbols. Internally, objects call `pd_findbyclass` to get another object from a symbol. Since every symbol can be bound to multiple objects with the same name (but with different classes) the user needs to pass a pointer to the expected class (`t_class*`). This is all fine as long as all objects are defined in the same module (e.g. in the Pd core or in a single-binary-library). However, as soon as the objects live in different binaries, you run into problems because you need to share the class pointer across DLL boundaries, which turns out to be very tricky.

To solve this problem, I've introduced a new function called `pd_findbyclassname`. This is the same as `pd_findbyclass()`, except that you pass the class name instead of the class pointer. This way you can share Pd objects without being forced to compile everything as a single binary. It even allows to share Pd objects between different external libraries, which hasn't been possible so far AFAICT!

### DIFF
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -6,6 +6,8 @@
 #include "m_imp.h"
 #include "g_canvas.h"   /* just for LB_LOAD */
 
+#include <string.h>
+
     /* FIXME no out-of-memory testing yet! */
 
 t_pd *pd_new(t_class *c)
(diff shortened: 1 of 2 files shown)


umlaeute:
`pd_findbyclassname` would probably be a better name.

PR #2538, opened 2025-03-04
support ctrl+click in pd window on macos

ben-wes:
this adds "right click" via ctrl+click via `<Control-Button-1>` in the Pd window for macOS similar to the canvas window behaviour:

https://github.com/pure-data/pure-data/blob/eaaf7991748a71412686a6d4a52972d67c4d4933/tcl/pd_bindings.tcl#L299-L300


umlaeute:
i got it (although, if i hadn't gat get it, i probably wouldn't have got your 2nd comment either...;-))

and yes: i was sloppy while reviewing

PR #2541, opened 2025-03-04
support tcl9 for pd window right click on macos

ben-wes:
support right click on error messages in Pd Window on macOS with Wish9.x


ben-wes:
@umlaeute btw ... i just learnt that all log lines with an object_id allow to find their sources. pretty cool. totally makes sense, but took me a few years and your hover cursor to see this. :)

(and makes me reconsider simply using `post()` in externals...)

PR #2542, opened 2025-03-06
vline~: delay segments with non-zero length by 1sp

### LINKED ISSUE #2534: vline~ seems to skip line start sample
testing develop build at https://github.com/pure-data/pure-data/commit/0a872a9aa47f2410323e8811c0f6006ee3c1165c on arm64 macOS.

i might be doing sth wrong - but this looks unexpected to me (sample rate 48000, block size 64):

```
[bang~]
| \
|  [0, 1 1.33333(    <-- ~64sp duration
|  |
|  [vline~]
| /
[tabwrite~ sig]

[array define sig 64]
```

the sample at index 0 has a value of `0.015625`, which is ~1/64. so the line seems to start with the 2nd sample here (getting the same result if i bang manually). if i add a [del 0.020833] before the message, the result looks "more correct" (~0.020833 being the ms/sp at 48k sample rate).

or is it intended that the 1st sample has the value after the 1st "step"?

ben-wes:
closes #2534

that's obviously a significant change in behavior - ~~therefore i set this as a draft~~. it would be effective from Pd >= 0.56 based on the `pd_compatibilitylevel` check.

i assume that this changeset might be the simplest way to achieve the expected (?) result - after first attempting to fix this in the perform function.

EDIT: adding a patch to test and compare for a few cases: [vline~-test.pd.zip](https://github.com/user-attachments/files/19189715/vline.-test.pd.zip)

### DIFF
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -295,6 +295,12 @@ static void vline_tilde_float(t_vline *x, t_float f)
     t_float inlet1 = (x->x_inlet1 < 0 ? 0 : x->x_inlet1);
     t_float inlet2 = x->x_inlet2;
     double starttime = timenow + inlet2;
+
+        /* add one-sample delay to segments with non-zero length to include
+        their start value */
+    if (inlet1 > 0 && pd_compatibilitylevel > 55)
+        starttime += x->x_msecpersamp;
+
     t_vseg *s1, *s2, *deletefrom = 0, *snew;
     if (PD_BIGORSMALL(f))
         f = 0;
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -216,11 +216,9 @@ typedef struct _vline
     t_vseg *x_list;
 } t_vline;
 
-    /* new perform function for Pd 0.56 to avoid incremented values in
-    ramps' start samples. vline_tilde_perform_old (below) is still used
-    for previous compatibility levels */
 static t_int *vline_tilde_perform(t_int *w)
 {
+    const int compat = (pd_compatibilitylevel < 56) ? 1 : 0;
     t_vline *x = (t_vline *)(w[1]);
     t_sample *out = (t_sample *)(w[2]);
     int n = (int)(w[3]), i;
@@ -240,22 +238,26 @@ static t_int *vline_tilde_perform(t_int *w)
     for (i = 0; i < n; i++)
     {
         double timenext = timenow + msecpersamp;
+            /* new mechanism for Pd 0.56 to avoid incremented values
+            in ramps' start samples. old mechanism is used for previous
+            compatibility levels */
+        double timecheck = compat ? timenext : timenow;
     checknext:
             /* new segment is present and starts before next sample */
-        if (s && s->s_starttime < timenext)
+        if (s && s->s_starttime < timecheck + msecpersamp)
         {
                 /* update value if targettime elapsed */
-            if (x->x_targettime <= timenow)
+            if (x->x_targettime <= timecheck)
                 f = x->x_target;
                 /* bash output value for zero-length segment */
             if (s->s_targettime <= s->s_starttime)
-                f = s->s_target, inc = 0;
+                f = s->s_target;
                 /* manage regular increment if starttime elapsed */
-            if (s->s_starttime <= timenow)
+            if (s->s_starttime <= timecheck)
             {
                 double incpermsec = (s->s_target - f) /
                     (s->s_targettime - s->s_starttime);
-                f = f + incpermsec * (timenow - s->s_starttime);
+                f = f + incpermsec * (timecheck - s->s_starttime);
                 inc = incpermsec * msecpersamp;
                 x->x_inc = inc;
                 x->x_target = s->s_target;
@@ -266,68 +268,7 @@ static t_int *vline_tilde_perform(t_int *w)
                 goto checknext;
             }
         }
-        if (x->x_targettime <= timenow)
-            f = x->x_target, inc = x->x_inc = 0, x->x_targettime = 1e20;
-        *out++ = f;
-        f = f + inc;
-        timenow = timenext;
-    }
-    x->x_value = f;
-    return (w+4);
-}
-
-static t_int *vline_tilde_perform_old(t_int *w)
-{
-    t_vline *x = (t_vline *)(w[1]);
-    t_sample *out = (t_sample *)(w[2]);
-    int n = (int)(w[3]), i;
-    double f = x->x_value;
-    double inc = x->x_inc;
-    double msecpersamp = x->x_msecpersamp;
-    double timenow, logicaltimenow = clock_gettimesince(x->x_referencetime);
-    t_vseg *s = x->x_list;
-    if (logicaltimenow != x->x_lastlogicaltime)
-    {
-        int sampstotime = (n > DEFDACBLKSIZE ? n : DEFDACBLKSIZE);
-        x->x_lastlogicaltime = logicaltimenow;
-        x->x_nextblocktime = logicaltimenow - sampstotime * msecpersamp;
-    }
-    timenow = x->x_nextblocktime;
-    x->x_nextblocktime = timenow + n * msecpersamp;
-    for (i = 0; i < n; i++)
-    {
-        double timenext = timenow + msecpersamp;
-    checknext:
-        if (s)
-        {
-            /* has starttime elapsed?  If so update value and increment */
-            if (s->s_starttime < timenext)
-            {
-                if (x->x_targettime <= timenext)
-                    f = x->x_target, inc = 0;
-                    /* if zero-length segment bash output value */
-                if (s->s_targettime <= s->s_starttime)
-                {
-                    f = s->s_target;
-                    inc = 0;
-                }
-                else
-                {
-                    double incpermsec = (s->s_target - f)/
-                        (s->s_targettime - s->s_starttime);
-                    f = f + incpermsec * (timenext - s->s_starttime);
-                    inc = incpermsec * msecpersamp;
-                }
-                x->x_inc = inc;
-                x->x_target = s->s_target;
-                x->x_targettime = s->s_targettime;
-                x->x_list = s->s_next;
-                t_freebytes(s, sizeof(*s));
-                s = x->x_list;
-                goto checknext;
-            }
-        }
-        if (x->x_targettime <= timenext)
+        if (x->x_targettime <= timecheck)
             f = x->x_target, inc = x->x_inc = 0, x->x_targettime = 1e20;
         *out++ = f;
         f = f + inc;
@@ -355,7 +296,6 @@ static void vline_tilde_float(t_vline *x, t_float f)
     t_float inlet1 = (x->x_inlet1 < 0 ? 0 : x->x_inlet1);
     t_float inlet2 = x->x_inlet2;
     double starttime = timenow + inlet2;
-
     t_vseg *s1, *s2, *deletefrom = 0, *snew;
     if (PD_BIGORSMALL(f))
         f = 0;
@@ -411,8 +351,7 @@ static void vline_tilde_float(t_vline *x, t_float f)
 
 static void vline_tilde_dsp(t_vline *x, t_signal **sp)
 {
-    dsp_add(pd_compatibilitylevel > 55 ? vline_tilde_perform : vline_tilde_perform_old,
-        3, x, sp[0]->s_vec, (t_int)sp[0]->s_n);
+    dsp_add(vline_tilde_perform, 3, x, sp[0]->s_vec, (t_int)sp[0]->s_n);
     x->x_samppermsec = ((double)(sp[0]->s_sr)) / 1000;
     x->x_msecpersamp = ((double)1000) / sp[0]->s_sr;
 }
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -238,6 +238,63 @@ static t_int *vline_tilde_perform(t_int *w)
     {
         double timenext = timenow + msecpersamp;
     checknext:
+            /* new segment is present and starts before next sample */
+        if (s && s->s_starttime < timenext)
+        {
+                /* update value if targettime elapsed */
+            if (x->x_targettime <= timenow)
+                f = x->x_target;
+                /* if zero-length segment bash output value */
+            if (s->s_targettime <= s->s_starttime)
+                f = s->s_target, inc = 0;
+                /* manage regular increment if starttime elapsed */
+            if (s->s_starttime <= timenow)
+            {
+                double incpermsec = (s->s_target - f) /
+                    (s->s_targettime - s->s_starttime);
+                f = f + incpermsec * (timenow - s->s_starttime);
+                inc = incpermsec * msecpersamp;
+                x->x_inc = inc;
+                x->x_target = s->s_target;
+                x->x_targettime = s->s_targettime;
+                x->x_list = s->s_next;
+                t_freebytes(s, sizeof(*s));
+                s = x->x_list;
+                goto checknext;
+            }
+        }
+        if (x->x_targettime <= timenow)
+            f = x->x_target, inc = x->x_inc = 0, x->x_targettime = 1e20;
+        *out++ = f;
+        f = f + inc;
+        timenow = timenext;
+    }
+    x->x_value = f;
+    return (w+4);
+}
+
+static t_int *vline_tilde_perform_old(t_int *w)
+{
+    t_vline *x = (t_vline *)(w[1]);
+    t_sample *out = (t_sample *)(w[2]);
+    int n = (int)(w[3]), i;
+    double f = x->x_value;
+    double inc = x->x_inc;
+    double msecpersamp = x->x_msecpersamp;
+    double timenow, logicaltimenow = clock_gettimesince(x->x_referencetime);
+    t_vseg *s = x->x_list;
+    if (logicaltimenow != x->x_lastlogicaltime)
+    {
+        int sampstotime = (n > DEFDACBLKSIZE ? n : DEFDACBLKSIZE);
+        x->x_lastlogicaltime = logicaltimenow;
+        x->x_nextblocktime = logicaltimenow - sampstotime * msecpersamp;
+    }
+    timenow = x->x_nextblocktime;
+    x->x_nextblocktime = timenow + n * msecpersamp;
+    for (i = 0; i < n; i++)
+    {
+        double timenext = timenow + msecpersamp;
+    checknext:
         if (s)
         {
             /* has starttime elapsed?  If so update value and increment */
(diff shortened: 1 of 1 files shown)


ben-wes:
not sure how compatibility level should be managed here if the next planned release is `0.55-3` - which would make this a fix from the future?

umlaeute:
> not sure how compatibility level should be managed here if the next planned release is 0.55-3 - which would make this a fix from the future?

easy: let's merge this only for 0.56

Spacechild1:
> easy: let's merge this only for 0.56

After all, a bug fix release should not contain breaking changes ;-)

porres:
> After all, a bug fix release should not contain breaking changes ;-)

I won't repeat what I said in https://github.com/pure-data/pure-data/issues/2534 but I don't see this as a bug, but a feature instead.

My suggestion, don't change the behaviour, don't add a compatibility flag.

If you want to add a new feature of setting a start value, we can think of a new 'set' message.

So you can have [set 0, 1 1000( or whatever, in a way that the first output sample is '0', then things start moving onward afterwards.

This is good cause: A) no breaking change; B) no weird/ugly initialization flag (which would probably need a message as well to change in between modes; C) it simply adds the new feature/behaviour with a simple message and allows users to choose in between, while keeping existing patches intact

Spacechild1:
> I won't repeat what I said in https://github.com/pure-data/pure-data/issues/2534 but I don't see this as a bug, but a feature instead.

No, it's clearly a bug. See my comment: https://github.com/pure-data/pure-data/issues/2534#issuecomment-2709353684

porres:
We can keep discussing over there, but I am quite aware of the behaviour of the objects, and I've been extensively trying to make my point. I'm not missing anything and I think this comes down to different interpretations. If it comes to that, at least the way I see it preserves the current behaviour and what I am proposing does not introduce any breaking changes, which is good right? :)

I mean, I have patches and tutorials that rely on this!

Spacechild1:
I totally get the backwards compatibility concerns (I have raised them myself), but a `[set(` message for `[vline~]` is neither a feature nor a solution. The ramp is *always* off by 1 sample, even without an explicit start value.

src/d_ctl.c
```diff
@@ -295,6 +295,12 @@ static void vline_tilde_float(t_vline *x, t_float f)
     t_float inlet1 = (x->x_inlet1 < 0 ? 0 : x->x_inlet1);
     t_float inlet2 = x->x_inlet2;
     double starttime = timenow + inlet2;
+
+        /* add one-sample delay to segments with non-zero length to include
+        their start value */
+    if (inlet1 > 0 && pd_compatibilitylevel > 55)
+        starttime += x->x_msecpersamp;
+
```
Spacechild1, line 303: Conceptionally, I don't think this is the right fix because it generates a "wrong" start time. Note that the code below tries to deal with the jump-and-slide case, but it assumes a correct start time.

I think the issue should be tackeld in the perform routine.

ben-wes: that was my impression as well, i admit. it's irritating to see `starttime` being rescheduled here (maybe that's a semantics issue). what it actually does (which you're certainly aware of) is postponing the subsample-corrected first increment since this is otherwise already applied at start time with the subsample-offset in
https://github.com/pure-data/pure-data/blob/eaaf7991748a71412686a6d4a52972d67c4d4933/src/d_ctl.c#L258

code-wise, i'm afraid that we (or at least i - since i tried and wasn't able to find an elegant solution) would add a bit of complexity to schedule the subsample-corrected first increment and the regular increment.

umlaeute:
> > easy: let's merge this only for 0.56
> 
> After all, a bug fix release should not contain breaking changes ;-)

as a sidenote: if the change is guarded with a `pd_compatibilitylevel > 55`, the merge could of course happen now, and it will only show up automatically after the 0.56 release...

as long as we can be *sure* that this doesn't introduce any regressions, I would be fine with that...

ben-wes:
> it will only show up automatically after the 0.56 release

i didn't check if anything is manipulated for this in official builds - but afaict, new builds automatically show behaviour "from the future" based on 

https://github.com/pure-data/pure-data/blob/b39c5bad6154e91fa93d7b758912e775c1f4f6d0/src/m_glob.c#L12

the compatibility only gets set through explicit messages otherwise (that's why i was surprised first when i saw the new behavior on a local `0.55-2` build despite the compatibility flag).

src/d_ctl.c
ben-wes: sounds right. i might give it another try in the perform function. not sure if i'll find a good solution though. so expect a cry for help and don't hesitate if you see a fix yourself. ;)

ben-wes:
added a commit that moves this to the perform function side. not sure i properly wrapped my head around this - but it behaves correctly now and didn't become more complex. there's one more condition checked now for every segment:
https://github.com/pure-data/pure-data/pull/2542/commits/0a1f03badbc81ab7b4d8473dea6e03174ded89ef#diff-c9d48c26b25e816b05420c6ce24301b5198001e7321e72e22c9eff62c6e6c182R251

the changeset is also rather ugly since i split this now into an old and new perform routine and the changes aren't visible that way. let me know if i should update this to make it more readable.

Spacechild1:
Yeah, I guess it's better to just make a new perform routine to keep things readable. Can you also add a comment to explain the difference between the old and the new perform routine?

umlaeute:
Urgh.

If you start DSP before switching the compat level, you will get the "wrong" behaviour, even for messages sent afterwards. 

Also it seems that this is quite a lot of code...

ben-wes:
> If you start DSP before switching the compat level, you will get the "wrong" behaviour, even for messages sent afterwards.

switching compat levels pauses dsp though, right? for me, switching with active dsp works.
https://github.com/pure-data/pure-data/blob/b39c5bad6154e91fa93d7b758912e775c1f4f6d0/src/m_glob.c#L53-L58

> Also it seems that this is quite a lot of code...

can't argue with that. true. but it would become messy to achieve the old behavior in the new version now (for me). and it also seems wrong to have condition checks in a perform function just for this compatibility feature?

Spacechild1:
>  and it also seems wrong to have condition checks in a perform function just for this compatibility feature?

That's what I think as well. The function is not short, but it's not terribly long either. If this was C++, we could use a `bool` template parameter together with `if constexpr`, but unfortunately we are stuck with C, so that's what we have to deal with :shrug:

On the other hand, if the check is done at the beginning of the function and the result cached in an int, I don't think that 1 or 2 additional if-statements would have a measurable performance impact. After all, the int value is constant for the whole duration of the function, so the branch predictor would have an easy time guessing the correct result. But it's not exactly pretty either.

ben-wes:
gave it another try. now combining both behaviors in one function. still need to test this more. no the differentiation is done here:

https://github.com/pure-data/pure-data/blob/aed2bc2829f406dd116c9807ce16c410fcc8d5ed/src/d_ctl.c#L243

ben-wes:
... aaand another try. now `compat` is a `const int` at the beginning with the compatibility check. and the differentiation for the mechanism is made here:

https://github.com/pure-data/pure-data/blob/1cd35f98ef3c483b0780cde84bf8ad1e41e9c23f/src/d_ctl.c#L244

Spacechild1:
Doesn't look too bad!

ben-wes:
slightly OT - but i wonder if there are objections to using `PD_MINOR_VERSION` here instead of `100000`? this could then be part of this PR to achieve expected behaviour for pre 0.56 releases?

https://github.com/pure-data/pure-data/blob/b39c5bad6154e91fa93d7b758912e775c1f4f6d0/src/m_glob.c#L12

millerpuckette:
I should have put something in the documentation years ago... Actually I think it's line~ that behaves wrong and vline~ and phasor~ behave correctly - they both react immediately to ramps, whereas line~ waits one sample before starting to move.  Of course this goes against the 0-centric style of much of Pd such as table-indexing.  But on the other hand, line~ reacts one sample slower than vline~ or phasor~.  It might sound odd to worry about a one-sample latency, but certainly in the case of using vline~ as an oscillator, it makes sense for it to start moving immediately.

So... would it be appropriate to offer (optionally) a pre-increment flag for line~ ot bring all three into sync?

porres:
> Actually I think it's line~ that behaves wrong and vline~ and phasor~ behave correctly - they both react immediately to ramps, whereas line~ waits one sample before starting to move. 

wow, I said the same thing so many times! :) 

> would it be appropriate to offer (optionally) a pre-increment flag for line~ ot bring all three into sync?

my suggestion is to add a new message to [vline~] that behaves like the desired functionality, which jumps to the desired value, then all scheduled messages afterwards happen later.

so [set 0, 1 1000 0( forces a jump to 0, then it starts a ramp afterwards and takes 1000ms to reach "1"

porres:
> wow, I said the same thing so many times! :)

I see now this is the PR, and my remarks are on the issue discussion.

So, I get the idea of start incrementing the line ramp right away.

The way I see it, [vline~] can add a line that jumps to a start point and then you can schedule another line afterwards. This is how you could set a start point, then a duration and end point. But what you have to do is set two lines, one line that jumps to the start value, and another that goes to destination.

Then if people do [0 0 0, 1 1000 0( the scheduler overwrites the first line and already starts incrementing, so the right way would be to delay the 2nd line by a sample.

Now, people are expecting the first line to hold its value and have all following lines to start a sample later. Instead of using a float (which actually represents a list with 2nd and 3rd arguments that default to '0', so '0' time and '0' delay), you can have a "set" message instead. And it sets a star line point for the whole thing. This is convenient because you don't have to specify a sample delay time, which is clumsy.

No patches are broken

No compatibility flag needed

No creation argument needed

Old and new functionalities cohexist perfectly

Documentation can make things clear

porres:
> The way I see it, [vline~] can add a line that jumps to a start point and then you can schedule another line afterwards. 

that would happen a sample later, so it'd be [0 0 0, 1 1000 0.0226757( just to make it sure

so, one line that jumps to '0', another that starts afterwards

>  This is convenient because you don't have to specify a sample delay time, which is clumsy.

so, yeah  [set 0, 1 1000 0( is better than [0 0 0, 1 1000 0.0226757(

Spacechild1:
> Actually I think it's line~ that behaves wrong and vline~ and phasor~ behave correctly - they both react immediately to ramps, whereas line~ waits one sample before starting to move.

@millerpuckette I'm not sure I follow. `[phasor~]` starts its ramp from 0 (and never reaches 1 because it wraps around). When I reset the phase to 0.5, it starts its ramp with 0.5.

Here's a comparison of the original `[phasor~]` and the *fixed* version of `[vline~]`: https://github.com/pure-data/pure-data/issues/2534#issuecomment-2711907451

Without that fix, the output of `[vline~]` would run ahead of `[phasor~]` by one sample. AFAICT `[vline~]` is really the outlier here.

Also note that `[vline~]` finishes the ramp 1 sample too early. When I ask for a ramp from 0 to 1 over 64 samples (e.g. 1.333 ms @ 48 kHz) at t=0, `[vline~]` would already reach its target at t=63. To me that's a duration of 63 samples, not 64 samples...

That's how I would expect `[vline~]` to behave:

| ramp length | target reached |
|---|---|
| 0 | t[0] |
| 1 | t[1] |
| ... | ... |
| 64 | t[64] |

porres:
> @millerpuckette I'm not sure I follow. [phasor~] starts its ramp from 0 
> (and never reaches 1 > because it wraps around). When I reset the phase
> to 0.5, it starts its ramp with 0.5.

I may be off, but I believe it's the fact that it reacts and jumps to 0.5 right away, and doesn't wait for the next sample to go there.

> When I ask for a ramp from 0 to 1 over 64 sample

we can't ask for a ramp from start position, to end position, and at a given length (that's 3 parameters), we can just tell it to "start going from whenever you are to this target" in 'x' lenght (2 parameters). But let's say that the current value is at '0', and you tell it to go to '1' in 64 samples.

> That's how I would expect [vline~] to behave:

You have t[0] and t[64], so you have (N+1) as the line length. So I think you're expecting the whole thing it to last 65 samples, not 64. But arguably, they are both the same line and the same length, you just add a sample delay in your expectation, where instead of the line to start moving, you sit and wait at value '0'. Miller seems to expect that it should just start moving, and I get that expectation. I actually agree it makes good sense.

Spacechild1:
> we can't ask for a ramp from start position, to end position, and at a given length

Obviously we can, if we don't start the line with an increment. I understand, it's a matter of perspective. I find the behavior of `[line~]` more useful and intuitive in this regard. You seem to disagree.

Anyway, I would really like to hear Miller's reply to my concerns regarding the line duration.

> You have t[0] and t[64], so you have (N+1) as the line length.

No, the length of a horizontal line is `x2 - x1`. In this case that's `64 - 0 == 64`.

I refuse to accept the notion that a line going from `t[0]` to `t[0]` has a length of 1. That's a zero-length line, or in other words: a jump.

porres:
FWIW, I can see what you mean now. And I also was just hoping to get a better answer from you ;)

Spacechild1:
Ok, I made a little drawing to better demonstrate the issue:

grafik

Let's say I'm at `t[1]` and I want to make a line from the current value (= 1) to 2 over a period of 2 seconds.

- Shouldn't the target value be reached exactly at `t[3]`? If not, where should it be?

- And shouldn't the value at `t[1]` still be 1? If not, what should it be?

If we turn this continuous function into a discrete function, the answers should still be the same.

porres:
yeah, I see it. I also see the other way around though (at time 'x' start incrementing right away), and I guess when I first analyzed this object, I just created this logical rule on how it operates. 

Here's a similar graph by feeding [`0 0 0, 1 0.362812 0.362812, 0 0.362812 0.725624`( into [vline~], in order to start a ramp up after a delay of 16 samples, then another ramp down of 16 samples as well. This is both current with the behaviour (top part) and this PR's behaviour (bottom)

Screenshot 2025-03-11 at 00 16 23

It's just all shifted by a sample to the right in this PR as we can see.

Now please have a look on what I came up with many years ago. When I designed [else/envgen~] I wanted to have both behaviours, cause you know, both seem valid! But the syntax here is different, it takes a single list with pairs of duration/destination. So this is the same as the current behaviour of [vline~]

Screenshot 2025-03-11 at 00 34 10

Now, if you want the behaviour you're expecting, you can add a '0' element in the start of the list, so the object also understands an odd number of elements in a list and treats it as a staring point.

Screenshot 2025-03-11 at 00 37 08

By the way, this is also effectively the same as adding a first pair of '0 0', which is "jump to point 0 immediately", then add the other segments. It's just much easier because the target times are added, and the problem is that [vline~] needs us to do all the math ourselves by incrementing delay times for each scheduled line.

Why am I showing you this? Just to make a point that the structure of [vline~] is very similar to [envgen~] but you make it both ways... so I propose a "set <float>" message, and what it does is that it adds a starting point for the whole thing... it just adds a sample and delays all the rest. And hey, maybe we can call it "start <float>"?

Anyway, it's easy to document this in depth and detail, and also how a new message should work.
We can just describe the phenomenon of how [vline~] operates, regardless of what people expect or think it should do and I'm not debating that. I just can't believe I'm the one now saying "hey folks, don't break this", cause it's usually the other way around :) 

So, bottom line, do whatever, just don't try to create a compatibility flag ;) if you can. I'll actually be ok if it comes down to that. I will be able to live with it, gladly document the change of behaviour and everything you all agree, so I'm off ;) 

Before I leave, I could throw this quite wild idea on the table: have this or a new object to adopt the [envgen~] syntax, seems much more natural and easier to deal with :)

Bye...

porres:
Ok, "set <float>" is a bad idea if you want to toggle ramps up and down, so I'm taking this back, sorry for the noise...

A new 'mode' then? a creation argument flag?

ben-wes:
while the discussion is still ongoing, i've added a `-i` flag for `line~` and `vline~` to this PR for now. this activates "instant increment mode".

ben-wes:
unfortunately, i found a bug in my version for delayed zero-length segments. will attempt to fix this ...

porres:
> A new 'mode' then? a creation argument flag?

Sorry for not being clearer, I meant doing this for the new behaviour, so patches don't break :)

> while the discussion is still ongoing, i've added a -i flag for line~ and vline~ to this PR for now. this activates "instant increment mode".

If you're gonna force people to change their patches anyway, I'd actually rather have a pd-compatibilty instead :) and not add too much complexity in the documentation and adding yet another flag to [line~] seems very unnecessary one way or another. If consistency is the ultimate desired goal, just keep it simpler, change [vline~] to make it like [line~] and have a pd-compatibility flag. It's a simpler solution if you're gonna force people to change the patch, and you can also just adapt and fix your patch with the new behaviour instead of using the compatibility flag anyhow.

> So, bottom line, do whatever, just don't try to create a compatibility flag ;) if you can. I'll actually be ok if it comes down to that. I will be able to live with it, gladly document the change of behaviour and everything you all agree, so I'm off ;)

just confirming I'm not going to be the only one who opposes to this :) 

>  I just can't believe I'm the one now saying "hey folks, don't break this", cause it's usually the other way around :)

yeah, I might be going through an identity crisis. Anyway, don't let me be this person, if I'm the only single voice, ignore me ;)

cheers

ben-wes:
> just keep it simpler

sounds good. just not sure how yet since my impression so far is that `line~` and `vline~` will both require both modes. the instant increment for `line~` according to @millerpuckette's comment above and start value preservation for `vline~` according to other opinions.

anyhow ... the current PR can't be merged anyway - and if the discussion is settled, i'd try to finalize the code according to the results (if i can - it's not unlikely that this will require a new PR by someone else). inverting the flags' functionality should be simple if defaults should stay as they are. and if we decide to close this and forget about it, i'm also fine. ;)

millerpuckette:
I confess I'm a bit overwhelmed by the volume and complexity of the code and of the potential trouble if we go in and change things.  I checked again and my previous comment was wrong about phasor~ - it is indeed post-increment as is line~.  But menawhile, the situation is now at least possible to document correctly with a comment in vline~ as to how it differs from the two others.  So I would suggest leaving the code alone and considering the difference in behavior an unfortunate design flaw.

porres:
> But menawhile, the situation is now at least possible to document correctly with a comment in vline~ as to how it differs from the two others

I can work with that for the time being, but I hope we can improve and fix the flaw somehow :) (it was probably not a good sign that I was the only one who agreed this made sense, hahhaha)

ben-wes:
> So I would suggest leaving the code alone and considering the difference in behavior an unfortunate design flaw

thanks for your comments on this, @millerpuckette! i hoped that it would be easier to adapt the code and also allow to maintain compatibility - but i'll gladly surrender for now, i admit. at least i learnt quite a bit while trying. :)

adding some details to the help file (and possibly also not calling it "deluxe [line~]" in `help-intro.pd`) will be a huge improvement for new users. i (like others that joined the discussion, assumingly) still think that it would be good to have a post-increment mode with `vline~`. but maybe that wouldn't need to be the same perform function ... and i wouldn't want to decide on it being the new default and breaking change.

porres:
Hey, I added many details and examples into the help file of [vline~]. Will revise it a couple more times but you all can check.

A screenshot

Screenshot 2025-03-12 at 14 58 01

check it in https://github.com/pure-data/pure-data/pull/2486/commits/7a2c014a89ee79a46945ba5d2c613508ac10646a

ben-wes:
thanks, @porres !

some typos:
* `Another known issue:` <- should be a colon imho
* `Unlke` -> `Unlike`
* `lenght` -> `length`
* `has a sub-sample accuracy` -> `has sub-sample accuracy`

... for the sub-sample accuracy and interpolation, we could add that line segments must not be overlapping for correct results.

and for completeness sake ... while this was not simple to fix in the code for me, i create an `xline~` abstraction now that fixes it on the patch level (at least, my tests all looked correct with this so far):
* [xline~.pd.zip](https://github.com/user-attachments/files/19219074/xline.pd.zip)

porres:
you could use 'receive pd-dsp-started' to update samplerate~

PR #2543, opened 2025-03-08
handle focus switch for <Modifier-Click> (fixes macOS behavior)

### LINKED ISSUE #2347: Edit/Perform mode switch not possible after Cmd+Click in other Pd window
this has already been an issue with Pd 0.54-1 - so it's not a regression.

tested and reliably reproduced many times now on `macOS sonoma` with `Pd 0.55.0` and `Pd 0.54` (also: Macbook Air M2 - but i assume that's not relevant):
1. create new patch window
2. create another patch window
3. create some random object
4. `Cmd`+`Click` in the other window (this obviously makes more sense if there's anything to interact with like a slider/bang/toggle/nbx - but for reproducing the issue, Cmd+Click on the empty canvas also works)
5. try to edit the previously created random object
6. try to switch edit/perform mode and see how window and cursor seem to be in "opposite" modes (window saying "[edit]" and cursor is regular mouse pointer ... or window without "[edit]" and hand cursor)

some more details:
* when creating a new object in this _failed_ state, it will ignore the mouse position and is created at the top left default (?) position
* if the text of an object was edited during `Cmd`+`Click`, entering text will still work afterwards (while mouse interaction doesn't)

... i didn't report this earlier since i got somewhat used to it, but it's quite annoying (since it breaks interaction flow) - and after another user reported something similar on the forum, i eventually tried to find a way to reproduce it.

ben-wes:
closes #2347

this fixes the macOS issue mentioned above where a `cmd`+`click` in another patch window changes the "interaction focus", but the previously focused window still looks like the one in focus and an actual focus change back to the front window requires a clicks outside and then inside the front window.

not sure if this is the proper way to fix this - but it does solve this rather annoying behavior by forcing the window focus for modifier+click.

currently only tested on macOS. ~~maybe this should be macOS-only code, since it doesn't seem to cause issues on other systems?~~ EDIT: forced focus only happens on macOS now.

### DIFF
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -221,17 +221,17 @@ proc ::pd_bindings::patch_bindings {mytoplevel} {
     bind $tkcanvas <Shift-ButtonPress-1> \
         "pdtk_canvas_mouse %W %x %y %b 1"
     bind $tkcanvas <$::modifier-ButtonPress-1> \
-        "pdtk_canvas_mouse %W %x %y %b 2"
+        "::pd_bindings::handle_modifier_click %W %x %y %b 2"
     bind $tkcanvas <$::modifier-Shift-ButtonPress-1> \
-        "pdtk_canvas_mouse %W %x %y %b 3"
+        "::pd_bindings::handle_modifier_click %W %x %y %b 3"
     bind $tkcanvas <$alt-ButtonPress-1> \
         "pdtk_canvas_mouse %W %x %y %b 4"
     bind $tkcanvas <$alt-Shift-ButtonPress-1> \
         "pdtk_canvas_mouse %W %x %y %b 5"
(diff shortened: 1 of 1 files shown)


umlaeute:
did you test on TclTk-9.0 as well?

ben-wes:
> did you test on TclTk-9.0 as well?

it shows the same issue and is working with this fix.

but i realize now when checking TclTk 8 vs. 9 again, that the cursor for the temporary run mode on interactive objects is switching to the hand when interacting. hmpf. that's a minor regression (although it fixes a worse issue imho). no idea where to fix that yet.

ben-wes:
fixed this now. it still doesn't handle cursors correctly for focus changes - but that's not a regression.

my previous version was also rather flawed since it reset the focus with every `<Modifier-Click>`. also, i added a macOS check for the forced focus - so other systems should not be affected (except for calling `pdtk_canvas_mouse` through the additional event handler).

PR #2552, opened 2025-03-16
add function that returns Pd's current color scheme

myQwil:
Most externals rely on Pd's built-in drawing methods for displaying its box and text, but in the case of a gui external, the colors of its text and borders are set independently from Pd, either by the user or hard-coded.

As someone who uses a slightly modified version of Pd with a dark theme, this creates a problem where new instances of a gui external will be virtually invisible unless the external was built to match the color scheme of a particular version of Pd.

This problem can be solved by adding a function that returns Pd's color scheme at run-time.


umlaeute:
I so wished that we could define our own colors in tk for canvas draw instruction (so rather than hardcoding the values in the C side or using tcl variables, instead us using `-fill blue` we could use `-fill color_selected`

but anyhow, this is practically a duplicate of 
- https://github.com/pure-data/pure-data/pull/2502
- https://github.com/pure-data/pure-data/pull/196

and of course
- https://github.com/pure-data/pure-data/pull/1765


though your solution uses a radically different approach (somewhat understandable, given that #196 has been lingering since 2017...)

myQwil:
These look like good solutions. They seem to be oriented more towards implementing color customizability in Pd, whereas this proposal is only concerned with providing outline colors at run-time, making it a much smaller change.

Maybe this proposal could be changed to reflect one component of one of these bigger proposals.

millerpuckette:
I think this is done by THISGUI->i_foregroundcolor (etc) now.  I agree that it's best to let externs see Pd's color scheme as C variables.

myQwil:
That works for me. Thank you for adding that :)

The only thing I would suggest is, maybe instead of symbols, they would be better suited as integers. That way, they would match the `x_fcol`, `x_bcol`, and `x_lcol` fields in `t_iemgui`.

I'll close this PR, but if you like that suggestion, I'd be happy to make those changes in a separate PR.

millerpuckette:
I think it's better to keep them with the "#" symbols so as not to get them confused with the "999" color convention used throughout the data/drawpolygon/etc. objects.

PR #2569, opened 2025-04-19
feat: Add support to `oscparse` for handling `d` type tag (double)

mitchmindtree:
The [Open Sound Control 1.0 spec](https://opensoundcontrol.stanford.edu/spec-1_0.html) mentions a (non-standard) `d` type tag for handling 64-bit floating point values. This adds support to the `oscparse` object for parsing osc args of this type.

Note that this implementation results in a potentially lossy conversion from `double` to `t_float` in the case that `t_float` is `float`.

This was motivated by some personal software that emits doubles for certain control values. Previously I was receiving these in supercollider without issue, but after experimenting with receiving the same data in puredata I noticed the `d` tag was unsupported, so thought I'd have a go at contributing it! If the non-standard flags are intentionally unsupported, feel free to disregard.

### DIFF
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -294,17 +294,21 @@ typedef struct _oscparse
                     ((((int)(((x)+2)->a_w.w_float)) & 0xff) << 8) | \
                     ((((int)(((x)+3)->a_w.w_float)) & 0xff) << 0)
 
-#define READDOUBLE(x) ({ \
-    uint32_t hi = READINT(x); \
-    uint32_t lo = READINT((x)+4); \
-    union { \
-        double d; \
-        struct { uint32_t lo; uint32_t hi; } i; \
-    } u; \
-    u.i.lo = lo; \
-    u.i.hi = hi; \
-    u.d; \
-})
+static double readdouble(t_atom *x)
+{
+    uint32_t hi = READINT(x);
+    uint32_t lo = READINT((x)+4);
+    union {
+        double d;
+        struct {
+            uint32_t lo;
+            uint32_t hi;
+        } i;
+    } u;
+    u.i.lo = lo;
+    u.i.hi = hi;
+    return u.d;
+}
 
 static t_symbol *grabstring(int argc, t_atom *argv, int *ip, int slash)
 {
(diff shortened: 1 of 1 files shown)


src/x_misc.c
```diff
@@ -294,17 +294,21 @@ typedef struct _oscparse
...
-    uint32_t lo = READINT((x)+4); \
-    union { \
-        double d; \
-        struct { uint32_t lo; uint32_t hi; } i; \
-    } u; \
-    u.i.lo = lo; \
-    u.i.hi = hi; \
-    u.d; \
-})
+static double readdouble(t_atom *x)
```
umlaeute, line 297: that should probably be `inline`  (e.g. using the `PD_INLINE` macro)

umlaeute:
to be honest, i don't think that OSC will officialize anything within the next few decades...[OSC-1.1](https://opensoundcontrol.stanford.edu/files/2009-NIME-OSC-1.1.pdf) has been proposed in 2009, and it is still only available as a PDF-download of a NIME paper. :shrug: )

however, if ever Pd64 is going official, it would be nice to be able to send numbers without precision loss...
(even if for backwards compatibility, the default type for `t_float` should probably stay `f`).

speaking of OSC-1.1, there are a couple of types that are now *required*.
the changes are

| type | explanation | possible Pd atom |
|---|---|---|
| `T` | *true* | `1` |
| `F` | *false* | `0` |
| `N` | *nil* | ??? |
| `I` | *impulse* | `bang`? |
| `t` | *timetag* | ??? |

danomatika:
I would consider the double tag canonical. Lots of environments use it and I would just cast it to float for non-double Pd.

danomatika:
For reference, [openFrameworks supports the following](https://github.com/openframeworks/openFrameworks/blob/master/addons/ofxOsc/src/ofxOscMessage.h):

```
i - int 32
h - int 64
f - float 32
d - float 64
s - string
S - symbol (usually handled like a string)
c - char
m - 4 byte midi message
T - bool true
F - bool false
N - nil/none
I - infinitum
t - 64-bit NTP time tag
b - blob, contains child messages
r - rgba 32
```

...but this is largely because it wraps oscpack which supports the full spec, more or less.

In practice, I'd argue that these are IMO canonical across most environments using OSC:
```
i - int 32
h - int 64
f - float 32
d - float 64
s - string
b - blob, contains child messages
```
String is often used for char and symbol. The boolean types end up being sent as number types. Sending a message without arguments is sufficient for a "bang" aka impulse.

millerpuckette:
OK, let's do it and not worry about OSC ever becoming a proper standard :)

PR #2574, opened 2025-05-15
Add [filtercoeff] abstraction to 'extra'

### LINKED ISSUE #1657: suggestion: add abstraction to calculate biquad coefficients
I wonder if we could have an object like else/biquad. All it does is implement some basic filters from the "Eq-cookbook". This is a compiled object but I also have a vanilla abstraction. Maybe such a vanilla abstraction could be included in the 'extra' folder as an extra object. This would make [biquad~] much more useful for the general public allowing Pd to have many built in 2nd order filters, namely: allpass, lowpass, highpass, bandpass, resonant, bandstop, eq, lowshelf, and highshelf (lowpass and highpass being resonant).

Screen Shot 2022-04-29 at 11 18 09

porres:
This is an abstraction that calculates biquad coefficients for 9 filter types

allpass, lowpass, highpass, bandpass, resonant, bandstop, peaking, lowshelf, and highshelf (lowpass and highpass being resonant).

I think this quite welcome and important for Vanilla, as [biquad~] becomes more useful and works out of the box for many projects.

I am calculating according to the "Cookbook formulae for audio EQ biquad filter coefficients" by Robert Bristow-Johnson.

this closes https://github.com/pure-data/pure-data/issues/1657

ps. I'll update the [biquad~] help file and also include this when/if this gets accepted.


porres:
A screenshot 

Screenshot 2025-05-15 at 17 00 08

porres:
Here's the new version taking a suggestion by @reduzent 

I also made this behave much more like a "classic" vanilla object and took away "bad influences" from my ELSE compiled version of this object. Moslty, we now only have a left hot inlet (they were all hot) and a bang message to output values.  

Screenshot 2025-05-19 at 13 58 34

millerpuckette:
I think this is too specific an object for the generality of the name.  And meanwhile, I've also written one with an additional ability to graph the frequency response... if I ever have time I'll release that, but _not_ as vanilla :)

PR #2596, opened 2025-05-26
Documentation updates

porres:
And here's some more, aiming to tackle all new things being included in 0.56


porres:
it seems the commit here was cherry picked. I will start a new PR later

Issue #2533, opened 2025-02-18
tabwrite~'s reference to graph array isn't updated when arrays are renamed


ben-wes:
testing `develop` build at 0a872a9aa47f2410323e8811c0f6006ee3c1165c on arm64 macOS.

#### to reproduce:
1. activate dsp
2. create graph array named `test`
3. create patch to write into `test`:
```
[noise~]
|
|  [bng]
| /
[tabwrite~ test]
```
4. click `bng` (works as expected)

the unexpected behavior starts here:
1. add another graph array and also name it `test`
2. rename old `test` to `temp`
3. click `bng` again --> no array gets redrawn for me now, but when clicking on the `temp` array, it becomes visible that the samples were actually written into this one instead of `test` (which would be the expected afaict)

PR #2598, opened 2025-05-26
Trigger error message in table DSP object if table has been renamed

Spacechild1:
This reverts parts of https://github.com/pure-data/pure-data/pull/2427 that attempted to fix https://github.com/pure-data/pure-data/issues/2533, specifically the following two commits:

https://github.com/pure-data/pure-data/pull/2427/commits/373df447046d54384bb2c208979ceabcc03b58ef
https://github.com/pure-data/pure-data/pull/2427/commits/03ccff4ca6110afcbce4c40ea773ff3b6e14c31c

First, this solution doesn't seem to work. If I rename a garray with "rename", any table DSP object, such as `[tabreceive~]`, that is bound to the old name just keeps referencing the same array.

Second, there is a severe problem with https://github.com/pure-data/pure-data/pull/2427/commits/373df447046d54384bb2c208979ceabcc03b58ef: by setting `garray_usedindsp` we cannot resize the garray anymore without causing a DSP update! The cool thing about `t_dsparray` (and `dsparray_get_array` in particular) is that we do *not* keep a direct reference to the table data and instead maintain a lose reference through a gpointer. This means the garray can be resized at runtime without DSP updates because `dsparray_get_array` automatically reaquires the array data. This will become more important in the future, when we hopefully can load soundfiles in a background thread. So let's please not go back to `garray_usedindsp`.

---

The proper fix is actually quite simple: we invalidate the array by incrementing `a_valid`, just like `array_resize` does. As a result, all DSP table objects that are currently bound to the array will post an error message and do nothing resp. output zeroes.

Properly fixes https://github.com/pure-data/pure-data/issues/2533

### DIFF
--- a/src/d_array.c
+++ b/src/d_array.c
@@ -63,33 +63,6 @@ static int dsparray_get_array(t_dsparray *d, int *npoints, t_word **vec,
     return 0;
 }
 
-static int dsparray_reset_array(t_dsparray *d) {
-    int size0 = 0, size1 = 0;
-    int phase = d->d_phase;
-    t_word *vec0 = 0, *vec1 = 0;
-    t_garray *a = 0;
-    if (d->d_symbol == &s_)
-        return 0;
-    a = (t_garray *)pd_findbyclass(d->d_symbol, garray_class);
-    if(!a)
-        return 0;
-    if (garray_getfloatwords(a, &size1, &vec1) && dsparray_get_array(d, &size0, &vec0, 0)) {
-        if(vec0 == vec1)
-                /* no change */
-            return 0;
-    }
-
-    gpointer_unset(&d->d_gp);
-#if 0
-        /*
-          JMZ: arrayvec_reset() calls dsparray_get_array() via arrayvec_testvec(),
-          so we skip it here
-        */
-    dsparray_get_array(d, &size1, &vec1, 1);
-#endif
-    return 1;
-}
-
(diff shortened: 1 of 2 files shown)


umlaeute (changes requested):
> Properly fixes https://github.com/pure-data/pure-data/issues/2533

only it doesn't :-/

with this PR, the behaviour is reverted exactly to the one described in the original #2533 report.

Spacechild1:
Aha, I see. https://github.com/pure-data/pure-data/issues/2533 is pretty an edge case, because it involves creating a duplicate array and then renaming the previous one. 

I only tested by renaming an array from "foo" to "bar" with the "rename" method. With this PR, this properly prints error messages like `foo: no such array` in table DSP objects that currently reference "foo". This does *not* happen with your fix.

However, I just noticed that this doesn't work if I rename the array manually in the array properties. This is strange. In fact, I even get a repeated error `consistency check failed: tabwrite_tilde_redraw` for `tabsend~`. Will have a look.

Either way, calling `garray_usedindsp` is not acceptable.

umlaeute (approved):

Spacechild1:
@millerpuckette Can you merge this? This PR will create conflicts (again) with https://github.com/pure-data/pure-data/pull/2595, which I will resolve once this PR merged.

millerpuckette:
ok, done :)

PR #2614, opened 2025-06-07
enable use of the numeric keypad to create boxes

### LINKED ISSUE #2613: digits of numpad aren't bound to object/message/(...) creation shortcuts
The `Put` menu mentions that Object can be created with shortcut `Mod+1`.

However the <kbd>1</kbd> key of the keyboard's numpad (when exists) cannot be used, which I've always  found odd.
The same for the other MOD+number shortcuts.

The fix is straightforward (add the five `<$::modifier-Key-KP_#>` to pd_bindings.tcl), but I don't know if this could hurt?

Ant1r:
fixes #2613

this simply corrects the inconsistency that currently prevents you, for example, from using the <kbd>1</kbd> key of the numeric keypad to create an object via the <kbd>Ctrl</kbd>+<kbd>1</kbd> shortcut.

### DIFF
--- a/tcl/pd_bindings.tcl
+++ b/tcl/pd_bindings.tcl
@@ -77,6 +77,11 @@ proc ::pd_bindings::global_bindings {} {
     bind all <$::modifier-Key-3>        {::pd_menucommands::scheduleAction menu_send_float %W floatatom 0}
     bind all <$::modifier-Key-4>        {::pd_menucommands::scheduleAction menu_send_float %W listbox 0}
     bind all <$::modifier-Key-5>        {::pd_menucommands::scheduleAction menu_send_float %W text 0}
+    bind all <$::modifier-Key-KP_1>     {::pd_menucommands::scheduleAction menu_send_float %W obj 0}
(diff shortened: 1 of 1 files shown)


millerpuckette:
Yes thanks... and a question: I just found myself on a Mac where command-3 wants to make a screen dump (and so you
can't use it to add a number box).  I'm not usually a Mac user... is this something that people just turn off in order to edit patches with Pd?  Or is this a problem for Mac users everywhere?

Ant1r:
> is this something that people just turn off in order to edit patches

I think so. But I'm not a regular MAC user either...
Perhaps it should be mentioned in the manual?

ben-wes:
i can report for macOS 15.5 (Sequoia) here. in this case, a fullscreen screen dump is made via `cmd`+`shift`+`3`, while `cmd`+`3` doesn't seem to be used by the OS itself.

just tested this PR with an external keyboard for the numpad. works perfectly well for me.

Ant1r:
it seems this was merged manually in f6c2e486b135b707ba93e3c8cb6afc37c7423ccb.

PR #2615, opened 2025-06-09
snake~: add more verbs and enhance 'in'

### LINKED ISSUE #1989: get number of channels in mc connections
This is something definitely not in the "that-would-be-nice" category as I see it. Here's a proof of concept, I am designing an output~ like abstraction for multichannel signals. I need to use [clone] just to include a simple [hip~ 3] object for all channels and I need to resize [clone] according to the number of channels and I need to get that information somehow.

I have a simple external for that right now clled [nchs~] (see picture below), it could be part of Vanilla, maybe as new [snake~] function, like [snake~ nchs] or built into [inlet~]. I like the [inlet~] idea cause it could also give us '0' for no signal connected, which also helps a lot (and it's not something I can do as an object, built in or not I guess). 

Screen Shot 2023-06-14 at 12 54 06

ping @millerpuckette

ben-wes:
- extends [snake~ in] to handle and join multichannel inputs
- adds [snake~ sum] for multichannel mixdown (with additional commit for bypass option)
- adds [snake~ pick] for picking specific channels from mc signal (passing the input through if no channel numbers are given)
- adds [snake~ split] for splitting input at given index

test patch: [test-snakes-minimal.pd.zip](https://github.com/user-attachments/files/30055905/test-snakes-minimal.pd.zip)

image

EDIT: removed a bunch of verbs that were part of this PR initially.

### DIFF
--- a/src/d_misc.c
+++ b/src/d_misc.c
@@ -252,6 +252,74 @@ static void *snake_sum_tilde_new(void)
     return (x);
 }
 
+/* ------------------------ snake_split~ ------------------------- */
+
+static t_class *snake_split_tilde_class;
+
+typedef struct _snake_split
+{
+    t_object x_obj;
+    t_sample x_f;
+    int x_index;        /* split index (0-based) */
+} t_snake_split;
+
+static void snake_split_tilde_dsp(t_snake_split *x, t_signal **sp)
+{
+    int i, nchans = sp[0]->s_nchans;
+    int left_chans, right_chans;
+
+        /* calculate output channel counts */
+    if (x->x_index <= 0) {
+        left_chans = 1;     /* single channel of zeros */
+        right_chans = nchans;
+    } else if (x->x_index >= nchans) {
+        left_chans = nchans;
+        right_chans = 1;    /* single channel of zeros */
+    } else {
+        left_chans = x->x_index;
+        right_chans = nchans - x->x_index;
+    }
+
+        /* set up output signals */
+    signal_setmultiout(&sp[1], left_chans);   /* left output */
+    signal_setmultiout(&sp[2], right_chans);  /* right output */
+
+        /* route channels to outputs */
+    if (x->x_index <= 0) {
+            /* left output: zeros, right output: all channels */
+        dsp_add_zero(sp[1]->s_vec, sp[1]->s_length);
+        dsp_add_copy(sp[0]->s_vec, sp[2]->s_vec, nchans * sp[0]->s_length);
+    } else if (x->x_index >= nchans) {
+            /* left output: all channels, right output: zeros */
+        dsp_add_copy(sp[0]->s_vec, sp[1]->s_vec, nchans * sp[0]->s_length);
+        dsp_add_zero(sp[2]->s_vec, sp[2]->s_length);
+    } else {
+            /* normal split */
+        for (i = 0; i < left_chans; i++)
+            dsp_add_copy(sp[0]->s_vec + i * sp[0]->s_length,
+                sp[1]->s_vec + i * sp[1]->s_length, sp[0]->s_length);
+        for (i = 0; i < right_chans; i++)
+            dsp_add_copy(sp[0]->s_vec + (x->x_index + i) * sp[0]->s_length,
+                sp[2]->s_vec + i * sp[2]->s_length, sp[0]->s_length);
+    }
+}
+
+static void snake_split_tilde_index(t_snake_split *x, t_floatarg f)
+{
+    x->x_index = (int)f;
+    canvas_update_dsp();
+}
+
+static void *snake_split_tilde_new(t_floatarg f)
+{
+    t_snake_split *x = (t_snake_split *)pd_new(snake_split_tilde_class);
+    x->x_index = (int)f;
+    inlet_new(&x->x_obj, &x->x_obj.ob_pd, &s_float, gensym("index"));
+    outlet_new(&x->x_obj, &s_signal);
+    outlet_new(&x->x_obj, &s_signal);
+    return (x);
+}
+
 /* ------------------------ snake_pick~ -------------------------- */
 
 static t_class *snake_pick_tilde_class;
@@ -349,6 +417,9 @@ static void *snake_tilde_new(t_symbol *s, int argc, t_atom *argv)
         else if (!strcmp(str, "sum"))
             pd_this->pd_newest =
                 snake_sum_tilde_new();
+        else if (!strcmp(str, "split"))
+            pd_this->pd_newest =
+                snake_split_tilde_new(atom_getfloatarg(1, argc, argv));
         else if (!strcmp(str, "pick"))
             pd_this->pd_newest =
                 snake_pick_tilde_new(s, argc-1, argv+1);
@@ -386,6 +457,16 @@ static void snake_tilde_setup(void)
     class_addmethod(snake_sum_tilde_class, (t_method)snake_sum_tilde_dsp,
         gensym("dsp"), 0);
     class_sethelpsymbol(snake_sum_tilde_class, gensym("snake-tilde"));
+
+    snake_split_tilde_class = class_new(gensym("snake_split~"),
+        (t_newmethod)snake_split_tilde_new, 0, sizeof(t_snake_split),
+            CLASS_MULTICHANNEL, A_DEFFLOAT, 0);
+    CLASS_MAINSIGNALIN(snake_split_tilde_class, t_snake_split, x_f);
+    class_addmethod(snake_split_tilde_class, (t_method)snake_split_tilde_dsp,
+        gensym("dsp"), 0);
+    class_addmethod(snake_split_tilde_class, (t_method)snake_split_tilde_index,
+        gensym("index"), A_FLOAT, 0);
+    class_sethelpsymbol(snake_split_tilde_class, gensym("snake-tilde"));
 
     snake_pick_tilde_class = class_new(gensym("snake_pick~"),
         (t_newmethod)snake_pick_tilde_new, (t_method)snake_pick_tilde_free,
--- a/src/d_misc.c
+++ b/src/d_misc.c
@@ -141,27 +141,43 @@ typedef struct _snake_in
 {
     t_object x_obj;
     t_sample x_f;
-    int x_nchans;
+    int x_nin;
+    t_sample *x_copybuf;
+    int x_copysize;
 } t_snake_in;
+
+static void snake_in_tilde_free(t_snake_in *x)
+{
+    if (x->x_copybuf)
+        freebytes(x->x_copybuf, x->x_copysize * sizeof(t_sample));
+}
 
 static void snake_in_tilde_dsp(t_snake_in *x, t_signal **sp)
 {
-    int i;
-        /* create an n-channel output signal. sp has n+1 elements. */
-    signal_setmultiout(&sp[x->x_nchans], x->x_nchans);
-        /* add n copy operations to the DSP chain, one from each input */
-    for (i = 0; i < x->x_nchans; i++)
-         dsp_add_copy(sp[i]->s_vec,
-            sp[x->x_nchans]->s_vec + i * sp[0]->s_length, sp[0]->s_length);
+    int i, nchans = 0, offset, copysize;
+    for (i = 0; i < x->x_nin; i++)
+        nchans += sp[i]->s_nchans;
+    copysize = nchans * sp[0]->s_length;
+    x->x_copybuf = (t_sample *)resizebytes(x->x_copybuf,
+        x->x_copysize * sizeof(t_sample), copysize * sizeof(t_sample));
+    x->x_copysize = copysize;
+        /* snapshot inputs before allocating output */
+    for (offset = 0, i = 0; i < x->x_nin; offset += sp[i]->s_nchans, i++)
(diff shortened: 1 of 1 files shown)


src/d_misc.c
```diff
@@ -146,13 +146,18 @@ typedef struct _snake_in
...
+    int i, j, outchan = 0;
+        /* count total channels across all (multichannel) inputs */
     for (i = 0; i < x->x_nchans; i++)
-         dsp_add_copy(sp[i]->s_vec,
-            sp[x->x_nchans]->s_vec + i * sp[0]->s_length, sp[0]->s_length);
+        outchan += sp[i]->s_nchans;
+        /* create output signal with total channel count. sp has n+1 elements. */
+    signal_setmultiout(&sp[x->x_nchans], outchan);
+        /* add copy operations to DSP chain, one per channel from each input */
+    for (outchan = 0, i = 0; i < x->x_nchans; i++)
```
Spacechild1: Actually, `x_nchans` is now a misnomer because it only determines the number of inlets, but not the number of output channels. I would suggest something like `x_nin`.

src/d_misc.c
```diff
@@ -205,6 +210,70 @@ static void *snake_out_tilde_new(t_floatarg fnchans)
...
+    t_sample x_f;
+    int x_nchans;
+} t_snake_count;
+
+static void snake_count_tilde_dsp(t_snake_count *x, t_signal **sp)
+{
+        /* store actual input channel count */
+    x->x_nchans = sp[0]->s_nchans;
+        /* output channel count on DSP reset */
+    outlet_float(x->x_obj.ob_outlet, x->x_nchans);
```
Spacechild1: *Never* call `outlet_` methods in a "dsp" method! Instead, you have to defer to the next scheduler tick with a clock.

Actually, I'm not sure this is a good design. I would prefer if the object works analogous to `[samplerate~]` in that you have to explicitly send a bang to get the value. You can then easily emulate your current behavior with `[r pd-dsp-started] -> [snake~ count]`. Since you already have a "bang" method, you only need to get rid of this line.

Spacechild1:
What's up with the changes in `g_canvas.c`? :) Also, feature PRs always target master, not develop.

src/d_misc.c
ben-wes: thanks for your review! true - renamed to `x_nin` with the last force-push.

src/d_misc.c
ben-wes: oops. i'll try to remember this, sorry! and i also think that `samplerate~` is a good reference. thanks! force-pushed the change.

ben-wes:
> feature PRs always target master, not develop.

makes sense. obviously my first feature PR. :) ... changed the base branch.

porres:
so, my [else/sum~] object works like this, you can change its state and tell it not to sum as well. It's useful for me as I found the need to sometimes decide wether I want to sum or not.

umlaeute:
> extends `[snake~ in]` to handle and join multichannel inputs

this probably should be removed from the title and the description, now that it is not implemented.

ben-wes:
> > extends `[snake~ in]` to handle and join multichannel inputs
> 
> this probably should be removed from the title and the description, now that it is not implemented.

maybe i'm missing something - but this should be implemented via fad9a5ee19df3e3cbb7cbe9f3f7e02bd7085a6c0 which is part of this PR

ben-wes:
@umlaeute i considered adding more commits to this PR, but felt like there are still some open decisions. otherwise, these would also be ready for review from my side:

https://github.com/ben-wes/pure-data/compare/feature/more-snakes...ben-wes:pure-data:feature/even-more-snakes?expand=1

... not sure how to proceed. does it make sense to add them here? or create another PR? or just wait? :)

umlaeute:
dunno.

i think multiple PRs just raise the probability of getting ignored :-/

so it's probably best to just add them in a single PR (and cleanup the commits (in case they are entangled) so that it is easy for @millerpuckette to just pick those he considers worthwhile).

umlaeute:
> maybe i'm missing something - but this should be implemented via fad9a5e which is part of this PR

oh. 
i did check the diffs (promise!) but I think I missed it (actually I was pretty sure it wasn't there, but on rechecking it seems it *is* there... sorry for the noise)

ben-wes:
> I think I missed it

no problem at all - that's a very small commit anyway! :)

but longer ones are incoming since i'll do as you mentioned before and add the other commits after cleaning them up a bit. should hopefully be possible to pick them individually then.

ben-wes:
also added a method now for filling up (by wrapping) or clipping a mc input signal.

i changed the name from "fit" to "shape" through a force-push (sorry for the indecision). wondering now whether this might need an option to switch the "fill" method from wrap to zero-fill. can still be added later - let's see if this verb is needed it all. :)

millerpuckette:
I think I want to go over these carefully, and I'm in rehearsals at the moment and would like to put 0.56 test out while I'm stuck with other tasks... can this be put off for 0.57?

ben-wes:
> can this be put off for 0.57?

sure. i didn't expect the whole PR to be accepted for 0.56 anyway! the commits are sorted by my rather personal priorities. these might probably be least controversial:
* `count`
* `sum`
* `split`
* `fromlist`

`in` with mc input support is also something i would gladly use in my patches. but this could also be a new verb (as discussed in #1996).

`pick` might need more discussion concerning design and verb (although it seems suitable to me).

the other commits are just suggestions so far and i would prefer further discussion there:
* `zip` / `unzip`
* `shape`
* bypass support for `sum`

good luck with the rehearsals!

ben-wes:
updated the order of commits so that the first 5 commits might be a simple addition that doesn't require much discussion (extending `in` to mc support and adding `sum`, `pick` and `split`).

`count` is probably not needed anymore now that `siginfo~` reports channel count so i moved that behind those. similarly, i moved the `bypass` message for `sum` behind those commits since i'm not sure whether that's needed.

millerpuckette:
I think the mc input and the sum, pick, and split verbs make sense.  I'm not sold on any of the rest of it.  In particular if I understand it, zip/unzip is a multiplexer which I can't think of a use case for.

ben-wes:
sounds good to me! let me know if i should remove the other commits from the PR or if you'd merge just the relevant ones?

i remember that we had some use case in mind for zip/unzip when discussing it with @porres last year - but didn't miss it recently, i admit. :)

umlaeute:
i guess *zip*/*unzip* and *shape* can be patched together with *pick*

umlaeute:
i totally fail to see a use case for *sum*'s `bypass` message.

i see that @porres wrote
> `[else/sum~]` object works like this, you can change its state and tell it not to sum as well. It's useful for me as I found the need to sometimes decide whether I want to sum or not.

but unfortunately he didn't come up with any examples how it would be useful to toggle between a a multichannel signal and a mono sum signal.

i could probably understand the need for a sum of all signals and a selected signal (which makes the output always mono), but that can be easily achieved with `[snake~ pick]` + `[snake~ sum]`, so ...


(I'm using "mono" for a single-channel signal)

porres:
For instance, I have used this feature in else/sum~ to toggle between an output in multichannels or not, as in [bpbank~]. The 'mc' message or flag outputs all filters as separate channels. Internally I have [clone] outputing multichannels but I have sum~ in the end. I once requested [clone] to be able to set the -d flags from messages, the idea was exactly this, but I solved it with an external. See:

Screenshot 2026-07-14 at 14 28 04

umlaeute:
but what is the use-case? (the use-case cannot just be "I made it work; and here's a help-patch demonstrating that it does work").

and I'm afraid you will have to explain your patch (I don't use ELSE; i guess that `[bpank~]` is a series of bandpass filters (most likely in parallel); i have no idea what `[xselect.mc~]` might do).

porres:
>  i guess that [bpank~] is a series of bandpass filters

It's actually a bank of filters in parallel, not in series, as in cascade. It's kinda like max's fffb~ https://docs.cycling74.com/legacy/max5/refpages/msp-ref/fffb~.html

but instead of multiple outlets, I have the output as either the sum or as a multichannel connection. 

[xselect.mc~] selects a single channel from a multiconnection signal with crossfade.

anyway, the use case is just to be able to choose if an output is either a sum or not

umlaeute:
> anyway, the use case is just to be able to choose if an output is either a sum or not

I'm seriously trying to understand what a use case for this is, and this answer is not helping.

to rephrase: "the use case for being able to choose if an output is either a sum or not, is just to be able to choose if an output is either a sum or not".

so I take it, there is none.

ben-wes:
i was just trying to patch a little bypass workaround based on `r~`/`s~`. but now i'm wondering:

what's the general policy for dsp updates? when i change the receiver to an `s~` with different channel count, pd logs "[...] changed number of channels; restart DSP to fix".

`snake~ pick` currently automatically updates dsp whenever it receives a `channels` message. i assume that this is expected - but maybe it's good to discuss it here now?

ben-wes:
probably not the cleanest bypass - triggers a warning with every receiver switch:

image

umlaeute:
i was going to suggest that it should be easy enough to build something like this with the `[channel(` message of `[send~]` resp `[catch~]`.

i have to admit that i wasn't aware of the big fat warning when changing `[send~]`'s number of channels though.

however, it can actually be easily avoided like this:
channelchange


alternatively, you could build something with `[throw~]`/`[catch~]` which happens to be less verbose (but admittedly is also less elegant for this specific purpose).

---

and of course, the example that @porres gave could be just as easily implemented with this:
image

ben-wes:
> it can actually be easily avoided

ouch ... of course, the warning can be avoided by just switching off dsp before changing the receiver! although in this example, the receiver only changes when dsp is on - but that's obviously easy to solve. thanks!

millerpuckette:
I thik it would be  better design if there were just a separate outlet for the sum - I really don't think it's good style to have a single outlet that changes its behavior drastically as a result of run-time messages.

ben-wes:
> I think it would be better design if there were just a separate outlet for the sum - I really don't think it's good style to have a single outlet that changes its behavior drastically as a result of run-time messages.

the commit that adds the sum verb is not adding that runtime message. it currently seems to me like that message is not needed anyway. a separate outlet for a "bypass" signal feels a bit useless to me though, too, since that can just be fetched from the previous object?

umlaeute:
> I think it would be better design if there were just a separate outlet for the sum - I really don't think it's good style to have a single outlet that changes its behavior drastically as a result of run-time messages.

for which object? `[else/bpank~]`? if so, I agree.

for `[snake~ sum]` a separate outlet makes little sense, as the sole purpose of this object is to *sum*. you can get the unsummed signals just by patching the input of the object to wherever you want.

millerpuckette:
... and since I think "bypass" is a bad idea to start with, perhaps the question already answered itself :)

umlaeute (commented):
while I really like the empty-list message of `[snake~ pick]`, I wonder whether it's a good idea to allow it for `[snake~ fromlist]`.

it's not really obvious that an empty list would do what it currently does.
and conversely, a single-channel with all zeros can be created with `[0(` much simpler than with `[list(`.

i would go for either
- just forbid empty lists
- have `[list(` create an multichannel signal with **0** channels (iirc this is possible; i don't know whether all objects will handle it correctly though; otoh, this object might be a good choice to create such a weird signal)

umlaeute:
@millerpuckette it's probably a good idea to just merge in the undisputed first 4 commits (77d2a23b44d56196179dbc14a3880952a183da61...f3d2a49997d9acc2ab47a319f1ebf69543649339) and leave the rest for further discussion...
(I guess you can only do this in a local clone, and not via the webinterface).

ben-wes:
now that we're in the middle of some discussion, i converted this to a draft to clarify the state. there's still a bad issue found by @umlaeute in https://github.com/pure-data/pure-data/issues/1996#issuecomment-4978188108 ... i assume that occurs because of reused buffers that i don't understand (but i should be able to fix it by copying to an intermediate signal first).

concerning the other features, i also think that those can be another PR then.

so please don't merge yet! :)

ben-wes:
to make it easier, i'll also just update this PR to just the "basic features".

porres:
> for [snake~ sum] a separate outlet makes little sense

I never proposed that and it's not what else/sum~ has. I thought we were talking about a message that controls the feature.

> you can get the unsummed signals just by patching the input of the object to wherever you want.

sure, but the point is to have a way to choose the kind of output via a message, since there is no way to do that unless you get into edit mode and change your patch... or use some crazy dynamic patching.

ben-wes:
updated this PR now to hopefully solve the issue reported in https://github.com/pure-data/pure-data/issues/1996#issuecomment-4978188108. and it only includes the first 5 commits now. still needs proper review though.

another PR will follow for more - but that doesn't seem urgent (although i do like `fromlist`, which is a quite nice option to create an mc signal of dynamic size. this can be achieved with some clone+resize abstraction as well though ...)

> there is no way to do that unless you get into edit mode and change your patch... or use some crazy dynamic patching.

@porres i think there are ways - see https://github.com/pure-data/pure-data/pull/2615#issuecomment-4978144801 :)

porres:
> @porres i think there are ways - see https://github.com/pure-data/pure-data/pull/2615#issuecomment-4978144801 :)

I think I can keep my external then :) I don't have a problem in not having it as obsolete

PR #2616, opened 2025-06-10
Doc updates

porres:
- closes https://github.com/pure-data/pddp/issues/277
- closes https://github.com/pure-data/pddp/issues/279
- closes https://github.com/pure-data/pddp/issues/280
- closes https://github.com/pure-data/pddp/issues/278
- closes https://github.com/pure-data/pddp/issues/283


millerpuckette:
Is it possible to get by without adding that stereo file?  I don;'t think it adds anything and then why don't we have a 13-channel one too?  (Same objection as for 8-bit soundfiles, which are NOT necessary for making cheap-sounding samples :)

porres:
I guess "it is" (and you can just reject that commit alone) - but I think it makes it really really clumsy when giving an example on how to play a stereo file :)

porres:
For reference, this is a screenshot of what's there now, and this is specially useful and good for a meaningful didactic example for [readsf~] and [soundfiler].

Screenshot 2025-06-17 at 13 46 36

the parent patch still has the example that overrides the header and I enjoyed the opportunity to also update the help file of [soundfiler] to show how to load a strereo file into 2 arrays - we were also missing a proper example for that, and an actual stereo file is needed for that.

Screenshot 2025-06-17 at 13 51 04

But again, your call, and tell me if I have to do something if you can't find a way to just reject that single commit

porres:
>  why don't we have a 13-channel one too?

I'd say we needed just one non-mono file, and a that a stereo one  should not only be sufficient, but makes better sense :)

millerpuckette:
yeah, actually the soundfiler example is one you really can't show withut a stereo file for it to chew on (I think for readsf~ showing a 2-channel readsdf~ reading a mono file would be adequate).  So I'll go ahead and include it as proposed here.

millerpuckette:
By the way - I think I should start doing squash-merges of the documentation branch - there's never much need to see the separate commits in the master branch and having so many commits makes it harder to figure out where one or another bug got introduced.  Any reason I shouldn't do that?  (I believe squash-merge makes a single commit out of the merge in place of all the commits in the branch that's merged from.)

Spacechild1:
> By the way - I think I should start doing squash-merges of the documentation branch

Yes, please! I think I have already suggested this in past. Having all these hundreds of tiny commits makes the history almost unreadable.

> there's never much need to see the separate commits in the master branch

That's what I think as well, in particular if it's only documentation changes.

porres:
I see this squashedly merged :) thanks!

PR #2622, opened 2025-06-17
Fix message box height after binbuf conversion

### LINKED ISSUE #2612: multiline msgbox appears broken on creation
steps to reproduce:
1. create a new patch
2. create a msgbox <kkd>Ctrl</kbd>+<kkd>2</kbd>
3. type `; pd dsp 1`
4. hit <kkd>Escape</kbd> to create the object

The result looks like this:
broken messagebox

Forcing a redraw of the window (eg. by switching desktops; or by minimizing and re-raising the window) or of the object (e.g. moving it), fixes the appearance.

This regression was introduced by da56218f697e9fe6b881c71974d652723ddde8f2 (merging https://github.com/pure-data/pure-data/pull/1928).
for whatever reasons, the PR itself (prior to the merge) does not show this behavior.

ben-wes:
invalidate pixel dimensions in `rtext_retext()` to force height recalculation

also in this case: not completely sure if this is a good solution - so needs proper review.

closes #2612

### DIFF
--- a/src/g_rtext.c
+++ b/src/g_rtext.c
@@ -648,6 +648,8 @@ void rtext_retext(t_rtext *x)
     x->x_buf = resizebytes(x->x_buf, x->x_bufsize, x->x_bufsize+1);
     x->x_buf[x->x_bufsize] = 0;
     rtext_findscreenlocation(x);
+        /* force dimension recalculation after text conversion */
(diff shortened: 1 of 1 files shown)


millerpuckette:
once again, merged but github doesn't see that it was....

ben-wes:
again, thanks and noted in the issue. :)

Issue #2623, opened 2025-06-17
FTBFS with recent settable devicename


umlaeute:
after merging https://github.com/pure-data/pure-data/pull/2466, the current master branch fails to build from source, due to losing constness of `jack_client_names`.

(the PR itself builds fine)

PR #2624, opened 2025-06-17
stable development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/2603

---

### Core
- Closes: https://github.com/pure-data/pure-data/issues/2623


millerpuckette:
This was the correct fix... I applied it by hand.

PR #2625, opened 2025-06-17
Fix struct's displace message

### LINKED ISSUE #2619: [struct]'s displace message only outputs 'x' displacement
I guess it should also report 'y'...

ben-wes:
y-value was missing due to argc being off by 1.

closes #2619


millerpuckette:
I believe this is now merged (even though github hasn't marked it as such)

ben-wes:
> I believe this is now merged

it certainly is. thanks! no idea why it's not displayed here. i commented in #2619 that it can be closed.

PR #2631, opened 2025-06-21
Windows: upgrade `pdprototype` with 'wish86'

### LINKED ISSUE #2630: Make Tcl/Tk 8.6 the default on Windows.
On Windows, the default Tcl/Tk version is still 8.5 in the source code, although the official Pd binaries by Miller already ship with Tcl/Tk 8.6. This is pretty annoying because Tcl/Tk 8.5 is pretty much unusable on modern high resolution screens.

To get the same executable as the official release, I currently need to do the following things:
1. manually build Tcl/Tk 8.6
2. run `configure` with `--with-wish="wish86.exe"`
3. execute `./msw/msw-app.sh -t ./msw/tcltk-8.6.13 <version>` after `make`

Note that 3. must be done *every time* I build Pd.

Ideally, I want to run `configure` without any extra options and simply type `make app`.

Can we take the upcoming Pd 0.56 release as a chance to "officially" upgrade to Tcl/Tk 8.6?

AFAICT, this is what we need to do:
- change the default definition of `WISH` in `s_inter.c` from `"wish85.exe"` to `"wish86.exe"`
- update `msw/pdprototype.tgz` so it contains the same Tcl/Tk libs and wish exe as the official Pd binaries

I'm terribly busy till end of this week. Is this something that maybe @Lucarda could take care of?

Lucarda:
This PR:

- upgrades 'pdprotoype' to ship the same tcl/tk as current Millers's win64 builds.
- move the old 32bit `pdfontloader.dll` for wish85 to `msw` folder and gets included on Miller's 32bit builds (untested)

Closes: #2630


Spacechild1:
Works like a charm, thank you so much!

@millerpuckette Can we still include this in Pd 0.56? It's such a big quality-of-life improvement for Windows developers like myself.

millerpuckette:
I think this is fixed in commit a8512585aaeee09d647fb50c6c08af8cc018a0f6
 ... am I missing something?

umlaeute:
nope.
afaict, https://github.com/pure-data/pure-data/commit/a8512585aaeee09d647fb50c6c08af8cc018a0f6 was just removing the "llibwinpthread-1.dll" from the pdprototype tarball.

*this* PR upgrades the precompiled binaries from 8.5.10 to Tcl/Tk-8.6.10

since the official downloads for Windows/amd64 already use Tcl/Tk-8.6.10, it makes sense to use the same version for Windows/i386

Lucarda:
> I think this is fixed in commit a851258
> ... am I missing something?


> since the official downloads for Windows/amd64 already use Tcl/Tk-8.6.10, it makes sense to use the same version for Windows/i386

this PR replaces `Tcl/Tk 8.5.10(i386)` with `Tcl/Tk-8.6.10(amd64)`. until now when we do standard builds we ended up using a `i386` tcl/tk with an `amd64` Pd. 

`i386` builds from @millerpuckette  uses his machine's precompiled `Tcl/tk 8.5.19`

Lucarda:
afaict `pdprototype` is not used on Miller's or IEM's builds. it is used for us to be able to build and test PRs.

millerpuckette:
It gets used in my own build scripts like pd/msw/build-msw.sh - we might need 2 versions, one for Windows XP and another for Windows 50 or whatever they're at now :)

umlaeute:
> afaict pdprototype is not used on Miller's or IEM's builds. it is used for us to be able to build and test PRs.

regarding the IEM's builds: yes, we are using our own pdprototype.
if miller isn't using it either, should we probably just dump the binary blob entirely?

it will make bootstrapping a new build environment a bit more complicated (but i think that the build scripts are easy enough to use; so the issue is mostly, that it takes longer).
and then the buildsystem should take care to use what is there (after the bootstrapping), basically obsoleting @Spacechild1 's steps `#2` and `#3` (in https://github.com/pure-data/pure-data/issues/2630)

i really dislike binary build artifacts in repositories :-)

Lucarda:
> It gets used in my own build scripts like pd/msw/build-msw.sh

i think it's not.

https://github.com/pure-data/pure-data/blob/3bba51379894f3b07987724d0286d210338677a9/msw/build-wxp-32.sh#L34-L37

and your builds do not use Tcl/Tk 8.5.10 from `pdprototype`

millerpuckette:
Oops, looks like the actial file is "msw-app.sh".  But I'd better check and see if the thing actually builds without it...

Lucarda:
> But I'd better check and see if the thing actually builds without it...

you can go without "msw-app.sh" that's when `pdprototype` jumps in and place a workable tcl/tk. (with this PR an amd64 tcl/tk)

the benefits of `pdprototype` is that you get a Windows testable app without all of the hurdle of having to compile or specify any tcl/tk Windows binaries.

Lucarda:
to explain my life as a Windows tester/debugger/dev I do

```
./autogen.sh
./configure
make app

```

then i get a nice folder with the app ready to run.

Spacechild1:
@Lucarda Exactly! That's what we used to do before we upgraded to Tcl/Tk 8.6. The problem is simply that we forgot to also update `prototype.tgz` accordingly.

Lucarda:
@Spacechild1  i had always used this even now. having a i386 tcl/tk did't bother to me as Pd (the C side) is immune to that so i could edit the c sources and just do again

`make app`

and i'm debugging or developing anyway.

Spacechild1:
> having a i386 tcl/tk did't bother to me

For me it's not really about i386 but about Tcl/Tk 8.5 because the scaling is all over the place. I prefer to develop with a GUI that looks the same as the official binaries.

Lucarda:
worth noting that the only thing that needs special care is `pdfonloader.dll`. this one has to match tcl/tk arch and version. we currently only have precompiled versions for:

- wish85 i386

- wish86 amd64

millerpuckette:
I'm going to have to get my WXP machine running again to see if this works for me... and/or, should 'pdprototype' be part of the git repo in the first place?  In case this doesn't work in XP, should we then have two versions, one for XP and the other for more modern windows-versions?

Spacechild1:
> should 'pdprototype' be part of the git repo in the first place?

Probably not. However, it would be great if we could just update it now and then think about better ways of handling this.

millerpuckette:
I just managed to test the version of Tcl/TK in Windows XP and... it doesn't run.  This might merely mean that it needs to be compiled differently -  I see on https://wiki.tcl-lang.org/page/On+What+Platforms+Does+Tcl+8.6+Run that it's supposed to work.  (Perhaps I'm being too stringent about backward compatibility, but I don't have any testing platforms newer than XP at the moment.)

Lucarda:
@millerpuckette  this was expected. are you saying that you develop for Windows on your xp machine?

i thought (based on your scripts) that you cross-compile on your linux machine (and suspect you run it with wine). the cross-compiled i386 version will work on your xp machine as `pdprototype` is not involved.

Lucarda:
also note that the most possible cause of `wish86` "not working" on xp surely is the arch. i changed to `amd64` and xp is most likely `i386`.

millerpuckette:
> also note that the most possible cause of `wish86` "not working" on xp surely is the arch. i changed to `amd64` and xp is most likely `i386`.

Unless I'm completely misunderstanding my own script, it bases the release on the files in pdprototype.  Pd itself is cross-compiled, but wish85.exe is taken directly from pdprototype.  (Also, the name "wish85.exe" is hard-coded in s_inter.c - this is stupid but probably should change to keep up with the new tcl/tk.

umlaeute:
to add a bit more confusion/datapoints/clarity:

a list of Tcl/Tk versions that are shipped with Pd can be found [here](https://zmoelnig.iem.sh/post/tclpd/)

it says:
| pd package | Pd-version | Tcl architecture | Tcl version |
|---|---|---|---|
| pd-0.55-2-i386.msw.zip | 0.55-2 | Windows/i386 | 8.5.19
| pd-0.55-2-i386.windows-installer.exe | 0.55-2 | Windows/i386 | 8.5.19
| pd-0.55-2.msw.zip | 0.55-2 | Windows/amd64 | 8.6.10
| pd-0.55-2.windows-installer.exe | 0.55-2 | Windows/amd64 | 8.6.10


so Pd-0.55-2 for Windows/i386 shipped a **Tcl/Tk-8.5.19** (Windows, i386).

Now the [`pdprototype.tgz` for the 0.55-2 tag](https://github.com/pure-data/pure-data/blob/0.55-2/msw/pdprototype.tgz) (which hasn't changed since 2018; and was only replaced after 0.55-2) contains (according to `[info patchlevel]`) a **Tcl/Tk 8.5.10** (Windows, i386).

So I do not think that miller has been using the msw/pdprototypes.tgz for quite a while (not since Pd-0.49-0).

Lucarda:
> So I do not think that miller has been using the msw/pdprototypes.tgz for quite a while (not since Pd-0.49-0).

yes i had say it in all possible ways since the starting of this thread.

millerpuckette:
Hmm... my script apparently unpacks pdprototype but then must be doing something crazy afterward.  I have no idea what's going on here.  Also no idea why, if the version of wish.exe isn't being used by anyone, why is it problematic for others who are trying to test Pd from source?

Lucarda:
it's not problematic. it's better and simpler (for us to test, debug or whatever) if we use the same tcl/tk that you ship in the amd64 version. it's also better for you if you are not in your machines and need to cross-compile the amd64 app and test (you don't need to compile any tcl/tk stuff) 

but i think i'm done. this PR was requested to me by @Spacechild1 in #2630 . if you think is better not to merge, we close it.

:)

umlaeute:
> it's also better for you if you are not in your machines and need to cross-compile the amd64 app and test 

no not really.
i think it is never better to include such a binary blob.

it might give you a bit of convenience, but really: if you don't want to compile things, you should download a Pd binary package, not the sources.
if otoh, you are prepared to build Pd yourself, then building Tcl/Tk as well is simple enough (and if it is not, then we should make it so; but I think these days it **is** simple).

as we have demonstrated (in this issue), this convenience copy creates actual problems.

so I'd vote for actual exclusion of the package "asap"

Lucarda:
no objections from my side.

Spacechild1:
> Hmm... my script apparently unpacks pdprototype but then must be doing something crazy afterward. 

@millerpuckette Which script? If you are talking about `msw/build-wxp-32.sh`, this doesn't mention `pdprototype` anywhere... The Tcl/Tk version isn't even the same, as @umlaeute showed.
 
 (As a side note: do we still have to provide Windows XP binaries? XP has been released 20 years ago and has been EOL for 11 years now...)

> Also no idea why, if the version of wish.exe isn't being used by anyone, why is it problematic for others who are trying to test Pd from source?

Again, the point of `pdprototype.tgz` is that Windows developers can simply do `make app` and get a working Pd bundle (without manually downloading and building Tcl/Tk), just like we can do on macOS.

The only issue is that the Tk version is not the same as the official (64-bit) Pd binaries. This is annoying because Tcl/Tk 8.5 has different scaling behavior and thus looks very different. That's why I have been asking to simply update the prototype (= this PR).

Spacechild1:
> as we have demonstrated (in this issue), this convenience copy creates actual problems.

@umlaeute Which problems? AFAICT nobody is using the prototype except for us Windows developers.

> if otoh, you are prepared to build Pd yourself, then building Tcl/Tk as well is simple enough 

It's not. You have to call an additional script (`tcltk-dir.sh`) buried in `msw`, which downloads the desired Tcl/Tk version. (How is the user supposed to know which version?) Then you need to manually call `msw/msw-app.sh --tk tcltk-<TKVERSION> <PDVERSION>` *every time*.

I really want to do `make app`, just like on macOS.*)

Also, Tcl/Tk takes several minutes to build on my machine. (Too many `configure` calls and no parallelization.)

And here comes the kicker: on my system, compiling Tcl/Tk 8.6.10 fails with a mysterious linker error:
```
tkWin32Dll.o:tkWin32Dll.c:(.text+0x5d): relocation truncated to fit: R_X86_64_32S against `.text'
```

> (and if it is not, then we should make it so; but I think these days it is simple).

Yes, but can we first update the prototype? It doesn't cost us anything.

---

*) we also have a prototype for macOS (`mac/stuff/wish-shell.tgz`).

umlaeute:
> we also have a prototype for macOS (mac/stuff/wish-shell.tgz).

and it's not used for the official macOS builds either. 

it should go away as well

umlaeute:
> Also, Tcl/Tk takes several minutes to build on my machine

you should only need to call this once. 

and `make app` could just detect the absence of a Tcl/Tk build, same trek the user to call whatever script in `msw/stuff/` or elsewhere.
the same for macOS.

as for the Tcl/Tk version: I do not consider this a problem. 
on Linux, we do not know the Tcl/Tk version used by Pd for about 2 decades. 
on the other systems, we have been using different versions (in the official builds) across Pass and architectures.
I do *not* consider this a problem, but rather a strength. 

if you do not know, which version to use, just pick the latest and greatest stable release.

(I come from a world where people actually cared about backward compatibility. I understand that this concept might be alien to people who grew up with things like JUCE,... ;-))

Spacechild1:
I have to say, this is getting a bit frustrating. All I'm trying to do is improve the UX for windows developers.

> and make app could just detect the absence of a Tcl/Tk build, same trek the user to call whatever script in msw/stuff/ or elsewhere.
> the same for macOS.

Yes, it could, but it currently doesn't. If you want to get rid of the prototypes, by all means make a PR and we can discuss this for Pd 0.57 or 0.58 or whatever.

*But I don't understand why we can't simply update the prototype in the meantime, which would improve things **now**, without any downsides AFAICT.*

> I do not consider this a problem, but rather a strength.

I'm repeating myself, but the issue is that Tcl/Tk 8.5 looks terrible on modern Windows systems with high-resolution displays. Everything is 25% larger (which means some patches don't fit into the screen anymore) and all the fonts and drawings look blurry (I think they are upscaled automatically).

When Windows developers build from source, they should get an app that looks the same as the recent official Pd binaries by default. This is currently not happening and this PR addresses this.

millerpuckette:
I simply don't understand what's going on.  If pdprototype.tgz doesn't have the same version of tcl/tk as I've been shipping with the binary, I don't know where it's coming from.

I've beenunder the imression that updating the prototype would affect my own builds (which I release, but I'm not sure they're considerend the 'official' release in this context.)

So (to spacechild1) - you mentioned that developer builds come out differently from the official build - do you think this PR ould change the official build to agree with developer builds, or would it change developer builds to agree with the official one?  I'm totally lost in this whole discussion I'm afraid.

And am I right that you only care about intel-64 builds and not i386 ones?  (in which case there could be two prototypes I guess)?

millerpuckette:
Aha - I now see that the tcl/tk app is overwritten in build.msw.sh with a private version I have stashed elsewhere.

So I still don't know what is being asked for - should my builds of Pd now include tcl/tk 8.6.x?  Or is everyone simply ignoring my builds in favor of the CI ones?  and which "official" and "developer" builds aren't coherent now and would be made that way by updating pdprototype.tgz?  I'm still lost.

umlaeute:
> I have to say, this is getting a bit frustrating. 

I understand that. 
and I also want to state (again, iirc): I do not at all oppose this PR.

(obviously I strongly believe that all the precompiled blobs shall be removed, as a source code repository is not the place for them; but as a quick fix, there is nothing wrong with replacing the current blob with something better. however, on the long run, it should go away; this will hopefully get rid of the confusion that miller finds himself in right now; and it shall also give our esteemed windows developers a better experience than having to fight outdated binaries)

Spacechild1:
> but as a quick fix, there is nothing wrong with replacing the current blob with something better. however, on the long run, it should go away

Ok, then we seem to agree anyway :) (What did we actually fight over!?)

Spacechild1:
> Aha - I now see that the tcl/tk app is overwritten in build.msw.sh with a private version I have stashed elsewhere.

Exactly!

> So I still don't know what is being asked for -

Just to merge this PR :) Nothing should change for you or the official builds!

> or would it change developer builds to agree with the official one?

This!

> And am I right that you only care about intel-64 builds and not i386 ones?

Yes, I think we only care about x86-64 for now. (If someone really needs/wants to build a x86 or ARM64 Pd version, they can always do that by building Tcl/Tk and then manually calling `msw-app.sh`.)

---

I see how this whole thing can be confusing if you're not familiar with our Windows dev workflow.

`pdprototype.tgz` simply contains a Tcl/Tk version (in binary form) that is used by default when you don't pass the `--tk <version>` argument to `msw/msw-app.sh`.

It is also used by `make app`, which we Windows developers typically use to build a Pd app for testing.

This PR simply updates `pdprototype.tgz` so that it contains the same Tcl/Tk version as the current official builds.

(I *think* the `pdprototype.tgz` is just the `bin` and `lib` content of some old (32-bit) Pd version, without the actual Pd binaries. At some point, you apparently stopped using the prototype yourself, but we developers kept using it through `make app`. When you upgraded Pd to Tcl/Tk 8.6, nobody upgraded the prototype, hence the discrepancy.)

millerpuckette:
OK, I think this is OK.  I finally got my goddam windows-XP VM back up and usable after some time away from it and was able to test this.  (My reason to test on XP is that I've hit snags in the past with surprising dependencies on libraries I had otherwise installed - the XP I'm using is an almost-clean install and therefore can catch problems with missing external dependencies.)

Incidentally, pdprototype contains hundreds of (ascii) tcl stcripts at least some of which are needed to run tcl/tk.  These don't seem to have been changed in this PR (so that it really _should_ change nothing in my own compiles :)

millerpuckette:
And... I agree with Hannes - let's work on getting rid of pdprototype altogether for next time.

PR #2632, opened 2025-06-21
Documentation updates

porres:
Some final ones for 0.56-0

- closes https://github.com/pure-data/pddp/issues/286
- closes https://github.com/pure-data/pddp/issues/282
- closes https://github.com/pure-data/pddp/issues/287
- closes https://github.com/pure-data/pddp/issues/285
- closes https://github.com/pure-data/pddp/issues/284
- closes https://github.com/pure-data/pddp/issues/289
- closes https://github.com/pure-data/pddp/issues/289
- closes https://github.com/pure-data/pddp/issues/291
- closes https://github.com/pure-data/pddp/issues/288


umlaeute:
sidenote: if you use markdown lists, the github will expand bug-numbers to their descriptions.

e.g.
```md
closes https://github.com/pure-data/pddp/issues/286
```

will render as 
```
closes pure-data/pddp#286
``` 

whereas
```md
- closes https://github.com/pure-data/pddp/issues/286
```

will render as 
```
closes "Document empty "set" message for table objects pddp#286"
```

i think the latter is much easier to read.

doc/5.reference/pd-messages.pd
```diff
@@ -361,14 +361,173 @@
...
+#X msg 156 620 colors \$1 \$2 \$3;
+#X msg 140 294 list #000000 #ffffff #0000ff, f 7;
+#X obj 140 264 loadbang;
+#X text 153 360 default;
+#X obj 156 584 list;
+#X obj 156 543 bng 21 250 50 0 empty empty empty 0 -10 0 12 #dfdfdf #000000 #000000;
+#X msg 70 161 colors gray80 purple magenta;
+#X msg 73 197 colors PeachPuff4 misty\ rose chocolate;
+#X msg 56 129 colors gray black red;
+#X text 27 15 You can set colors for background. You can use common color names of RGB values in hexadecimal starting with "#". Basic colors are possible as well as fancy named ones (refer to a tcl/tk manual) and gray scale (gray0 to gray100)., f 65;
```
umlaeute: i think it is not a good idea to promote named colors.

if we will ever switch to a different GUI toolkit (want to get rid of Tcl/Tk for something fancy?), this will add an additional technical debt for no good reason.

i think it is best if people are encouraged to use hexcolors.

feel free to mention that *currently* Tcl/Tk color names can be used (possibly linking to https://www.tcl-lang.org/man/tcl/TkCmd/colors.htm for a list), but that one should be aware that this might not be the most portable thing to do.

(e.g. right now you can already use this to make your colors incompatible across platforms, e.g. by using names like `systemLabelColor`).

i also see zero reason why you should want to document color names with spaces (`misty rose`). what benefit do the users get from this information? (apart from everybody triggering https://github.com/pure-data/pure-data/issues/2637#issuecomment-2996096875)

doc/5.reference/pd-messages.pd
```diff
@@ -361,14 +361,173 @@
...
+#X text 212 96 default values;
+#X text 38 422 Iemguis and externals with set colors are the only ones not affected., f 15;
```
umlaeute: technically correct, but this be made more generic?
e.g. "Many GUI objects (built in, and external) will probably just ignore these settings."

umlaeute (changes requested):

doc/5.reference/pd-messages.pd
umlaeute: ah i see that i was commenting on an old commit, and the new wording is better.

doc/5.reference/pd-messages.pd
umlaeute: ah i see that i commented on an old commit, and colors with spaces are no longer suggested (you now use `MistyRose`).

the remaining points still stand.

porres: see https://github.com/pure-data/pure-data/pull/2632/commits/181efdd5b9d49e5fed27ec29d317b3faab08bb1a

doc/1.manual/resources/3.3.3.png
umlaeute, line 1: i wonder: shouldn't we use some settings using English device names? (esp. the broken rendering of `Kopfhörer` is a bit disturbing)

doc/1.manual/resources/1.2.1.png
umlaeute, line 1: should we use a screenshot from macOS?
Apple rewrites the actual menu, adding new elements ("Send Pd Feedback to Apple") and renaming others ("HTML Manual..." becomes "Pd Help")

doc/1.manual/resources/3.3.3.png
ben-wes: that's on me. i'll try to make a better screenshot.

doc/1.manual/resources/1.2.1.png
ben-wes: sounds better, yes. all those additional elements are a bit distracting. i can currently only contribute macos and windows screenshots though. will replace with windows ...

EDIT: just realized that i had some other menu in mind (and have no idea what it was). so while the changes are not so many (feedback entry, search field instead of "About Pd" and "Pd Help" instead of "HTML Manual..."), i think i'll combine both in one image and explain in the caption. although i'd hope that it's quite self-explanatory.

doc/1.manual/resources/3.3.3.png
ben-wes: created https://github.com/porres/pure-data/pull/93/files ... but if we want it completely in English, someone else will have to provide that. changing system language doesn't help and since it's a work laptop with limited rights, i can't change anything about the drivers. :/

doc/1.manual/resources/1.2.1.png
ben-wes: decided to add a remark in the text and stick with only a new windows screenshot: https://github.com/porres/pure-data/pull/93/files

Issue #2626, opened 2025-06-18
allow all settings to be saved with the array in the patch


porres:
As discusses in https://github.com/pure-data/pure-data/pull/2427

I think that other settings should be saved with the array when saving the patch. That'd be
- edit status
- show/hide array

Spacechild1:
IMO we should (only) save settings that are available via the array properties. So I would only support this together with https://github.com/pure-data/pure-data/issues/2618

porres:
> IMO we should (only) save settings that are available via the array properties. So I would only support this together with https://github.com/pure-data/pure-data/issues/2618

Well, so far, we get to save colors and line width, which aren't in the properties yet. So I wouldn't see a reason not to allow this to also be saved with the patch. On the other hand, I agree it all should be also in the properties and I wouldn't mind to see this all implemented at once.

Spacechild1:
@umlaeute is already working on this: https://github.com/pure-data/pure-data/pull/2635

Issue #2618, opened 2025-06-10
 improve array properties


porres:
we have many settings in arrays that can only be set via messages, including recently added ones (see https://github.com/pure-data/pure-data/pull/2427 )

I think it makes sense that ALL settings should and could be settable via properties. Specially now that arrays will store such settings, so just asking for properties would be sufficient instead of setting up a message into a send object or something (since arrays don't even have inlets). This would really make things more convenient.

cheers

porres:
> I think it makes sense that ALL settings should and could be settable via properties.

namely, they are:
- show/hide array toggle ('vis' message)
- show/hide array name toggle ('visname' message)
- edit toggle ('edit' message)
- line width
- color picking

Spacechild1:
I fully endorse this feature request :)

PR #2635, opened 2025-06-22
garray dialog for new features

umlaeute:
extends the array properties to all the new settable features
- Closes: https://github.com/pure-data/pure-data/issues/2618 
- Closes: https://github.com/pure-data/pure-data/issues/2626

## BUGS
- [ ] "Create in last graph" is currently broken

## TODO
- *Undo*
  - [x] garray creation in new graph
  - [ ] garray creation in last graph
  - [ ] properties during array creation
  - [ ] properties of already created array

### DIFF
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -364,9 +364,15 @@ void canvas_menuarray(t_glist *canvas)
         if (!pd_findbyclass(gensym(arraybuf), garray_class))
             break;
     }
-    pdgui_stub_vnew(&x->gl_pd,
-        "pdtk_array_dialog", x, "siii",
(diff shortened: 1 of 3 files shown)


umlaeute:
hmm. it turns out that proper *undo* for the array properties is more complicated than expected.

the good thing is, that "array undo" is currently (`0.56-0test1`) not fully implemented anyhow (state changes are ignored; only *add* and *delete* are supported), so the regression is small (and implementing the missing parts to not-regress is probably simple enough).

even without *Undo*, my current implementation has substantial changes (more than i would like, for now).

PR #2644, opened 2025-06-23
fix crash when deleting scalars with text fields

### LINKED ISSUE #2588: deleting scalar causes segfault
How to reproduce:

1. Start Pd, load attached test patch
2. Switch to edit mode
3.  Select scalar
4. Hit 'Delete'
5. Move mouse cursor in canvas
6. Segfault

Here is the gdb backtrace:

```
#0  0x00005555556255b7 in pd_checkobject ()
#1  0x00005555555b6a8b in gobj_shouldvis ()
#2  0x00005555555ebae8 in rtext_findhit ()
#3  0x00005555555d0949 in canvas_doclick ()
#4  0x00005555555d2f52 in canvas_motion ()
#5  0x000055555561c434 in binbuf_eval ()
#6  0x000055555562bd26 in socketreceiver_read ()
#7  0x000055555562a8cc in sys_domicrosleep ()
#8  0x000055555562d475 in sys_pollgui ()
#9  0x00005555556273af in sched_idletask ()
#10 0x000055555562778d in m_mainloop ()
#11 0x0000555555634219 in sys_main ()
#12 0x00007ffff7a2a1ca in __libc_start_call_main (main=main@entry=0x55555556fa80 <main>, argc=argc@entry=3, 
    argv=argv@entry=0x7fffffffdbb8) at ../sysdeps/nptl/libc_start_call_main.h:58
#13 0x00007ffff7a2a28b in __libc_start_main_impl (main=0x55555556fa80 <main>, argc=3, argv=0x7fffffffdbb8, 
    init=<optimized out>, fini=<optimized out>, rtld_fini=<optimized out>, stack_end=0x7fffffffdba8)
    at ../csu/libc-start.c:360
#14 0x000055555556fad5 in _start ()
```

This is with current master (cf09b3e3b3800411c79e90b48529814c31852a74). Deleting a scalar consisting only of a [filledpolygon] doesn't crash. A scalar consisting only of a [drawnumber] cannot be deleted. Only a combined scalar exhibits the crash. Probably there is a conflict between [drawnumber] being not deletable and [filledpolygon] being deletable.

Test patch:

```
#N struct combined float x float y float cat;
#N canvas 501 40 517 200 12;
#X obj 236 104 namecanvas thus;
#X obj 26 134 s thus;
#N canvas 1162 521 450 300 combined 0;
#X obj 25 146 struct combined float x float y float cat;
#X obj 25 174 print combined;
#X obj 19 67 filledpolygon 436 436 0 0 0 0 15 100 15 100 0;
#X obj 20 38 drawnumber cat 2 0 877 cat=;
#X connect 0 0 1 0;
#X restore 236 128 pd combined;
#X msg 26 106 scalar combined 20 20 436;
#X scalar combined 20 20 436 \;;
#X connect 3 0 1 0;
```

ben-wes:
add rtext cleanup when deleting scalars with text fields (`drawsymbol`, `drawnumber`) to prevent use-after-free crash.

not completely sure whether there might be a simpler approach without a new `rtext_cleanupforscalar()` (and without traversing all canvases and rtexts)?

- closes #2588

### DIFF
--- a/src/g_rtext.c
+++ b/src/g_rtext.c
@@ -135,6 +135,34 @@ void rtext_free(t_rtext *x)
     freebytes(x, sizeof *x);
 }
 
+static void glist_dortextcleanupforscalar(t_glist *glist, t_scalar *sc)
+{
+    t_rtext *rtext, *next;
+    t_gobj *g;
+        /* clean up rtexts in current canvas */
+    if (glist->gl_editor)
+        for (rtext = glist->gl_editor->e_rtext; rtext; rtext = next)
+        {
+            next = rtext->x_next;
+            if (rtext->x_scalar == sc)
+                rtext_free(rtext);
+        }
+        /* recursively clean up rtexts in child canvases */
+    for (g = glist->gl_list; g; g = g->g_next)
+    {
+        if (g->g_pd == canvas_class)
+            glist_dortextcleanupforscalar((t_glist *)g, sc);
+    }
+}
+
+    /* clean up rtexts that reference a scalar being deleted */
+void rtext_cleanupforscalar(t_scalar *sc)
+{
+    t_glist *glist;
+    for (glist = pd_this->pd_canvaslist; glist; glist = glist->gl_next)
+        glist_dortextcleanupforscalar(glist, sc);
+}
+
 void rtext_setcolor(t_rtext *x, int color)
 {
     char buf[80];
--- a/src/g_canvas.h
+++ b/src/g_canvas.h
@@ -497,6 +497,7 @@ EXTERN void rtext_untype(t_rtext *x);
 EXTERN void rtext_getrect(t_rtext *x, int *x1p, int *y1p, int *x2p, int *y2p);
 EXTERN void rtext_retextforscalar(t_rtext *x, char *buf, int len,
     int xpix, int ypix);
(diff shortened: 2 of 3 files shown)


src/g_rtext.c
```diff
@@ -135,6 +135,21 @@ void rtext_free(t_rtext *x)
...
+    t_rtext *rtext, *next;
+    for (glist = pd_this->pd_canvaslist; glist; glist = glist->gl_next)
+        if (glist->gl_editor)
+            for (rtext = glist->gl_editor->e_rtext; rtext; rtext = next)
+            {
+                next = rtext->x_next;
+                if (rtext->x_scalar == sc)
+                    rtext_free(rtext);
+            }
+}
```
Spacechild1, line 164: This only traverses the top level canvases. What about the children?

Note: if we recursively traverse all canvases, this could be pretty expensive, particularly since we need to do this for every scalar.

ben-wes: thanks! force-pushed an update. but i converted this to a draft now since it doesn't seem like a good solution.

millerpuckette: I think I found a more local way to fix this one (commit ca5f336d7fd686477ad354f8db4395918f46a748)

ben-wes: works for me and certainly looks better. :) thanks!

Issue #2642, opened 2025-06-23
issues regarding setting colors with the new "colors" message to pd


porres:
A reboot of https://github.com/pure-data/pure-data/issues/2637
01) Cords are still drawn as black after changing colors. 
02) GOP and array names aren't created with foreground colors
03) The selection rectangle is always black. It could be the same as the selection color (makes more sense to me).
04) Displacing GOPs sets the outline color back to foreground (as well as the array name), this is an old related issue https://github.com/pure-data/pure-data/issues/2500
05) Iemguis must change their outline and selection color as well.
06) Pd should restrict color names to hex numbers preceded by "#". This is needed by iemguis as well as other forks (which won't need to bother to support fancy tcl/tk color names like "mistyrose"). Pd should also not give tcl/tk errors for invalid colors. We should check and give a normal error instead and this is easier with hex colors.
-------------
07) By default, iemguis shoud take background and foreground color the same as Pd's (reacting to color changes is out of this scope and requires deeper changes).
08) The text editing bar marker is always black, should be the foreground color.
09) The Pd console window does not change colors (we may also have to consider how to deal with the red error messages).
11) Changing colors don't recreate the windows in the current order.

porres:
I took care of the first four items in https://github.com/pure-data/pure-data/pull/2646

I had a hard time dealing with iemguis, but someone more familiar and comfortable with the code can easily do what has to be done 

As for the rest, I don't thinj I can manage. I see that the Pd console window is purely tcl/tk code, so something more complicated would need to be done. I'm guessing the editing bar marker is also on the tcl/tk side.

porres:
https://github.com/pure-data/pure-data/pull/2659 took care of the 5th issue and was merged into https://github.com/pure-data/pure-data/pull/2646

porres:
- I took care of the first seven items in https://github.com/pure-data/pure-data/pull/2646

porres:
issues 1-6 are taken care of in https://github.com/pure-data/pure-data/pull/2668

porres:
Most of these have been fixed, so I'm rebooting it.

Issue #2637, opened 2025-06-22
issues when setting colors with the new "colors" message to pd


porres:
I'm having two issues when changing colors (edit, I'm finding much more new ones in the comments below). 

- First and more problematic, the cords are still drawn as black, even if you changed it. One can select the cord and unselect it so the new color applies. This completely ruins patching when you can't see the cords.

- The second is just really annoying. When you change the color within a patch, other patches that were opened are brought to front on top of it. The current order should be respected and reconstructed I think.

porres:
I see yet a new issue. The color of the bounds rectangle of GOP do not change their color. You have to select and unselect them for the colors to apply. (This includes arrays)

porres:
By the way, we have another issue related to this that is still happening of course, see https://github.com/pure-data/pure-data/issues/2500

porres:
Ok, another one. I see we can set any colors with names that "tck/tk" support, see https://www.tcl-lang.org/man/tcl/TkCmd/colors.htm

I was able to use a symbol with escaped spacebars for the background color, but not the other ones (I get an tcl/tk error as they don't understand escaped characters so it seems).

porres:
By the way, I see that the Pd window doesn't change its color, and to me it's clear that it should also!

porres:
When selecting a portion of yout patch, the selection rectangle is always black, but this is bad when you change the colors as you may not see it. The color must be the same as the seted foreground then!

porres:
IEMGUIS must be updates as well. 

I know they have a background color and I don't mean that. Although I have a related issue/request - see https://github.com/pure-data/pure-data/issues/2639

What I mean here is the rectangle outline color, which should be the same as other boxes. Also, the selection color must conform to the new setting. Right now they are hardcoded to black (outline) and blue (selection).

porres:
> When selecting a portion of yout patch, the selection rectangle is always black, but this is bad when you change the colors as you may not see it. The color must be the same as the seted foreground then!

I think it makes sense that the "selecting rectangle" should have the "select color" (blue by default).

umlaeute:
> I was able to use a symbol with escaped spacebars for the background color, but not the other ones (I get an tcl/tk error as they don't understand escaped characters so it seems).

yes. but why would you want to use a color with a space? all Tcl/Tk colours (e,g, `cornflower blue`) that have a space also have an alternative name without a space (e.g. `CornflowerBlue`)-

[...]

PR #2646, opened 2025-06-24
fix color issues

porres:
In this PR I will try to address some of the issues in https://github.com/pure-data/pure-data/issues/2642

This PR fixes
- [x] Cords were still drawn as black after changing colors, which was bad because it wouldn't be visible in darker backgrounds.
- [x] The selection rectangle is now the same color as the selection color. This was hardcoded to black, which equally bad.
- [x] Create GOP and array names with foreground color. This was also problematically drawn as black.
- [X] Fix displacing GOPs with the select color. It would change to black when doing so.
- [x] Add colors to iemguis' outline and select colors (via @myQwil's PR). They were still hardcoded to black and blue. The GOP color is also used now for things like the active number in the number box, which was hardcoded to red.
- [x] make iemguis load Pd's colors as default. The background color and foreground color are taken from Pd's run time colors now.
- [x] Restric Pd's color to hex numbers preceded by "#" (which is what iemguis understand and other forks don't have to deal with fancy tcl/tk color names  like "mistyrose")
- [x] Check if given colors are valid before applying (this prevents tcl/tk errors). Closes https://github.com/pure-data/pure-data/issues/2663


porres:
Ok, I got stuck quickly... made less progress than I hoped, and wasted a lot of time dealing with iemguis, which I thought it was the easiest to handle and I just can't deal with its complexity.

Can someone take over please? :)

src/g_graph.c
```diff
@@ -963,6 +963,11 @@ static void graph_displace(t_gobj *z, t_glist *glist, int dx, int dy)
         if (glist_isvisible(glist)) {
             glist_redraw(x);
             canvas_fixlinesfor(glist, &x->gl_obj);
+            char tag[80];
+            sprintf(tag, "graph%lx", (t_int)z);
+            pdgui_vmess(0, "crs rr",
+                glist_getcanvas(glist), "itemconfigure", tag,
+                "-fill", THISGUI->i_selectcolor->s_name);
```
umlaeute, line 973: i'm not entirely sure whether we should use the `r` specifier (instead of `s`) for the `THISGUI->i_selectcolor->s_name`).

- using `s`, any bad characters in the string will be escaped (so you could use a color like `alice blue`)
- using `r`, the string value is trusted and not escaped. this allows you to use `$sometclvar` (and the GUI will expand this value¹). otoh, if you use a color name with a space (`alice blue`) you will get an error.

this is related to the https://github.com/pure-data/pure-data/issues/2637#issuecomment-2996096875

i'm not sure about the correct way (which i take as an opportunity to say that i think that the current design of the `colors` message is not very good :-/)

i *am* pretty sure that the escaping of colours should be consistent.

this is not a criticism of this PR. just pointing out the issues.

¹ iff there *is* a `sometclvar` defined in the GUI; otherwise you will get an error

porres: oh yeah, this is a mistake, it was supposed to be "s", thanks for catching that

porres: While we're here, let me ask. how come the displace requires a complete redraw? Seems inefficient! I guess there's a good reason not to just "move" the GUI with a tag like others, but I'd like to know which...

porres: and oops, yeah, I changed to 's'

porres:
@umlaeute do you think these could be incorporated into develop? or is it "too much"?

porres:
Not sure why this is failing to check... but the same is true for the current master. I believe nothing is wrong with my latest commit. I tested and it runs fine

myQwil:
Check the output. For example, makefile_gnu is saying:
```
g_graph.c:473:68: error: expected ‘)’ before ‘x’
  473 |         if ((x->gl_pixheight = atom_getfloatarg(3, argc, argv) < 1)
      |            ~                                                       ^
      |                                                                    )
  474 |             x->gl_pixheight = 1;
      |             ~                                                       
g_graph.c:475:5: error: expected expression before ‘}’ token
  475 |     }
      |     ^
```

porres:
> Check the output. For example, makefile_gnu is saying:

I did, and that makes no sense to me... as 1) the doc updates also have that fail (and I never touched this code) and 2) I can't even find this code snippet for real, seems like a hallucination

porres:
and of course 3) I built without such an error :)

myQwil:
Ah ok, my bad. And it's as you say, the same error is in master.

porres:
> Ah ok, my bad. And it's as you say, the same error is in master.

that would be number 4 :)

src/g_all_guis.c
```diff
@@ -1051,10 +1051,11 @@ t_iemgui *iemgui_new(t_class*cls)
...
     iem_inttosymargs(&x->x_isa, 0);
     iem_inttofstyle(&x->x_fsf, 0);
-
-    x->x_bcol = 0xFCFCFC;
-    x->x_fcol = 0x00;
-    x->x_lcol = 0x00;
+    t_atom colors[3];
+    SETSYMBOL(colors+0, THISGUI->i_backgroundcolor);
+    SETSYMBOL(colors+1, THISGUI->i_foregroundcolor);
+    SETSYMBOL(colors+2, THISGUI->i_backgroundcolor);
```
ben-wes: while this solves the expected way to display iemguis for users with custom colors, it will also create a mixture of default colored patches with this update (`#FCFCFC #000000 #000000` vs `#FFFFFF #000000 #000000`).

not adapting the display of iemguis obviously discourages users to change colors without an update like this - but maybe #2648 should be further discussed before making such changes?

src/g_canvas.c
```diff
@@ -2323,6 +2323,22 @@ static void glist_dorevis(t_glist *glist)
...
+        return;
+    }
+    if (sel->s_name[0] != '#'){
+        post("selection color name \"%s\" not supported", fg->s_name);
+        return;
+    }
+    if (*gop->s_name && gop->s_name[0] != '#'){
+        post("got  color name \"%s\" not supported", fg->s_name);
+        return;
+    }
```
ben-wes: aren't these actually error type messages? and the text sounds a bit like a specific name is not supported, while actually *no* name is supported? so maybe users should get a hint to use hex colors in that case?

also, more validation might actually be useful (but could be an issue report on its own): using invalid hex colors like `#0000` currently causes tcl errors.

* EDIT: created #2663

millerpuckette:
And... there's another show-stopper bug I have to deal with... so there will be time for me to look at this more carefully now.

src/g_canvas.c
umlaeute: it's probably better to actually *parse* the colorsymbol and return a normalized version.

e.g. like this (the parser is the naive parser used by the iemgui)

```C
static t_symbol* normalize_color(t_symbol*s) {
        if ('#' == s->s_name[0])
        {
            char colname[MAXPDSTRING];
            int col = (int)strtol(s->s_name+1, 0, 16);
            pd_snprintf(colname, MAXPDSTRING-1, "#%06x", col & 0xFFFFFF);
            return gensym(colname);
        }
        pd_error(0, "invalid color '%s'", s->s_name);
        return 0;
}

void glob_colors(void *dummy, t_symbol *fg, t_symbol *bg, t_symbol *sel, t_symbol *gop)
{
    fg = normalize_color(fg);
    bg = normalize_color(bg);
    sel = normalize_color(sel);
    gop = normalize_color(gop);
    if(!(fg && bg && sel && gop)) return;
    // ...
```

src/g_canvas.c
```diff
@@ -2323,6 +2323,22 @@ static void glist_dorevis(t_glist *glist)
 void glob_colors(void *dummy, t_symbol *fg, t_symbol *bg, t_symbol *sel,
     t_symbol *gop)
 {
+    if (fg->s_name[0] != '#'){
+        post("foreground color name \"%s\" not supported", fg->s_name);
+        return;
+    }
+    if (bg->s_name[0] != '#'){
+        post("background  color name \"%s\" not supported", fg->s_name);
```
umlaeute: this (like the remaining errors)  displays the wrong color name.

ben-wes:
i tested this a bit and there are still a few issues (colors of these objects were not changed after creation, pd was started with `-font 10 -send "pd colors #b4b8ce #24272d #66aaff #77aa22"`):

* label colors (including vu scale) don't adapt to custom palette (therefore all invisible)
* `toggle` border color and label seem to be some random color on creation and after selection (not on redraw though - for example when zooming in/out)
* border of vu meter is not correctly displayed on redraw through zoom in/out or on creation (stays black)

default colors (all labels are set to "test"):

image

after changing label and foreground color:

image

myQwil:
@ben-wes I think that's me forgetting to change a couple of `rk` format specifiers to `rs`. Should be an easy enough fix

myQwil:
This should fix the toggle and vumeter outlines
```diff
diff --git a/src/g_toggle.c b/src/g_toggle.c
index 23b58a6f..30b4b673 100644
--- a/src/g_toggle.c
+++ b/src/g_toggle.c
@@ -114,9 +114,9 @@ void toggle_draw_select(t_toggle* x, t_glist* glist)
         sprintf(lcol_buf + 1, "%06x", x->x_gui.x_lcol);
 
     sprintf(tag, "%pBASE", x);
-    pdgui_vmess(0, "crs rk", canvas, "itemconfigure", tag, "-outline", col);
+    pdgui_vmess(0, "crs rs", canvas, "itemconfigure", tag, "-outline", col);
     sprintf(tag, "%pLABEL", x);
-    pdgui_vmess(0, "crs rk", canvas, "itemconfigure", tag, "-fill", lcol);
+    pdgui_vmess(0, "crs rs", canvas, "itemconfigure", tag, "-fill", lcol);
 }
 
 void toggle_draw_update(t_toggle *x, t_glist *glist)
diff --git a/src/g_vumeter.c b/src/g_vumeter.c
index 0b101ca7..644e6245 100644
--- a/src/g_vumeter.c
+++ b/src/g_vumeter.c
@@ -172,9 +172,9 @@ static void vu_draw_config(t_vu* x, t_glist* glist)
     pdgui_vmess(0, "crs iiii", canvas, "coords", tag,
         xpos - hmargin, ypos - vmargin,
         xpos+x->x_gui.x_w + hmargin, ypos+x->x_gui.x_h + vmargin);
-    pdgui_vmess(0, "crs ri rk", canvas, "itemconfigure", tag,
-        "-width", zoom,
-        "-fill", x->x_gui.x_bcol);
+    pdgui_vmess(0, "crs ri rk rs", canvas, "itemconfigure", tag,
+        "-width", zoom, "-fill", x->x_gui.x_bcol,
+        "-outline", THISGUI->i_foregroundcolor->s_name);
 
     sprintf(tag, "%pSCALE", x);
     if(x->x_gui.x_fsf.x_selected)
```

ben-wes:
> This should fix the toggle outline

@myQwil thanks!

I think we'll also need the adapted mechanism for color validation and normalization as discussed above ...

src/g_all_guis.c
porres: I think this is unrelated to that discussion

porres:
> And... there's another show-stopper bug I have to deal with... so there will be time for me to look at this more carefully now.

@millerpuckette While there are some things still being debated, there are some "no-brainers" in here and @danomatika suggests we split this and move the "hot" iemguis topics to a separate PR. 

For reference, the no-brainer ones would be all up to https://github.com/pure-data/pure-data/pull/2646/commits/7cb0ad2658fc805c9fb7139bb153f625d09b5830 and then taking https://github.com/pure-data/pure-data/pull/2646/commits/f6aeb4327d30d70e7634b9f721d611a0c1f3e7a8 as well.

This would leave us testing and dealing with iemguis separately and I guess I don't think it's the end of the world that iemguis are completely unaware of Pd's color for now (I mean for their default background/foreground/label colors).

porres:
- I'm rebooting this to a much simpler and easier to manage PR, see https://github.com/pure-data/pure-data/pull/2668

Issue #2561, opened 2025-03-31
Using scrolling to adjust sliders and number boxes


sirewix:
Is it possible to use mousewhell or touchpad scrolling events to adjust sliders (and maybe other things like number boxes)? This would be extremely convenient and I'm surprised that it's not a feature already. Maybe I'm missing something, but googling and searching this repo I did not found any mentions or proposals of such functionality.

Spacechild1:
Unfortunately, it's currently not possible. I agree that this would be a great feature!

I don't know if Tk supports touch events, but mousewheel events should definitely be possible.

porres:
Look, I don't know how ceammc do it, but that is possible on their GUI objects.

@Spacechild1 and @sirewix  try installing them in Vanilla and test objects like ui.slider and ui.knob - I can scroll with my trackpad on my macbook air

ben-wes:
i made a PR where you can test scroll wheel and touchpad behavior. i kept it as a draft for now since i'm not sure whether it's the right approach to change the bindings this way. but it feels quite intuitive (short demo video is in the PR description):
* https://github.com/pure-data/pure-data/pull/2651

(not yet tested on Linux and Windows)

porres:
And I just added "wheel scrolling" to [else/mouse] :) 

for reference:
https://github.com/porres/pd-else/commit/5330870e5083a5b8e430c730ac0a855890886462

I'm using it successfully in an endless rotary knob when it is active...

<img width="285" alt="Image" src="https://github.com/user-attachments/assets/1a1ba423-c36f-46df-91fe-e2ff8e779ffb" />

I will try and see how to add this internally to knob

PR #2651, opened 2025-07-01
scroll wheel support for sliders, numbox, float and list boxes

ben-wes:
adds scrollwheel and touchpad interaction for sliders, numbox, float and list boxes.
currently only tested on macOS with touchpad and some logitech mouse with TclTk9.0.1 and TclTk8.6.16.

tested features:
* proportional step sizes depending on ranges of sliders and float/list boxes (if range is limited) 
* adapts to logarithmic scale
* uses vertical direction (also for horizontal sliders - not most intuitive with touchpad, but makes sense with mouse wheel)
* interaction only possible in run mode (otherwise scrolls canvas) 

* closes https://github.com/pure-data/pure-data/issues/2561

demo showing all supported objects with TclTk9:

https://github.com/user-attachments/assets/8d066ceb-aa25-4969-b3d9-a253b848569d

known issues:
* might be nice to support horizontal touchpad interaction for hsl.
* if you interact with a float in a list gatom and its width changes, you might accidentally interact with another float since there's no interaction state (like for dragging), but only single events. i wouldn't know how to change this ...
* similarly, when you're in run mode and scroll the canvas, the wheel event will be caught as soon as there's an object behind the mouse pointer. unless we manage some state here, this is expected as well.

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -3082,7 +3082,7 @@ void canvas_mouseup(t_canvas *x,
 
     /* helper function to try wheel event on a specific object with coordinate transform */
 static int canvas_trywheelobj(t_gobj *y, t_canvas *x, int xpix, int ypix,
-    t_float dx, t_float dy, t_symbol *axis, int mod)
+    t_float dx, t_float dy, int is_touchpad, int mod)
 {
     int x1, y1, x2, y2;
     if (canvas_hitbox(x, y, xpix, ypix, &x1, &y1, &x2, &y2, 0)
@@ -3090,8 +3090,8 @@ static int canvas_trywheelobj(t_gobj *y, t_canvas *x, int xpix, int ypix,
     {
         t_float objx = xpix - x1, objy = ypix - y1;
         t_float deltax = dx, deltay = dy;
-            /* if not xy (touchpad), normalize to unit steps */
-        if (axis != gensym("xy"))
+            /* if not touchpad, normalize to unit steps */
+        if (!is_touchpad)
         {
             deltax = (dx > 0) ? 1.0f : (dx < 0) ? -1.0f : 0.0f;
             deltay = (dy > 0) ? 1.0f : (dy < 0) ? -1.0f : 0.0f;
@@ -3105,7 +3105,7 @@ static int canvas_trywheelobj(t_gobj *y, t_canvas *x, int xpix, int ypix,
 
     /* try wheel event on any object that can handle it, including GOP subpatches */
 static int canvas_trywheel(t_canvas *x, int xpix, int ypix,
-    t_float dx, t_float dy, t_symbol *axis, int mod)
+    t_float dx, t_float dy, int is_touchpad, int mod)
 {
     t_gobj *y;
     for (y = x->gl_list; y; y = y->g_next)
@@ -3120,13 +3120,13 @@ static int canvas_trywheel(t_canvas *x, int xpix, int ypix,
                 if (canvas_hitbox(x, y, xpix, ypix, &x1, &y1, &x2, &y2, 0))
                 {
                         /* delegate to GOP subpatch */
-                    if (canvas_trywheel(gl, xpix, ypix, dx, dy, axis, mod))
+                    if (canvas_trywheel(gl, xpix, ypix, dx, dy, is_touchpad, mod))
                         return 1;
                 }
             }
         }
             /* then check regular objects */
-        if (canvas_trywheelobj(y, x, xpix, ypix, dx, dy, axis, mod))
+        if (canvas_trywheelobj(y, x, xpix, ypix, dx, dy, is_touchpad, mod))
             return 1;
     }
     return 0;
@@ -3137,12 +3137,12 @@ void canvas_wheel(t_canvas *x, t_symbol *s, int argc, t_atom *argv)
     t_float dx = atom_getfloat(argv), dy = atom_getfloat(argv+1);
     int mod = atom_getint(argv+2);
     int xpix = atom_getint(argv+3), ypix = atom_getint(argv+4);
-    t_symbol *axis = atom_getsymbol(argv+5);
+    int is_touchpad = atom_getint(argv+5);
     if (!x->gl_editor) return;
         /* run mode or edit mode with cmd/ctrl - check for objects */
     if (!x->gl_edit || (mod & 2))
     {
-        if (canvas_trywheel(x, xpix, ypix, dx, dy, axis, mod))
+        if (canvas_trywheel(x, xpix, ypix, dx, dy, is_touchpad, mod))
             return;  /* event consumed by object */
     }
         /* fall back to canvas scrolling: edit mode, or no object found.
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -447,21 +447,23 @@ proc ::pdtk_canvas::handle_wheel {tkcanvas x y axis amount {modifier 0}} {
     set canvas_y [$tkcanvas canvasy $y]
     set deltaX 0
     set deltaY 0
+    set is_touchpad 0
     switch $axis {
-        xy {lassign [tk::PreciseScrollDeltas $amount] deltaX deltaY}
+        xy {
+            set is_touchpad 1
+            lassign [tk::PreciseScrollDeltas $amount] deltaX deltaY
+        }
         x {set deltaX $amount}
         y {set deltaY $amount}
     }
     # apply platform-specific scroll direction for object manipulation
-    set obj_deltaX $deltaX
-    set obj_deltaY $deltaY
     if {$::windowingsystem ne "aqua"} {
-        set obj_deltaX [expr {-$deltaX}]
-        set obj_deltaY [expr {-$deltaY}]
+        set deltaX [expr {-$deltaX}]
+        set deltaY [expr {-$deltaY}]
     }
     # send wheel event to core to check if we hit any objects
     # and send the canvas scroll event back if we didn't
-    pdsend "$mytoplevel wheel $obj_deltaX $obj_deltaY $modifier $canvas_x $canvas_y $axis"
+    pdsend "$mytoplevel wheel $deltaX $deltaY $modifier $canvas_x $canvas_y $is_touchpad"
 }
 
 #------------------------------------------------------------------------------#


Spacechild1:
Awesome! Long overdue!

> similarly, when you're in run mode and scroll the canvas, the wheel event will be caught as soon as there's an object behind the mouse pointer. unless we manage some state here, this is expected as well.

This is a *big* problem. Here's a compromise: scrolling individual widgets always requires a keyboard modifier. This means we'd need to press, say, Cmd/Ctrl for "normal" scrolling. "fine" scrolling already uses "Shift", so there is no problem.

> proportional step sizes depending on ranges of sliders and float/list boxes (if range is limited)

Personally, I'm not a too big fan of this. IMO it would be better if we could explicitly set the step size (large and small step).

Side note: In the future we might also support stepping values with the keyboard (which I also think is long overdue). In that case, the scroll wheel and arrow keys are just two ways of achieving the same thing.

Also, is there a particular reason why you're doing this for sliders and gatoms, but not for IEM numberboxes? They are all just different representations of the same thing. I can somehow see why it makes sense for sliders (because of the linear graphical representation) but not so much for gatoms.

That being said, I can imagine the proportional step size being a useful default mechanism in case the user did not specify the step size. I just don't think that `(x->a_draghi - x->a_draglo) / 127.0` is good formula. What about `<next power of 10 of the high limit> * 0.1`:

| Range | Next power of 10 | (Large) step size |
|---|---|---|
| 0 - 0.05 | 0.1 | 0.01 |
| 0 - 1 | 1 | 0.1 |
| 1 - 8 | 10 | 1 |
| 1 - 80 | 100 | 10 |

Another side note: Qt has a really nice option that is called "adaptive decimal step". The idea is that the step size depends on the number of decimals of the current value. Examples: 1 -> 1.1, 10 -> 11, 100 -> 110, etc.  See https://doc.qt.io/qt-6/qdoublespinbox.html#setStepType. Just something to keep in our minds as a (future) option.

> if you interact with a float in a list gatom and its width changes, you might accidentally interact with another float since there's no interaction state (like for dragging), 

TIL that you can drag individual items in a list atom! 😮 I would have never expected this.

Spacechild1:
Hi, you've replied to a deleted message ;-) I deleted my message because I first wanted to actually try you PR. Turns out it immediately crashes Pd when I move the mouse wheel, even on an empty canvas. I'm currently investigating.

I've saved my original answer and wanted to repost it once I can actually get it to run here on Windows. Maybe it's better if you (temporarily) delete your answer as well, because it's rather confusing without the original post :)

src/g_editor.c
```diff
@@ -57,6 +57,8 @@ static t_glist *glist_finddirty(t_glist *x);
 static void canvas_zoom(t_canvas *x, t_floatarg zoom);
 static void canvas_displaceselection(t_canvas *x, int dx, int dy);
 void canvas_setgraph(t_glist *x, int flag, int nogoprect);
+void canvas_wheel(t_canvas *x, t_floatarg fxpos, t_floatarg fypos,
+    t_symbol *axis, t_floatarg famount, t_floatarg fmodifier);
```
Spacechild1: Unnecessary forward declaration.

src/g_editor.c
```diff
@@ -3080,6 +3082,45 @@ void canvas_mouseup(t_canvas *x,
     x->gl_editor->e_onmotion = MA_NONE;
 }
 
+void canvas_wheel(t_canvas *x, t_floatarg fxpos, t_floatarg fypos,
+    t_symbol *axis, t_floatarg famount, t_floatarg fmodifier)
```
Spacechild1: That's a tricky one! When you use fixed arguments (not `A_GIMME`), you must first list *all* symbol/gpointer arguments (in their respective order) and then all float arguments (in their respective order). This is because of the way `pd_typedmess` works. (The idea is to keep the number of possible argument combinations within reasonable limits.)

The current code actually crashes here on Windows 10. I'm surprised it does not crash for you. (I have an explanation, but it's too technical and worth getting into.)

This means you must change the signature as follows:
```
void canvas_wheel(t_canvas *x, t_symbol *axis, t_floatarg fxpos, t_floatarg fypos,
    t_floatarg famount, t_floatarg fmodifier)
```

NOTE: you can keep the actual method arguments as they are!

I would also suggest to add a comment to inform future readers.

(Personally, I tend to use `A_GIMME` in such situations, even if the number of arguments is known ahead, simply because I want to avoid this weirdness.)

Spacechild1 (changes requested):

Spacechild1:
Awesome! Long overdue!

Some things I've noticed:

1. the scroll direction should likely be configurable. It also should be consistent between canvas scrolling and number scrolling. For me, the natural scrolling direction for increasing a value would *upwards* whereas you have currently hardcoded it the other way. This is even more confusing because the canvas scrolling still works as expected (wheel up -> scroll up). Are there different platform defaults/expectations on Windows, Linux and macOS?

2. Similarly, I would expect to scroll *up* to move a slider to the right. This is how volume and transport sliders work on all my media apps here on Windows. Either way, horizontal scrolling should follow the logic of vertical scrolling (up -> increase VS up -> decrease).

3. I would use the small step by default and use a modifier for a large step. Also, 0.01 seems to be a rather small (default) multiplier.

> similarly, when you're in run mode and scroll the canvas, the wheel event will be caught as soon as there's an object behind the mouse pointer. unless we manage some state here, this is expected as well.

This is a *big* problem and it would make scrolling almost unusable for me.

Solution Nr. 1: scrolling individual widgets always requires a keyboard modifier. This means we'd need to press, say, Cmd/Ctrl for "normal" scrolling. "fine" scrolling already uses "Shift", which is already used for horizontal canvas scrolling, so this would have to be "Cmd+Shift".

Solutions Nr. 2: only scroll if the element has the focus. Disadvantage: you first have to click on the object before you can scroll it. Also, objects like sliders don't have a notion of focus (yet).

Any ideas for other solutions?

> proportional step sizes depending on ranges of sliders and float/list boxes (if range is limited)

Personally, I'm not a too big fan of this. IMO it would be better if we could explicitly set the step size (large and small step) as part of the GUI properties.

Also, is there a particular reason why you're doing this for sliders and gatoms, but not for IEM numberboxes? They are all just different representations of the same thing. I can somehow see why it makes sense for sliders (because of the linear graphical representation) but not so much for gatoms.

That being said, I can imagine the proportional step size being a useful default mechanism in case the user do not resp. cannot specify the step size. I just don't think that `(x->a_draghi - x->a_draglo) / 127.0` is good formula. What about `<next power of 10 of the high limit> * 0.1`:

| Range | Next power of 10 | (Large) step size |
|---|---|---|
| 0 - 0.05 | 0.1 | 0.01 |
| 0 - 1 | 1 | 0.1 |
| 1 - 8 | 10 | 1 |
| 1 - 80 | 100 | 10 |

Side note: In the future we might also support stepping values with the keyboard (which IMO is also long overdue). In that case, the scroll wheel and arrow keys are just two ways of doing the same thing.

Another side note: Qt has a really nice option that is called "adaptive decimal step". The idea is that the step size depends on the number of decimals of the current value. Examples: 1 -> 1.1, 10 -> 11, 100 -> 110, etc.  See https://doc.qt.io/qt-6/qdoublespinbox.html#setStepType. Just something to keep in our minds as a (future) option.

> if you interact with a float in a list gatom and its width changes, you might accidentally interact with another float since there's no interaction state (like for dragging), 

TIL that you can drag individual items in a list atom! 😮 I would have never expected this.

ben-wes:
@Spacechild1 thanks a lot for the feedback on this!

> Are there different platform defaults/expectations on Windows, Linux and macOS?

coming from Windows, i was quite irritated about the inverted scrollwheel direction on macOS first, but got used to it. for the sliders and number boxes, it felt natural to me that an upward movement also rises the value. i guess, i'll need test with more systems to be sure it is consistent. seems like this needs OS specific handling (since your expectation about the slider interaction is similar to mine). and it certainly makes sense to make it configurable!

> horizontal scrolling should follow the logic of vertical scrolling

so `right -> higher`, i guess? that might become a bit complicated with tcltk8. i'll need to look into this more - but it seemed to me like tcltk8 is sending shift+scroll for right swipe, which collides with the fine grained control

> I would use the small step by default and use a modifier for a large step. Also, 0.01 seems to be a rather small (default) multiplier.

i also wasn't sure about the 0.01. maybe 0.1 is enough. but do we really want to invert the logic here compared to the drag interaction with sliders and number boxes?

> Solution Nr. 1: scrolling individual widgets always requires a keyboard modifier.

i hadn't thought of this! sounds like a good option to avoid unexpected behaviour if the user's intent is to just scroll in run mode (as learnt with previous versions). i'm a bit hesitant though since i also like the mouse-only interaction with objects in run mode (thinking of performance setups for example).

checked some random web demo for mousewheel/slider interaction now that seems to work with a timeout for the element interaction. if i continuously scroll up down there with the cursor passing the slider, it keeps scrolling. only if i stop scrolling (or move the mouse), it switches to slider interaction afterwards. for (macbook) touchpad, it seems to be handled in an even more dedicated way and check the 2-finger touch. not sure if this is applicable in our context. timeout might certainly be an option - but obviously also add complexity.
https://owlcarousel2.github.io/OwlCarousel2/demos/mousewheel.html

> Solutions Nr. 2: only scroll if the element has the focus. Disadvantage: you first have to click on the object before you can scroll it. Also, objects like sliders don't have a notion of focus (yet).

i thought about this, too. we'll probably need it in some way for keyboard interaction anyway, right? and there should be a visual feedback as well then ... i'm not sure if i'll manage to handle all of that though.

> why you're doing this for sliders and gatoms, but not for IEM numberboxes

i mainly didn't like the check for the rather arbitrary default "unlimited" limit of `-1e+37..1e+37`. so i decided to always treat it as unlimited. also remembered the help info stating "As of Pd 0.52, the plan is to stop supporting this object in favor of the atom number box". but true - would be good anyway to have it behave consistently.

> What about `<next power of 10 of the high limit> * 0.1`:

sounds good to me!

> TIL that you can drag individual items in a list atom! 😮 I would have never expected this.

wait until you can use these to dynamically create a multichannel signal! :)

besides those points, i'm currently not completely sure about the event path:
1. wheel handler in gui, sending message to pd
2. hitbox check and object handling in pd
3. otherwise message back to gui for canvas scroll handling

... does that seem right?

i'm also considering to get rid of the duplicate `xy` unpacking (once in tcl for canvas scroll and once in c for objects). might be better if the wheel messages were directly created with a `deltax` and `deltay`. any thoughts on this?

Spacechild1:
> so right -> higher, i guess? that might become a bit complicated with tcltk8. i'll need to look into this more - but it seemed to me like tcltk8 is sending shift+scroll for right swipe, which collides with the fine grained control

With horizontal scrolling I meant scrolling horizontal widgets, like HSliders, not actual horizontal scrolling. Sorry for the confusion! I meant that horizontal widgets should logically behave the same as vertical widgets.

> but do we really want to invert the logic here compared to the drag interaction with sliders and number boxes?

Good point! I didn't think of that. So now I know where the 0.01 multiplier comes from. And from that perspective it also makes sense that the Shift-modifier activates the smaller step size.

> if i continuously scroll up down there with the cursor passing the slider, it keeps scrolling. only if i stop scrolling (or move the mouse), it switches to slider interaction afterwards.

I would still find this potentially annoying. I don't always scroll continuously. If I scroll through a patch, I would often stop to look at something. If the mouse happens to be over a slider, I would have to move the mouse to continue scrolling. Even worse: I might not notice and accidentally change the slider value.

>> Solutions Nr. 2: only scroll if the element has the focus. Disadvantage: you first have to click on the object before you can scroll it. Also, objects like sliders don't have a notion of focus (yet).

> i thought about this, too. we'll probably need it in some way for keyboard interaction anyway, right?

That's what I was thinking, yes.

> and there should be a visual feedback as well then ...

Definitely. This would be important for keyboard navigation.

>  i'm not sure if i'll manage to handle all of that though.

I can understand that. However, scroll wheel support should always be designed with keyboard navigation in mind because it's really two sides of the same picture.

> 1. wheel handler in gui, sending message to pd
> 2. hitbox check and object handling in pd
> 3. otherwise message back to gui for canvas scroll handling
>
> ... does that seem right?

Looks right to me!

porres:
> Solutions Nr. 2: only scroll if the element has the focus. Disadvantage: you first have to click on the object before you can scroll it. Also, objects like sliders don't have a notion of focus (yet).

That's my preferred idea. Something I've been looking into for [else/knob]. 

I agree sliders should be able to show us they have focus, like number boxes. Arguably radio buttons as well

By the way, [else/knob] can be either activated by a click or an "active" message input as well.

src/g_editor.c
ben-wes: removed with a325b061dad3c73dd3b9c0c5781ef1f7bd911c8d

src/g_editor.c
ben-wes: changed signature to use `A_GIMME` with a325b061dad3c73dd3b9c0c5781ef1f7bd911c8d, also passing the final deltas for x and y now. unfortunately, ~~the `axis` symbol~~ a touchpad indicator is currently still required since i'm using unit deltas for the value manipulation if they are not coming from touchpad input (but the original values for scrolling).

ben-wes:
the current version should fix the unexpected scroll wheel inversion for value manipulation on non-macos systems. i had a preference added for this, but decided to remove it again with the last commit.

it now also works for GOP objects which wasn't the case before.

focus handling is not part of this yet. neither is the adaption of steps to powers of 10 (i'll gladly give both a try though). more considerations on these though:

> I just don't think that `(x->a_draghi - x->a_draglo) / 127.0` is good formula.

the reasoning here was that it makes sense for the default range of sliders. so it's not completely arbitrary. i don't say this is good - but just wanted to go sure that it's clear where this came from.

another thought on how the object interaction interferes with canvas scrolling: while the handling of an active state is certainly an option, i'm still not sure whether it's good to require an additional interaction (click) prior to the scroll wheel interaction for each manipulation of values. after all, smooth scrolling is possible in edit mode. but i also admit that it's annoying to scroll in run mode and i guess we can't have both. :)

btw: touchpad interaction feels a lot more natural with tcl/tk 9.

ben-wes:
> scroll wheel support should always be designed with keyboard navigation in mind because it's really two sides of the same picture.

another consideration: (future) keyboard interaction should probably be possible for the active object regardless of the mouse position? i wonder if the same should be true then for mouse wheel interaction? that would mean that canvas scrolling is no more possible while an object is active. i don't say this is bad - but is that the expectation?

porres:
> That's my preferred idea. Something I've been looking into for [else/knob].

For the record. I just implemented mouse wheel scrolling into [knob]. It just listens to mouse info and works if the object is active/clicked. Seems like a good mechanism to me. The only issue is that even so, other canvas interactions still happen so you can scroll things up and down. I never have messy patches like that, but not anyone is nice and clean like me.

PR #2667, opened 2025-07-08
keep "-libs "from the cmdline separate from those loaded via preferences

### LINKED ISSUE #2665: Pd ignores arguments after a crash
sometimes your preferences are bonkers and cause a crash of Pd.
to counter this, Pd tries to detect such a previous crash, and will try to prevent another crash by ignoring  the preferences:
> skipping loading preferences... Pd seems to have crashed on startup

that's great.

however, I think it is overdoing: when a crash was detected, Pd seems to not only ignore the (persistent) preferences, but also the arguments on the cmdline.

i think this is the wrong thing to do: cmdline arguments should *always* be honoured.

e.g. i'm currently debugging a crash in an external.
i have a simple test-patch that can trigger the crash, automatic (e.g. via loadbang, or by receiving some symbol)

I'm starting Pd like so:
```command
wine "${PDPATH}/bin/pd.com" -noprefs -lib oopsie -open crashtest.pd -send "test 1"
```

now, if this crashes Pd (which it does), the next time I issue the same command, Pd just starts without trying to load my "oopsie" library and does not it open the `crashtest.pd` patch.
instead I have to start Pd without arguments (just to be sure it doesn't delete the original - working - preferences), save the preferences and quit Pd.
only then I can run the next iteration of my crashing command.

(as you can see, I'm starting Pd with `-noprefs`, because for debugging I actually do not want any persistent preferences. but I need my arguments)

I'm currently observing this under Wine. I usually do *not* see this behaviour, hence tagging is "Windows"

umlaeute:
this is an alternative implementation on how to keep loading externals after a crash was detected.

(my "best way forward" as discussed in https://github.com/pure-data/pure-data/issues/2665#issuecomment-3049124630))

libraries passed via `-lib` cmdline flags, are now stored in a separate namelist (`tmp_externlist`) which is only used during startup.

the externals in this list are *always* loaded (regardless of a detected prior crash).
they are loaded after the ones provided in the preferences (which is the original behaviour).

the only functional difference is, that we now might call `sys_load_lib` twice for a library that is mentioned both in the preferences and on the cmdline.
for libraries that were successfully loaded, this shouldn't make much of a difference (as `sys_load_lib()` will drop duplicate successful requests), but if a library cannot be loaded (e.g. because it doesn't exist) you will get the error "foobar: can't load library" twice.

i was thinking about adding the cmdline list of externals to `STUFF->st_externlist` as well (e.g. while loading),
but refrained from doing so, as I actually find it annoying that cmdline libraries can get promoted to preference libraries (when saving the prefs).

---

- Closes: https://github.com/pure-data/pure-data/issues/2665

### DIFF
--- a/src/s_main.c
+++ b/src/s_main.c
@@ -75,8 +75,9 @@ typedef struct _patchlist
     char *pl_args;
 } t_patchlist;
 
-static t_patchlist *sys_openlist;
-static t_namelist *sys_messagelist;
+static t_namelist *tmp_externlist = 0;
+static t_patchlist *sys_openlist = 0;
+static t_namelist *sys_messagelist = 0;
 static int sys_version;
 int sys_oldtclversion;      /* hack to warn g_rtext.c about old text sel */
 
@@ -435,6 +436,12 @@ int sys_main(int argc, const char **argv)
                 post("%s: can't load library", nl->nl_string);
         sys_oktoloadfiles(1);
     }
+    for  (nl = tmp_externlist; nl; nl = nl->nl_next)
+    {
+        if (!sys_load_lib(0, nl->nl_string))
+            post("%s: can't load library", nl->nl_string);
+    }
+
         /* open patches specifies with "-open" args */
     for (pl = sys_openlist; pl; pl = pl->pl_next)
         openit(cwd, pl->pl_file, pl->pl_args);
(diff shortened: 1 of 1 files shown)


umlaeute:
ping @millerpuckette

Issue #2663, opened 2025-07-08
`colors` message: invalid hex codes or color names create tcl errors


ben-wes:
macOS, arm64, master at 3bba513

when a user sends an invalid `colors` message to `pd`, this currently creates TCL errors when trying to draw objects. should probably be caught.

and it seems to distinguish at least these 2 cases:

invalid hex codes:

```
(Tcl) UNHANDLED ERROR: invalid color name "#ff000000"
    (processing "-outline" option)
    invoked from within
".x600001204800.c itemconfigure {0x132008200BASE} -width 1 -fill {#ffffff} -outline {#ff000000} "
    ("uplevel" body line 6)
    invoked from within
"uplevel #0 $docmds"
```

invalid names (that don't look like hex codes?)

```
(Tcl) UNHANDLED ERROR: unknown color name "foo"
    (processing "-outline" option)
    invoked from within
".x600001204800.c itemconfigure {0x132008200BASE} -width 1 -fill {#ffffff} -outline {foo} "
    ("uplevel" body line 6)
    invoked from within
"uplevel #0 $docmds"
```

porres:
In the case of symbols, I think it might be impossible to catch the error before it happens. Checking if it is a hex number of a given digit length is easier. I'm already restricting to hex symbols in this PR https://github.com/pure-data/pure-data/pull/2646 - so it could be improved.

But then, on the dev list, people maybe still want symbols, so I don't know... people would need to know what they're doing and check the docs :)

porres:
hey, I see you have a fix for this :)

umlaeute:
> But then, on the dev list, people maybe still want symbols, so I don't know... people would need to know what they're doing and check the docs :)

do you have a reference? i seem to have missed such an email.

porres:
> do you have a reference? i seem to have missed such an email.

@danomatika mentioned on the dev-list he "wasn't sure about the #". What I gathered from this is that we must discuss this more.

danomatika:
I would pull the fixes & "no brainers" out into a separate PR @millerpuckette can consider, while leaving the others either here or in separate issues. I know it's easy to tackle everything at once, but we also don't want to rush into things which would be much harder to later revert.

ben-wes:
> I see you have a fix for this

[...]

PR #2668, opened 2025-07-08
Fix Pd colors

porres:
Fixes some issues with Pd's colors. This is a reboot of https://github.com/pure-data/pure-data/pull/2646

It's made simpler and packed into 3 commits:

- [x] Some basic items didn't conform with Pd's colors: Cords weren't created with the foreground color. GOPs and array names didn't retain foreground colors when reopening the patch. The selection rectangle was hardcoded to black and now is the same as the selection color. Displacing GOPs changed the selection color to foreground (also applied to array name and this was an old bug prior to adding colors).
- [x] Iemguis now change their outline color to Pd's foreground, as well as selection color to Pd's selection color. Scalars also use the selection color.
- [x] Pd should restrict color names to hex numbers preceded by "#". This makes it easier to catch illegal symbols that trigger tcl/tk errors. It also makes life easier for other forks (which won't need to bother to support fancy tcl/tk color names like "mistyrose"). In the end, it's also desirable if we want to set iemguis's background and foreground colors to Pd's
- [ ] Unrestrict color names to also allow "blue", "red" and whatnot... but store these as hex symbols. This process also allows validating and preventing tcl/tk errors. It also makes things work with iemguis. Pd forks can also find a way to support tcl/tk colors, after all, they provide all the table values for all supported symbols.

------

- closes https://github.com/pure-data/pure-data/issues/2663
- partially closes items in https://github.com/pure-data/pure-data/issues/2642

### DIFF
--- a/src/g_bang.c
+++ b/src/g_bang.c
@@ -37,22 +37,28 @@ static void bng_draw_config(t_bng* x, t_glist* glist)
     sprintf(tag, "%pBASE", x);
     pdgui_vmess(0, "crs iiii", canvas, "coords", tag,
         xpos, ypos, xpos + x->x_gui.x_w, ypos + x->x_gui.x_h);
-    pdgui_vmess(0, "crs ri rk", canvas, "itemconfigure", tag,
-        "-width", zoom, "-fill", x->x_gui.x_bcol);
+    pdgui_vmess(0, "crs ri rk rs", canvas, "itemconfigure", tag,
+        "-width", zoom, "-fill", x->x_gui.x_bcol,
+        "-outline", THISGUI->i_foregroundcolor->s_name);
 
     sprintf(tag, "%pBUT", x);
     pdgui_vmess(0, "crs iiii", canvas, "coords", tag,
         xpos + inset, ypos + inset,
         xpos + x->x_gui.x_w - inset, ypos + x->x_gui.x_h - inset);
-    pdgui_vmess(0, "crs ri rk", canvas, "itemconfigure", tag,
-        "-width", zoom, "-fill", (x->x_flashed ? x->x_gui.x_fcol : x->x_gui.x_bcol));
+    pdgui_vmess(0, "crs ri rk rs", canvas, "itemconfigure", tag,
+        "-width", zoom, "-fill", (x->x_flashed ? x->x_gui.x_fcol : x->x_gui.x_bcol),
+        "-outline", THISGUI->i_foregroundcolor->s_name);
 
     sprintf(tag, "%pLABEL", x);
     pdgui_vmess(0, "crs ii", canvas, "coords", tag,
         xpos + x->x_gui.x_ldx * zoom, ypos + x->x_gui.x_ldy * zoom);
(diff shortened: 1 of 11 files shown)


porres:
- please consider https://github.com/pure-data/pure-data/pull/2669 if this gets completely merged

porres:
> This becomes quite hard if instead of a nicely formatted hex symbol I get "blue", "magenta" or even "chocolate" instead...

Ok, I figured out how to keep using these values, at least for the colors message sent to Pd. The thing is that there is a quite simple way to validate the input and convert these symbols supported by tcl/tk to hex symbols.

This way, we can give an error if the color is invalid and not do anything, and also store something like "THISGUI->i_foregroundcolor " as a hex symbol, that can be used in iemguis and what not.

I actually like  being able to use colors like "red/magenta" and stuff. It's friendly as well... and "chocolate", who doesn't like "chocolate"?

I am testing this for my [knob] external and can make a new commit for this later on.

Here's how it goes, we need a proc like this.

```
sys_gui("\n"
"proc color2hex {name} {\n"
"    if {[catch {winfo rgb . $name} rgb]} {\n"
"        return \"\"\n"
"    }\n"
"    set r [format %02x [expr {[lindex $rgb 0] / 256}]]\n"
"    set g [format %02x [expr {[lindex $rgb 1] / 256}]]\n"
"    set b [format %02x [expr {[lindex $rgb 2] / 256}]]\n"
"    return \"#$r$g$b\"\n"
"}\n"
"\n")
```

then I have something like

```
static void knob_bgvalid(t_knob *x, t_floatarg v, t_symbol *c){
    int valid = (int)v;
    if(v){
        x->x_bg = c;
        knob_config_bg(x);
    }
    else
        post("[knob] invalid color (%s)", c->s_name);
}

static void knob_bgcolor(t_knob *x, t_symbol *s, int ac, t_atom *av){
    x->x_ignore = s;
    if(!ac)
        return;
    sys_vgui("set color_name {%s}\n", knob_getcolor(ac, av)->s_name);
    sys_vgui("set color_escaped [string map {{ } {\\ }} $color_name]\n");
    sys_vgui("set hex_color [color2hex $color_name]\n");
    sys_vgui("if {$hex_color eq \"\"} {pdsend \"%s _bgvalid 0 $color_escaped\"} else {pdsend \"%s _bgvalid 1 $hex_color\"}\n", x->x_bindname->s_name, x->x_bindname->s_name);
}
```
These even supports colors with spaces, such as "Misty\ Rose" (my favorite)

porres:
> Ok, I figured out how to keep using these values, at least for the colors message sent to Pd. The thing is that there is a quite simple way to validate the input and convert these symbols supported by tcl/tk to hex symbols.
> 
> This way, we can give an error if the color is invalid and not do anything, and also store something like "THISGUI->i_foregroundcolor " as a hex symbol, that can be used in iemguis and what not.

The issue with Pd Forks remain... but they can also find a way to support tcl/tk color symbols, after all, they provide all the table values for all supported symbols.

porres:
really not confident on how to do this, just storing this conversation here for future reference
https://github.com/copilot/share/c81c409c-0860-8c07-a840-a84804d42012

PR #2669, opened 2025-07-09
Allow iemguis to load Pd's runtime colors as the default

porres:
This uses Pd's background color for iemguis' background colors. While iemgui's foreground color and label colors take the value of Pd's foreground color.

This is just a simple change and more changes should be required to make iemguis adapt to Pd's color when they're set/changed by the user, but dealing with that is outside the scope of this PR!

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -1052,9 +1052,11 @@ t_iemgui *iemgui_new(t_class*cls)
     iem_inttosymargs(&x->x_isa, 0);
     iem_inttofstyle(&x->x_fsf, 0);
 
-    x->x_bcol = 0xFCFCFC;
-    x->x_fcol = 0x00;
-    x->x_lcol = 0x00;
+    t_atom colors[3];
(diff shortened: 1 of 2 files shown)


porres:
> Also note that iemguis won't understand color names as "black", "white", etc... so this PR requires that https://github.com/pure-data/pure-data/commit/9a94a67c1860ba9c13e1e8e5de64e63ab6068e7d from https://github.com/pure-data/pure-data/pull/2668 is accepted and merged.

I can imagine it'd be quite possible to add support for color symbols in iemguis if the above PR/commit is not considered. I wouldn't like to be the person who'd try and do that.

porres:
hmmm, [cnv] still has its own default colors for background and label. Having the exact same color as the background is not a good idea indeed, as it makes it invisible. So I thought about setting the background color to something close but a bit darker or lighter depending on the color. I did this in https://github.com/pure-data/pure-data/pull/2669/commits/74b9e9cff2ce8c8e3cd98e1abceee2ed6a56735b

porres:
this needs rebooting

PR #2670, opened 2025-07-09
add '-h' flag for hexadecimal colors in Data Structures and add Hex colors to Garrays

porres:
This add hex color support to  [drawtext]/[drawnumber]/[drawsymbol],  [plot] and the other shapes (polygons). It also adds support for garrays.

On top of this I hope to allow the color of garrays to initialize with Pd's foreground color.


porres:
Ok, I added a new commit with what I had as an attempt to add hex colors to arrays, even though it failed, but I thought I could put this back as a draft, like I first had... I can't seem to do it.

porres:
Ok, I had a silly mistake, but now it works!!! And here's a test patch...

[test.pd.zip](https://github.com/user-attachments/files/21145862/test.pd.zip)

For the record, I feel this is rather clumsy, but I was first worried about making things work. Now it's time to review and polish.

umlaeute:
> Now it's time to review and polish.

so we agree that this is not ready for inclusion into Pd-0.56?

src/g_array.c
```diff
@@ -1195,24 +1195,70 @@ static void garray_width(t_garray *x, t_floatarg width)
...
-static void garray_color(t_garray *x, t_floatarg color)
+static int rangecolor(int n)    /* 0 to 9 in 5 steps */
+{
+    int n2 = (n == 9 ? 8 : n);               /* 0 to 8 */
+    int ret = (n2 << 5);        /* 0 to 256 in 9 steps */
+    if (ret > 255) ret = 255;
+    return (ret);
+}
+
+static void symboltocolor(t_symbol *sym, char *s)
```
umlaeute: so a "color" is a `(char*)`.

of which size? if the size is fixed, you should say so in the declaration (use `char s[8]`)
if it is not fixed, then you should pass the length as an argument.

src/g_array.c
```diff
@@ -1195,24 +1195,70 @@ static void garray_width(t_garray *x, t_floatarg width)
...
+{
+    if ('#' == sym->s_name[0])
+    {
+        int col = (int)strtol(sym->s_name+1, 0, 16) & 0xFFFFFF;
+        pd_snprintf(s, 8, "#%06x", col);
+    }
+    else
+    {
+        sprintf(s, "#%6.6x", 0);
```
umlaeute: why do you use `pd_snprintf()` in one case, and `sprintf()` in the other?

src/g_array.c
```diff
@@ -1195,24 +1195,70 @@ static void garray_width(t_garray *x, t_floatarg width)
...
+    }
+}
+
+static int numbertocolor(int n)
```
umlaeute, line 1219: and now "color" is an `int`?

src/g_array.c
```diff
@@ -1195,24 +1195,70 @@ static void garray_width(t_garray *x, t_floatarg width)
...
     t_template *scalartemplate;
     if (!(scalartemplate = template_findbyname(x->x_scalar->sc_template)))
     {
         pd_error(0, "array: no template of type %s",
             x->x_scalar->sc_template->s_name);
         return;
     }
-    colorwas = template_getfloat(
+/*    colorwas = template_getfloat(
         scalartemplate, gensym("color"), x->x_scalar->sc_vec, 1);
```
umlaeute, line 1252: why has this been commented out?
any reason to redraw the array with the same color?

src/g_template.c
```diff
@@ -1313,6 +1347,19 @@ static int numbertocolor(int n)
     return color;
 }
 
+static void symboltocolor(t_symbol *sym, char *s)
```
umlaeute, line 1426: duplicate code

src/g_template.c
```diff
@@ -2028,15 +2103,19 @@ static void plot_vis(t_gobj *z, t_glist *glist,
         }
         else
         {
-            int outline = numbertocolor(
-                fielddesc_getfloat(&x->x_outlinecolor, template, data, 1));
+            char colorstr[8];
+            if(x->x_outlinecolor.fd_type == A_FLOAT){
+                int color = numbertocolor(
+                    fielddesc_getfloat(&x->x_outlinecolor, template, data, 1));
+                pd_snprintf(colorstr, sizeof(colorstr), "#%06X", color);
```
umlaeute, line 2207: yet another way to convert the numeric color to a string.

src/g_template.c
```diff
@@ -2129,11 +2208,11 @@ static void plot_vis(t_gobj *z, t_glist *glist,
                 }
             ouch:
 
-                pdgui_vmess(0, "crr ri rk rk ri rS",
+                pdgui_vmess(0, "crr ri rs rs ri rS",
                     glist_getcanvas(glist), "create", "polygon",
```
umlaeute, line 2309: sidenote: i actually quite like the idea to have a color specifier `k` in the vmess.

how about using an `int32` as the internal representation and keeping to use `k`?

src/g_template.c
```diff
@@ -2747,11 +2827,15 @@ static void *drawtext_new(t_symbol *classsym, int argc, t_atom *argv)
             fielddesc_setfloat_const(&x->x_vis, 0);
             argc--; argv++;
         }
+        else if (!strcmp(firstarg->s_name, "-h"))
```
umlaeute: i thought that most people would prefer to not introduce the `-h` flag, and instead autodetect whether it's a symbol or a float?

src/g_template.c
```diff
@@ -2933,8 +3023,14 @@ static void drawtext_vis(t_gobj *z, t_glist *glist,
             basex + fielddesc_getcoord(&x->x_xloc, template, data, 0));
         int yloc = glist_ytopixels(glist,
             basey + fielddesc_getcoord(&x->x_yloc, template, data, 0));
-        int color = numbertocolor(
-            fielddesc_getfloat(&x->x_color, template, data, 1));
+        char colorstr[8];
+        if(x->x_color.fd_type == A_FLOAT){
+            int color = numbertocolor(
+                fielddesc_getfloat(&x->x_color, template, data, 1));
+            pd_snprintf(colorstr, sizeof(colorstr), "#%06X", color);
```
umlaeute, line 3180: duplicate code

src/g_array.c
```diff
@@ -806,12 +808,10 @@ static void garray_save(t_gobj *z, t_binbuf *b)
...
-        t_float fval;
-        style = template_getfloat(
+        t_symbol *color = template_getsymbol(
             scalartemplate, gensym("color"), x->x_scalar->sc_vec, 1);
-        binbuf_addv(b, "ssi;", gensym("#A"), gensym("color"), style);
-
-        fval = template_getfloat(
+        binbuf_addv(b, "sss;", gensym("#A"), gensym("color"), color);
+        t_float fval = template_getfloat(
             scalartemplate, gensym("linewidth"), x->x_scalar->sc_vec, 1);
```
umlaeute, line 815: this  breaks compatibility with older Pds.

umlaeute (changes requested):

porres:
> so we agree that this is not ready for inclusion into Pd-0.56?

Depends on the review :) but I'm not that hopeful and  wouldn't bet on it, haha. I bet I have to improve some things.

It might be approved with easy changes and rather quickly, but as I see it, this is a big feature addition and requires lots of documentation updates and it seems we're already fully packed with goodies for 0.56, so I think it's more about that.

I'd say this is high priority for 0.57 on the other hand, as it's necessary to add support for garrays to load Pd's foreground color.

src/g_array.c
porres: this is for the case when you have a float to set the color, like we do now. It then calls this 'numbertocolor' function to turn, say "666" it into an RGB int, then use that to create a string.

src/g_array.c
porres: This became tricky when I changed to symbols, and I just wanted to test if things worked at all, but yeah, I should write new code for that. I'm still not sure though.

I was gonna ask about compatibility issues, if patches are saved with float colors, how do we manage that. I'm trying to figure this out right now when also managing to initialize the array with Pd's colors. I think this is the main issue now and I'm afraid there might be a case where we could be comparing a float to a symbol, so I just commented out for now.

src/g_template.c
porres: yup, I need to figure out how to use just one. This is not the only duplication. I just injected some functions in here as well so I could test, but this is definitely one of the things I was referring to about "polishing"

src/g_template.c
porres: another duplicate actually, there's yet another one. All functions I added to g_array are duplicate

src/g_template.c
porres: But how would it be if we have a symbol input? As I see it, we could actually keep the code as it was, with 'k', and use an "if" statement to deal with the case of symbols. That would make the code much larger and my strategy was to convert floats to strings and just use strings... but I'm happy with any solution/suggestion.

src/g_template.c
porres: yeah, that's a damn good idea. I just never thought of it

src/g_array.c
porres: I see. I guess you're the one to come up with a good solution. How was it for iemguis when it started saving colors as symbols? 

Could we get the symbol and save it as an int? That wouldn't give errors but it would open the patch on an older version with the wrong color. So I don't know...

porres:
> Depends on the review :) 

so... definitely no... haha. Thanks for the reviews!

src/g_array.c
porres: fixed with https://github.com/pure-data/pure-data/pull/2670/commits/ad2878353db1f3b2c7ab19ee43de062e5cb3a02b

src/g_array.c
porres: fixed with https://github.com/pure-data/pure-data/pull/2670/commits/ad2878353db1f3b2c7ab19ee43de062e5cb3a02b

porres:
Hi @millerpuckette since you're doing lots of work related to data structures, would you please check this out? I had to work on solving conflicts a few times already with latest changes. 

I think this is a neat feature that many people want, the ability to be able to set colors with more resolution in RGB channels, what do you think?

porres:
damn, I don't know what changed, but now creating arrays crash Pd... probably some recent change in the code as this wasn't happening before, but the rest (change in g_template.c) seem to work just fine and could be taken.

millerpuckette:
I think having a hex color as a symbol isn't a very good idea.  I've tried to think of a clean way to actually have a triple of floats define a color but it would require adding a creation flag to every drawing command - probably add 100s of lines of ugly code. Meanwhile I'm still trying to make data structures into something more usable fundamentally, which will entail making more changes.  So my inclination is to wait on this until things are more stable in other respects.

porres:
> I think having a hex color as a symbol isn't a very good idea.

I also like 3 floats better, but this is also more convenient and easy. Another thing is that we already have symbols to set colors in Pd, not only in Pd Colors, but also in iemguis.

PR #2671, opened 2025-07-11
Fix Pd colors a bit more

umlaeute:
this builds on (and includes) https://github.com/pure-data/pure-data/pull/2668/
but adds the following:

- use `unsigned int` for the internal representation of colors (rather than a symbol) (ffafeaa10526466e912b6dd6ce5d5876cb5ae633)
  - as a consequence, we can get back to using `k` (the *color* specifier) in `pdgui_vmess()` for sending colors to Pd-GUI again (rather than a weird mixture of `k`, `s` (escaped strings) and `r` (raw strings).
- the 2nd commit (929212f9b831e58c9d37edd7dafb56ee9ed38521) allows iemgui's to understand a color `default` which just uses whitish for the background, and black for label/foreground.
  - this is mean **only** as a smallish change to allow future versions of Pd to use the `default` color in a more meaningful way (as in: use the colors from Pd's color scheme)
  - it will get the "default" colors for `[vu]` and `[cnv]` wrong (as they don't typically have a `#FFFFFF` background), but i think the overall experience is better than making everything `#000000` (which is what is currently happening)
  - this is *not* for the users to be used yet (and probably shouldn't be documented yet)

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -926,11 +926,11 @@ static void iemgui_draw_iolets(t_iemgui*x, t_glist*glist, int old_snd_rcv_flags)
     sprintf(tag, "%pOUT%d", x, 0);
     pdgui_vmess(0, "crs", canvas, "delete", tag);
     if(!x->x_fsf.x_snd_able) {
-        pdgui_vmess(0, "crr iiii rs rs rS",
+        pdgui_vmess(0, "crr iiii rk rk rS",
             canvas, "create", "rectangle",
             xpos, ypos + x->x_h + zoom - ioh, xpos + iow, ypos + x->x_h,
-            "-fill", THISGUI->i_foregroundcolor->s_name,
-            "-outline", THISGUI->i_foregroundcolor->s_name,
+            "-fill", THISGUI->i_foregroundcolor,
+            "-outline", THISGUI->i_foregroundcolor,
             "-tags", 2, tags);
         /* keep label above outlet */
         pdgui_vmess(0, "crss", canvas, "lower", tag, tag_label);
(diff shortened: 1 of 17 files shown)


porres:
https://github.com/pure-data/pure-data/pull/2668 also has plans to build on top of https://github.com/pure-data/pure-data/pull/2668 so I guess I'll hold...
 
My concern to still allow users to set pd's colors via symbols and use a tcl/tk proc to validate and normalize to an int value (in the context of this PR). What do you think about this? I got stuck and need help anyway, so if you like the idea we can include this in here.

porres:
The https://github.com/pure-data/pure-data/pull/2669 also deals with iemguis and also builds on top of  https://github.com/pure-data/pure-data/pull/2668

the way I'm dealing with  [cnv]'s default color for background is to set something close but a bit darker or lighter depending on whatever color is the 'default'. See https://github.com/pure-data/pure-data/commit/74b9e9cff2ce8c8e3cd98e1abceee2ed6a56735b

umlaeute:
>  use a tcl/tk proc to validate and normalize to an int value (in the context of this PR). What do you think about this?


to be honest: not very much. 

it feels a bit like asking some random webservice... which I wouldn't do in a real-time application for something so trivial.

umlaeute:
> the way I'm dealing with [cnv]'s default color for background is [...]


good idea, but.

this PR is targeting 0.56, so I deliberately tried to keep things to a minimum. 

my gut feeling says that adjusting colours to shades of the main colour, it's on the realm of "features", and I don't want to go there at this stage. 

OTOH, not doing something excitedly stupid for a special colour "default" is (meant to be") *future proof*.

millerpuckette:
I think these are good ideas.  I think they can be changed later without breaking things - but I wouldn't be against just throwing them into 0.56 since IEM GUI's color-stuff can always be advertized as possibly to be changed in the future.

Issue #2673, opened 2025-07-13
also allow color names for "pd colors" message (like "blue", "red", etc).


porres:
So we're now restricting color names to a hex symbol format, which was needed.

I am now thinking of a way to allow users to set color by names as before, but validate them first and convert to hex symbols. This process also allows validating and preventing tcl/tk errors and also works with iemguis. Pd forks can also find a way to support tcl/tk colors, after all, they provide all the table values for all supported symbols (for reference: https://www.tcl-lang.org/man/tcl9.0/TkCmd/colors.html and https://wiki.tcl-lang.org/page/Colors+with+Names ).

I find it friendly and useful for users to use color names without having to master conversions. 

I am successfully using this method for my externals, like [else/colors], which can convert symbols to hex/RGB formats. We just need a tcl/tk 'proc' like this.

```
sys_gui("\n"
"proc color2hex {name} {\n"
"    if {[catch {winfo rgb . $name} rgb]} {\n"
"        return \"\"\n"
"    }\n"
"    set r [format %02x [expr {[lindex $rgb 0] / 256}]]\n"
"    set g [format %02x [expr {[lindex $rgb 1] / 256}]]\n"
"    set b [format %02x [expr {[lindex $rgb 2] / 256}]]\n"
"    return \"#$r$g$b\"\n"
"}\n"
"\n")
```

Then, for [else/colors], I have something like

```
static void colors_valid(t_colors *x, t_floatarg v, t_symbol *c){
    int valid = (int)v;
    if(v){
        strncpy(x->x_color, c->s_name, MAXPDSTRING);
        colors_bang(x);
    }
    else
        post("[colors] invalid color (%s)", c->s_name);
}

static void colors_symbol(t_colors *x, t_symbol *sym){
    sys_vgui("set color_name {%s}\n", sym->s_name);
    sys_vgui("set color_escaped [string map {{ } {\\ }} $color_name]\n");
    sys_vgui("set hex_color [color2hex $color_name]\n");
    sys_vgui("if {$hex_color eq \"\"} {pdsend \"%s _valid 0 $color_escaped\"} else {pdsend \"%s _valid 1 $hex_color\"}\n",
        x->x_id->s_name, x->x_id->s_name);
}
```

Doing such form of callback is usual in Pd, with internals and externals. The [openpannel] object does something like that, for instance. I'm just not sure how to do this for Pd, as we're not really using a particular object to communicate with.
[...]

PR #2675, opened 2025-07-13
allow users to set Pd colors with color names as before

porres:
This uses a tcl/tk function to validate color symbols (like "blue", "red", etc) and returns a hex symbol to be saved, like we're currently doing

closes https://github.com/pure-data/pure-data/issues/2673

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -2331,38 +2331,48 @@ static void glist_dorevis(t_glist *glist)
             glist_dorevis((t_glist *)g);
 }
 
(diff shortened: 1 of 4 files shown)


umlaeute:
i feel very uncomfortable with this.

```
[0   \
|
[colors grey$1 white blue(
|
[s pd]
```

porres:
now I'm surious, why?

porres:
Well, anyway, miller is not keen on this, here's a quote

" I like the idea of restricting colors to something that another
future GUI can understand."

Issue #2639, opened 2025-06-23
request: allow iemguis/array colors to be set with Pd's colors


porres:
I'm thinking of an option to iemguis and probably externals to be able to change according to a color theme, now possible with the 'colors' message to Pd.

Maybe something like a "theme" toggle, that would allow the background and foreground color to be loaded and changed accordingly.

Arrays should also be into this category as the foreground is setable.

porres:
there's a new related comment at https://github.com/pure-data/pure-data/issues/2637#issuecomment-2994622312

so let me clarify that by themeable background and foreground colors do not apply to the selection or outline color.

porres:
I see that we can do all of this with externals as it is, so no new change is needed. I have tested this with my [knob] object and I think it works great!

I'm using a "theme" message and it just uses the color themes for background and foreground. Maybe "theme" is not the best name, but I don't know what else to call this.

<img width="356" alt="Image" src="https://github.com/user-attachments/assets/a28d96a5-cf1e-43ea-8313-26714ece82aa" />
<img width="316" alt="Image" src="https://github.com/user-attachments/assets/80dce5e3-c860-49c0-b6bf-d4b8d144d42b" />

umlaeute:
even with the new `color` message, we do not have *themes* in Pd - we only have the possibility to set a few colours.

while this can be a building block for "themes", it lacks "meta information" (that is: there's no *group* of colours, just individual ones).

i'm not sure whether i like the current implementation (i tend towards "no"), but well... this is what we have for now.

in any case: if we think of using `color` to set a theme by setting individual colors, setting a theme for iemguis would also just set individual colours for all of the iemguis.
currently this means setting the colours for *each object* (which is tedious; but doable).

so: how would you imagine such a theming to work?
e.g. i could imagine a generic "colour transform" function that gets applied to all iemguis - or even better: to all (builtin) objects. but then this would somewhat replace the current `color` message, and we are back at field 0.


i'm inclined to close this as a duplicate of #931

porres:
> i'm inclined to close this as a duplicate of https://github.com/pure-data/pure-data/issues/931

No, that issue is something else :) that one is to add proper "themes", but I guess I'm not calling it right... so, in order to make it straight, that one is about adding some "presets" for colors, in the GUI. We can make that when we add the ability to set colors via preferences or something.

Now, this issue is about letting iemguis and arrays to be aware of Pd's background/foreground/select colors and change accordingly.

[...]

PR #2679, opened 2025-07-14
load Pd's colors as iemguis default

porres:
This uses Pd's background color for iemguis' background colors. While iemgui's foreground color and label colors take the value of Pd's foreground color.

This is just a simple change and more changes should be required to make iemguis adapt to Pd's color when they're set/changed by the user, but dealing with that is outside the scope of this PR!

[cnv] had its own default colors for background and label. Having the exact same color as the background is not a good idea as it makes it invisible. So I'm setting the background color to something close but a bit darker or lighter depending on the color.


porres:
here's a test screenshot

Screenshot 2025-07-14 at 09 02 23

porres:
@umlaeute I have to say I don't if this gets in the way of your plans for setting 'default' colors in iemguis. I just couldn't get the idea at all...

Other than that, I think this is supposed to be quite simple, just use Pd's default colors when creating iemguis...

umlaeute:
the "issue" here is that it will store the current colors in the patch.

afaik, this is what Plug Data already does, and people seem to be [less than happy](https://github.com/pure-data/pure-data/issues/2639#issuecomment-3005307465).

my idea for the `default` color is to be able to always get a color based on the **current** colors (instead of the what was current when the patch was saved)

porres:
> the "issue" here is that it will store the current colors in the patch.

yeah, but if I understand correctly, your accepted PR is now just storing color values as well.

> afaik, this is what Plug Data already does, and people seem to be less than happy.

It is not ideal indeed, but it is at least a first step in that direction. I'm up for a way to make iemguis load Pd's color when opening the patch of course, but that needs more discussion, right?

Also, it's be important to make iemguis react to color changes, not only at load time.

The way I did this was by adding a new boolean argument and checkbox on properties. If "on", it reacts to changes and loads Pd's color. I'm not sure yet how you want to achieve this.

One way or another, I guess this "doesn't get in the way", huh?

ben-wes:
> [...] your accepted PR is now just storing color values as well.

if a user touches the color values, we can assume that the colors should *not* follow any theme, but be custom colors. this should probably also stay the case (and is even more likely) if iemguis are "themed" on creation.

> One way or another, I guess this "doesn't get in the way", huh?

not sure about this. it's certainly not great if users with dark mode (or disco mode) get the "old" default colors for iemguis. and it's not unlikely that they'll mainly be annoyed and change the colors anyway ...

as mentioned in the comment linked above, i would like to avoid the state that you get with plugdata, where you open patches in your theme and still see those guis in the colors of other users' themes. the only way i see for storing and interpreting a backward compatible "follow the theme" (or "this is the default") state is by not touching and detecting the old default colors. actual "default" symbols would obviously be much better - just not sure how that can be achieved in a backward compatible way?

https://github.com/pure-data/pure-data/discussions/2648#discussioncomment-13652258 is trying to contribute to this discussion.

porres:
> your accepted PR is now just storing color values as well.

So, I tested, and I can see that one can alter the saved file to include "default" and it'll load the default color #fcfcfc, but this color gets stored in the file... I can see that a "color default default default" message will also work, to load so called default colors, but the patch still stores the color names.

I guess I see where this is going, but only the ground is laid so far.... nothing really happens. And we can use the 'default' argument to load the themes instead. So I guess I can build on top of this idea...

the thing is that right now this isn't much different, as it stores the theme colors. Only when we have an actual support to save the argument as 'default' is when we can start offering the color scheme instead of hard coded defaults.

The question is, why don't we start saving the colors with the 'default' symbolic argument already? What's in the way?

ben-wes:
> What's in the way?

opening patches with `default` colors in previous Pd versions?

umlaeute:
> What's in the way?

the pending release

porres:
> So, I tested, and I can see that one can alter the saved file to include "default" and it'll load the default color #fcfcfc, but this color gets stored in the file... I can see that a "color default default default" message will also work, to load so called default colors, but the patch still stores the color names.

This was misleading. This actually happens to ANY symbol that is not in hex format... so I'm really not sure about the future plans for the transition phase. Just to be clear, we'd start saving the argument as "default" in 0.57 and this would result in loading the 'old' default colors in 0.56... but then, saving it would also break and not work in 0.57, right?

>  this is what Plug Data already does, and people seem to be less than happy.

it's still better than ignoring color schemes altogether, before we do something better with it. More importantly, if the user wants to adapt the GUIs to the same color scheme, it's just more work to manually set it up and it'll also be saved with the patch anyway...

So we might as well just let these be loaded cohesively along other patch elements (number boxes and stuff).

umlaeute:
> actual "default" symbols would obviously be much better - just not sure how that can be achieved in a backward compatible way?

with plenty of time, as we did with other incompatible changes on the patch-level. (have Pd accept a new format at release X, but don't write that format until X+n, where n is some smallish number (1<=n<=4).

ben-wes:
> with plenty of time

ok, that's what i assumed - slowly growing the range of "backward". :)

while the current version (Pd0.56-0test2) at least doesn't display black guis from future patches that store "default" colors, the next steps should then be these, i assume?:

* make "default" colors display based on palette's foreground/background
* save "default" if colors stay untouched (not overwriting with hex colors)
* allow to reset to default

... possibly, there could be some feature (plugin?) in the future that allows to detect and set the current "#fcfcfc #000000 #000000" to "default default default"?

umlaeute:
> while the current version (Pd0.56-0test2) at least doesn't display black guis from future patches that store "default" colors, the next steps should then be these, i assume?:

yes.

> ... possibly, there could be some feature (plugin?) in the future that allows to detect and set the current "#fcfcfc #000000 #000000" to "default default default"?

or just a simple one-liner sed-script to explicitly convert your patches to use `default`.
e.g.
```sh
find . -name "*.pd" -exec sed -e 's| #fcfcfc #000000 #000000 | default default default |' -i {} +
```

(this is totally untested and should probably be refined to prevent accidental replacement in false-positives)

src/g_mycanvas.c
```diff
@@ -252,9 +252,27 @@ static void *my_canvas_new(t_symbol *s, int argc, t_atom *argv)
...
+    unsigned int g1 = (bg >> 8)  & 0xFF;
+    unsigned int b1 =  bg        & 0xFF;
+
+    unsigned int r2 = (fg >> 16) & 0xFF;
+    unsigned int g2 = (fg >> 8)  & 0xFF;
+    unsigned int b2 =  fg        & 0xFF;
+
+    unsigned int r = (unsigned int)(r1 + (r2 - r1) * factor);
+    unsigned int g = (unsigned int)(g1 + (g2 - g1) * factor);
+    unsigned int b = (unsigned int)(b1 + (b2 - b1) * factor);
```
ben-wes: i guess that the unsigned types are causing issues here for negative differences (for example if `r2 < r1`)?

umlaeute: there's really no reason why r1,...b2 should be `unsigned int`. just make them a `float` type. (you will get intermediate floats anyhow via the `factor`).

and `r`, `g`, `b` should not be unsigned to avoid undefined behaviour, *unless* you do a range-check before assigning to them - but if you are range checking, then you could as well use `unsigned char`.

in any case, before reassembling the final colour from the 3 channels, you must ensure that no channel exceeds the `0x00..0xFF` range (otherwise colors will bleed into the other channels).

:-/

umlaeute:
i take the quality of the color interpolator as an indication that this branch is not ready for inclusion into Pd-0.56 and thus converted it to a draft.

src/g_mycanvas.c
Spacechild1: > and r, g, b should not be unsigned to avoid undefined behaviour

Nitpick, but *unsigned* integer overflow is defined behavior by the standard. E.g. `unsigned int i = -1` is guaranteed to be `UINT_MAX`. Only *signed* integer overflow is undefined behavior!

porres: I welcome any of you to help and collaborate with a PR, as I'm really busy until the end of the month now.

porres:
Ok, that seemed like an easy/quick review I could handle in a few minutes... sorry I'm in such a hurry and just asked AI to write lousy code... 

Some screenshots:

Screenshot 2025-07-18 at 13 09 45

Screenshot 2025-07-18 at 13 09 33

Screenshot 2025-07-18 at 13 09 06
Screenshot 2025-07-18 at 13 09 18

Screenshot 2025-07-18 at 13 08 51

src/g_mycanvas.c
porres: please check if it's good acceptable for 'not a draft' now (I'm still in a haze)

umlaeute:
>
>just asked AI to write lousy code... 
>

please do not do this. 

apart from generating more or less lousy code (which I have about zero intention to waste time reviewing), the license situation is totally unclear. 

I think we should proactively ban AI generated code.

umlaeute:
>
>just asked AI to write lousy code... 
>

please do not do this. 

apart from generating more or less lousy code (which I have about zero intention to waste time reviewing), the license situation is totally unclear. 

I think we should proactively ban AI generated code.


in any case, I will stop reviewing those PRs now, until after the release.

I think that rushed code created with the help of copilot  is actually harmful for getting the release out.

porres:
To be clear and for the record I was referring to the last bad code (and yeah, I was counting on reviewing, sorry). So... this means https://github.com/pure-data/pure-data/pull/2679/commits/95af75f9400470b53a058bf89bdcc984501bdaa1 is not AI! And it's the best I can do... (which isn't that great, I know, but it all looks fine now).

I fixed it taking into consideration the feedbacks, and I tested it as you can see with the screenshots above. It's ready for review/testing/appreciation/more feedback, etc... so not a "draft", but whatever :) this is not urgent and can wait anyway... (the "highlighting" issue should be more important though...)

Issue #2680, opened 2025-07-14
giving invalid colors to Pd turns it into black


porres:
I tried the now unsupported color names and eveything just turned black. Hence, it seems that it didn't skip setting the color. 

A screenshot

<img width="1044" height="685" alt="Image" src="https://github.com/user-attachments/assets/ccd7f1ea-80ba-4920-a8a6-5e3c66cd021f" />

porres:
tested with https://github.com/pure-data/pure-data/commit/66e038a2a0fd45ad9c3a9b5c4e5e81350dbed602

porres:
fixed with https://github.com/pure-data/pure-data/commit/c09ef5b79b47f1e87629a2dec7c74407dc863001

PR #2681, opened 2025-07-14
Fix Pd Colors yet a bit more

porres:
- closes https://github.com/pure-data/pure-data/issues/2680
- closes https://github.com/pure-data/pure-data/issues/2686

### DIFF
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -508,88 +508,12 @@ proc ::pdtk_canvas::cleanname {name} {
     return $name
 }
 
-proc ::pdtk_canvas::rgb_to_hsv {r g b} {
-    set rf [expr {$r / 255.0}]
-    set gf [expr {$g / 255.0}]
-    set bf [expr {$b / 255.0}]
-
-    set max $rf
-    if {$gf > $max} { set max $gf }
-    if {$bf > $max} { set max $bf }
-
-    set min $rf
-    if {$gf < $min} { set min $gf }
-    if {$bf < $min} { set min $bf }
-
-    set v $max
-    set d [expr {$max - $min}]
-    if {$max == 0} {
-        set s 0.0
-    } else {
-        set s [expr {$d / $max}]
-    }
-
-    if {$d == 0} {
-        set h 0.0
-    } elseif {$max == $rf} {
-        set h [expr {60 * fmod((($gf - $bf) / $d), 6)}]
-    } elseif {$max == $gf} {
-        set h [expr {60 * ((($bf - $rf) / $d) + 2)}]
-    } else {
-        set h [expr {60 * ((($rf - $gf) / $d) + 4)}]
-    }
-
-    if {$h < 0} { set h [expr {$h + 360}] }
-
-    return [list $h $s $v]
-}
-
-proc ::pdtk_canvas::hsv_to_rgb {h s v} {
-    set c [expr {$v * $s}]
-    set x [expr {$c * (1 - abs(fmod($h / 60.0, 2) - 1))}]
-    set m [expr {$v - $c}]
-
-    if {$h < 60} {
-        set rf $c; set gf $x; set bf 0
-    } elseif {$h < 120} {
-        set rf $x; set gf $c; set bf 0
-    } elseif {$h < 180} {
-        set rf 0; set gf $c; set bf $x
-    } elseif {$h < 240} {
-        set rf 0; set gf $x; set bf $c
-    } elseif {$h < 300} {
-        set rf $x; set gf 0; set bf $c
-    } else {
-        set rf $c; set gf 0; set bf $x
-    }
-
-    set r [expr {int(255 * ($rf + $m))}]
-    set g [expr {int(255 * ($gf + $m))}]
-    set b [expr {int(255 * ($bf + $m))}]
-    return [list $r $g $b]
-}
-
-proc ::pdtk_canvas::adjust_color {hex} {
-    scan $hex "#%2x%2x%2x" r g b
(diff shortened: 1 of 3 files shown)


porres:
I guess the sane thing to do is to create a new internal color for the non-highlighted cords and calculate the adjusted color just once, huh?

The way I'm doing now is that every time someone creates a connection, the color is calculated.

Not that I think this will choke Pd, it's just not elegant I suppose.

Also, we can use such color for the Pd window in place of the light grey color for "Log level > 2" and [cnv].

porres:
> Also, we can use such color for the Pd window in place of the light grey color for Log level > 2

Oh, and [cnv] as well :)

porres:
testing [cnv] with the same logic...

Screenshot 2025-07-17 at 08 59 12
Screenshot 2025-07-17 at 08 59 22
Screenshot 2025-07-17 at 08 59 38
Screenshot 2025-07-17 at 09 01 11

it does look like a darker gret for the plain black/white (see below), but that doesn't look bad  I think, and it looks good overall in most cases. I wasn't fully satisfied with one of the examples above, but if you go too crazy on colors, you might end up with something kinda funny. For highlighting cords it really did look better anyway and the idea is to offer a simple solution (not overkill) that works best overall...

Screenshot 2025-07-17 at 09 08 08

tcl/pdtk_canvas.tcl
```diff
@@ -508,15 +508,34 @@ proc ::pdtk_canvas::cleanname {name} {
...
+            set v [expr {max(0, $v - 0x1F)}]
+        }
+        set $val $v
+    }
+    
+    # Format back to hex string
+    format "#%02x%02x%02x" $r $g $b
+}
+
+proc ::pdtk_canvas::cords_to_foreground {mytoplevel {state 1} fg bg sel} {
```
umlaeute, line 520: hmm, i think that when doing such calculations on the GUI side (besides the question whether this is a good idea or not), then the GUI should have an a-priori knowledge of the colours involved.

that is: do **not** send the fg/bg/sel colors along with the `cords_to_foreground` command, but instead tell the GUI about the default colors when they are set (that is, in `glob_colors`).


---
also, i think that the proc signature `{mytoplevel {state 1} fg bg sel}` does not make any sense (the `state` variable becomes *required*).
this of course is not a problem if the additional arguments go away again :-)

ben-wes:
> does look like a darker grey for the plain black/white

i think there's an issue with your algorithm. linear interpolation between black and white always results in shades of grey.

tcl/pdtk_canvas.tcl
porres: how do I tell the GUI about the default colors when they are set?

porres: "also, i think that the proc signature {mytoplevel {state 1} fg bg sel} does not make any sense (the state variable loses becomes required)."

I didn't really understand this... but I have to say I just tried something that seemed to make sense to me and it worked

porres:
> i think there's an issue with your algorithm

mine? I just took your file :)

tcl/pdtk_canvas.tcl
porres: > how do I tell the GUI about the default colors when they are set?

I wonder if this means the GUI can then also change the Pd Console window. 

Any guidance in how to do that is highly appreciated. I'd like to try and do that here.

Also, we still have the issue with the edit text blinking bar, which is always black. This PR could expand and fix all these issues

ben-wes:
> I just took your file :)

AFAICT, these [cnv] colors above are calculcated in the core via 3309349325e42fd8fd7a577367721810a409fef4. i only tested this interpolation with a tcl version.

ben-wes:
oh ... wtf? i just realize that my comment here (and the quote - no idea how that's possible) ended up in this PR instead of #2686, where i was referring to your cnv screenshots. sorry, i guess, i got a little confused with the different color discussions in PRs and issues.

tcl/pdtk_canvas.tcl
umlaeute: > I didn't really understand this... 

yea sorrty, i think my mobile's autocompletion got in the way again. i've corrected my original sentence and it should be more clear now.


> but I have to say I just tried something that seemed to make sense to me and it worked

i'll take this as: "I'm just experimenting a bit with ideas, and see what comes out of it".
this is fine by itself, but i think it marks this PR is **not ready** for inclusion in the pending 0.56 release

umlaeute:
converted to draft, so it doesn't get accidentally included in Pd-0.56.
(this branch is still at the stage of trying things out)

tcl/pdtk_canvas.tcl
porres: @umlaeute at least https://github.com/pure-data/pure-data/pull/2681/commits/64d2152f25b5c02c908f89bffae5105c074210f4 is a simple fix that should go into develop.

As for the rest, I have to sign off for the rest of the month. I wish I could take care of more stuff but I'm just really busy now. 

I think that fixing the highlight bug is important for 0.56 and if I can't nail if for now, I hope someone else gets to make a proper fix. One way or another, I think the current solution is good enough as a fix, at least in my tests... and things can be improved later.

porres: >  i've corrected my original sentence and it should be more clear now.

Cool, but I still don't why the state variable becomes required, how it wasn't before and why this is bad. One way or another, the main issue is: how do we tell the GUI about the default colors when they are set?

porres:
closing this in favor of https://github.com/pure-data/pure-data/pull/2692

But I am telling @millerpuckette to cherry pick or apply this simple fix at https://github.com/pure-data/pure-data/pull/2681/commits/64d2152f25b5c02c908f89bffae5105c074210f4

PR #2682, opened 2025-07-15
audio settings: add refresh button for device list

ben-wes:
this is an attempt to allow hotplugging of audio devices for portaudio.

currently, the button generically closes and reopens audio (if it was open) - this might be restricted to portaudio only? and also not sure if it's actually necessary to handle the `was_open` state - or just do close/reopen?

for the preferences window, it repopulates the previous content. the standalone dialog is completely recreated.

### DIFF
--- a/src/m_glob.c
+++ b/src/m_glob.c
@@ -26,6 +26,7 @@ void glob_start_preference_dialog(t_pd *dummy, t_symbol*s);
 void glob_audio_properties(t_pd *dummy, t_floatarg flongform);
 void glob_audio_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_audio_setapi(t_pd *dummy, t_floatarg f);
+void glob_audio_refresh_devicelist(t_pd *dummy);
 void glob_midi_properties(t_pd *dummy, t_floatarg flongform);
 void glob_midi_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv);
 void glob_midi_setapi(t_pd *dummy, t_floatarg f);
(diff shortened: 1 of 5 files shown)


ben-wes:
i was hoping that a comparable mechanism might also be applicable for midi device refresh - but close->reinit->reopen doesn't seem trivial in that case.

will keep this as is for now.

Spacechild1:
This works nicely here on Windows!

The only caveat is that we can't reinitialize portaudio if we are in "callback" mode and DSP is on because it would deadlock. I think we just have to detect this case, post a warning message to the console and return.

```
void glob_audio_refresh_devicelist(t_pd *dummy)
{
    int was_open;
    if (sched_get_using_audio() == SCHED_AUDIO_CALLBACK))
    {
        pd_error(NULL, "Cannot refresh device list in 'callback' mode when audio is running. Please turn off DSP and try again.");
        return;
    }
    // ...    
}
```

> and also not sure if it's actually necessary to handle the was_open state - or just do close/reopen?

No, that's perfectly fine! We only want to reopen the audio device if it has been open.

Spacechild1:
Miller has already implement something similar with 954bb49. However, I would strongly prefer this PR because it doesn't interrupt audio when opening the audio settings dialog.

ben-wes:
> Miller has already implement something similar with [954bb49](https://github.com/pure-data/pure-data/commit/954bb49d816404b93b7aad878780a83f6ba52b06). However, I would strongly prefer this PR because it doesn't interrupt audio when opening the audio settings dialog.

oh, i didn't see that! somehow, i get notifications for almost everything here, but not for changes on master (and no idea how to change that). :)

... i wanted to avoid the audio interruption when opening the settings dialog as well - but it admittedly reduces complexity quite a bit to do it that way.

Spacechild1:
For me the audio interruption while opening the dialog is a no-go. I sometimes need to check the audio settings during a performance. On my system, reinitialising portaudio can take up to 1.5 seconds, which is unacceptable in a live setting.

Spacechild1:
I can't really review the Tcl code. It obviously works, but @umlaeute might have some comments.

ben-wes:
rebased to master and adopted `pa_reinitialize()` now to only add the refresh button mechanism.

Issue #2683, opened 2025-07-16
portmidi reset creates errors on macOS


ben-wes:
system: Pd0.56-0test2, macOS, arm64

portmidi reinit (triggered when opening the MIDI settings or tabbed preferences) creates loads of `PortMidi: 'Bad pointer'` errors for me (completely flooding the log). also, the device list is not updated here.

see https://github.com/pure-data/pure-data/commit/816e5d70c37b1214bf4fb0322add37a4700e58d2#commitcomment-162190136

ben-wes:
the error is gone for me if i do `sys_close_midi();` before `Pm_Terminate();`. but it doesn't help much: the device list still doesn't receive any updates.

Spacechild1:
> the error is gone for me if i do sys_close_midi(); before Pm_Terminate();

For me too! As I noted in https://github.com/pure-data/pure-data/pull/2685#discussion_r2210590762, you also have to call `sys_reopen_midi()` afterwards.

>  the device list still doesn't receive any updates.

On Windows this actually works!

ben-wes:
> On Windows this actually works!

that's cool! if it also works for linux, i might have to reconsider my OS decisions. :)

(reopening midi is included after a force-push to that PR)

Spacechild1:
>  if it also works for linux, i might have to reconsider my OS decisions. :)

This sounds more like a portaudio bug. I don't see why reinitializing the library shouldn't update the device list.

ben-wes:
> This sounds more like a portaudio bug. I don't see why reinitializing the library shouldn't update the device list.

i think so, too. seems wrong that it gets completely reinitialized and still "remembers" the old device list. anyway - i'll close this issue here since we're now dealing with another issue ...

PR #2685, opened 2025-07-16
Add "Rescan Devices" button to tabbed audio preferences

### LINKED ISSUE #2684: Rescanning audio devices not possible in tabbed preferences
system: Pd0.56-0test2, macOS, arm64

Not really a bug - but the inconsistency might be irritating for users:

The "Rescan devices" button is only available in the audio settings, not in the audio tab of the preferences.

ben-wes:
this makes the "Rescan Devices" button part of the audio settings frame. that way, the layout is not as nice, but the button is present in the audio dialog and the tabbed preferences for audio.

the ui update mechanism is obviously (and unfortunately) more complex here since it checks what to update on the tcl side (instead of just calling `glob_audio_properties(0)`).

* closes #2684
* makes sure that we don't reinitialize portaudio in "callback" mode when audio is running (would deadlock)
* prevents the console flooding reported in #2683 (but doesn't solve the device list update issue)

### DIFF
--- a/src/s_audio.c
+++ b/src/s_audio.c
@@ -788,21 +788,26 @@ void glob_audio_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
     for which we must deinit and reinit the library to see new devices. */
 void glob_rescanaudio(t_pd *dummy)
 {
-#ifdef USEAPI_PORTAUDIO
-    t_audiosettings as;
-    sys_get_audio_settings(&as);
-    if (as.a_api == API_PORTAUDIO)
+    if (sched_get_using_audio() == SCHED_AUDIO_CALLBACK)
     {
-        int restart = canvas_dspstate;
-        sys_close_audio();
-        pa_reinitialize();
-        if (restart)
-            sys_reopen_audio();
+        pd_error(0, "Cannot refresh device list in 'callback' mode when audio is running. "
+            "Please turn off DSP and try again.");
+        return;
     }
-    glob_audio_properties(0);
+#ifdef USEAPI_PORTAUDIO
+    if (audio_nextsettings.a_api == API_PORTAUDIO)
+    {
+        int was_open = audio_isopen();
+        if (was_open) sys_close_audio();
+        pa_reinitialize();
+        if (was_open) sys_reopen_audio();
+    }
 #else
-    post("skipping rescan because it's not defined for this API");
+    post("Skipping rescan because it's not defined for this API");
 #endif
+    sys_gui_audiopreferences();
+        /* refresh audio dialog (if it's open) */
+    pdgui_vmess("::dialog_audio::refresh_ui", "");
 }
 
 void sys_listdevs(void)
--- a/src/s_midi_pm.c
+++ b/src/s_midi_pm.c
@@ -125,8 +125,10 @@ void sys_close_midi(void)
 
 void sys_reinit_midi()
 {
+    sys_close_midi();
     Pm_Terminate();
     Pm_Initialize();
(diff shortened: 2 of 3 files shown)


src/s_audio.c
```diff
@@ -788,21 +788,25 @@ void glob_audio_dialog(t_pd *dummy, t_symbol *s, int argc, t_atom *argv)
...
-        sys_close_audio();
-        pa_reinitialize();
-        if (restart)
-            sys_reopen_audio();
+        pd_error(0, "Cannot refresh device list in 'callback' mode when audio is running. "
+            "Please turn off DSP and try again.");
+        return;
     }
-    glob_audio_properties(0);
+    was_open = audio_isopen();
```
Spacechild1: I would only close and reopen audio right before/after `pa_reinitialize()` because it's only necessary there.

src/s_midi_pm.c
```diff
@@ -125,6 +125,7 @@ void sys_close_midi(void)
 
 void sys_reinit_midi()
 {
+    sys_close_midi();
     Pm_Terminate();
     Pm_Initialize();
```
Spacechild1, line 130: You also need to reopen MIDI, otherwise MIDI would stay closed if you cancel the dialog!

Spacechild1:
> (but doesn't solve the device list update issue)

It does on Windows at least! (Didn't test on macOS or Linux.)

src/s_midi_pm.c
ben-wes: indeed that didn't make sense. i only tested with confirming the dialog. thanks! - fixed.

Spacechild1:
Can you maybe also add the missing "callback" scheduler check in `glob_rescanaudio`? You had this in your original PR, but Miller forgot about it. See my comment https://github.com/pure-data/pure-data/commit/524a079a339b9d09bc426a11ff411400bb8c5a91#r162213762.

src/s_audio.c
ben-wes: should be fixed. thanks!

ben-wes:
> Can you maybe also add the missing "callback" scheduler check

should be here:
https://github.com/pure-data/pure-data/pull/2685/files#diff-92f4680bf795c4719b7bfec186f43b459c4c3fdd726b8378dd883c73110389c8R791-R796

Spacechild1:
D'oh, I need to get my glasses!

I just went ahead and updated your PR description.

danomatika:
I just tested and it works nicely. I feel like we could improve the workflow such as not having to close/reopen the dialog as much, but it's a great improvement form having to close/reopen Pd itself. Thanks!

umlaeute:
> I feel like we could improve the workflow such as not having to close/reopen the dialog as much,

you shouldn't need to manually close/re-open the dialog. instead the dialog will re-create itself (which creates a short flicker and could be improved somewhen in the future).

i cannot actually test this (as i don't have changing devices), but I'm pretty sure it works as advertised...

danomatika:
I meant more regarding the need to hit the Ok button on mac to have the device change applied. This is not an issue for you since you have the Apply button on Linux.

umlaeute:
ah i see.
but that would be another issue.

danomatika:
That's what I meant: it should be noted. I wouldn't be adverse to showing the apply button when the audio tab is shown on mac or simply show it always. Ideally to match Apple HIG, selecting the audio device would change it automatically however plenty of other software has an "Apply" metaphor on the platform so in this case it would save the step of closing the dialog to affect the device change. I can blame myself since I was the one to bring that platform distinction to the tk gui.

PR #2688, opened 2025-07-17
Update portmidi script and Portmidi to v2.0.6

danomatika:
This is #2305 + PortMidi v2.0.6. I'm mainly opening the PR to see that the CI builds on Linux and Windows.


danomatika:
Doh there are bunch of line ending check fails. I have to get going to work (busy today). Can someone check this?
@umlaeute etc? Maybe the script should fix line endings when copying the sources.

danomatika:
Oh, Miller's makefile.mac needs to remove references to the following:

~~~
pm_mac/finddefault.c
pm_mac/readbinaryplist.c
~~~

danomatika:
Ok, only `makefile.mac` needed that change.

I also noticed that it builds `porttime/ptmacosx_cf.c` while portmidi/Makefile.am builds `porttime/ptmacosx_mach.c`. I suppose there should be no real difference so probably not an issue.

ben-wes:
> Can someone check this?

2357e5861c45216a742816e4a277fb48b95580d6 ... but this is obviously just fixing the imported files (hopefully), not touching any scripts.

umlaeute:
> I'm mainly opening the PR to see that the CI builds on Linux

note that we are not using PortMidi on linux (just plain ALSA resp OSS)


> Doh there are bunch of line ending check fails. I have to get going to work (busy today). Can someone check this? @umlaeute etc? Maybe the script should fix line endings when copying the sources.

should we just ignore style errors in the portaudio/portmidi sources?

umlaeute:
i've fixed line-endings and permissions in this repository, updated the `update.sh` script to do this automatically, also and forwarded the fixes to upstream https://github.com/PortMidi/portmidi/pull/95

i then thought about upstreaming [our local patches](https://github.com/pure-data/pure-data/tree/b8b479f549e1bf192c6e8e020f733b653b396824/portmidi/patches), but it seems it got broken with the v2.0.6 import.

the upstream code only does an `#ifdef RATE_LIMIT` check, and we are ensuring that `RATE_LIMIT` is always defined to something (1, 0, anything)

---
**EDIT** @danomatika , @Spacechild1 as i do not feel comfortable with updating the `RATE_LIMIT` patch myself, maybe you would want to have a look?

i also wonder, why this has never been upstreamed (possibly because it originally targeted a portmidi version that was not available in a public git repository)

danomatika:
The whole backstory is from 2019 (see #581). The issue from what I remember is that they did *not* want to allow for it to be changeable at compile time via a define even though there is one, hence the patch. You could try convincing them again.

Since it's a simple `#define LIMIT_RATE 1`... there is no way to override it via `-DLIMIT_RATE 0` without the ifndef block.

Spacechild1:
> The issue from what I remember is that they did not want to allow for it to be changeable at compile time via a define

I just had a look at the PR thread and Roger seemed to be open for it: https://github.com/pure-data/pure-data/issues/581#issuecomment-494827499

I actually promised to send a patch: https://github.com/pure-data/pure-data/issues/581#issuecomment-494830059. Maybe the issue is that I never actually sent it? I can't remember because that was 6 years ago. (Time flies!) But now that development has moved to GitHub, we simply make a PR :)

danomatika:
> i've fixed line-endings and permissions in this repository, updated the `update.sh` script to do this automatically, also and forwarded the fixes to upstream [PortMidi/portmidi#95](https://github.com/PortMidi/portmidi/pull/95)

Thanks, looks good.

> the upstream code only does an `#ifdef RATE_LIMIT` check, and we are ensuring that `RATE_LIMIT` is always defined to something (1, 0, anything)

As mentioned, the upstream code simply sets the define with no way to override it. I am comfortable with this tiny but required change for us, hence the patch.

> **EDIT** @danomatika , @Spacechild1 as i do not feel comfortable with updating the `RATE_LIMIT` patch myself, maybe you would want to have a look?

It's fine with v2.0.6. I just ran the script again and the patched log message says it worked ok. The change looks fine in the c file.

danomatika:
Hah it could be that we just forgot and the impetus slipped. Yeah, do a PR... it's such a tiny change and doesn't need to be documented or advertised ATM.

danomatika:
@umlaeute The script's patching mechanism will try the all patches in the `patches` directory and only actually apply the ones that don't fail. This makes it possible to add patches to newer versions while keeping old ones around. We just have to read the output to notice if anything actually fails when it shouldn't.

Spacechild1:
> it's such a tiny change and doesn't need to be documented or advertised ATM.

When looking at the patch, we do quite a bit more than just add a `#ifndef`. Some parts of the original code didn't compile if `LIMIT_RATE` is 0 because they forgot a couple of `#if`s. Nevertheless, the patch is pretty small and I will try to make a PR.

danomatika:
ah could be... I can update it.

danomatika:
Actually, the ifdefs are fine in v2.0.6 and it's only on [pmmacosxcm.c line 719](https://github.com/PortMidi/portmidi/blame/master/pm_mac/pmmacosxcm.c#L719). This change was made 4 years ago after our initial discussion, so I think the patch is fine as it is. If we still need to support a 4+ year older version, we can make a new patch.

umlaeute:
that's what I was trying to say :-)

Spacechild1:
> When looking at the patch, we do quite a bit more than just add a #ifndef

Ah, sorry, the PR contains a patch of a patch and I got confused about the changes...

> it's only on [pmmacosxcm.c line 719](https://github.com/PortMidi/portmidi/blame/master/pm_mac/pmmacosxcm.c#L719)

It's also here: https://github.com/PortMidi/portmidi/blame/master/pm_mac/pmmacosxcm.c#L776

And both are actually wrong because it should be `#if LIMIT_RATE` and not `#ifdef LIMIT_RATE` :)

danomatika:
Oi, you're right. In this case, it would be better for us to drop the `-DLIMIT_RATE=0`from the build with the current patch.

Spacechild1:
> In this case, it would be better for us to drop the -DLIMIT_RATE=0from the build with the current patch.

Why? How should we disable the rate limiting otherwise?

I think we just have to change these two lines and update the patch. It's still a trivial fix.

danomatika:
Yes, you're right, I wasn't thinking it needs to be `#if`. It's such a trivial fix but it seems I have to do it. In any case, the branches should now have the updated patch and sources.

danomatika:
Ok, PR submitted upstream: https://github.com/PortMidi/portmidi/pull/96

Spacechild1:
Cool, thanks!

danomatika:
Closing this in favor of #2930 and PortMidi v2.0.8

PR #2690, opened 2025-07-17
stable (already merged) development branch

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/2666

---

### GUI
- update both standalone audio dialog and preferences on device-rescan
   (rather than only updating one of both; it's possible to have both open...)
- disable device popup menu, if there's nothing to select from

### DIFF
--- a/tcl/dialog_audio.tcl
+++ b/tcl/dialog_audio.tcl
@@ -94,6 +94,12 @@ proc ::dialog_audio::state2widgets {state args} {
         set s "disabled"
     }
     foreach w $args {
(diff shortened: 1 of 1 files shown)


tcl/dialog_audio.tcl
ben-wes, line 1: damn. i thought i had checked this, sorry. with tabbed preferences open, the audio settings won't open. but the other way around is possible (although it's not obvious here on macos since the audio settings get hidden behind the pd window in that case). anyway: thanks.

PR #2692, opened 2025-07-19
reflect palette in gui, use for cords_to_foreground

### LINKED ISSUE #2686: Issue with Pd colors and "Highlight active cord while connecting".
This setting changes the color of the cord connection, which works correctly for a black foreground but loses effect if you have a greyish one.

When making the connection, it turns all connections to black, so it doesn't respect the current foreground color.

Not sure what is the best way to change the cord color in this case. I guess we can take the background color and use a relative color to that one.

ben-wes:
this is adding a tcl interpolation function for an attenuated foreground color and reflects the palette in the gui, updated by the `colors` message.

* closes #2686 
  * fixes the hardcoded colors for the cords_to_foreground feature, which obviously don't align well with custom palettes.

known issues:
the redundancy of setting the initial color values on the gui side is a bit ugly. it might be an option to send an initial `colors` message? but i didn't want to do this here

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -2371,10 +2371,13 @@ void glob_colors(void *dummy, t_symbol *fg, t_symbol *bg, t_symbol *sel,
         pd_error(0, "skipping color update");
         return;
     }
-    THISGUI->i_foregroundcolor = c_fg;
-    THISGUI->i_backgroundcolor = c_bg;
-    THISGUI->i_selectcolor = c_sel;
-    THISGUI->i_gopcolor = c_gop;
+        /* forward palette to gui */
+    pdgui_vmess("::pd_colors::set_palette", "kkkk",
+        THISGUI->i_foregroundcolor = c_fg,
+        THISGUI->i_backgroundcolor = c_bg,
+        THISGUI->i_selectcolor = c_sel,
+        THISGUI->i_gopcolor = c_gop
+    );
     for (gl = pd_this->pd_canvaslist; gl; gl = gl->gl_next)
         glist_dorevis(gl);
 }
--- a/tcl/pdtk_canvas.tcl
+++ b/tcl/pdtk_canvas.tcl
@@ -510,13 +510,14 @@ proc ::pdtk_canvas::cleanname {name} {
 
 proc ::pdtk_canvas::cords_to_foreground {mytoplevel {state 1}} {
     if {$::pdtk_canvas::enable_cords_to_foreground} {
-        set col black
+        set col $::pd_colors::palette(foreground)
         if { $state == 0 } {
-            set col lightgrey
+            set col [::pd_colors::interpolate \
+                $::pd_colors::palette(background) $::pd_colors::palette(foreground) 0.17]
         }
         foreach id [$mytoplevel find withtag {cord && !selected}] {
-            # don't apply backgrouding on selected (blue) lines
-            if { [lindex [$mytoplevel itemconfigure $id -fill] 4 ] ne "blue" } {
+            # don't apply backgrounding on selected lines
+            if { [lindex [$mytoplevel itemconfigure $id -fill] 4 ] ne $::pd_colors::palette(selected) } {
                 $mytoplevel itemconfigure $id -fill $col
             }
         }
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -241,6 +241,36 @@ namespace eval ::pdgui:: {
 }
 
 #------------------------------------------------------------------------------#
+# color palette and utilities
+
+namespace eval ::pd_colors {
+    variable palette
+    array set palette {
+        foreground "#000000"
+        background "#FFFFFF"
+        selected   "#0000FF"
+        gop        "#FF0000"
+    }
+}
+
+proc ::pd_colors::set_palette {fg bg sel gop} {
+    set ::pd_colors::palette(foreground) $fg
+    set ::pd_colors::palette(background) $bg
+    set ::pd_colors::palette(selected) $sel
+    set ::pd_colors::palette(gop) $gop
+}
+
+proc ::pd_colors::interpolate {color1 color2 amount} {
+    set amount [expr {max(0, min(1, $amount))}]
+    scan $color1 "#%2x%2x%2x" r1 g1 b1
+    scan $color2 "#%2x%2x%2x" r2 g2 b2
+    set r [expr {int($r1 + ($r2 - $r1) * $amount)}]
+    set g [expr {int($g1 + ($g2 - $g1) * $amount)}]
+    set b [expr {int($b1 + ($b2 - $b1) * $amount)}]
+    return [format "#%02x%02x%02x" $r $g $b]
+}
+
+#------------------------------------------------------------------------------#
 # coding style
 #
 # these are preliminary ideas, we'll change them as we work things out:


porres:
> do we want to add derived values to the palette?

- I think it's a good idea, to be be used in [cnv], cord highlighting and pd console window.

> the redundancy of the default values on the gui side is a bit ugly

- Doesn't seem like a big deal to me

> is this the proper way (and context) to reflect these colors?

- No idea, but as long as we're doing this, it'd be good to (as soon as colors are set) also configure colors for other items on the GUI side like the pd console window (see https://github.com/pure-data/pure-data/issues/2698 )

porres:
> (and the pd console window)

I just realized that the text window of [text] is another element we missed, and this is taken care of in pdtk_textwindow.tcl , see
https://github.com/pure-data/pure-data/blob/master/tcl/pdtk_textwindow.tcl#L28

It's hardcoded to white and, for the record, if we were to also hardcode the text caret and text color we simply need to add "-foreground black -insertbackground black" before "-background white"

I can't seem to find how to set the text caret for the boxes in a canvas though (EDIT: it was easy and I have a new PR for it)

porres:
The text window background is set in 
https://github.com/pure-data/pure-data/blob/master/tcl/pdtk_textwindow.tcl#L28

It's hardcoded to white and, for the record, if we were to also hardcode the text caret and text color we simply need to add "-foreground black -insertbackground black -background white"

I can't seem to find how to set the text caret for the boxes in a canvas though

src/g_canvas.c
```diff
@@ -2346,6 +2347,20 @@ static unsigned int normalize_color(t_symbol *s)
...
+    int r1 = (col1 >> 16) & 0xFF;
+    int g1 = (col1 >>  8) & 0xFF;
+    int b1 =  col1        & 0xFF;
+    int r2 = (col2 >> 16) & 0xFF;
+    int g2 = (col2 >>  8) & 0xFF;
+    int b2 =  col2        & 0xFF;
+    unsigned int r = r1 + (r2 - r1) * factor;
+    unsigned int g = g1 + (g2 - g1) * factor;
+    unsigned int b = b1 + (b2 - b1) * factor;
+    return (r << 16) | (g << 8) | b;
```
umlaeute: could we still mask the 3 channels?
the current solution will bleed into other channels if the factor is out-of-bounds.

tcl/pdtk_canvas.tcl
```diff
@@ -508,15 +519,23 @@ proc ::pdtk_canvas::cleanname {name} {
     return $name
 }
 
+proc ::pdtk_canvas::set_color_palette {fg bg sel gop att} {
```
umlaeute: i wonder whether we shouldn't use explicit names like so:
```
::pdtk_canvas::set_color_palette foreground #000000
::pdtk_canvas::set_color_palette background #FFFFFF
```

the advantage would be, that it can be easily expanded to future "colortypes".

the advantage of  your proposal is, that it kind of mimicks the `colors` message on the core side.
(but then it adds a new color, so it breaks this symmetry; probably it is indeed better to keep the message/proc strictly symmetric and do the interpolation on the GUI side)

tcl/pdtk_canvas.tcl
```diff
@@ -508,15 +519,23 @@ proc ::pdtk_canvas::cleanname {name} {
...
+    set ::pdtk_canvas::colors(fg) $fg
+    set ::pdtk_canvas::colors(bg) $bg
```
umlaeute: the older i get, the more i prefer long variable names. e.g. `foreground` instead of `fg`.

tcl/pdtk_canvas.tcl
```diff
@@ -508,15 +519,23 @@ proc ::pdtk_canvas::cleanname {name} {
...
+    set ::pdtk_canvas::colors(sel) $sel
+    set ::pdtk_canvas::colors(gop) $gop
```
umlaeute: and finally i wonder whether these should go into the `pdtk_canvas` namespace, or into a more global namespace. (at least the setting proc itself)

src/g_canvas.c
ben-wes: thanks for the feedback! the reasoning here was: masking (wrapping) is also not very intuitive. so i thought we'd just rely on `0 <= factor <= 1` - but when writing this here now i guess that clamping would be the *right* solution?

tcl/pdtk_canvas.tcl
ben-wes: feels to me like this is part of more general decisions concerning the "gui awareness" of the core? i honestly don't feel qualified to contribute to that. mirroring the `colors` message seemed logical - but then it should probably really not add more and let the gui take care of handling colors derived from the palette itself.

tcl/pdtk_canvas.tcl
ben-wes: certainly the better idea if colors are later also used in non-canvas contexts (are they?). currently not sure if that will require a new namespace or does `pd_guiprefs` seem right?

src/g_canvas.c
umlaeute: i think so.

tcl/pdtk_canvas.tcl
umlaeute: indeed (i agree to everything you say)

tcl/pdtk_canvas.tcl
umlaeute: well, it seems there are people clamouring about the PdWindow not honouring the colors, so...

`pd_guiprefs` is definitely wrong. this is the namespace for the preference system itself, not for the preferred values.

src/g_canvas.c
ben-wes: moved this to tcl for now and included the clamping there ...

tcl/pdtk_canvas.tcl
ben-wes: kept the mirrored function setting the whole palette. no preferences there though.

tcl/pdtk_canvas.tcl
ben-wes: used the full names for the tcl palette.

EDIT: and not sure about the `select` keyword ...

tcl/pdtk_canvas.tcl
ben-wes: moved code to `pd-gui.tcl` and new `pd_colors` namespace.

tcl/pdtk_canvas.tcl
umlaeute: "selection"? "selected"?

ben-wes: selected it is.

porres:
Oh, I this is doing things I was attempting in other PRs, so I'm closing those. I also see you're acting on https://github.com/pure-data/pure-data/issues/2700 and edited the first comment.

And I also also see this is not a draft anymore. I say fixing cord highlighting is important for the next release, as well as the others. I hope this gets considered.

ben-wes:
removed the commit changing `colors` to `color` message and swapping `foreground` and `background` arguments. i assume that these shouldn't be changed anymore after the release of Pd0.56-0.

porres:
Is this good for 0.56-1? :) I say fixing cord highlighting is an important bug fix. I hope this gets considered - ping @millerpuckette

millerpuckette:
I worry about how this is going to complicate getting the gtk version 
off the ground...
is it possible to put the logic in the C code instead of the tk?

porres:
> is it possible to put the logic in the C code instead of the tk?

Sure. If I remember correctly, that's how I first did this on another PR

umlaeute:
please do not.

right now, the core does not know *anything* about the "cords to foreground" - it is entirely implemented on the GUI side.
if a prospective GTK version wants to have the same functionality, it can easily implement it all by itself. or ignore it.


---

i'm really opposed to moving GUI-only stuff over to the core, only to save a bit of implementation for a potential GUI written in another toolkit. in the end this all boils down to: "if you want to write a GUI in GTK/JUCE/FooWdiget, you just have to implement your own TclTk layer for your fancy toolkit". but:
- implementing Tcl/Tk (or just Tk) is a non-trivial operation, and a very easy way to introduce very serious bugs
- these toolkits already have their own idiosyncratic ways to handle things
- adding another (complext) layer on top of an existing toolkit is *very likely* to destroy all the benefits you ever get from the new toolikt. if your brandnew JUCE-based GUI looks identical to the 90ies-style Tcl/Tk GUI **and** performs worse (because of the additional translation layer, that effectively circumvents all performance improviements), why would you want to re-implement such a GUI in the first place?

umlaeute:
because @porres is re-iterating this:
> Is this good for 0.56-1? :) I say fixing cord highlighting is an important bug fix.

i think not.

- there are "millions" of background/foreground combinations where the currently hard-coded color for backgrounded cords is perfectly visible.
- if you happen to pick a combination where this is not true, you can easily switch off the "Highlight active cord" functionality in the GUI properties
- if you cannot live without the "hightlight active cord", you can probably live with the fact that you do not see the already-connected (non-active) cords while connecting

i totally agree that it is ugly, and annoying.
and that it should be fixed.
i just don't see how this is an **important** issue. instead i would label it as "very minor".

Issue #2674, opened 2025-07-13
More issues regarding setting colors with the new "colors" message to pd


porres:
Some more issues not yet fixed.
- [x] The text editing bar marker (aka 'caret') is always black, should be the foreground color (fixed with https://github.com/pure-data/pure-data/pull/2694 )
- [x] Changing colors don't recreate the windows in the current order.
- [ ] The Pd console window does not change colors (we may also have to consider how to deal with the red error messages). See https://github.com/pure-data/pure-data/issues/2698 and https://github.com/pure-data/pure-data/issues/2700
- [ ] By default, iemguis shoud take background and foreground color the same as Pd's (also discussed in https://github.com/pure-data/pure-data/issues/2639 )
- [ ] Also allow arrays to be created with Pd's color (first we need to be able to use hex colors for DS, which is being worked on in this PR https://github.com/pure-data/pure-data/pull/2670 )
- [ ]  Allow both iemguis and arrays to react to color changes.
- [ ] not sure what this color is: https://github.com/pure-data/pure-data/blob/66e038a2a0fd45ad9c3a9b5c4e5e81350dbed602/src/g_graph.c#L780 but maybe we should use Pd's colors?
- [ ] Text windows from the [text] object also need to apply to Pd's colors
- [ ] https://github.com/pure-data/pure-data/issues/2686

sebshader:
https://github.com/pure-data/pure-data/pull/196#issuecomment-3094707394

PR #2694, opened 2025-07-21
fix text caret with foreground color

porres:
This applies Pd's foreground color to the text caret.

Right now it defaults to black, so it becomes invisible with a black background

This issue is mentioned amongst others in https://github.com/pure-data/pure-data/issues/2674

-------

It was a small change that required also sending THISGUI->i_foregroundcolor (besides THISGUI->i_backgroundcolor) to "pdtk_canvas_new" and use "-insertbackground" to configure it

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -1917,10 +1917,11 @@ void canvas_vis(t_canvas *x, t_floatarg f)
                 case we're talking to an older GUI version (so that
                 pureVST can work with Pd 0.55 as its GUI) */
             if (THISGUI->i_backgroundcolor != 0xFFFFFF)
-                pdgui_vmess("pdtk_canvas_new", "^ ii si k", x,
+                pdgui_vmess("pdtk_canvas_new", "^ ii si kk", x,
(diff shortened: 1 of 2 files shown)


porres:
Some screenshots 

Screenshot 2025-07-20 at 23 26 22
Screenshot 2025-07-20 at 23 26 39
Screenshot 2025-07-20 at 23 27 17

millerpuckette:
OK, ok, _really_ the last one :)

umlaeute:
i'm pretty sure that *this* is not the best solution.

- instead of setting the colour for each new canvas, we should just tell the GUI that the insertbackground has a given colour (or let it figure out the correct one by itself)
- afaik, the canonical way to set an option for *all* new widgets would be `option add *insertBackground {olive drab}`

i will submit a PR once 0.56 is out.

millerpuckette:
OTOH everything else gets its colors set by Pd telling TK what color to use for each object... keeping things local in this way keeps the GUI layer thin (and so, less hard to envision swapping out for GTK or QT or whatnot).

Issue #2696, opened 2025-07-21
Inconsistent color argument order in pd colors compared to IEM objects


charlesneimog:
Hi, the feature of colors is very cool hehe.

But I thing that it is weird to use 

`pd colors <foreground> <background> <selection_colors>`

instead of 

`pd colors <background> <foreground> <selection_colors>`

because the iem objects use `<background> <foreground> <selection_colors>`

<img width="1421" height="789" alt="Image" src="https://github.com/user-attachments/assets/4eec1eba-9d4d-415d-8423-15ca4aa89c1e" />

Obs: maybe I am missing something, then please close this!

porres:
> maybe I am missing something, then please close this!

I don't think so and, to be honest, I also think it makes better sense to start with background color!

ben-wes:
certainly a good moment to question this. ;)

(i agree that it's nice to strive for consistency - in which case `color` vs. `colors` would be another question. and i didn't see any of this.)

porres:
> certainly a good moment to question this. ;)

yeah, I brought this to the pd-list by the way

> in which case color vs. colors would be another question. and i didn't see any of this.

oops, I have to admit I noticed the inconsistency of "colors" x "color" but was quiet about it. And I'm usually the annoying guy claiming for consistency... so yeah, while we're at it, I also don't oppose. I guess I was so worried about thinking more crucial bugs and issues so I just let this go :) 

On the list, I'm also raising the question whether we should promote the oblivion of GOP color... I won't open another issue, but I think it makes sense to just keep it as an optional arg and not change it to selection color if we only have 3 args

porres:
> , I'm also raising the question whether we should promote the oblivion of GOP color...

the fact that charles didn't even mention 'GOP' color seems to point to the fact that this is indeed being promoted :)

porres:
> because the iem objects use background, foreground, selection_colors

actually background, foreground and label colors ;)

umlaeute:
one thing to note though, is that for the iemguis, the "background" colour is the colour of the object, whereas for Pd it is the colour of the window. 
just telling that the names might not mean the same thing

there are also different views, e.g that for both "background" is a *filled* area, whereas "foreground" is for a linish thing.

porres:
but also in this case, if iemguis has the same color of pd's background color, it would just blend in and reflect the canvas background.

umlaeute:
[...]

PR #2697, opened 2025-07-22
make pd and iemguis "color" message consistent

porres:
- closes https://github.com/pure-data/pure-data/issues/2696

- This incorporates changes from https://github.com/pure-data/pure-data/pull/2695 and also already updates the docs to reflect the changes in this PR


src/m_glob.c
```diff
@@ -45,7 +45,7 @@ void glob_fastforward(t_pd *ignore, t_floatarg f);
 void glob_settracing(void *dummy, t_floatarg f);
 void glob_vis(void *dummy, t_symbol *s);
 void glob_closesubs(void *dummy);
-void glob_colors(void *dummy, t_symbol *fg, t_symbol *bg, t_symbol *sel,
+void glob_color(void *dummy, t_symbol *fg, t_symbol *bg, t_symbol *sel,
```
ben-wes: not critical - but argument names should be switched here, too.

porres: good catch

porres:
I'll reboot this based on the new changes in the doc so it's not too messy

porres:
reboot in https://github.com/pure-data/pure-data/pull/2701

PR #2708, opened 2025-07-26
Add multi-instance pd support for fftw

timothyschoen:
Currently, multiple instances will share an fftw plan and buffers, with unchecked concurrent access. This PR splits the fftw plans per pd instance, and also locks around fftw plan creation (which only used to happen for the complex fft) because that cannot be done concurrently, according to:

http://www.fftw.org/fftw3_doc/Thread-safety.html

### DIFF
--- a/src/d_fft_fftw.c
+++ b/src/d_fft_fftw.c
@@ -46,7 +46,17 @@ typedef struct {
     fftwf_complex *in,*out;
 } cfftw_info;
 
-static cfftw_info cfftw_fwd[MAXFFT+1 - MINFFT],cfftw_bwd[MAXFFT+1 - MINFFT];
+typedef struct {
+    cfftw_info info[MAXFFT+1 - MINFFT];
+} cfftw_plans;
+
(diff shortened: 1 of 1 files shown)


timothyschoen:
I guess it would be preferable to not call pd_globallock on every entry into cfftw_getplan/rfftw_getplan, but I think that can only be safe if we make cfftw_ninstances an atomic_int. That would seem like a good solution to me, if pd's code guidelines allow that.

Edit: pd_ninstances would also need to be atomic_int for that to be safe

timothyschoen:
Ah, so that's why the double if-statement used to be there, so it doesn't constantly lock! Unfortunately that doesn't help for this solution as much, because I'm pretty sure doing that trick for `(pd_ninstances > cfftw_ninstances)` is still unsafe, since we write a new value to cfftw_ninstances inside the lock.

millerpuckette:
I merged [2780](https://github.com/pure-data/pure-data/pull/2780), which might make it possible  to make FFTW instance-safe more simply?

PR #2726, opened 2025-08-08
update cords for goprect message

### LINKED ISSUE #2724: goprect not fixing lines in opened GOP
If you have a GOP with an outlet and change the graph size with goprect, the connections are not "fixed". See below

Image

This works with "donecanvasdialog" though, see below

Image

When using the goprect in this case, I can easily crash Pd afterwards when opening/closing windoes, so I think it's related

ping @Spacechild1

ben-wes:
redraw cord lines on visible parent patch for `goprect` message.

closes #2724

### DIFF
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -475,9 +475,15 @@ static void graph_goprect(t_glist *x, t_symbol *s, int argc, t_atom *argv)
             x->gl_pixheight = 1;
     }
     if (x->gl_havewindow)
+    {
         glist_redraw(x);
+            /* possibly fix patch cords on parent canvas */
+        if (x->gl_owner && !x->gl_isclone && glist_isvisible(x->gl_owner))
+            canvas_fixlinesfor(x->gl_owner, &x->gl_obj);
+    }
     else
-        /* glist_redraw() won't remove "ghost objects" */
+            /* we have to redraw the parent canvas because glist_redraw()
+            won't remove "ghost objects" or update patch cords */
         canvas_redraw(glist_getcanvas(x));
 }
 


src/g_graph.c
```diff
@@ -479,6 +479,10 @@ static void graph_goprect(t_glist *x, t_symbol *s, int argc, t_atom *argv)
     else
         /* glist_redraw() won't remove "ghost objects" */
         canvas_redraw(glist_getcanvas(x));
+
+        /* always update cords on parent if visible. */
+    if (x->gl_owner && !x->gl_isclone && glist_isvisible(x->gl_owner))
+        canvas_fixlinesfor(x->gl_owner, &x->gl_obj);
```
Spacechild1: As you said, this is only an issue if the abstraction window is open, so I think the correct solution would be:
```
if (x->gl_havewindow)
{
    glist_redraw(x);
    /* possibly fix patch cords on parent canvas */
    if (x->gl_owner && !x->gl_isclone && glist_isvisible(x->gl_owner))
        canvas_fixlinesfor(x->gl_owner, &x->gl_obj);
}
else
    /* we have to redraw the parent canvas because glist_redraw() won't remove "ghost objects"
    or update patch cords */
    canvas_redraw(glist_getcanvas(x));
```
This seems to work fine for me.

(Note that `canvas_redraw(glist_getcanvas(x));` already redraws the whole parent canvas, including the patch cords.)

Spacechild1 (changes requested):

src/g_graph.c
ben-wes: ah, so the line fix was redundant that way - thanks! updated.

PR #2734, opened 2025-08-12
report xruns to user

Spacechild1:
Currently, xruns only flash the "Audio I/O Error" message. Now we also send the number of dropped samples to "pd-xrun", so the user can handle it.

Xruns are reported from the audio callback with the `sys_reportxrun` function, which just increments an atomic counter. In `sched_idletask` we atomically check and reset the counter and take the appropriate action.

### DIFF
--- a/src/m_sched.c
+++ b/src/m_sched.c
@@ -6,6 +6,7 @@
 
 #include "m_pd.h"
 #include "m_imp.h"
+#include "m_private_utils.h"
 #include "s_stuff.h"
 #ifdef _WIN32
 #include <windows.h>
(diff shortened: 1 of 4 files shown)


reduzent:
> This PR makes sure that the "Audio I/O Error" label is flashed whenever an xrun occurs 

That sounds really good, but raises the question: What does it indicate now? I mean, why is your PR necessary?

Spacechild1:
> What does it indicate now? I mean, why is your PR necessary?

Sorry, I was not very clear. The flashing of "Audio I/O Error" hasn't changed, the only new thing is that the dropped samples are reported to the user as a Pd message. I've updated the PR description accordingly.

PR #2735, opened 2025-08-13
Make bng's circle outline the same color as its flash.

myQwil:
If `THISGUI->i_foregroundcolor` is set to white, and a bng's background color is also set to white, the circle's outline becomes impossible to see. The tgl object, on the other hand, doesn't have this problem with its X outline because that uses the user-set foreground color.

### DIFF
--- a/src/g_bang.c
+++ b/src/g_bang.c
@@ -47,7 +47,7 @@ static void bng_draw_config(t_bng* x, t_glist* glist)
         xpos + x->x_gui.x_w - inset, ypos + x->x_gui.x_h - inset);
     pdgui_vmess(0, "crs ri rk rk", canvas, "itemconfigure", tag,
         "-width", zoom, "-fill", (x->x_flashed ? x->x_gui.x_fcol : x->x_gui.x_bcol),
-        "-outline", THISGUI->i_foregroundcolor);
+        "-outline", x->x_gui.x_fcol);
 
     sprintf(tag, "%pLABEL", x);
     pdgui_vmess(0, "crs ii", canvas, "coords", tag,
@@ -96,8 +96,6 @@ static void bng_draw_select(t_bng* x, t_glist* glist)
         col = lcol = THISGUI->i_selectcolor;
 
     sprintf(tag, "%pBASE", x);
-    pdgui_vmess(0, "crs rk", canvas, "itemconfigure", tag, "-outline", col);
-    sprintf(tag, "%pBUT", x);
     pdgui_vmess(0, "crs rk", canvas, "itemconfigure", tag, "-outline", col);
     sprintf(tag, "%pLABEL", x);
     pdgui_vmess(0, "crs rk", canvas, "itemconfigure", tag, "-fill", lcol);


porres:
I just wanna point that the same is of course true if the background is black and you have pd's default colors...

so it's nothin to do with changing Pd's colors.

But one way or another, it just really does make sense that the circle color is the same as the bang's foreground color.

myQwil:
That's true. The white-on-white scenario is what happened to me, but something similar could happen with any other color.

porres:
I thought this was an issue, and went ahead and made a PR hahaha, since it's better than yours, as it handles the select/unselect, I'll post it

porres:
anyway, even if you're going ahead and making a PR, it's nice to have an issue to relate it to I think

myQwil:
Select/unselect looks fine to me. Should the circle outline adopt the select color when the bng is selected? The tgl doesn't do that for the X outline.

porres:
> Should the circle outline adopt the select color when the bng is selected? 

Oh, I see you deleted altering the color of it. No, please, no... I think we should stick to what was happening before, which is change to the select color and back, for sure. That's how I did it

porres:
> I think we should stick to what was happening before, which is change to the select color and back, for sure. That's how I did it

ok, as discussed on the other PR, I see this has a point... it just still feels weird to me. The label color changes to the select color, so why not the foreground?

Also... the number in an atom number box also changes to the select color... but the, this is a big debate that would require bigger changes

porres:
for the record, some screenshots

Screenshot 2025-08-13 at 01 18 36

Screenshot 2025-08-13 at 01 18 51
Screenshot 2025-08-13 at 01 23 20

myQwil:
I'm not against the idea of the circle outline changing to selection color when selected. I'm just trying to stay consistent with the toggle element in its current form.

I think that if that change is made for one element, a similar change should be made for the other one as well.

porres:
well, I now think all foreground elements should change to the select color :) but that's something else

ben-wes:
i'm not sure about this, i admit. imho, this is trying to partly solve the problem that these gui objects currently don't adapt to the custom palette.

to me, the circle outline of the bang object is a similar kind of structural graphical element as the separating lines in hradio/vradio. i think it's fine if they align with the global palette's foreground color (and i also think it's good if these non-interaction-related gui parts adapt to the selection color - which should ideally include iolets as well).

the question then is how these objects can easily adapt to the palette. this has been discussed before, but maybe it would be good to agree on some kind of path for this ... i'm currently [against](https://github.com/pure-data/pure-data/issues/2639#issuecomment-3005307465) simply setting all colors of an iemgui to the palette like #2679 does.

0.56 already includes a [commit](https://github.com/pure-data/pure-data/commit/b38ef00ebd9761cdbc744f198ee0ff05a09ce59c) to handle a literal "default" color. this currently defaults to the iem default colors and not the custom palette though - so imho, that should be changed and it would be a good thing to do that rather soon.

the problem with the "default" symbol as a color is that previous pd versions will parse it as `#000000`. so it will require some transition period before actually writing this to pd files unless we want to cause a lot of confusion when users with different pd versions exchange patches. maybe we can get an idea of a reasonable transition period based on the [deken trends](https://deken.puredata.info/trends). :)

i might be wrong with my argument against automatically setting iemgui colors to the palette on creation - but otherwise, we'd need some global "Set all GUI object colors to palette" function (in the "Tools" menu?) - in the future, this could then set all to the literal "default".

and maybe, for future compatibility with older versions, a "Reset all GUI object colors to defaults" might also be useful (in this case, this would mean: black on white).

(sorry for this wall of text!)

myQwil:
Now that you mention it, hradio and vradio face the same issue as the circle outline. They should probably be part of this PR as well.

Having palette-based default colors on creation sounds reasonable, but this is still an issue for gui objects that aren't newly created. For example, I first noticed it by opening a patch with bang objects already inside of it.

ben-wes:
> this is still an issue for gui objects that aren't newly created

sure, but the same could be said for the object outline and iolets. it's just generally not very useful to have "light themed" objects (with their own bright background) on a "dark themed" background. i'm not strongly against it - but to me, it doesn't make sense to now introduce differing colors for these inner line elements and the actual object outline (and iolets).

it also affects the display of old patches, where people might have designed abstractions with black and white [bng]s that just display a colorful foreground when actually receiving a bang. this would change the appearance quite strongly, right? same for [*radio] ... i think it would be bad if all these lines here were also red - imho, we should preserve the option to use the object's foreground color as an actual signal color:

image

myQwil:
I see what you mean.

What if, instead of the outlines using the gui's foreground color, they instead used the inversion of the gui's background color? Maybe there could be an exception to that rule if the color is too close to grey.

ben-wes:
> the inversion of the gui's background color?

hmm ... that can get pretty colorful if people like to patch on a blue canvas or so. :)

i still think that the main challenge here is to find a good path for the next updates so that guis align with the global palette by default.

porres:
this is why I say it's nice to also open an issue, so we can discuss things on the issue itself and not flood the PR.

> hradio and vradio face the same issue as the circle outline.

True, but in this case, more than the bang's circle, it should adapt to the select color :) 

> imho, this is trying to partly solve the problem that these gui objects currently don't adapt to the custom palette

it's got nothing to do with the "custom palette"... it's a problem with the historical and default colors palette in Pd 0.55 and before. Just make the background black. Let's not mix things. But I see this problem might happen more if Pd color changes, but we're discussing that elsewhere and I see it as a separate thing.

The problem to me here is that there is a design flaw that we don't have control of the outline color in iemguis, so we can't have the background color the same as the outline color, which is bad for bangs and radio buttons. 

We can expand this issue and mention how inlets/outlets also become invisible and I've mentioned that already... 

Actually, same applies to label color of course, as we can't also have it the same as the background color :) but for white or any other light color labels, a dark canvas for contrast was also possible.   

So let's focus on outline as this is more of an issue for bang and radios... since they're not simply used by the actual rectangular outline, but also a part of the internal drawing. I see this indeed as a design flaw.

I guess the simplest fix is to have the same as foreground color for sure. But now that we're talking about radios, it'd be really bad and weird to have the actual outline with a different color, so it'd need to only the cell splitting lines...

> this would change the appearance quite strongly, right?

I can see some people getting unhappy :) so maybe we just don't do nothing... I was a bit annoyed, to be honest. But I think I would adapt with time, as it's the required improvement/fix.

myQwil:
I'll close this as it doesn't seem like the proper solution anymore.

porres:
> I'll close this as it doesn't seem like the proper solution anymore.

should we open an issue or a [discussion](https://github.com/pure-data/pure-data/discussions)?

myQwil:
> should we open an issue or a [discussion](https://github.com/pure-data/pure-data/discussions)?

Maybe it could be added to #2674
Something like "Outline colors more prone to blend in with background colors of gui objects"

porres:
> Maybe it could be added to https://github.com/pure-data/pure-data/issues/2674

no :) 

I'll just create one...

PR #2736, opened 2025-08-13
make bang's circle color the same as the bang's foreground color

porres:
The circle is not visible when the background color is the same as the foreground color - as in black background with the default black foreground. 

this is an improved version of https://github.com/pure-data/pure-data/pull/2735

Some screenshots.

Screenshot 2025-08-13 at 01 18 36
Screenshot 2025-08-13 at 01 18 51
Screenshot 2025-08-13 at 01 23 20

### DIFF
--- a/src/g_bang.c
+++ b/src/g_bang.c
@@ -47,7 +47,7 @@ static void bng_draw_config(t_bng* x, t_glist* glist)
         xpos + x->x_gui.x_w - inset, ypos + x->x_gui.x_h - inset);
     pdgui_vmess(0, "crs ri rk rk", canvas, "itemconfigure", tag,
         "-width", zoom, "-fill", (x->x_flashed ? x->x_gui.x_fcol : x->x_gui.x_bcol),
-        "-outline", THISGUI->i_foregroundcolor);
+        "-outline", x->x_gui.x_fcol);
 
     sprintf(tag, "%pLABEL", x);
     pdgui_vmess(0, "crs ii", canvas, "coords", tag,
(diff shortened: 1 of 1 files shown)


porres:
I wonder if we should make inlets of iemguis the same as the foreground as well...

myQwil:
I wouldn't say your PR is necessarily better. I want bng's circle outline to behave more similarly to the tgl's X outline. You want the circle outline to adopt the selection color when selected, despite that not being the case for tgl's X outline.

porres:
> I wouldn't say your PR is necessarily better. 

I thought you completely ignored the selection, I didn't see you deleted it. So I thought you had a bug, and that when you'd select and unselect, than you'd get the foreground color back.

But I also disagree on what you did, it's inconsistent to all other elements... so I think this is better because it retains the current and consistent behaviour.

myQwil:
Like I said, it's consistent with the toggle element. Does toggle's X change color when selected?

porres:
hmm, I see, you have a point after all :) I guess I'm just used to seeing the button area selected and it feels weird to me otherwise. Actually, having the foreground color also feels a bit unsettling and unnatural. So yeah, I guess I'm just not used to the idea yet...

porres:
it's one of those things you never realize or think too much, but now I think that not changing the foreground color to the select color is inconsistent.

The label color changes to the select color, so why not the foreground?

Also... the number in an atom number box also changes to the select color... but the, this is a big debate that would require bigger changes

porres:
ok, closing in favor of https://github.com/pure-data/pure-data/pull/2735

PR #2752, opened 2025-08-21
datastructures: fix saving unrealized dollar symbols to file

### LINKED ISSUE #2749: datastructures: realized versions of dollar symbols are wrongly saved to file
Pd current master (b940ac21607d0a8eb0af41ac9a295c15dea32c9f), Ubuntu 20.04

In the following example, the `$0-data` subpatch is populated with one instance of the `$0-element` structure.
Unfortunately, the `$0-element` member is encoded to file as `1003-element`, which can lead to subtle bugs and crashes.

For instance:
- open the patch twice
- on the second instance of the patch, clear `$0-data` and add it a new `$0-element` member (the needed stuff is in `[pd template-element]`)
- save the second instance; check the file: now the name of the member is e.g  `1004-element`
- close everything, reopen the patch you saved
- move the mouse above the `$0-data` window: here it crashes!
Explanation: `$0-data` loads with a `1004-element` member, which doesn't exists (only `$0-element`, aka `1003-element` exists).

```pd
#N struct 1003-element float x;
#N canvas 174 646 450 300 12;
#N canvas 666 687 450 300 template-element 1;
#X obj 19 120 struct \$0-element float x;
#X obj 19 169 drawnumber x;
#X obj 213 120 s pd-\$0-data;
#X msg 213 95 clear;
#X msg 19 95 add \$1;
#X obj 19 45 bng 20 250 50 0 empty empty empty 0 -10 0 12 #fcfcfc #000000 #000000;
#X obj 19 70 symbol pd-\$0-data;
#X connect 3 0 2 0;
#X connect 4 0 0 0;
#X connect 5 0 6 0;
#X connect 6 0 4 0;
#X restore 68 25 pd template-element;
#N canvas 135 179 595 408 \$0-data 1;
#X scalar 1003-element 0 \;;
#X restore 70 90 pd \$0-data;
```

Ant1r:
fixes #2749

currently a patch (here a simplified version of the example in #2749) is saved like this:
```pd
#N struct 1003-element float x;
#N canvas 174 646 450 300 12;
#N canvas 666 687 450 300 template-element 1;
#X obj 19 120 struct \$0-element float x;
#X restore 68 25 pd template-element;
#N canvas 135 179 595 408 \$0-data 1;
#X scalar 1003-element 0 \;;
#X restore 70 90 pd \$0-data;
```

You can see that:
- the template is initially created with name `$0-element` by `[struct \$0-element float x]`
- when opening the patch, Pd *realizes* the template name to `1003-element`
- when saving the patch, Pd wrongly saves the pre-declaration of the template **and** the scalar with the realized name:
`#N struct 1003-element float x;`
`#X scalar 1003-element 0 \;;`

This leads to various bugs and crashes, this is #2749 issue.

The goal of this PR is to save (and correctly reload) the file like this:
```pd
#N canvas 174 646 450 300 12;
#N struct \$0-element float x;
#N canvas 666 687 450 300 template-element 1;
#X obj 19 120 struct \$0-element float x;
#X restore 68 25 pd template-element;
#N canvas 135 179 595 408 \$0-data 1;
#X scalar \$0-element 0 \;;
#X restore 70 90 pd \$0-data;
```
Note that:
- the correctly *unrealized* template name is used everywhere
- the top-level canvas is declared before the template;
We need this, for being able to *realize* the dollar symbols of the template (the template name itself and the potential array template names) at load time: we need to resolve the value of `$0` and other `$` values (if the patch is used as an abstraction).


Ant1r:
ping @millerpuckette
this is fixing a crash.

Ant1r:
@millerpuckette may be good to have for 0.56.2

Ant1r:
yeah well it should never happen in the future, and in past versions it was crashing anyway...
so I think it's OK to keep a trace that something isn't right with this patch, and that it should be cleared and re-populated.

ben-wes:
ok. i was thinking of cases where one might be restructuring a patch and temporarily cut the [struct] to insert it elsewhere. but that's certainly a corner case (since the user would also need to focus the patch with scalars during this operation).

src/g_readwrite.c
```diff
@@ -698,31 +704,16 @@ static void canvas_savetemplatesto(t_canvas *x, t_binbuf *b, int wholething)
...
-        }
-        binbuf_addsemi(b);
+
+        t_binbuf *bb = template_get_creation_binbuf(template);
+        if (bb) {
+            binbuf_addv(b, "s", &s__N);
+            binbuf_addbinbuf(b, bb);
+            binbuf_addsemi(b);
+        } else bug("canvas_savetemplatesto template_get_creation_binbuf");
     }
```
millerpuckette: It would be easier for me if you adhered to pd's stylistic conventions so for instance instead of this:
```
       if (bb) {
            binbuf_addv(b, "s", &s__N);
            binbuf_addbinbuf(b, bb);
            binbuf_addsemi(b);
        } else bug("canvas_savetemplatesto template_get_creation_binbuf");
```
this:
```
        if (bb)
         {
              ....
        }
       else ....
```

millerpuckette: also, again conventional... "if (condition)"   -- spaces separate keywords like 'if' from what follows.

I can always go in a fix this by hand but then it causes confusion because github thinks your code didn't get adopted :)

Ant1r: @millerpuckette sorry for these rule violations!
I corrected these and a few other ones, hoping there aren't any more.

millerpuckette:
I made one minor tweak which I can't even explain clearly, otherwise all good, thanks.

PR #2757, opened 2025-08-25
preserve and dynamically handle "default" colors for iemguis

### LINKED ISSUE #2639: request: allow iemguis/array colors to be set with Pd's colors

ben-wes:
this is an attempt for dynamically handling the `"default"` color that was introduced with b38ef00. this draft should provide a basis for testing and further discussion.

* includes #2692
* includes #2702
* supersedes #2679
* closes #2639

to discuss (list probably not exhaustive):
1. ~is it reasonable to handle the `"default"` color via the `-1`/ `0xffffffff` sentinel value internally?~ EDIT: used additional flags now
2. ~similarly: is it a good approach to send additional "default" flags to the properties dialog (and back)?~ EDIT: assuming that the flags are a good idea, this question is obsolete
3. there is now an "Adapt to theme" ~button~ toggle in the iemgui properties, which represents and is applied for the selected color type (bg, fg, label). probably makes sense - but needs quite a few clicks to set all 3 colors that way.
4. for [cnv] and [vu], the background color adapts to the current palette with an interpolated value between bg and fg colors. this doesn't always work too well (see video). maybe, [vu] should generally have a dark background. not so sure about [cnv]'s background color ... a [different interpolation](https://observablehq.com/@zanarmstrong/comparing-interpolating-in-different-color-spaces) might give better results - but also seems overkill.
5. since `"default"` colors will appear all black (parsed to `0`) in Pd versions < 0.56 and obviously, patches with old discrete colors don't adapt to a custom palette, there is now a new submenu in the ~"Tools"~ patch window's "Edit" menu to set all iemguis of a patch to the old standard colors or the new adaptive colors (aka `"default"`). ~the functionality probably makes sense - but not so sure about the placement of the actions.~
6. these menu actions trigger a confirmation dialog - might not be needed, but since this is currently not undoable and potentially impactful, i thought it might be good to protect this action.
7. ~might also (or alternatively) be good to have a mode to generally set iemguis to adaptive colors on creation (or when loading a patch - at least if they have the standard colors).~ EDIT: added a global `iemgui-default-colors` message for this now.
8. also not sure about all these terms and where to use what (towards users) - adaptive, standard, default, theme, palette, GUI object, iemgui, ...

in some distant future, iemguis could/should be generally initialised with the `"default"` colors on creation.

---

here's a demo:

[test-patch-colors.pd.zip](https://github.com/user-attachments/files/21989645/test-patch-colors.pd.zip)

https://github.com/user-attachments/assets/1b5f9c35-1af4-421f-b0dd-58458ee03494

"Colors" section in properties dialog with additional "Adapt to theme" toggle:

image

### DIFF
--- a/po/Makefile.am
+++ b/po/Makefile.am
@@ -43,6 +43,7 @@ TCLFILES = \
     pd_docsdir.tcl \
     pd_guiprefs.tcl \
     pd_i18n.tcl \
+    pd_iemgui_colors.tcl \
     pd_menucommands.tcl \
     pd_menus.tcl \
     pdtcl_compat.tcl \
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -164,7 +164,16 @@ int iemgui_modulo_color(int col)
 
 unsigned int iemgui_getcolor_background(t_iemgui *x)
 {
-    return x->x_bcol_default ? THISGUI->i_backgroundcolor : x->x_bcol;
+    if (!x->x_bcol_default)
+        return x->x_bcol;
+    const char *class_name = class_getname(x->x_obj.te_pd);
+        /* interpolated color for cnv background */
+    if (!strcmp(class_name, "cnv"))
+        return interpolate_colors(THISGUI->i_backgroundcolor, THISGUI->i_foregroundcolor, 0.12);
+        /* fixed color for vu background */
+    else if (!strcmp(class_name, "vu"))
+        return 0x404040;
+    return THISGUI->i_backgroundcolor;
 }
 
 unsigned int iemgui_getcolor_foreground(t_iemgui *x)
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -164,20 +164,17 @@ int iemgui_modulo_color(int col)
 
 unsigned int iemgui_getcolor_background(t_iemgui *x)
 {
-    return (x->x_bcol == IEM_GUI_COLOR_DEFAULT_SENTINEL) ?
-        THISGUI->i_backgroundcolor : x->x_bcol;
+    return x->x_bcol_default ? THISGUI->i_backgroundcolor : x->x_bcol;
 }
 
 unsigned int iemgui_getcolor_foreground(t_iemgui *x)
 {
-    return (x->x_fcol == IEM_GUI_COLOR_DEFAULT_SENTINEL) ?
-        THISGUI->i_foregroundcolor : x->x_fcol;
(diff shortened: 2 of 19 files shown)


ben-wes:
> 1. is it reasonable to handle the `"default"` color via the `-1`/ `0xffffffff` sentinel value internally?

probably not if we want to support transparent colors in the future. in that case, `#FFFFFFFF` in the [planned](https://github.com/pure-data/pure-data/discussions/2648#discussioncomment-13756558) `#AARRGGBB` syntax would be a fully opaque white. but then again, `#00FFFFFF` is currently a fully opaque white. so would the alpha channel need to reflect transparency instead of opacity?

(side note: i'm not sure whether the feature discussion really makes sense in a PR. i thought it might be good to have some stuff testable here and technically improve it in the same context - but maybe there's a better place and maybe others already have a better solution in mind. if so, i'll gladly close this again.)

EDIT: added flags to the struct and removed the sentinel value mechanism with the last commit.

Ant1r:
> would the alpha channel need to reflect transparency instead of opacity

Seems a good idea at first sight!

ben-wes:
updated description and screen recording to current state and demo the `iemgui-default-colors` message.

porres:
> "Adapt to theme" button

should be a toggle, huh?

ben-wes:
> should be a toggle, huh?

it is! you can pass 0 or 1. EDIT: it's also visible in the video.

porres:
how does it look if it's on? so if you have it on and load the properties, is it active? Is there also a message to configure this?

ben-wes:
ah, sorry, my bad. shouldn't respond from the phone. i thought you were referring to the new message. but you're right - that button in the properties should be a toggle! will change that. EDIT: done

ben-wes:
> Is there also a message to configure this?

if you set colors to `default`, the effect is that this flag is active.

ben-wes:
> would the alpha channel need to reflect transparency instead of opacity

actually, it shouldn't be as problematic as i assumed since we can obviously adapt `iemgui_getcolorarg()` to differentiate between `#AARRGGBB` and `#RRGGBB` syntax. in the latter case, we can set colors to fully opaque ourselves (by ORing with `0xff << 24`).

Issue #2760, opened 2025-08-30
Bug when using array set and visual arrays with large lists: segmentation fault


zepadovani:
Hello,

I need to visualize list of numbers in a graphic array and I'm using 'array set' to do this. The lists are generated on every block, by a bang~ and a simple until/random list generator. Obviously, this is only to reproduce concisely the bug: I'm working with spectral information in my real situation here.

The idea works well, until... i turn DSP off: then I get (almost aways) a segmentation fault / crash.

This happens also if i try to use a metro with a very small interval to work as a "shutter" to control a [list append] object: if the metro is turned on, allowing [array set] to attempt to write lists in the array while I turn the DSP off, ocasionally (very frequently!) I will also have a segmentation fault and crash.

The bug does not happen if the array is not being visualized ([array define] or [table] when they are not open to be visualized).

I'm attaching three simple abstractions that show what I say above. (They are attached as a single zip)

1. bug_graph_array.pd => just tries to write a list of numbers (smaller or equal to the arrays size).
2. also_buggy.pd => uses a metro as a "shutter" to trigger lists stored in a list append object.
3. notbuggy.pd => when not visualizing, things work better.

Also, the bug does not happen in plugdata, although the visualization is strangely slower there.

I'm on macOS 15.6.1 / PD 0.56-1 / Apple M4

[visarray_set_bug.zip](https://github.com/user-attachments/files/22058388/visarray_set_bug.zip)

ben-wes:
reproduced this here (macos 15.7, m2, pd0.56-1) also with a static signal:

```
[bang~]
|
[tabwrite~ a]

[array define a 64]
```

... crashes with open array view as described above. seems to be some timing issue of the drawing update. with the static signal, i don't manage to reproduce it at [metro 10]. quickly happens with [metro 5] though (which is obviously still way slower than [bang~] at 48kHz and blocksize 64).

backtrace:

```
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=1, address=0x1c00)
    frame #0: 0x0000000183171368 libsystem_platform.dylib`_platform_memmove + 168
libsystem_platform.dylib`_platform_memmove:
->  0x183171368 <+168>: stp    q2, q3, [x0]
    0x18317136c <+172>: subs   x2, x2, #0x40
    0x183171370 <+176>: b.ls   0x18317138c    ; <+204>
    0x183171374 <+180>: stp    q0, q1, [x3]
[...]

PR #2762, opened 2025-09-01
Fix possible crash in sched_send_dacs()

Spacechild1:
`sched_idletask()` may turn off DSP or change audio settings. In this case, we must return early with `SENDDACS_NO`!

Fixes https://github.com/pure-data/pure-data/issues/2760, which is a serious regression in Pd 0.56 caused by 73acd66c8a04a151404ef4a6c2f0edae0214f10d.


Spacechild1:
ping @millerpuckette

PR #2763, opened 2025-09-02
remove ghost objects in gop window on 'goprect'

### LINKED ISSUE #2759: gprect message produces ghosts objects in subpatch
when a GOP subpatch is open, if we `[goprect(` it, then all its objects are re-created, but old drawings remain:

Image

Pd current master 65e1fedcc595f57491a1ac55cc31eabd67d54387
It doesn't happen on tag 0.56.0 .

```pd
#N canvas 704 66 291 171 12;
#N canvas 788 84 252 153 \$0-goprect 1;
#X text 128 5 comment1;
#X text 128 37 comment2;
#X obj 44 103 s pd-\$0-goprect;
#X msg 44 76 goprect 10 10 100 50;
#X connect 3 0 2 0;
#X coords 0 -1 1 1 100 50 1 10 10;
#X restore 38 41 pd \$0-goprect;
```

ben-wes:
this completely redraws the subpatch now if it's open

* closes #2759

### DIFF
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -477,14 +477,16 @@ static void graph_goprect(t_glist *x, t_symbol *s, int argc, t_atom *argv)
     }
     if (x->gl_havewindow)
     {
-        glist_redraw(x);
-            /* possibly fix patch cords on parent canvas */
+        canvas_redraw(x);
+            /* possibly fix patch cords and resized goprect on parent canvas */
         if (x->gl_owner && !x->gl_isclone && glist_isvisible(x->gl_owner))
+        {
+            graph_vis(&x->gl_gobj, x->gl_owner, 0);
+            graph_vis(&x->gl_gobj, x->gl_owner, 1);
             canvas_fixlinesfor(x->gl_owner, &x->gl_obj);
+        }
     }
     else
-            /* we have to redraw the parent canvas because glist_redraw()
-            won't remove "ghost objects" or update patch cords */
         canvas_redraw(glist_getcanvas(x));
 }
 


src/g_graph.c
```diff
@@ -477,10 +477,10 @@ static void graph_goprect(t_glist *x, t_symbol *s, int argc, t_atom *argv)
     }
     if (x->gl_havewindow)
     {
-        glist_redraw(x);
-            /* possibly fix patch cords on parent canvas */
+            /* redraw GOP window to clear any ghost drawings */
+        canvas_redraw(x);
         if (x->gl_owner && !x->gl_isclone && glist_isvisible(x->gl_owner))
-            canvas_fixlinesfor(x->gl_owner, &x->gl_obj);
+            glist_redraw(x->gl_owner);
```
Spacechild1: Just a question: why do we need `glist_redraw` instead of `canvas_fixlinesfor`?

Also, I would keep the comment that this is about fixing patch cords on the parent canvas. (Or is there more to it?)

ben-wes: the selected goprect isn't redrawn in the parent canvas that way if the subpatch is open - that's also the reason why i omitted the previous comment. but maybe it would be good to mention both (goprect redraw and line fixes). admittedly, it would be nice to only redraw what's actually needed in all cases ...

Spacechild1: > the selected goprect isn't redrawn in the parent canvas that way if the subpatch is open

AFAICT the behavior on the parent canvas (if the subpatch is open) was fine, the problem was just with ghost objects in the subpatch itself. So why do we need to change that line as well? Is it because we replace `glist_redraw` with `canvas_redraw` above?

To be honest, I don't really know the difference between `glist_redraw` and `canvas_redraw`. When should I use which? There are some comments for the former, but none for the latter...

ben-wes: i force pushed an update which might have less redundancy:
* it uses `canvas_redraw` for the subpatch to avoid these duplicated objects that were obviously caused by `glist_redraw`
* it only redraws goprect (via `graph_vis` toggle) and fixes the lines for the parent in case of the open subpatch

maybe it would be good to review the actual redraw mechanisms. it feels a bit buggy that `glist_redraw` caused these artifacts. `glist_redraw` also updates the goprect in the parent, which is why we need to do that additionally now when using `canvas_redraw` for the subpatch.

Ant1r:
I didn't really study the effects of this PR, but I think the reason for #2759 happens sooner in the code, likely in 
https://github.com/pure-data/pure-data/blob/65e1fedcc595f57491a1ac55cc31eabd67d54387/src/g_graph.c#L470
see https://github.com/pure-data/pure-data/issues/2759#issuecomment-3250960079

ben-wes:
> see [#2759 (comment)](https://github.com/pure-data/pure-data/issues/2759#issuecomment-3250960079)

thanks, @Ant1r! i should have read your report more carefully, which clearly states that this worked well before (in 0.56-0) and which also explains why this wasn't found when adding the cords fix.

closing this PR since it's a rather brute force way of addressing the problem. feel free to open another one!

PR #2764, opened 2025-09-02
free rtext on deletion to prevent later hit-test crash

### LINKED ISSUE #2761: regression : crash with mouse move
Testing automatonism (a vanilla only lib) can crash Pd.
step to reproduce : 
Install automatonism via deken, start main_patch(template) and add an object (euclid for example), save the patch. open the patch and move the cursor over the patch. 
Pd 0.56-1 will crash, pd 0.55-2 did not crash. 

you can also open this patch with automatonism installed via deken to trigger the crash.
[testautomatonism.pd.zip](https://github.com/user-attachments/files/22073750/testautomatonism.pd.zip)

ben-wes:
* closes #2761

... this seems to avoid stale rtexts during hit-testing by forcing to free them. not completely sure if there might be bad side effects though. tested with the example of the related issue.

### DIFF
--- a/src/g_graph.c
+++ b/src/g_graph.c
@@ -126,7 +126,7 @@ void glist_delete(t_glist *x, t_gobj *y)
     if (glist_isvisible(canvas))
         gobj_vis(y, x, 0);
     if (glist_getcanvas(x)->gl_editor && (ob = pd_checkobject(&y->g_pd)))
-        rtext = glist_getrtext(x, ob, 0);
+        rtext = glist_getrtext(x, ob, 1);
     if (x->gl_list == y) x->gl_list = y->g_next;
     else for (g = x->gl_list; g; g = g->g_next)
         if (g->g_next == y)


umlaeute:
hmm, on first glance this doesn't really look like a proper fix.

afaiu, what it does is it forces the creation of an rtext, just so we can delete it a couple of lines later.

ben-wes:
@umlaeute I assume so, too, and am about to close this one. I experimented quite a bit more some weeks ago, but I have to admit that I don't understand and can't control the full rtext life cycle. it would be great if all related issues (performance, ghost objects, deletion with intermediate additional creation, donecanvasdialog message, coords message,...) would be tackled for the next bugfix release. but I'm afraid i can't contribute the proper solutions there.

umlaeute:
as mentioned in #2761, it apparently *does* make all the memory issues go away, so I wouldn't close it yet (until we have a proper fix)

millerpuckette:
commended upon in https://github.com/pure-data/pure-data/issues/2761 - I think this is indeed a good idea whether or not it completely fixes the problem.

millerpuckette:
commended -> commented

Spacechild1:
Comments can be edited ;-)

ben-wes:
closing this PR, which was added as 98e76d85c74ed1731ca223ecb1904cc6729e9175

PR #2765, opened 2025-09-03
Add 'native' pipewire audio backend

charlesneimog:
Hello,

@umlaeute, could you take a look?

I’m proposing this because using JACK with PipeWire introduces some limitations, while ALSA and PortAudio give me even more issues.

With JACK, the main problems are:

- *Changing block size is not straightforward*: This is crucial for me: sometimes I need a reactive system, but other times—especially with [pd-upic](https://github.com/charlesneimog/pd-upic)), I need a huge block size to listen properly without rendering offline.

- *Output device switching is neither persistent nor fast*: (Possibly a configuration issue, but still.) I work mainly on laptops and frequently switch between Bluetooth headphones and audio interfaces. Globally, this is straightforward with tools like [pwvucontrol](https://flathub.org/apps/com.saivert.pwvucontrol), but for some reason, Pure Data isn’t recognized as a client when using ALSA or JACK. Either on `pwvucontrol` or `pavucontrol`, and `Helmut` is again not persistent.
 
This PR solves this issues for me!

But I yet have some problems to understand the `ringbuffers`, so this is my best provide a good solution for my problem hehe.

### DIFF
--- a/src/s_audio_paring.c
+++ b/src/s_audio_paring.c
@@ -107,8 +107,10 @@ long sys_ringbuf_init(sys_ringbuf *rbuf, long numBytes,
 ** Return number of bytes available for reading. */
 long sys_ringbuf_getreadavailable(sys_ringbuf *rbuf)
 {
-    long ret = atomic_int_load(&rbuf->writeIndex)
-        - atomic_int_load(&rbuf->readIndex);
+    long write_idx = atomic_int_load(&rbuf->writeIndex);
+    long read_idx = atomic_int_load(&rbuf->readIndex);
+
+    long ret = write_idx - read_idx;
     if (ret < 0)
         ret += 2 * rbuf->bufferSize;
     if (ret < 0 || ret > rbuf->bufferSize)
@@ -116,6 +118,7 @@ long sys_ringbuf_getreadavailable(sys_ringbuf *rbuf)
             "consistency check failed: sys_ringbuf_getreadavailable\n");
     return (ret);
 }
+
 /***************************************************************************
 ** Return number of bytes available for writing. */
 long sys_ringbuf_getwriteavailable(sys_ringbuf *rbuf)
--- a/src/s_audio_pipewire.c
+++ b/src/s_audio_pipewire.c
@@ -1,0 +1,605 @@
+/* Copyright (c) 1997-2003 Guenter Geiger, Miller Puckette, Larry Troxler,
+* Winfried Ritsch, Karl MacMillan, and others.
+* For information on usage and redistribution, and for a DISCLAIMER OF ALL
+* WARRANTIES, see the file, "LICENSE.txt," in this distribution.  */
+
+/* this file inputs and outputs audio using the pipewire API available on linux. */
+
+/* support for pipewire 0.3 by Charles K. Neimog <charlesneimog@outlook.com> */
+
+#ifdef USEAPI_PIPEWIRE
+
+#include "m_pd.h"
+#include "s_stuff.h"
+#include <stdio.h>
+#include <unistd.h>
+#include <stdlib.h>
+#include <string.h>
+
+#include <pipewire/pipewire.h>
+#include <spa/param/audio/format-utils.h>
+#include <spa/param/latency-utils.h>
+#include <spa/param/param.h>
+
+#include "m_private_utils.h"
+#include "s_audio_paring.h"
+
+/* enable thread signaling instead of polling (like JACK backend) */
+#if 1
+#define THREADSIGNAL
+#endif
+
+#define MAX_BUFFERS 4
+#define MAX_ALLOCA_SAMPLES  (16*1024)
+
+/* ------------------------------- Backend state ------------------------------- */
+typedef struct _pw_state {
+    int samplerate;
+    int blocksize;
+    int in_n_devices;
+    int out_n_devices;
+
+    unsigned in_channels;
+    unsigned out_channels;
+
+    unsigned desired_block;
+
+    struct pw_thread_loop *tloop;
+    struct pw_stream *stream_in;
+    struct pw_stream *stream_out;
+
+    struct spa_hook stream_in_listener;
+    struct spa_hook stream_out_listener;
+
+    /* sys_ringbuf-based FIFOs (interleaved t_sample audio) */
+    sys_ringbuf inring;         /* PW -> Pd */
+    sys_ringbuf outring;        /* Pd -> PW */
+    char *inbuf;                /* backing storage for inring */
+    char *outbuf;               /* backing storage for outring */
+
+#ifdef THREADSIGNAL
+    t_semaphore *sem;           /* wake Pd scheduler when audio thread has work */
+#endif
+
+    int running;
+    volatile int dio_error;     /* set on over/underruns */
+} t_pw_state;
+
+static t_pw_state pw_state;
+static int pw_inited = 0;
+
+/* ------------------ PipeWire callbacks ------------------ */
+static void pw_in_param_changed(void *data, uint32_t id, const struct spa_pod *param)
+{
+    (void)data; (void)id; (void)param;
+    logpost(0, PD_DEBUG, "[pipewire] input param changed");
+}
+
+static void pw_in_process(void *data)
+{
+    t_pw_state *st = (t_pw_state *)data;
+    if (!st || !st->stream_in) return;
+
+    struct pw_buffer *b = pw_stream_dequeue_buffer(st->stream_in);
+    if (!b) return;
+
+    struct spa_buffer *buf = b->buffer;
+    if (!buf || buf->n_datas <= 0 || buf->datas[0].data == NULL) {
+        pw_stream_queue_buffer(st->stream_in, b);
+        return;
+    }
+
+    void *base = buf->datas[0].data;
+    size_t maxsize = (size_t)buf->datas[0].maxsize;
+    size_t chunk_offset = 0;
+    size_t chunk_size = 0;
+    if (buf->datas[0].chunk) {
+        chunk_offset = (size_t)buf->datas[0].chunk->offset;
+        chunk_size   = (size_t)buf->datas[0].chunk->size;
+    }
+    if (chunk_offset > maxsize) {
+        pw_stream_queue_buffer(st->stream_in, b);
+        return;
+    }
+    size_t nbytes = chunk_size ? chunk_size : (maxsize - chunk_offset);
+    if (nbytes == 0) {
+        pw_stream_queue_buffer(st->stream_in, b);
+        return;
+    }
+
+    const void *src = (const char *)base + chunk_offset;
+
+    /* Write interleaved float/t_sample bytes into input ring */
+    long written = sys_ringbuf_write(&st->inring, src, (long)nbytes, st->inbuf);
+    if (written < (long)nbytes) {
+        /* overflow: drop tail */
+        st->dio_error = 1;
+    }
+
+#ifdef THREADSIGNAL
+    if (st->sem) sys_semaphore_post(st->sem);
+#endif
+    pw_stream_queue_buffer(st->stream_in, b);
+}
+
+static const struct pw_stream_events stream_in_events = {
+    PW_VERSION_STREAM_EVENTS,
+    .param_changed = pw_in_param_changed,
+    .process       = pw_in_process,
+};
+
+static void pw_out_param_changed(void *data, uint32_t id, const struct spa_pod *param)
+{
+    (void)data; (void)id; (void)param;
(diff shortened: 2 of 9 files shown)


src/s_stuff.h
```diff
@@ -100,10 +100,11 @@ typedef struct _audiosettings
...
 #define API_JACK 5
-#define API_SGI 6           /* gone */
-#define API_AUDIOUNIT 7
-#define API_ESD 8           /* no idea what this was, probably gone now */
-#define API_DUMMY 9
+#define API_PIPEWIRE 6
+#define API_SGI 7           /* gone */
+#define API_AUDIOUNIT 9
+#define API_ESD 9           /* no idea what this was, probably gone now */
+#define API_DUMMY 10
```
Spacechild1, line 107: You can't just change the values of existing constants (even though the constants in questions are probably not used anymore)

src/s_audio_paring.c
```diff
@@ -107,15 +107,18 @@ long sys_ringbuf_init(sys_ringbuf *rbuf, long numBytes,
 ** Return number of bytes available for reading. */
 long sys_ringbuf_getreadavailable(sys_ringbuf *rbuf)
 {
-    long ret = atomic_int_load(&rbuf->writeIndex)
-        - atomic_int_load(&rbuf->readIndex);
+    long write_idx = atomic_int_load(&rbuf->writeIndex);
+    long read_idx = atomic_int_load(&rbuf->readIndex);
+
+    long ret = write_idx - read_idx;
```
Spacechild1, line 113: Why this change?

src/s_audio_pipewire.c
```diff
@@ -0,0 +1,616 @@
...
+    }
+    if (pw_state.outbuf){
+        free(pw_state.outbuf);
+        pw_state.outbuf = NULL;
+    }
+
+    pw_state.running = 0;
+    pw_state.dio_error = 0;
+
+    pw_deinit();
```
Spacechild1, line 515: Why do you deinitialize the PipeWrite library?

src/s_stuff.h
charlesneimog: @Spacechild1 Thank you for the review. This adds the definition of `API_PIPEWIRE`. I placed it after `API_JACK` because both are Unix-specific APIs, and for me, it doesn’t make sense to put it after `API_DUMMY`. 

From my review of the code, these constants are used like an enum, so adding them here seems more appropriate and shouldn’t cause any issues. 

However, I also do not see a problem with putting it after `API_DUMMY`. I just change one line instead!

src/s_audio_paring.c
charlesneimog: That's why I said that:

> But I still have some problems understanding the ring buffers

Without this change, I have consistency check issues, and I really can't understand why; my limited C knowledge stops here. This is because I am also sharing the code as a draft, probably you or umlaeute will understand this problem more clearly and faster than I. Sorry for that!

src/s_audio_pipewire.c
charlesneimog: In my understanding, I need to deinitialize PipeWire to safely stop its threads, destroy streams, free buffers, and release resources, preventing leaks or crashes when Pd closes.

charlesneimog: I know that this is called when I turn off the audio, but I don't see another way to call `pw_deinit()` when I close Pd and not when I turn off the audio.

src/s_stuff.h
Spacechild1: The thing is that these constants are used as API identifiers which are stored in the settings. If you change the values, you mess up existing settings. (Although in practice it probably doesn't matter because the APIs in question are not used.)

> From my review of the code, these constants are used like an enum, so adding them here seems more appropriate and shouldn’t cause any issues.

The same thing would be true of enums. If the values are part of the public API or stored in settings, you can't change them.

I guess you could also replace `API_SGI` with `API_PIPEWIRE`. I don't think anyone is still running Pd on SGI machines :)

src/s_audio_paring.c
Spacechild1: The code is functionally equivalent. All it does is use temporary variables. What did you think this change would do?

> Without this change, I have consistency check issues

What kind of issues exactly? And you are really claiming that this change solves these issues? I have a hard time believing this.

If there are issues, they are certainly somewhere else.

src/s_audio_pipewire.c
Spacechild1: I see! Yes, there is currently no way to deinitialize the audio libraries. We have the same issues with portaudio.

At the moment, we don't really cleanup any resources at all when Pd is shut down, but this is certainly on my to-do list. 

Anyway, the proper place to do this would be in `pd_term_systems` (which is currently empty :), probably by adding a function `pd_term_audio` to `s_audio.c` which would then call `pipewire_term` in `s_audio_pipewire.c`.

Spacechild1:
Thanks for the effort! I'm not really a Linux user and I have no practical experience with PipeWire, so I can't really tell if it would make sense to add a PipeWire - or if your issues could be solved in another way.

Similarly, I can't really study the implementation, but I skimmed over the whole PR and have some general comments:

1. make sure that the code adheres to Pd's style guides. Most importantly:
    1. Allman-style for braces (every opening and closing brace has its own line)
    2. indentation is always 4 spaces, also after line breaks
    3. only use `/* */` for comments
    
2. Pd still uses C89, so all variable declarations must go to the beginning of the current block. 

3. This PR doesn't implement callback mode. You can just copy the relevant code from the Jack backend.

However, before you do any more work, I guess we should wait for Linux people like @umlaeute to comment.

PS: Please remove the CMakeLists.txt from your PR description. It takes a lot of space and it's not really relevant for us.

src/s_audio_paring.c
charlesneimog: Yes, I know, this is why I am saying that I don't understand the `ringbuffers` and all this part. This must be some stupid thing I did, but I can't see it. But see:

https://github.com/user-attachments/assets/534e137d-fa5c-4f8e-9570-5921dd5ba756

I will review this again!

Spacechild1: The video shows that you also get a segfault. Have to tried running it in a debugger to see where it actually crashes?

Spacechild1: Also, have to tried and check the actual value of `ret` that causes the failed consistency check?

Spacechild1:
> Changing block size is not straightforward: 

AFAICT it would absolutely possible to set the buffer size from Pd with `jack_set_buffer_size` (https://jackaudio.org/api/group__ServerControl.html#gad96bda0901b924f5996e02e41a979685). We just don't do this by default so that users don't accidentally change the Jack server settings. However, we could introduce an option to the GUI, e.g. a checkbox "force buffer size". (In fact, that's what Reaper does in its audio settings for ASIO because many ASIO drivers use a global buffer size (and sample rate), similar to the Jack server.) 

> Output device switching is neither persistent nor fast:

Don't the devices show up as jack ports in the audio settings? When I want to switch devices on Windows, I just change the device in the audio settings. I don't see why that would be different with Jack.

reduzent:
> * _Changing block size is not straightforward_: This is crucial for me: sometimes I need a reactive system, but other times—especially with [pd-upic](https://github.com/charlesneimog/pd-upic)), I need a huge block size to listen properly without rendering offline.

I'm using PipeWire since I upgrade my system to Ubuntu 24.04. Since then I'm using the JACK shim provided by PipeWire for Pd. This setup allows me to change blocksize and samplerate on the fly while Pd is connected as JACK client. I consider this problem of yours tackled. However, I had to tweak the system to make all JACK clients (and GUIs like QJackCtl) use the PW implementation of JACK:

```
$ cat /etc/ld.so.conf.d/pipewire-jack-x86_64-linux-gnu.conf 
/usr/lib/x86_64-linux-gnu/pipewire-0.3/jack/
```

reduzent:
> Don't the devices show up as jack ports in the audio settings? When I want to switch devices on Windows, I just change the device in the audio settings. I don't see why that would be different with Jack.

It's not that straight forward, from what I understand:

https://jackaudio.org/faq/multiple_devices.html

And even if you manage to make JACK server to connect to many interfaces, they need to be able to share a clock. There is no clock drift adjustment in the JACK server. It's definitely different from PipeWire where you simply can plug'n'play with soundcards.

charlesneimog:
> AFAICT it would absolutely possible to set the buffer size from Pd with jack_set_buffer_size

But that still wouldn’t make it possible to change outputs. And beyond my own issues, PipeWire is excellent and has become the new default on Linux, because that, I yet think that would be good a 'native' implementation!

charlesneimog:
But this:

> This setup allows me to change blocksize and samplerate on the fly while Pd is connected as a JACK client.

From Pd? What exactly are you using?

I don’t have `jack` installed, so it can only be PipeWire’s JACK replacement:

```
❯ pacman -Qs jack
local/pipewire-jack 1:1.4.7-1
    Low-latency audio/video router and processor - JACK replacement
```

umlaeute:
@reduzent wrote:
> I'm using PipeWire since I upgrade my system to Ubuntu 24.04.

while i'm using Debian rather than Ubuntu, this is what I have been doing for a while on my day-to-day laptop as well.
(there are a few dedicated performance-machines, that still run jack2; but those are a dying breed).

@charlesneimog wrote:
> From Pd? What exactly are you using?

while this question was not directed at me, this is what I do to use a specific blocksize in pipewire:
```sh
# run the JACK-shim at 44.1kHz with a blocksize of 128 samples:
export PIPEWIRE_RATE=128/44100
pd
```

I figure that this is "fast enough switching" for me (of course it is not really; but i guess that the use-cases for different blocksizes are distinct enough that you do not want to switch within a Pd session; and for a start-script this method works good enough)

@charlesneimog also wrote:
> Output device switching is neither persistent nor fast

i'm not sure that i understand this problem. like @reduzent (I guess), using the JACK shim and a persistent patchbay (e.g. qjackctl) should do the trick, no?

@reduzent wrote
> I consider this problem of yours tackled.

this was also my stance on needing a pipewire implementation (but of course, this could be changed ;-))

reduzent:
> > This setup allows me to change blocksize and samplerate on the fly while Pd is connected as a JACK client.
> 
> From Pd? What exactly are you using?

For blocksize:
```
pw-metadata -n settings 0 clock.force-quantum 1024
```

For samplerate
```
pw-metadata -n settings 0 clock.force-rate 44100
```

New values are applied immediately. Pd (connected as JACK client) follows suite immediately, too (You can check with [samplerate~ ]).

You can also check with `jack_bufsize`  and `jack_samplerate`.

It's really amazing how fluid PipeWire behaves.

reduzent:
If you really want to change those settings from Pd, you could issue those commands through `command` external (available on Deken).

reduzent:
BTW: I am not speaking against a PW backend in Pd. I simply like to point out that things are already really good in those regards.

Spacechild1:
> For blocksize:
>
> ```pw-metadata -n settings 0 clock.force-quantum 1024```
>
> For samplerate
>
> ```pw-metadata -n settings 0 clock.force-rate 44100```

I'm not sure I would consider this as particularly user-friendly. However, as I've pointed out above, it would be totally possible to set the block size and/or sample rate within Pd, we just have to "protect" each with a checkbox. By default, we would use the current Jack block size and sample rate, but the user could override it if needed. (Ideally, that's also what we should do with ASIO; currently, Pd forces the ASIO block size and sample rate, but often we just want to use the current driver settings.)

charlesneimog:
Yes, the current approaches (JACK shim, `pw-metadata`, etc.) solve the problems, but I still think a native PipeWire backend would be better, simpler, and more accessible, making Pd on Linux feel as straightforward as it should be.

My main concern is always for beginners, advanced tools like `qjackctl` can be very intimidating. I know it is powerful, but as @Spacechild1 pointed out, it is not really “user-friendly", it feels like too much. In my case, a very common task is simply switching Pd’s output from speakers to headphones — but this is what I see when I open qjackctl. I probably need to read the entire documentation just to understand what is going on:

image  

Compare this with `pwvucontrol` **when using this PR** (note that Pd does not show up here when using JACK):

image  

This is a huge difference in usability. Notice also that I can change the default output device, and all new PipeWire apps will follow it automatically.

Of course, users can script `pw-metadata` or use the command external, and we could even create a dedicated external for this. But I don’t see the point of not making the experience easier out of the box.

Finally, this PR does not change anything for JACK users, but it makes Pd behave like a first-class citizen in PipeWire environments, which are now the default on most modern Linux distros. Since the JACK and PipeWire APIs are very similar, I don’t think this would add much complexity to the audio backend or create maintainability issues. That is why I see a native backend as the better long-term solution.

Issue #496, opened 2018-09-24
feature request, add initbang / closebang


porres:
I'm basically just formally bringing a longstanding feature request to github, based also on a discussion from the pd-dev list, see https://www.mail-archive.com/pd-dev@lists.iem.at/msg01831.html

In short, new object [savestate] tangentially includes [initbang] functionality, following its inclusion, the request is to also add  [initbang] /  [closebang] objects.

danomatika:
I agree. [closebang] is one of those objects I need with vanilla libpd projects and, so far, I've used workarounds by sending a message to #closebang.

porres:
This is now also useful for PlugData, quoting timothy "People that make audio plugins with plugdata want to delete automation parameters when the patch closes, so that they don't persist in the DAW after opening another patch (with its own params)".

I could add it to ELSE, but it'd be better if we had it in Vanilla ;)

umlaeute:
@timothyschoen could also include `iemguts`

Spacechild1:
Side note: while I can totally see why one could have concerns about `[initbang]`, `[closebang]` seems entirely harmless to me, as it's just the counterpart to `[loadbang]`.

danomatika:
I also need closebang in PdParty and basically send a bang to an internal #closebang message for now. It's mainly needed to close spawned patches which were opened by a main patch.

_Hah, I just realized I already wrote the above in a previous post... from 2018._

porres:
>  I can totally see why one could have concerns about [initbang],
>  [closebang] seems entirely harmless to me

I guess I could make a new PR that only adds [closebang] :)

timothyschoen:
Small note here: I wrote my own solution for now using pdlua, which works, but one thing that's tricky is that the index of objects matters a lot. If [closebang] runs after the attached object is already deleted, nothing will happen (and perhaps there is even a crash risk there). Ideally, a [closebang] implementation would always run before any other objects get freed.

EDIT: never mind, I see this is already implemented correctly in porres' PR. Sorry for the noise

Spacechild1:
> EDIT: never mind, I see this is already implemented correctly in porres' PR. Sorry for the noise

Not really, but the ordering issues are addressed separately in https://github.com/pure-data/pure-data/pull/596

PR #2767, opened 2025-09-12
Add [closebang]

porres:
Will update documentation accordingly if this gets merged

closes https://github.com/pure-data/pure-data/issues/496


Spacechild1:
Ideally, this should go hand in hand with https://github.com/pure-data/pure-data/pull/596

Issue #2773, opened 2025-10-01
Crash when closing two patch instances via menuclose


danomatika:
I found this debugging a project using libpd which uses libpd but it also crashes desktop Pd

I often open and close patches in a playlist programmatically using `;pd open` and `;menuclose`. If you open two instances of the same patch, then try to close them using menuclose, Pd crashes. This also happens if you open the instances manually (via menu or double-clicking file), then use menuclose.

Example patches in zip: [menuclose_crash.zip](https://github.com/user-attachments/files/22648903/menuclose_crash.zip)

<img width="492" height="464" alt="Image" src="https://github.com/user-attachments/assets/cf5833d8-909f-4d22-b0e4-4cd4923d8e0d" />

danomatika:
ping @umlaeute @millerpuckette

umlaeute:
it doesn't crash here (Debian/GNU linux), but just closes one of the two abstractions (the 2nd one).

**however**, clicking on the crash-[bng], valgrind gives me an error (triggered by the close):
```
==619228== Invalid read of size 8
==619228==    at 0x40B6C01: bindlist_anything (m_pd.c:113)
==619228==    by 0x40AE754: binbuf_eval (m_binbuf.c:764)
==619228==    by 0x4098888: message_symbol (g_text.c:423)
==619228==    by 0x40B56BB: outlet_symbol (m_obj.c:619)
==619228==    by 0x40D2023: makefilename_symbol (x_connective.c:1438)
==619228==    by 0x40B56BB: outlet_symbol (m_obj.c:619)
==619228==    by 0x40AE754: binbuf_eval (m_binbuf.c:764)
==619228==    by 0x40B5488: outlet_bang (m_obj.c:582)
==619228==    by 0x40D19A7: trigger_list (x_connective.c:1048)
==619228==    by 0x40B5488: outlet_bang (m_obj.c:582)
==619228==    by 0x4052B45: bng_bout2 (g_bang.c:212)
==619228==    by 0x4052C23: bng_loadbang (g_bang.c:293)
==619228==    by 0x4052C23: bng_newclick (g_bang.c:288)
==619228==  Address 0x5333a48 is 8 bytes inside a block of size 16 free'd
==619228==    at 0x49D787F: free (vg_replace_malloc.c:989)
==619228==    by 0x40B700B: pd_unbind (m_pd.c:178)
==619228==    by 0x4057F86: canvas_unbind (g_canvas.c:749)
[...]

PR #2775, opened 2025-10-04
defer freeing canvas in canvas_menuclose()

Spacechild1:
This avoids memory issues when programmatically closing multiple canvasses bound to the same name or closing a canvas from within itself.

Fixes https://github.com/pure-data/pure-data/issues/2773.

~~For a more general solution, see https://github.com/pure-data/pure-data/pull/2776.~~

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -32,6 +32,7 @@ struct _canvasenvironment
 typedef struct _canvas_private
 {
     t_undo undo;
+    t_clock *free_deferred; /* see canvas_free_deferred() */
 } t_canvas_private;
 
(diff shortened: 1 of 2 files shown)


danomatika:
A couple typos on the comments:
* without (running) into
* freeing "no" demand

Spacechild1:
Turns out this is not necessary for closing multiple instances of the same abstraction, see https://github.com/pure-data/pure-data/issues/2773#issuecomment-3377151058.

Closing a canvas from within itself is really an edge case and can be done with externals like `[iemguts/canvasdelete]`. Therefore I'm closing this issue.

PR #2776, opened 2025-10-05
Add canvas_defer() method

Spacechild1:
`canvas_defer()` safely defers a canvas method call to the end of the current clock tick. If the canvas is destroyed, all pending method calls are automatically cancelled. In this case, the function is called with the first argument set to NULL, so it can still free the data object (if used).

This PR also applies `canvas_defer()` to the "menuclose" and "clear" canvas messages to prevent potential crashes during message passing (e.g. by accidentally deleting/clearing the containing canvas or a parent canvas).

Note that "clear" still deletes all scalars synchronously, otherwise we would break existing data structure patches. This is always safe because scalars don't participate in message passing and are protected by gpointers.

Here's a test patch:
[canvas-defer-test.zip](https://github.com/user-attachments/files/22712096/canvas-defer-test.zip)

Fixes https://github.com/pure-data/pure-data/issues/2773. See also the issue discussion for more context.

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -29,9 +29,20 @@ struct _canvasenvironment
     int ce_dollarzero;     /* value of "$0" */
     t_namelist *ce_path;   /* search path */
 };
+
+typedef struct _defer_item
+{
+    t_canvas *d_owner;
+    t_clock *d_clock;
+    void *d_data;
+    t_messfn d_fn;
+    struct _defer_item *d_next;
+} t_defer_item;
+
 typedef struct _canvas_private
 {
     t_undo undo;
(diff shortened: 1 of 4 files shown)


Spacechild1:
As @ben-wes correctly pointed out, the new behavior of "clear" breaks dynamic patching idioms like the following:

> clear, obj 100 100 bng, obj 100 75 loadbang, connect 1 0 0 0, vis 1

I guess there is just no way to make "clear" safe without breaking backwards compatibility... But that's probably ok since the unsafe part is confined to the realm of dynamic patching (which is inherently unsafe).

I'm closing this PR in favor of https://github.com/pure-data/pure-data/pull/2775 (which only fixes "menuclose")

Issue #2784, opened 2025-10-27
Crash when Jack server is stopped (Linux)


solipd:
Pd 0.56-1  and  0.56 freeze with `watchdog: signaling pd...` 
when I stop the Jack server on Debian 13 (same on Ubuntu 24.04), even when DSP is switched off.
Previous versions of pd don't crash. 

Output of the QjackCtl console when server is stopped: 
```
Jack main caught signal 15
Released audio card Audio0
audio_reservation_finish
```

Spacechild1:
I've made some experiments and Pd indeed blocks in `jack_client_close()`. I don't really understand why.

The shutdown handler only sets the `jack_didshutdown` flag and posts to a semaphore:
https://github.com/pure-data/pure-data/blob/12343d6f2d9e4753653319f0e6bcb75cec393d9f/src/s_audio_jack.c#L200-L207

In `jack_send_dacs` we check the `jack_didshutdown` flag:
https://github.com/pure-data/pure-data/blob/12343d6f2d9e4753653319f0e6bcb75cec393d9f/src/s_audio_jack.c#L661-L667

If the flag is set, we call `jack_reopen_audio()`, which actually just closes the Jack client:
https://github.com/pure-data/pure-data/blob/12343d6f2d9e4753653319f0e6bcb75cec393d9f/src/s_audio_jack.c#L628-L638

This is where it finally blocks forever in `jack_client_close`, although we are following the advice in the Jack documentation:
https://jackaudio.org/api/types_8h.html#a1f461fe99711af7edcb1b73b1315fbb3

> Note that after server shutdown, the client pointer is not deallocated by libjack, the application is responsible to properly use [jack_client_close()](https://jackaudio.org/api/group__ClientFunctions.html#ga405646705e600d8bf66327bdbab8e363) to release client ressources.

> Warning: [jack_client_close()](https://jackaudio.org/api/group__ClientFunctions.html#ga405646705e600d8bf66327bdbab8e363) cannot be safely used inside the shutdown callback and has to be called outside of the callback context.

We are indeed calling `jack_client_close()` outside the shutdown callback and also outside the process callback.

Spacechild1:
> Previous versions of pd don't crash.

I can confirm this. Pd 0.55-1 seems to handle a server shutdown just fine.

PR #2785, opened 2025-10-27
Jack: prevent deadlock when closing audio after a server shutdown

Spacechild1:
When `jack_client_close()` is called after a Server shutdown, it triggers the error callback, which in turn would cause a deadlock. This PR prevents this from happening by guarding the error callback function, similar to the `jack_srate()` callback.

Fixes https://github.com/pure-data/pure-data/issues/2784


Spacechild1:
@solipd Please test!

solipd:
Seems to work, no issues for me so far.

PR #2788, opened 2025-10-30
Fix/gop transforms

### LINKED ISSUE #2758: regression bug in 0.56-1 and 'coords' message or is it intentional?
Maybe I missed it, but I've been having issues with my externals. I then narrowed it down to [display] which used 'coords' message. It seems ok now with the new 'goprect' message. I just want to make sure if this was intentional or if we don't want old patches ruining...

Here's an example where I'd get things scrambled in the patch quite harshly

Image

### LINKED ISSUE #2759: gprect message produces ghosts objects in subpatch

ben-wes:
this is mainly a change based on the discussion in #2759 and the findings by @Ant1r.

again, not sure about possible side effects - but for me, it ...

* fixes #2758
* fixes #2759

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -567,6 +567,9 @@ static void canvas_coords(t_glist *x, t_symbol *s, int argc, t_atom *argv)
         x->gl_ymargin = atom_getfloatarg(8, argc, argv);
         canvas_setgraph(x, atom_getfloatarg(6, argc, argv), 0);
     }
+        /* redraw parent to remove "ghost objects" */
+    if (x->gl_owner && !x->gl_havewindow)
+        canvas_redraw(glist_getcanvas(x->gl_owner));
 }
 
 


reduzent:
Works for me. Thanks!

millerpuckette:
The first commit seems to fix the problems (unless of course I'm missing something)... is the second one a bug fix or just an improvement?  If the latter, I'd like to take it on for 0.57 :)

ben-wes:
i added the second one to avoid ghost objects for the `coords` message. not sure if that's the proper place for a redraw though (and these redraws feel a bit brute force in general). and also not sure if that was a regression from 0.55-2 - i guess that would be the criteria for fix or improvement? can't test right now. but could do so tonight...

millerpuckette:
OK, ghost objects would be a bug worthy of fixing.  I had thought that it was always 100% safe to just nuke rtexts anytime, but on thinking about it now, I can imagine that could leave ghosts if things didn't get erased first.  SO... I'll await your tests but I think we'll discover that we need both commits.

Ant1r:
I can confirm that the 2nd commit is needed in order to fix #2759 (current master 36d7be0e171b954e9772c83e2dc87cfd58087f58 still exhibits the ghost objects bug).

However, neither commit fixes one of the roots of the problem: the management of `t_rtext`/`pdtk_text` lifecycle is currently buggy, and https://github.com/pure-data/pure-data/issues/2759#issuecomment-3258622392 is still true:
when a `comment` is created, the associated `pdtk_text` is ***set*** before ***created***:
```tcl
(...)
pdtk_text_set .x559aeddd7ce0.c {.x559aeddd7ce0.t559aeddfd160} {test} ;
(...)
pdtk_text_new .x559aeddd7ce0.c {{.x559aeddd7ce0.t559aeddfd160} {text} } 95 101 {test} 12 #000000 ;
(...)
```

ben-wes:
@Ant1r https://github.com/pure-data/pure-data/compare/master...ben-wes:pure-data:fix/rtextset tries to prevent this. i didn't create a PR for that though since i'm a bit confused with the initial `-1` for width/height that didn't seem to get checked later.

Ant1r:
~~Finally I think that the 2nd commit is likely to involve more unnecessary redraw.
I think #2790 is better. I checked it solves #2758 too.~~

Ant1r:
Sorry I was wrong, #2790 doesn't fix #2758.
The current PR should be merged  as a whole.
#2790 will be rewritten only to fix the rtext lifecycle.

millerpuckette:
OK, merged... I'll watch for your 2790-replacement :)

Ant1r:
Cool!
#2790 has been replaced with #2791.

Issue #2766, opened 2025-09-11
DSP Spikes when mousing over GOP's within GOP's


bradysharp2:
There appears to still be a spike in DSP used when mousing over GOP Abstractions in pd-0.56-1 release.

Running latest Manjaro with XFCE (X11) on an old laptop (Dual Core I5-6300U, 8GB RAM)
Using Jack w/ Pipewire, 48k/64 block size.  Delay in Audio Settings is set to 15ms, but these settings don't seem to matter.

The Patch:
[DSP-test.pd.zip](https://github.com/user-attachments/files/22267250/DSP-test.pd.zip)
64 GOP subpatches, with 4 GOP subpatches within each on.  These are signal-only and relatively small, with no audio processed.

In PD 0.56-1 release:
The patch idles at around 16-18% peak, 14-15% mean DSP.
When moving the mouse across the 64 GOP subpatches, each with 4 GOP subpatches within them, The DSP spikes to about 88% peak, 50% mean, and I get audio IO errors and crackling.
There is several seconds delay when turning on DSP for the first time as well.  It toggles a little faster after the first time.
I can actually drag a GOP subpatch with the mouse, and the DSP is not affected.  So it's only when idly moving the cursor, and only within the PD Canvas.

In PD 0.55-2:
Same patch, Peak is 18%-ish, Mean is 14%-ish, no matter what I do with the mouse. DSP toggle is very responsive.  No IO Errors.

Christof Ressi confirmed a similar issue on the listserv, on a much faster PC running Windows, and notes a similarity to a bug that was resolved in 0.56-1 release.

From Christof:
> There has been a performance regression in Pd 0.56-test1 that sounds very similar to yours: (https://github.com/pure-data/pure-data/issues/2654)). However, that one has been resolved since the final 0.56 release. (I've just tested to make sure.)
> [...]
> I just made a patch with 64 GOP subpatches, each containing 4 instances of your abstraction. Then I tried to quickly move the mouse over the canvas. Here are my results:
> 
> Pd 0.56-1: [iemlib/dsp~] reports an average load of 7-10% and occasional peak loads of up to 28%.
>
> Pd 0.55-2: [iemlib/dsp~] reports an average load of 0.5-1% and occasional peaks of up to 3%.
>
> In both cases, the average idle load (with DSP turned on) is 0.5-1%.
> (Tested on a Windows 10 laptop with an AMD Ryzen 7 PRO 5850U.)"

Thanks,

Brady Sharp

[...]

Issue #2779, opened 2025-10-15
cursor feedback missing for editable scalar text in GOP


ben-wes:
`0.56-1` on macOS, arm64

hovering over editable scalar text fields inside GOP subpatches doesn't change the cursor to indicate they're clickable, unlike scalars on the main canvas.

test patch: [bug-mousefeedback-scalar-text.pd.zip](https://github.com/user-attachments/files/22931710/bug-mousefeedback-scalar-text.pd.zip)

this is a regression - still worked with `0.55-2`.

PR #2789, opened 2025-10-30
check mouse above rtext only for mouse click

ben-wes:
sorry for the PR spam from my side! - but maybe some of it actually helps. this here is also something i checked a few weeks ago. for me, it improves CPU usage for the test patch in #2766 quite a bit more (by another ~50%) compared to the current master at b876394.

i wonder if anything relevant is missing when restricting this to the `doit` case though.

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -2381,8 +2381,8 @@ static void canvas_doclick(t_canvas *x, int xpix, int ypix, int mod, int doit)
     }
 
         /* is the mouse over a text, in either an object box or a scalar? */
-    rtext = rtext_findhit(x, xpix, ypix, &hitobj, &hitscalar,
-        &hitwords, &hitdrawtext);
+    rtext = doit ? rtext_findhit(x, xpix, ypix, &hitobj, &hitscalar,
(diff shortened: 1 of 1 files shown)


millerpuckette:
I think this will result in mouse cursor not always changing to correct state when the mouse moves into an editable area.  I'm leaning toward simply throttling the mouse motin messages at the Tcl/TK level... but nt for 0.56-2 - I don't think it's urgent and I've already seen many innocent-looking changes break things late in the release cycle :)

ben-wes:
i only tested this on macos, where cursor changes in run mode look right to me (as you mentioned somewhere else, cursor changes need mouse movement though).

but i was honestly also a bit surprised that the `doit` condition doesn't seem to cause missing cursor changes. that's why these PRs also need serious review. ;)

concerning cursor changes, i think that #2779 is quite relevant since those changes were missing completely for 0.56 - ##2786 is trying to address that.

millerpuckette:
cool - I went and grabbed https://github.com/pure-data/pure-data/pull/2786.
Still unsure about cursor biz, will think about it.

PR #2796, opened 2025-11-12
Documentation updates

### LINKED ISSUE #30: update Shahrokh Yadegari's webpage on expr and the same html file we ship with Pd
Not something we can fix here, but we're now shipping a reference html file as the reference of expr and it is updated, but there's a webpage in the wild on Shahrokh Yadegari's webpage which is outdated. This is more of a reminder for me to try and send our new html to him so he updates it and then we can close this. 

I have made yet more updates and revised the html file, let's wait until it gets in and the new 0.52 version is out so we can replace this document online as well.

### LINKED ISSUE #239: about paf (F12 and F13)
So, F13.paf.control in audio examples says 

"_Note however that there's a [paf~] extern available which solves this problem better and, moreover, runs much faster._"

Now, this external used to be part of 'extra', but it isn't there anymore. 

we can revive it and I remember I discussed this on the list at some point, or forget about it...

ping @millerpuckette

### LINKED ISSUE #296: move chapters out of resources/ folder
this has been bothering me for ages (since the manual overhaul).

conceptually, i think that the manual consists of a collection of HTML pages.
these HTML pages are then enhanced by some assets, that is: images and CSS

but when i navigate to [pd/doc/1.manual](https://github.com/pure-data/pure-data/tree/0.56-0/doc/1.manual), then i do not see these chapters.
instead i only see the `index.htm` site, and a catch-all `resources/` directory (which I take to include the *assets*)

i understand that as per https://github.com/pure-data/pddp/issues/240 this is fully intentional:
> So let's call them "chapter1", "chapter2", etc, and also move it to the 'resources' folder.

but i really don't agree with "chapters" being mere *resources*: they are the actual manual.
as such they should be in 

(you could argue that the `index.htm` page is more of a *resource*: it helps us navigate the the chapters, but that's about it.)

i understand that the idea for the re-organization was:
- cleanup the directory structure (so the users do not get overhelmed)
- pick meaningful names (`x2.htm` -> `chapter5.htm`)

but if we put the `chapter`s into the `resources/` directory, little is won: they are again buried in files (there are 6 html files, and 114 other files in that directory).

i would like to suggest to move the `chapter*.htm` up one level into the `pd/doc/1.manual/`.

i guess the only reason to put them into the `resources/` folder was to have `index.htm` as the first entry in the directoy (and `chapter` sorts alphabetically before `index`)

### LINKED ISSUE #300: minor DS wording issues.
- *doc/4.data.structures/01.scalars.pd* -> *[pd template1]*
  > "graph no parent"

  i think this should read **on** rather than *no*

- various help-patches in *doc/5.reference/* mention
  >the 4.Data.Structure section

  but the docs only have a directory `4.data.structures` (notice capitalization and plural)

- a subset of these help-patches says:
  >the 4.Data.Structure section of the Pd's tutorials

  i think this should read `of Pd's tutorials` or `of the Pd tutorials`

  (and maybe use singular `tutorial` or even `documentation`, but that's probably a matter of taste)

### LINKED ISSUE #301: non-conformant HTML in chapter5 (release notes)
there's a couple of closing `</p>` tags missing.

browsers probably render this just fine, but when trying to extract meaningful information, it becomes a problem...

```
doc/1.manual/resources/chapter5.htm:2216: HTML parser error : Unexpected end tag : p
</UL></P>
         ^
doc/1.manual/resources/chapter5.htm:2233: HTML parser error : Unexpected end tag : p
</UL></P>
         ^
doc/1.manual/resources/chapter5.htm:2281: HTML parser error : Unexpected end tag : p
</UL></P>
         ^
doc/1.manual/resources/chapter5.htm:2297: HTML parser error : Unexpected end tag : p
</UL></P>
         ^
doc/1.manual/resources/chapter5.htm:2321: HTML parser error : Unexpected end tag : p
</P>
    ^
doc/1.manual/resources/chapter5.htm:2336: HTML parser error : Unexpected end tag : p
</P>
    ^
doc/1.manual/resources/chapter5.htm:2359: HTML parser error : Unexpected end tag : p
</P>
    ^
doc/1.manual/resources/chapter5.htm:2368: HTML parser error : Unexpected end tag : p
</P>
    ^
doc/1.manual/resources/chapter5.htm:2392: HTML parser error : Unexpected end tag : p
</P>
    ^
doc/1.manual/resources/chapter5.htm:2608: HTML parser error : Unexpected end tag : p
</P>
    ^
```

### LINKED ISSUE #303: abstraction reloading
looking up `doc/2.control.examples/12.PART2.subpatch.pd` I read:

> If you change one copy of an abstraction the change isn't automatically made on any other copies. You must keep track, save the changes, and cause Pd to reload the other copies (for example, by closing and reopening the containing patch.)

i think this is somewhat confusing.

there's no need to "close and reopen the containing patch" to reload the "other copies" of my saved abstraction.
simply saving the absrtraction will update all it's copies (but not if you open the file as a standalone patch).

### LINKED ISSUE #304: $args in abstractions
looking up `doc/2.control.examples/12.PART2.subpatch.pd` I read:
> Note that "$1", etc, has a different meaning in object boxes.

a **different meaning** compared to what?

this is the first mention on `$1` in this patch. we cannot assume that the user just came over from `doc/2.control.examples/10.more.messages.pd` where `$1` is explained differently.

### LINKED ISSUE #305: manual section 2.5.1 - there's clipping on macOS
someone made this remark at a video, see 4:23 https://www.youtube.com/watch?v=G3CIJ3l5Fd8

Apparently, macOS doesn't clip the output and we should note it.

### LINKED ISSUE #306: manual §6.6: copy-paste error
there's a copy-paste error in §6.6 of the manual, [where it reads](https://github.com/pure-data/pure-data/blob/b35ff302ab3bc422881654fc3dc934694671e7a3/doc/1.manual/resources/chapter6.htm#L759)

```html
<p>Then install to the default location (C:32 or C:64) and follow the setup/update info on the Msys2 webpage.</p>
```

this was introduced when converting the INSTALL.txt to html by @Lucarda in [1c44e3ef](https://github.com/pure-data/pure-data/commit/1c44e3eff2ddd65b47c47c4e9d8e4f1ce2aaf0eb).

however the [original INSTALL.txt](https://github.com/pure-data/pure-data/blob/1c44e3eff2ddd65b47c47c4e9d8e4f1ce2aaf0eb/INSTALL.txt#L304) reads:
```txt
Then install to the default location (C:\msys32 or C:\msys64) and follow the
```

### LINKED ISSUE #308: framp~-help.pd and throw~-catch~-help.pd cannot turn off DSP
both help-patches for `[framp~]` and `[throw~]`/`[catch~]` have a DSP-toggle, that only turns *on* (and never off):

```
[tgl]
|
[; pd dsp 1(
```

obviously this should be `[; pd dsp $1(`

porres:
The usual doc updates

- closes https://github.com/pure-data/pddp/issues/296
- closes https://github.com/pure-data/pddp/issues/301
- closes https://github.com/pure-data/pddp/issues/304
- closes https://github.com/pure-data/pddp/issues/303
- closes https://github.com/pure-data/pddp/issues/300
- closes https://github.com/pure-data/pddp/issues/239
- closes https://github.com/pure-data/pddp/issues/305
- closes https://github.com/pure-data/pddp/issues/30
- closes https://github.com/pure-data/pddp/issues/306
-  closes https://github.com/pure-data/pddp/issues/308


doc/1.manual/2.theory.of.operation.htm
```diff
@@ -2249,24 +2249,32 @@ <H4> <A href="#s2.11.1" id="s2.11.1"> 2.11.1. Differences between Pd and MAX</A>
...
+than one element. In Pd it is only implicit if the list starts with a float. So
+if you have a list that starts with a symbol, you need the 'list' selector in Pd.
+MAX has no 'symbol' selector, but a one element symbol message is considered a
+symbol and not a list. Pd explicitly needs to define a symbol data type with a
+symbol selector.</P>
+
+<P> Moreover, in Pd, a one element list is effectively a float or a symbol
+(or even a pointer for that matter), while MAX can differentiate a one element
+list with a list selector and a float/int or symbol. Also, in Pd, a list
+without any elements (hence, an empty list) is treated as a bang message.</P>
```
umlaeute: this sounds like Pd "can **not**" differentiate a one element list from e.g. a `float` message, which is not true.

porres: Pd can, but in effect (this is where "effectively" comes in) it cannot, right? This has been a big conversation between me and @Spacechild1 by the way...

I was even asking whether and where would a "list 1" message be requires instead of a float message, so one could make use of a "list 1" message being generated by the listbox, for instance. My reasoning is that we should just convert to float instead. But the rule seems to be that there's no real distinction between the two.

Anyway, so now what? And how would you suggest to better phrase this?

thanks for the review

umlaeute: my point was less that "one element lists are effectively floats", but more the following "*while* Max *can differentiate*".

it's easy to understand the 2nd part ("max is able to..."), but the first part of the dichotomy is less clear (what does "effectively" really mean here?)...

since the "while" turns the two parts into opposites, the only remaining interpretation (for me, as a non-native speaker), is: Ps ist bot alle to differentiate. 

which is just wrong. 

Pd *can* differentiate, but it doesn't have to.

porres: Well, I see, I will think about it... and as I see it, Pd can differentiate, but it never really has to. That's more like it.

porres: I made a review in https://github.com/pure-data/pure-data/pull/2796/changes/88b08c659e0019d432dec16a3312747814729950  hope it's ok now

umlaeute: dunno.
to be honest, I'm unsure whether the new wording will help anybody to understand what you mean. 

anyhow, I'll rest my case for now...

porres: I think we also need to explain this in the manual itself and then refer to it, see https://github.com/pure-data/pddp/issues/307

doc/5.reference/text-object-help.pd
```diff
@@ -413,6 +416,9 @@
 #X text 134 16 - convert from list.;
 #X text 45 68 [text fromlist] converts a list such as [text tolist] would output and fills the text with it. Whatever the text had previously contained is discarded.;
 #X msg 45 156 1 2 \\\; 3 4 5 \\\; 7 8 9 10 \\\; cis boom bah \\\, walk the dog \\\;;
+#X obj 306 432 savestate;
+#X text 264 433 see:;
+#X text 105 373 This is a good way to import the data from a stored list \, qhich can also be recalled from a [savestate] in an abstraction., f 42;
```
umlaeute: typo "qhich"

PR #2797, opened 2025-11-13
adds a basic theme editor

ThomasAUB:
Adds a basic color editor for "foregreound", "background", "selection" and "graph on parent" colors.

The colors can be applied without closing the editor window and are saved on "OK".

screenshot


umlaeute (changes requested):
I think there is no reason for all the dialog on the core side.

such a theming editor should be a GUI only thing (and could be implemented as a GUI plugin, distributed separately from Pd).

(the only exception might be telling the GUI about the current colour theme, if the user programmatically sets it via a `[; pd colors ...(` message; see https://github.com/pure-data/pure-data/pull/2757 or similar)

umlaeute:
also, the theming editor obviously should integrate into the preference editor.

ThomasAUB:
thanks for the review

PR #2799, opened 2025-11-17
force rtext coordinate recalculation when closing subpatch

### LINKED ISSUE #2798: gatoms in GOP are drawn at subpatch-relative position after closing subpatch
test patch: [bug-gatom-subpatch.pd.zip](https://github.com/user-attachments/files/23592584/bug-gatom-subpatch.pd.zip)

Image

*(number box is supposed to be inside the goprect)*

i don't understand the exact conditions in which this happens (can't always reproduce it for slower metros for example, which doesn't make sense to me).

EDIT: this is a regression introduced with 0.56-0

ben-wes:
simple attempt to close #2798


ben-wes:
while this fixes it, it doesn't fix the root cause. closing and trying to avoid yet another redraw ...

PR #2812, opened 2025-12-31
add initbang

### LINKED ISSUE #2811: add initbang
this kinda relates to https://github.com/pure-data/pure-data/issues/496 which requests to add a related object. I was first also making the request for initbang over there but I'm now formally splitting this request into a new separate issue. This is motivated by yet another and new request/discussion on the pd list.

porres:
will include and update the docs later if/when this gets merged

closes https://github.com/pure-data/pure-data/issues/2811

### DIFF
--- a/src/x_misc.c
+++ b/src/x_misc.c
@@ -128,6 +128,35 @@ static void loadbang_setup(void)
         gensym("loadbang"), A_DEFFLOAT, 0);
 }
 
+/* -------------------------- initbang ------------------------------ */
+static t_class *initbang_class;
+
+typedef struct _initbang
+{
+    t_object x_obj;
+} t_initbang;
+
+static void *initbang_new(void)
+{
+    t_initbang *x = (t_initbang *)pd_new(initbang_class);
+    outlet_new(&x->x_obj, &s_bang);
+    return (x);
+}
+
+static void initbang_initbang(t_initbang *x, t_floatarg action)
+{
+    if (action == LB_INIT)
+        outlet_bang(x->x_obj.ob_outlet);
+}
+
+static void initbang_setup(void)
+{
+    initbang_class = class_new(gensym("initbang"), (t_newmethod)initbang_new, 0,
+        sizeof(t_initbang), CLASS_NOINLET, 0);
+    class_addmethod(initbang_class, (t_method)initbang_initbang,
+        gensym("loadbang"), A_DEFFLOAT, 0);
+}
+
 /* ------------- namecanvas (delete this later) --------------------- */
 static t_class *namecanvas_class;
 


porres:
somewhat related to https://github.com/pure-data/pure-data/pull/2767

src/x_misc.c
```diff
@@ -128,6 +128,35 @@ static void loadbang_setup(void)
...
 }
 
+/* -------------------------- initbang ------------------------------ */
+static t_class *initbang_class;
+
+typedef struct _initbang
+{
+    t_object x_obj;
+} t_initbang;
+
```
umlaeute, line 138: no need for this structure. just use `t_object` directly.

(also, if the code is directly copied from iemguts, you might add proper credits in the git commit message)

porres: What I actually did was just copy the loadbang code and change the names to 'initbang'... 

And I guess I just copied the struct from loadbang

"typedef struct _loadbang
{
    t_object x_obj;
} t_loadbang;"

I wouldn't know how to not have the struct or how it works, and I wonder if the one from loadbang is also unnecessary then.

ben-wes: @porres the code below works for me, avoiding the struct.
@umlaeute: the [old iemguts version](https://github.com/iem-projects/pd-iemguts/blob/main/src/initbang.c#L43-L46) looks quite similar though. :) 

``` c
/* -------------------------- initbang ------------------------------ */
static t_class *initbang_class;

static void *initbang_new(void)
{
    t_object *x = (t_object *)pd_new(initbang_class);
    outlet_new(x, &s_bang);
    return (x);
}

static void initbang_initbang(t_object *x, t_floatarg action)
{
    if (action == LB_INIT)
        outlet_bang(x->ob_outlet);
}

static void initbang_setup(void)
{
    initbang_class = class_new(gensym("initbang"), (t_newmethod)initbang_new, 0,
        sizeof(t_object), CLASS_NOINLET, 0);
    class_addmethod(initbang_class, (t_method)initbang_initbang,
        gensym("loadbang"), A_DEFFLOAT, 0);
}
```

EDIT: i guess, this should be consistent with loadbang though (either way).

ben-wes: and maybe it would be good to name it `initbang_loadbang` instead of `initbang_initbang` similar to how iemguts does it - after all, the action triggers on loadbang and is named `LB_INIT` ... ?

ben-wes:
@porres it looks to me like the `initbang_setup()` call is missing btw.

porres:
> initbang_setup() call is missing btw.

@ben-wes you meant in x_misc_setup, right? Just added it...

PR #2815, opened 2026-01-05
don't draw iolets for objects inside GOP

### LINKED ISSUE #2768: ghost iolets of GOP sub-subpatches remain when deleting subpatch
found this while testing `0.56-1` on macos in the context of
* #2766

... when deleting a GOP-subpatch with sub-subpatches having iolets, these are not cleared from the canvas on deletion (right one deleted in this screenshot):

Image

test patch: [bug-ghost-iolets.pd.zip](https://github.com/user-attachments/files/22295527/bug-ghost-iolets.pd.zip)

this does not seem to happen with `0.55-2`.

ben-wes:
imho, this is a valid approach to address the issue of ghost iolets remaining from objects inside a deleted GOP. after all, displaying these iolets doesn't really make sense since cords aren't drawn either.

it can obviously change the appearance of patches - but they should look cleaner. :)

* closes #2768

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -1035,8 +1035,7 @@ static void iemgui_draw(t_iemgui *x, t_glist *glist, int mode)
         break;
     default:
         /* skip drawing iolets if we're inside a GOP subpatch */
-        if (glist_isgraph(glist) && !glist->gl_havewindow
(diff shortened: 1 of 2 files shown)


reduzent:
> After all, displaying these iolets doesn't really make sense since cords aren't drawn either.

👍️

PR #2816, opened 2026-01-05
catch crashes for 'dialog' messages to iem objects

### LINKED ISSUE #2813: Sending a `dialog` message without arguments to any `iemgui` object crashes Pd
Pd crashes when you send a "dialog" message with no arguments to any of the IEM gui objects. Try it out here: [dialog-crash.zip](https://github.com/user-attachments/files/24409324/dialog-crash.zip) (just click on any "dialog" message and 💥). Sometimes Pd freezes, but clicking again would lead to a crash.

Tested on latest pd on macos and windows, also on debian (compiled from source, as well as on Pd.53.1 from apt (0.53.1+ds-2+deb12u1)

#### Details
I got this with `valgrind` using pd compiled from source, clicking on the "dialog" connected to "radio" (full output here: [valgrind-pd-dialog-crash.txt](https://github.com/user-attachments/files/24409675/valgrind-pd-dialog-crash.txt)):
```
==390214== Conditional jump or move depends on uninitialised value(s)
==390214==    at 0x14D752: srl_is_valid (g_all_guis.c:50)
==390214==    by 0x14FC96: iemgui_dialog (g_all_guis.c:802)
==390214==    by 0x181548: radio_dialog (g_radio.c:308)
==390214==    by 0x1B3C70: pd_typedmess (m_class.c:1091)
==390214==    by 0x1B71F7: outlet_anything (m_obj.c:641)
==390214==    by 0x19A4B0: messresponder_anything (g_text.c:404)
==390214==    by 0x1B3F0E: pd_typedmess (m_class.c:1158)
==390214==    by 0x1AC08A: binbuf_eval (m_binbuf.c:764)
==390214==    by 0x19A533: message_float (g_text.c:416)
==390214==    by 0x19A990: message_click (g_text.c:504)
==390214==    by 0x19DDBA: text_click (g_text.c:1416)
==390214==    by 0x157D02: gobj_click (g_canvas.c:167)
...
```

#### Questions
- On a regular patching or performing session, one would not send an empty "dialog" message like this, but the method is exposed on every iemgui, AFAICT for good reason, so there could be some protection. Or, should this method not be exposed at all?
- Maybe I'm missing a fix on some branch, has this been fixed already? 
- If not, some basic solutions could be:
  - simple fix across all iemguis: early return if no args are received by `iemgui_dialog` 
  - default some values (the code already tries to do so). But in this case, how do we pick the defaults like size from the more general `iemgui_dialog` function? We risk ending up with more code in the individual radio/toggle/bng `_dialog` methods.

#### More specific questions about `iemgui_dialog`
I'm trying to understand the root of the crash and the logic of the design here, so that I can come up with a PR that addresses the problem in a more complete way, rather than just patching it up. So, I'm looking into this function more closely. A PR here could refactor some stuff on this function, for example, the `sr_flags` that is returned and not used. Why is that `sr_flags` not being used? This unused return appears on every `_dialog` method, so I'm wondering, should this return be used for its purpose, or even hijacked to return `-1` and do something different? And likewise, `srl` is passed as a reference and populated in iemgui_dialog, but I'm not seeing it used after it has been populated, so why is it passed in the first place?

ben-wes:
not sure if this change of logic is valid - honestly don't really understand why there's a check for `IS_A_FLOAT` before falling back to `gensym("empty")`.

anyway, this would resolve #2813 - although i'm also not sure whether protection is needed for messages that are not documented?

- Closes https://github.com/pure-data/pure-data/issues/2813

### DIFF
--- a/src/g_all_guis.c
+++ b/src/g_all_guis.c
@@ -777,25 +777,17 @@ int iemgui_dialog(t_iemgui *iemgui, t_symbol **srl, int argc, t_atom *argv)
         oldsndrcvable |= IEM_GUI_OLD_RCV_FLAG;
     if(iemgui->x_fsf.x_snd_able)
         oldsndrcvable |= IEM_GUI_OLD_SND_FLAG;
-    if(IS_A_SYMBOL(argv,7))
-        srl[0] = atom_getsymbolarg(7, argc, argv);
-    else if(IS_A_FLOAT(argv,7))
-    {
-        srl[0] = gensym("empty");
-    }
-    if(IS_A_SYMBOL(argv,8))
-        srl[1] = atom_getsymbolarg(8, argc, argv);
-    else if(IS_A_FLOAT(argv,8))
-    {
-        srl[1] = gensym("empty");
-    }
-    if(IS_A_SYMBOL(argv,9))
+    srl[0] = (argc > 7 && IS_A_SYMBOL(argv,7)) ? atom_getsymbolarg(7, argc, argv) : gensym("empty");
+    srl[1] = (argc > 8 && IS_A_SYMBOL(argv,8)) ? atom_getsymbolarg(8, argc, argv) : gensym("empty");
+    if(argc > 9 && IS_A_SYMBOL(argv,9))
         srl[2] = atom_getsymbolarg(9, argc, argv);
-    else if(IS_A_FLOAT(argv,9))
+    else if(argc > 9 && IS_A_FLOAT(argv,9))
     {
         sprintf(str, "%g", atom_getfloatarg(9, argc, argv));
         srl[2] = gensym(str);
     }
+    else
+        srl[2] = gensym("empty");
     if(init != 0) init = 1;
     iemgui->x_isa.x_loadinit = init;
     for(i=0; i<3; i++)
--- a/src/g_radio.c
+++ b/src/g_radio.c
@@ -308,6 +308,10 @@ static void radio_dialog(t_radio *x, t_symbol *s, int argc, t_atom *argv)
     sr_flags = iemgui_dialog(&x->x_gui, srl, argc, argv);
     x->x_gui.x_w = iemgui_clip_size(a) * IEMGUI_ZOOM(x);
     x->x_gui.x_h = x->x_gui.x_w;
+    if(num < 1)
+        num = 1;
+    if(num > IEM_RADIO_MAX)
+        num = IEM_RADIO_MAX;
     if (num != x->x_number && glist_isvisible(x->x_gui.x_glist))
     {
         /* we need to recreate the buttons */


porres:
> i'm also not sure whether protection is needed for messages that are not documented?

yeah, I don't think there's a bug in an undocumented message that is not supposed to be used :)

ben-wes:
> I don't think there's a bug in an undocumented message that is not supposed to be used :)

true. and as my comment implied, i'm not really sure about what to do here. but preventing a crash still doesn't feel wrong if it's not adding much complexity.

Ant1r:
> don't really understand why there's a check for `IS_A_FLOAT`

me neither ;-) I think your changes are legitimate.

src/g_radio.c
```diff
@@ -308,6 +308,10 @@ static void radio_dialog(t_radio *x, t_symbol *s, int argc, t_atom *argv)
     sr_flags = iemgui_dialog(&x->x_gui, srl, argc, argv);
     x->x_gui.x_w = iemgui_clip_size(a) * IEMGUI_ZOOM(x);
     x->x_gui.x_h = x->x_gui.x_w;
+    if(num < 1)
+        num = 1;
+    if(num > IEM_RADIO_MAX)
+        num = IEM_RADIO_MAX;
```
umlaeute, line 314: nitpick: this seems to be unrelated to the other fix (which is a more serious out-of-bounds array access). i would have put this into separate commits...

fdch:
> there's a check for IS_A_FLOAT

I think floats are valid send, receive, and label names, aren't they?


> i'm not really sure about what to do here. but preventing a crash still doesn't feel wrong

Well, the method could be nice(r) and report an error if no arguments are passed. That way, if your gui accidentally drops the arguments, or they collide with ethernet dust and disappear, pd lets the dev know.


Anyway, kidding aside, I had no rush here but I'm glad that a PR is already in place (and reviewed). Though, I would have liked to see what people thought about this function.

ben-wes:
> I think floats are valid send, receive, and label names, aren't they?

no. only symbols are accepted. floats fall back to `empty`.

> Well, the method could be nice(r) and report an error if no arguments are passed

i'm not sure how you found this method in the first place? :) ... AFAICT, its intended use is only for Pd's actual properties dialogs.

if you want to change properties from within your patch, you're supposed to use the other messages which were invented for this very purpose and which are documented in the help files of each object ... or are you missing any specific options there?

src/g_all_guis.c
```diff
@@ -777,21 +777,11 @@ int iemgui_dialog(t_iemgui *iemgui, t_symbol **srl, int argc, t_atom *argv)
...
-    {
-        srl[1] = gensym("empty");
-    }
-    if(IS_A_SYMBOL(argv,9))
+    srl[0] = (argc > 7 && IS_A_SYMBOL(argv,7)) ? atom_getsymbolarg(7, argc, argv) : gensym("empty");
+    srl[1] = (argc > 8 && IS_A_SYMBOL(argv,8)) ? atom_getsymbolarg(8, argc, argv) : gensym("empty");
+    if(argc > 9 && IS_A_SYMBOL(argv,9))
         srl[2] = atom_getsymbolarg(9, argc, argv);
-    else if(IS_A_FLOAT(argv,9))
+    else if(argc > 9 && IS_A_FLOAT(argv,9))
```
fdch, line 784:
```c
    if(argc > 9 && IS_A_SYMBOL(argv,9))
        ...
    else if(argc > 9 && IS_A_FLOAT(argv,9))
    {
        ...
    }
```

The problem here is that `argc > 9` is always false, so `srl[2]` will be uninitialized. I'd initialize it to "empty" before that `if/else if` statement and that would be it.


```c
    srl[2] = gensym("empty");
    if(argc > 9 && IS_A_SYMBOL(argv,9))
        srl[2] = atom_getsymbolarg(9, argc, argv);
    else if(argc > 9 && IS_A_FLOAT(argv,9))
    {
        sprintf(str, "%g", atom_getfloatarg(9, argc, argv));
        srl[2] = gensym(str);
    }
```

fdch:
> no. only symbols are accepted. floats fall back to empty.

this is relatively new, though, see https://github.com/pure-data/pure-data/commit/cb210b4cec992a7384a79355487f419cdf7ee538

>  its intended use is only for Pd's actual properties dialogs

It's the message to set all the iemgui's properties. But if you can send a message to it, and the message can crash it, it will crash.

More importantly, I don't think this fix addresses the linked issue (see my comment [here](https://github.com/pure-data/pure-data/pull/2816/commits/d78034cfa46c32312758b8bbb672542ed6d6dbc9#r2671797013)). Have you tried testing it with the attached patch? From valgrind, I get the `uninitialized value` is on line 792:
```
==112013== Conditional jump or move depends on uninitialised value(s)
==112013==    at 0x15284F: srl_is_valid (g_all_guis.c:50)
==112013==    by 0x15284F: iemgui_dialog (g_all_guis.c:792)
```

ben-wes:
> this is relatively new, though, see [cb210b4](https://github.com/pure-data/pure-data/commit/cb210b4cec992a7384a79355487f419cdf7ee538)

ah, thanks! i should have checked git blame to understand why there was this float check before falling back. that explains it (as a leftover from the previous code). ok - i assume there was a reasoning though for preventing floats as sender/receiver names. i prefer not to touch that for this fix here.

> Have you tried testing it with the attached patch?

damn - no. i tested with my own patch, but missed `[cnv]`. added a fallback now.

fdch:
Nice, the crash is gone; or at least valgrind does not complain.

PR #2819, opened 2026-01-08
dialog_gatom: escape blank entries

myQwil:
If entries for numbers are completely blank, parameters will be misinterpreted as intended for the parameter that's supposed to come before it, and values will get assigned to the wrong properties.

Example: if width entry is blank, and lower limit is set to 1, width will be set to 1 instead, while lower limit will be set to whatever non-blank parameter happens to come next.

### DIFF
--- a/src/g_text.c
+++ b/src/g_text.c
@@ -990,12 +990,21 @@ static int gatom_doclick(t_gobj *z, t_glist *gl, int xpos, int ypos,
     return (1);
 }
 
+static t_float atom_floatelse(int which, int argc, const t_atom *argv,
+    t_float fallback)
+{
+    if (argc <= which) return (fallback);
+    argv += which;
+    if (argv->a_type == A_FLOAT) return (argv->a_w.w_float);
+    else return (fallback);
+}
+
     /* message back from dialog window */
 static void gatom_param(t_gatom *x, t_symbol *sel, int argc, t_atom *argv)
 {
-    t_float width = atom_getfloatarg(0, argc, argv);
-    t_float draglo = atom_getfloatarg(1, argc, argv);
-    t_float draghi = atom_getfloatarg(2, argc, argv);
+    t_float width = atom_floatelse(0, argc, argv, x->a_text.te_width);
+    t_float draglo = atom_floatelse(1, argc, argv, x->a_draglo);
+    t_float draghi = atom_floatelse(2, argc, argv, x->a_draghi);
     t_symbol *label = gatom_unescapit(atom_getsymbolarg(3, argc, argv));
     t_float wherelabel = atom_getfloatarg(4, argc, argv);
     t_symbol *symfrom = gatom_unescapit(atom_getsymbolarg(5, argc, argv));
--- a/tcl/dialog_gatom.tcl
+++ b/tcl/dialog_gatom.tcl
@@ -40,9 +40,9 @@ proc ::dialog_gatom::apply {mytoplevel} {
     global gatomlabel_radio
 
     pdsend "$mytoplevel param \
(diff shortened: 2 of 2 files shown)


umlaeute:
thanks.

i don't really like defaulting empty values to zero, though.
my gut feeling tells me that an empty value in a number field is invalid, and the dialog should not be allowed to commit its data.

alternatively, an empty field could just mean "the previously set value" (but i guess this would benefit from some visual feedback (like a "placeholder", as is available in Tk>=9)

myQwil:
Okay, I've taken a new approach where the text boxes are escaped using the regular escape function, so blank entries should return a single dash. This will come back as a symbol atom, and because it's not a float atom, the check will fail and the previous value will get used as a fallback.

src/g_text.c
```diff
@@ -990,12 +990,21 @@ static int gatom_doclick(t_gobj *z, t_glist *gl, int xpos, int ypos,
...
 
+static t_float atom_floatelse(int which, int argc, const t_atom *argv,
+    t_float fallback)
+{
+    if (argc <= which) return (fallback);
+    argv += which;
+    if (argv->a_type == A_FLOAT) return (argv->a_w.w_float);
+    else return (fallback);
+}
+
```
ben-wes, line 1001: it feels a bit off to me to have this kind of `atom_` helper function here and not in `m_atom.c`. but i'm admittedly not familiar with all the conventions.

wondering though whether simple conditional expressions might suffice here - even though they're obviously getting a bit long and redundant:

``` c
t_float width = argv[0].a_type == A_FLOAT ? atom_getfloatarg(0, argc, argv) : x->a_text.te_width;
```

either way, i appreciate the fix!

myQwil:
There's a problem with this solution regarding negative numbers. The escape function checks if the first character is a dash (minus sign), so the placeholder character would need to be something other than dash for this to work.

umlaeute:
good point. 

I refrained from commenting some more as I'm afraid I didn't like the new implementation that much anyhow, but didn't want to create busywork...

but in the light of the new problem...

- the problem is on the GUI side, so I think it should be solved on the GUI side (alone). no need to add cruft to the core
- the GUI already knows the previous values (as this is what is initially displayed)

so my "simple" solution (just fall back to the original values) would look like this: 
- in the create_dialog store the original values on some array (with a namespaced name; and we need to be able to distinguish the values for multiple dialogs, so the widget name (of the dialog) ought to be somehow part of this. either of the array name or the array key) 
- in the apply proc, check for empty values and fall back to the values in the array. update the values in the array with new (valid) values (in case we want to apply again) 
- when destroying the dialog, don't forget to clean up the array

the more complicated solution would be: 
- have a boolean per-dialog array with flags, that indicate whether a given field is valid 
- when creating the entries, add a validator proc that sets the valid-flag for the given value depending on the output of `string is double` (or similar) 
- within the validator, check whether *all* flags (for the dialog) are true. if yes, enable the `OK` and `Apply` buttons, otherwise disable them. 

finally I think that passing a proper list to pdsend (`pdsend [list ...]`) rather than a string (`pdsend "..."`), would simply avoid disappearing arguments altogether. I guess this is equivalent to your first version (setting invalid values to 0).
in general I think that there are too many string arguments that should be lists (but there might be subtleties, when it comes to weird arguments with spaces and such)

anyhow, I'm currently afk, so the above is just an overview (and please excuse me, if I got all the names wrong)

umlaeute:
i've gone ahead and implemented a fixes (on the GUI side) for (afaict) all dialogs in #2845

myQwil:
Thanks, I tested your PR and it fixes things on my end, so I'll close this.

Issue #2821, opened 2026-01-09
`mouse` command on startup or without GUI


giuliomoro:
This [simple patch](https://github.com/user-attachments/files/24535257/_main.pd.zip)
<img width="290" height="228" alt="Image" src="https://github.com/user-attachments/assets/e3d191dd-1312-4dcc-9eb7-48495b15b7d1" />

Until before 1b928e6769bb31d3ab1220a13facb9d7f0e70406, when running the patch without an editor (e.g.: inside libpd) I would get `error: consistency check failed: editor`. As of 1b928e6769bb31d3ab1220a13facb9d7f0e70406 that turns into a segmentation fault in g_editor's `canvas_doclick`:
```
    x->gl_editor->e_lastmoved = 0;
```
because `x->gl_editor` is null

giuliomoro:
on further investigation, it also happens (at least on macos) when starting a full-fledged gui application but this patch is pased on the command line.
E.g.:
```
make
./bin/pd ~/_main.pd
```
bam, segmentation fault. `lldb` points to the very same line as the culprit

PR #2822, opened 2026-01-09
canvas_doclick(): exit if we don't have an editor and thus prevent a segfault

giuliomoro:
Hotfix for https://github.com/pure-data/pure-data/issues/2821 that at least prevents the crash. This fixes the symptoms, but there are probably better, more thoughtful solutions,


umlaeute:
i've cherry-picked this into the `develop` branch as 77b3b19a9f128dbcfbbd5d97ed4188be714ed0da

PR #2824, opened 2026-01-14
IRIX support

umlaeute:
this started as a fun project based on a user-request on the mailinglist to get Pd running on an SGI Octane.

there's a couple of general improvements (mostly to the build system), that I think  - while required to build on IRIX-6.5 - might be a good addition for Pd proper.
of course there's also highly specific SGI stuff (like re-introducing the s_audio_sgi.c and s_midi_sgi.c), which might be better left outside (given that practically nobody is using that anymore, we might consider it simply as cruft).

anyhow, I'm opening this WIP here, so there's a central place to document the changes (and allow for some discussion).

### build system
- autotools
  - use regular expressions for word-delimiters, rather than the non-POSIX `-w` flag
  - check whether `pthread_join()` requires linking against some pthread library
  - check whether the `strnlen()` function exists (apparently this is POSIX.1-2008, so relatively new) and provide (and eventually use) a fallback implementation
  - check whether `isnan()`/`isinf()`/`isfinite()` (C99) are usable, and otherwise provide fallback implementations
- i18n
  - more robust calculation of the `POFILES` to be generated (use `MSGFILES` (which are generated from the `POFILES`) as the base, rather than the arbitrary `ALL_LINGUAS`)
  - fix calculation of the `MSGFILES` list from `ALL_LINGUAS` in case some weird *make* does not properly handle the (undefined) `$(empty)` placeholder variable

### core
- replaced a leftover manual define for IRIX6 https://github.com/pure-data/pure-data/blob/5fc935fe725d88b3b54e4fa1c66eb60dcbf3d988/src/m_pd.h#L804 with defines that are automatically defined by the compiler
- fix detection of endianness on MIPS CPUs , in case there's no headerfiles to help us

### SGI specific stuff
- autotools: test for SGI-specific audio and MIDI headers
- re-introduce `s_audio_sgi.h` and `s_midi_sgi.h`
  - they were dropped around Pd-0.38
  - fix compilation with newer Pd
  - made it work with ADAT
- pd-gui.tcl: do not attempt to register a windows icon if `$::tcl_platform(os)` starts with `IRIX` (apparently Xsgi doesn't like that)

### other
- fix a simple typo for the (totally unused) `ESD` backend: https://github.com/pure-data/pure-data/blob/5fc935fe725d88b3b54e4fa1c66eb60dcbf3d988/src/s_audio.c#L436-L437
  (this should of course read `API_ESD`)

# remaining issues
- [x] ~~I haven't been able to test whether the SGI audio/MIDI system actually works. at least it doesn't crash~~ Audio OUT has been confirmed [to work on an SGI/Octane](https://lists.iem.at/hyperkitty/list/pd-list@lists.iem.at/message/JTFWMNSBMBYT3EF54A5HKN4J6VBAWZGH/)
- [x] ~~my Indy VM cannot properly print floats with `"%f"`; I do yet not know whether this is a problem specific to my VM, a problem due to compiling with `gcc` (rather than `MIPSpro`) or simply a bug in the libc installed on the system~~ this seems to be indeed a problem of the [emulator](https://forums.irixnet.org/thread-4793.html), and apparently does not happen on [real systems](https://lists.iem.at/hyperkitty/list/pd-list@lists.iem.at/message/SNH2FGJIYIKKYFKFONYAVO5JNGZKHQZ3/).
- [x] ~~IPv6 networking is currently broken and prevents a socket from being created correctly. the simple fix is to force the socket flags to `AI_PASSIVE` instead of the dual-stack code that is currently been used in https://github.com/pure-data/pure-data/blob/5fc935fe725d88b3b54e4fa1c66eb60dcbf3d988/src/s_net.c#L61-L68 but of course we do want to keep the current code (the fix is simple, *unless* you want to avoid adding cruft to the source code for a niche architecture)~~ added a `--disable-ipv6` flag to autotools, that disables the IPv6 stack.
- [ ] allow building with `MIPSpro` (probably this is illusory)

### DIFF
--- a/tcl/pd-gui.tcl
+++ b/tcl/pd-gui.tcl
@@ -311,7 +311,8 @@ proc init_for_platform {} {
             # some platforms have a menubar on the top, so place below them
             set ::menubarsize 0
             # trying loading icon in the GUI directory
-            if {$::tcl_version >= 8.5} {
+            # the X-server on IRIX (and IRIX64) cannot handle this, so we exclude that platform
+            if {$::tcl_version >= 8.5 && [ string first IRIX $::tcl_platform(os) ] == -1} {
                 set icon [file join $::sys_guidir pd.gif]
                 if {[file readable $icon]} {
                     catch {


umlaeute:
# build instructions

## pre-requisites
So far we know that it works on IRIX/6.2 (the last IRIX release).
You will also need a compiler (currently only the *GNU C Compiler* (`gcc`) is known to work; `MIPSpro` seems to **not** work). 

You will also need *GNU make* (typically invoked as `gmake`)

afaik, both are available on the "Freeware" set of CDs.

## Tcl/Tk

For a GUI we need a recent enough Tcl/Tk (at least 8.5; ideally 8.6 or higher).
IRIX only comes with Tcl-7.4p1, and Tk-4.0 (apparenty this is from a time, when the two versions were not synched yet).

Get both  Tcl and Tk from https://www.tcl-lang.org/software/tcltk/download.html (I used Tcl/Tk-6.6.17; at the time of writing Pd's Tcl/Tk-9.x is still somewhat experimental), and extract the two tarballs.
compile and install them.
**NOTE** it seems that some add-on packages, like `sqlite` fail to compile; as they are not needed, i just removed them

```sh
# remove Tcl packages we don't need
rm -rf tcl8.6.17/pkgs/tdbc* tcl8.6.17/pkgs/sqlite*
# build & install Tcl
cd tcl8.6.17/unix
./configure
gmake
gmake install
# build & install Tk
cd ../../tk8.6.17/unix
./configure
gmake
gmake install
```

## the sources

obviously get this PR (or just [branch](https://github.com/umlaeute/pure-data/tree/irix)).

## preparing the build system

The build system relies on `autotools`, `automake` and `libtool`, but IRIX does not have recent enough versions of these.

The simplest way to get usable sources is to prepare a dist-tarball on another host (e.g. a recent Linux machine), like so:

```sh
# instead of autogen.sh, you can simpyl run 'autoreconf -fiv'
./autogen.sh
./configure
# generate a dist tarball with the full build system
make dist
```

then copy the generated tarball (something like `pd-0.56.2.tar.gz`) to the IRIX machine and unpack it.

## configuring the build
```sh
./configure --without-local-portaudio --without-local-portmidi \
   --disable-ipv6 \
   --with-wish=/usr/local/bin/wish8.6 \
   --with-external-extension=pd_irix6 \
   --with-deken-os=Irix --with-deken-cpu=mipsn32
```

#### flags explained

- `--without-local-portautio --without-local-portmidi` -- I'm not sure whether this is actually needed; but PortAudio (resp. PortMidi) cannot handle the SGI subsystem, so there's no use to even try building it (and it involves C++, which probably won't work with the very old compilers on the system
- `--disable-ipv6` disables IPv6 networking (in theory IRIX supports IPv6, but it seems that it is not functional)
- `--with-wish=/usr/local/bin/wish8.6` -- Tcl/Tk's "gmake install" installed wish to `/usr/local/bin/wish8.6`, and Pd tries to just open "wish" by default, hence the "--with-wish" flag.
  alternatively you could just make sure that /usr/local/bin is in your PATH, and symlink "wish8.6" to "wish".
- `--with-external-extension=pd_irix6` -- by default, Pd would use `.so` as the filename extension for externals; traditionally, Pd uses `pd_irix6` on IRIX/N32.
- `--with-deken-os=Irix --with-deken-cpu=mipsn32` -- totally optional and actually only used for deken (there probably aren't any Irix specific packages uploaded yet, but who knows)

## building

```
gmake
```

sometimes, I get failures when building the `po/` directory (some unknown flag to `xgettext`). just running `gmake` again seems to do the trick.

## start Pd
```
./bin/pd
```

(if things go awry, use "./bin/pd -verbose -verbose -stderr" to see what's wrong)

umlaeute:
@millerpuckette I do not know your position about supporting platforms like IRIX (with *supporting* I mean "having code that in theory should work", rather than "being committed that this platform actually works").

i believe that most of the changes in this PR are actually general improvements (mostly catching some corner cases), and should be included.

the changes that are really SGI specific come last (starting with 404314403969cc78b4e490a57f0f2b615f82db76 "Re-introduce s_audio_sgi.c and s_midi_sgi.c"), so if have strong feelings about not including mostly dead code, you could omit these (but then: apart from the single line in pd-gui.tcl, the changes are really only the s_audio_sgi.c and s_midi_sgi.c, so at least the mostly dead code is contained)

millerpuckette:
yeah.  Time to merge it :)

wogaut:
Yes, please merge the IRIX code, it works well on my SGI Octane (tested on actual hardware)! Really appreciate the effort, time to bring IRIX back!

umlaeute:
it is already merged! 

thanks miller

Issue #652, opened 2019-06-08
proposal for a new hybrid iemgui: [sradio]


fdch:
I have been working on this GUI external that is pretty messy and has a lot of copied code inside that could be better and more efficiently loaded. But I thought I share it here and see if anyone might be interested in something like this. It works pretty well and afaict it is pretty robust.
![sradio-help](https://user-images.githubusercontent.com/13335792/59152467-bc1df580-8a44-11e9-98a3-1a87d383be5d.jpg)
 

It is a combined hack of `[hradio]` and `[text]`, and it acts as a sequence of toggles that can be set by clicks or by set message passing a list. With <key>shift</key> + click you open its text file for further edits, and it has a store/recall methods for quickly switching between patterns. Next/prev messages advance the sequence by one step, and the output is the stored value (0 or 1, as in a toggle)

It is a lot of code to move around, so I thought I should ask first if this is something that would be of interest for anyone other than me. It is a nice tool to make patterns on the fly and to read them in sequence.

The code for now is here https://github.com/fdch/fd_lib/tree/master/sradio

Cheers!

Spacechild1:
Hi, I think it would be more interesting to actually have a 2D grid of cells (so your [sradio] would be just a special case of a grid with a single row) + methods to set/get individual cells as well as whole rows and columns. At least this is what I would want to do immediately.

fdch:
I agree. In fact, that is exactly what it is, only that you can only see/access one row at a time. I have been using it as you say, just stacking different same-length [sradio] to have that matrix-like GUI and synch all of them with the same metro, for instance

porres:
I have thees abstractions as a way to deal with 2D toggle arrays 

https://github.com/porres/pd-else/blob/master/Classes/Abstractions/drum.seq.pd
https://github.com/porres/pd-else/blob/master/Classes/Abstractions/mtx.ctl.pd

jmmmp:
You can also check [jmmmp/matrixctrl] and use them with only 1 row. Not exactly the purpose of that object, but can work.

fdch:
The object in question [is included on my fd_lib library](https://github.com/fdch/fd_lib/blob/main/src/sradio.c), so I'll close this.

porres:
> The object in question [is included on my fd_lib library](https://github.com/fdch/fd_lib/blob/main/src/sradio.c), so I'll close this.

it crashes Pd for me....

sebshader:
[...]

PR #2826, opened 2026-01-16
Iemgui object extension proposal: Radio Matrix

fdch:
This is a new object belonging to the IEMGUI `radio` family that instantiates a 2d grid on the canvas.
It stems from a discussion on an external in https://github.com/pure-data/pure-data/issues/652,
but now it can be available as an internal.

### Concept
The concept is pretty straightforward: to enable a 2d grid of toggles (i.e., crosses) with individual access.
The look of the object is basically that: a grid of crosses, with the filled square representative of the `radio` family.
In the 2d grid, the filled square represents the *index* of either the *columns* or the *rows* depending on orientation.

### Screenshot:
image

### Compatibility with old versions:
The PR itself introduces significant changes to `src/g_radio.c`. I know that's not great for reviewing, but I hope it's not to painful to see the diff (I apologize in advance ;). Anyhow, `radio` should remain completely backwards compatible with `hradio` and `vradio` (even to pd version <0.36). Or, that's at least what I was able to verify manually on current hardware.

### Help wanted for testing:
I've developed and tested it extensively on debian. But, I only compiled and minimally test it on macos and windows. So, I still need help to make sure this is 100% backwards compatible, *and that it actually works*. I added a reference file to the doc directory, hoping this gets accepted and moved properly into the radio help file. On the reference file, you can find also an *example* using this grid object as a basic step sequencer.

#### Radio Matrix (new object)
- Create it by typing `radio` in an object.
- Shift-Click on squares to toggle them ON or OFF.
- Click on columns (horizontal orientation) or rows (vertical orientation) to move the index
- Right-click + Properties to change dimensions or orientation
- From `hradio` or `vradio` you can now go back and forth to `radio`.
- Same messages as `hradio` or `vradio` (number, orientation, iemgui messages, etc)
- New Messages:
  - "resize": 2 floats, columns and rows: resize the matrix to the new size (columns, rows)
  - "cell": 1 float: get the value at that index
  - "readline": 2 floats, axis and index: read a line along an axis (1|0) at the given index.
  - "get": 2 floats, column and row: get the value at that location (column, row)
  - "dump": output the entire matrix as a list to the outlet (or to the provided symbol if provided)
  - "fill": a float with which to populate the entire matrix
  - "clear": no args: identical to "fill 0"
  - "list": list of floats (1|0) with which to fill the matrix.
  - "writeline": 3 floats, axis, index and value: writes a value to the line along an axis (1|0) at the given index.
  - "column": 1 arg: get column at index / 2 args: set value along a column at index
  - "row": 1 arg: get row at index / 2 args: set value along a row at index
  - "put": 3 floats, column, row and value: set the provided value at that location (column, row)
  - "step": 1 float: increment the index (filled square) by the provided value
  - "next": no args: identical to "step 1"
  - "prev": no args: identical to "step -1"
  - "border_mode": a float to define action at border (clip: 0, modulo: 1)
  - "output_mode": if 1, first send out a list at the current index, otherwise output is always the index

#### Properties Dialog:
- 2 arguments are added for iemgui dialogs, so that Radio can change Columns and Rows, and Orientation ("order").
This affects calls on all iemgui objects, so their calls were also adjusted to be offset by 2.
- radio: Columns is the old "num" and Rows is the new "num2".
- slider: Width/Height dimension logic was kept, but Orientation was also added to the properties dialog.
- hot swapping columns/rows or width/height when orientation is selected

#### Menu entry
- Created via `Put > Matrix`

#### Key binding
- Created via <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>X</kbd> or <kbd>Cmd</kbd>+<kbd>Shift</kbd>+<kbd>X</kbd> (macos)

### Small demo
example-radio-matrix


umlaeute:
cool, but why is it called `[radio]`? isn't the the toggle-matrix functionality decidedly "non-radio"?

also i'm not sure about the radio-functionality being 1-dimensional only, whereas the toggle-functionality is 2d.

for me it would make somewhat more sense to split the functionality into a toggle-matrix and a radio-matrix.
i understand that you want to be able to overlay them (and overlying two independent GUI objects will make it a bit complicated to interact with both).

fdch:
> but why is it called [radio]? 

Because the orientation is encoded in the size, not the name, and `matrix` or `mtx` were too zexy :). I thought of `grid` as well, or `mradio`, but it made more sense to just call it like the filename as it is meant to be a more general version of the original. 

> isn't the toggle-matrix functionality decidedly "non-radio"?

As it is now, you can still draw toggles on `hradio` or `vradio`, and you could get some use out of that as a single list.

> split the functionality into a toggle-matrix and a radio-matrix.

I'm curious about this distinction, could you elaborate?

> overlying two independent GUI objects will make it a bit complicated to interact with both

Yes, but we don't need to interact with both at the same time; it's the same object just in a different state.

fdch:
Ok, I think my previous replies did not address your concerns at all. Sorry for the noise! Here's a better reply, but please correct me if I'm wrong:

Let me see if I understood correctly. You are proposing to split into radio and toggle functionality. And you also suggest both can be implemented in 1 and 2 dimensions (I think). 

As I see it, your suggestion would allow for these configurations (I might be missing some):

1D radio line:
- mutually exclusive filled square along a line (current hradio/vradio)

2D radio:
- stacked 1d radio lists (per-row radio):
  - so you can only select one option per row (or column if vertical)
  - we'd have as many filled squares as rows, they would all initially be set to the first column (unless init'ed), and the user can click to change its position to another column. 
- single 2d radio matrix (global index)
  - we'd have one filled square that can be anywhere and outputs (column,row) or index

1D toggle line:
- single list of independent toggles:
  - we click on the toggles to change them
  - we bang it to get the whole list out of it. 
  - we send a number and use it as an index to the list to get a stored value out of it 

2D toggle matrix:
- stacked 1d lists of independent toggles:
  - pretty much what the current proposal is, but without the filled square 

In this scenario, the current radio proposal would be a combination of the 2D toggle matrix and the 1D radio line :S

I get your point then, that interacting with such a hybrid object might be complicated and rather unintuitive. Also, that the name does not tune in well with what the user would expect of a radio. 

I'm open to suggestions for changing, improving (*), and renaming as needed, as I think it would be good to have a native matrix in pd.

(*) I'm also thinking that the toggle matrix does not necessarily have to be binary, so values can easily be `t_float` but shown as crosses when filled.

fdch:
I've updated the object to contain both stacked radio and toggle matrix functionality. The included pd file shows all messages and has audio examples for both cases.

Issue #1010, opened 2020-05-15
request: resize guis


porres:
Pd allows us to resize objects, graphs/arrays, messages and comments when in edit mode. I'd like to expand this functionality to resize GUIs (iemguis+gatom).

Also, we could add vertical resizing for graphs/arrays, and then also iemguis. Moreover, on the bottom right corner, one could resize both horizontal and vertical axis.

gatom would just resize its width / a bang and toggle would only resize on the bottom right corner (for both width/height). A canvas could resize its visible rectangle (and not its "actual - selectable - size", or maybe both, but never only its "actual size")

Issue #2656, opened 2025-07-05
Provide an API for resizing external objects


porres:
Right now I can resize an object in a rather clumsy way in my oppinion, by creating a clickable window in edit mode in objects like [cyclone/comment] and [cyclone/scope~].

Some Pd native objects have a more intuitive and clean way of doing this, but it's hardcoded and not exposed by the API. This is taken care in "canvas_doclick" 9 https://github.com/pure-data/pure-data/blob/2a30fe0fbebdf9068bb518ecd1e02449529488db/src/g_editor.c#L2313 ) by the way. See https://github.com/pure-data/pure-data/blob/2a30fe0fbebdf9068bb518ecd1e02449529488db/src/g_editor.c#L2507 for instance...

It's certainly easy to work with clicking on GUI externals in run mode. In edit mode all it happens is that the object gets selected. It'd be nice if we could also work with clicking and decide wether we can change the cursor and resize the object.

This may also be related to https://github.com/pure-data/pure-data/issues/2655 which is a request for an easy way to react to edit status change

This is related to https://github.com/pure-data/pure-data/issues/1010 as we could use this system for iemguis as well.

Also related is a long and very confusing discussion of me trying to find my way and figure things out in the dev list - see https://lists.iem.at/hyperkitty/list/pd-dev@lists.iem.at/thread/QQRHHKMZDJORHSWJ63Z4NKAORFHNTDEL/

PR #2827, opened 2026-01-18
Allow mouse triggered IEMGUI resizing

fdch:
Here's a smaller PR that allows click+drag resizing of IEMGUI objects. Slider, bang, and toggle use the same general function. The rest of the objects needed more specific functions. Resizing together with <kbd>Shift</kbd> clamps the size to a square.

Please let me know what you think or would like to improve, I'm open to anything.

#### Demo without zoom
iemgui-resize

#### Demo with zoom
Note how the handle remains at the center. This is somewhat weird, but manageable.
iemgui-resize-zoomed

### DIFF
--- a/src/g_numbox.c
+++ b/src/g_numbox.c
@@ -25,9 +25,21 @@ static void my_numbox_clip(t_my_numbox *x)
         x->x_val = x->x_max;
 }
 
+static int my_numbox_fontsize_increment(t_my_numbox *x)
+{
+    int f = 31;
+
+    if(x->x_gui.x_fsf.x_font_style == 1)
+        f = 27;
+    else if(x->x_gui.x_fsf.x_font_style == 2)
+        f = 25;
+
+    return (f);
+}
+
 static void my_numbox_calc_numwidth(t_my_numbox *x)
 {
-    const int fi = IEM_NBX_FS_PROPORTIONS[x->x_gui.x_fsf.x_font_style];
+    const int fi = my_numbox_fontsize_increment(x);
     const int zoom = IEMGUI_ZOOM(x);
     int nw = x->x_gui.x_w / zoom - 4 - (x->x_gui.x_h/2)/zoom;
     nw *= IEM_NBX_FS_FACTOR;
@@ -37,7 +49,7 @@ static void my_numbox_calc_numwidth(t_my_numbox *x)
 
 static void my_numbox_calc_fontwidth(t_my_numbox *x)
 {
-    const int fi = IEM_NBX_FS_PROPORTIONS[x->x_gui.x_fsf.x_font_style];
+    const int fi = my_numbox_fontsize_increment(x);
     const int zoom = IEMGUI_ZOOM(x);
     int w = x->x_gui.x_fontsize * fi * x->x_numwidth;
     w /= IEM_NBX_FS_FACTOR;
--- a/src/g_all_guis.h
+++ b/src/g_all_guis.h
@@ -82,7 +82,6 @@
 #define IEM_MAX_SYM_LEN 127
 
 #define IEM_NBX_MINDIGITS 1
-#define IEM_NBX_FS_PROPORTIONS ((const int[]){31, 27, 25})
 #define IEM_NBX_FS_FACTOR 36
 
 #define IEM_GUI_DRAW_MODE_UPDATE 0


porres:
Awesome, way overdue, and closes https://github.com/pure-data/pure-data/issues/1010

I just don't think we want a bang that can be resized bi dimensionally, looks like an extra feature. I'd have a double sized diagonal arrow to set the size only. Same with toggle. Exclusively horizontal and vertical resizing would also be nice for something like a slider, so if you drag from the right handle, you'd get just horizontal resizing, bottom is vertical and bottom right corner would be both.

umlaeute:
i agree that `[tgl]` and `[bng]` are currently forced to square shapes , and changing this is a separate issue.

(since both objects only store a single dimension in their binbuf, the non-square shapes are only valid for the just-changed object. if you save the patch and reload (or just copy the object), the 2nd dimension is lost)

fdch:
Nice! Thanks for pointing out that issue.

Right, I forgot toggle and bang have only one size. I'll change it so that it's always square for these.

Also `cnv` has one size for the clickable box and looks weird after saving when the width is larger than the height. It wouldn't be such a pain to slide a specific fix for `cnv` so that the box goes back to size 20 if that happens after saving... or to clamp it to the size of the vis.

> I'd have a double sized diagonal arrow to set the size only
> ....  
> Exclusively horizontal and vertical resizing 

This sounds super nice and more like an advanced gui interaction that I suspect I cannot really do here without changing quite a lot. 

Maybe I can clamp `hslider` resizing to horizontal dragging if shift is pressed (or vertically in `vsliders`). What do you think?

fdch:
Linking https://github.com/pure-data/pure-data/pull/1039 here, because it would be great to merge that first and then adapt this PR to those changes.

fdch:
OK, here it goes. The PR grew a bit, but I think for good reason. Please let me know what you think.

Here's new demo gifs showing the changes:

### Radio
- resize changes the size of the square (and hence the total size)
- <kbd>ctrl|cmd</kbd> + resize: changes the internal number

resize-radio

### Slider
- resize is locked to orientation
- <kbd>shift</kbd> + resize: makes it a square resize
- <kbd>ctrl|cmd</kbd> + resize: freeform resize

resize-slider

### Numbox
- resize is locked horizontally, but changes the number of digits
- <kbd>shift</kbd> + resize: locks number of digits, but changes height and fontsize
- <kbd>ctrl|cmd</kbd> + resize: (_alla_ freeform) changes height, fontsize, and number of digits

resize-numbox


### MyCanvas
- resize is by default freeform
- <kbd>shift</kbd> + resize: makes it a square resize
 
resize-canvas


### Toggle, Bang, Vumeter
(no modifier keys)
- toggle and bang: resize defaults to square resize
- vu: resize changes width and height, but height is clamped to the scale (as it was)
 
resize-others

porres:
this is somewhat related and while we're at it, we could think of a way to also expose a mechanism and allow externals to be resized as well. See https://github.com/pure-data/pure-data/issues/2656

It'd be great if I could utilize the same procedure for my GUI externals to be resized. Do you think you could also help me out and work on that separately @fdch ? thanks

fdch:
> expose a mechanism and allow externals to be resized as well. 

Can you give me an example I can use for an external that would need resizing?

My 2cents here: I guess the external would need a method for handling the resize, and pd would need to know that it is an external that can be resized... Right now, I made it so that if the clicked object is one of the iemgui widgets, it runs its resize function. And I made sure each iemgui has its own resize function. Maybe we need to do this more dynamically, so if the object exposes a resize function pd detects it, but I'm not sure how much of a can of worms that would open.


> Do you think you could also help me out and work on that separately 

Sure!

src/g_numbox.c
```diff
@@ -30,30 +27,22 @@ static void my_numbox_clip(t_my_numbox *x)
...
-
-    if(x->x_gui.x_fsf.x_font_style == 1)
-        f = 27;
-    else if(x->x_gui.x_fsf.x_font_style == 2)
-        f = 25;
-
-    w = x->x_gui.x_w / IEMGUI_ZOOM(x) - 4 - x->x_gui.x_h / 2 * IEMGUI_ZOOM(x);
-    w *= 36;
-    x->x_numwidth = w / (x->x_gui.x_fontsize * f);
+    const int fi = IEM_NBX_FS_PROPORTIONS[x->x_gui.x_fsf.x_font_style];
```
umlaeute: this looks like a super-easy way to get an out-of-bounds array access. i guess it should at least do some array size checking before (which would unfortunately reduce readability back to the original code, or beyond...)

fdch: Yes, indeed! Good catch. I'll put the if/else as it was, but into its own function.

porres:
> Maybe we need to do this more dynamically, so if the object exposes a resize function pd detects it, but I'm not sure how much of a can of worms that would open.

I don't really understand well the resize mechanism in Pd, but I know can be aware of the area when hovering so you can set the cursor when within the resizing part - and can capture clicks in this case. So i'm thinking of something general like that.

I already tried asking about how it works in https://lists.iem.at/hyperkitty/list/pd-dev@lists.iem.at/thread/QQRHHKMZDJORHSWJ63Z4NKAORFHNTDEL/#QQRHHKMZDJORHSWJ63Z4NKAORFHNTDEL

fdch:
From IOhannes on the thread you mention:

> it's not easy to change it to make more objects do something class-specific in edit mode, without cluttering rather unrelated code (g_editor), or - possibly - resorting to even uglier hacks

I'm afraid that's exactly what I did: clutter the `g_editor` code :) 

Nevertheless, my point is that currently pd has a big if/else on the `g_editor` that allows it to change the cursor and run an object's resize method. What I did is add another case on that if/else, one that is specific for iemguis. I don't know how an external would be detected so that we can add another case and run its resize function. That's because the widget behavior needs to be present on the object: regular objects have a `text_widgetbehavior` and iemgui ones have `slider_widgetbehavior` (for example, in the slider case). 

To allow for some kind of external use:

Maybe we can have a new `external_widgetbehavior` type of pointer that can be defined by externals and that would allow `g_editor` to trigger its resize method (which should also be defined by the external). 

Another way would be to actually look for a `resizefn` and if it is defined, pd runs it. That would be a more dynamical approach and would make that big if/else a bit leaner. But, it would be a bit more involved as a change.

millerpuckette:
I think this would eventually be a good thing to add... but perhaps a bit later once the gui-independence (aka the gtk option) is more settled.  For example, since any gui-bearing extern would have to be adapted, perhaps that's when the size of a widgetbehavior could be changed conditionally on not being tk - then the changes become 'standard' for non-tk implementations.

Issue #2831, opened 2026-02-09
Request: A new object to handle dictionaries


porres:
This has been discussed on the Pd list and it seems [text] is not powerful enough to handle this new data type. And I'm thinking of MAX's [dict] as a reference here.

Somewhat related: This would also allow us to have nested lists, but maybe we could think of yet another object to handle it? See https://github.com/pure-data/pure-data/issues/505#issuecomment-1400748312 / also also somehat related --> https://github.com/pure-data/pure-data/pull/875

tremblap:
nested dicts is also important, and api access to that memory structure for externals. JSON parsing for files to save/read... so this is not trivial a job, but https://github.com/nlohmann/json is great for the latter.

Issue #2794, opened 2025-11-06
[text] adds a backslash to visualize brackets/curlies


porres:
I don't understand why this happens, feels like a bug to me.

```
list [1.5 [1 2]] [1 2] [ringo] paul
```

becomes

```
list \[1.5 \[1 2\]\] \[1 2\] \[ringo\] paul
```

this is not visually present in list boxes, etc...

<img width="353" height="194" alt="Image" src="https://github.com/user-attachments/assets/8726b80d-250a-475e-9e7c-03782b640b95" />

ben-wes:
this report looks a bit funny since markdown interprets `\` as an escaping character and doesn't display it in your text (you might replace `\` with `\\` - although it's clear on the screenshot anyway). :)

umlaeute:
(updated the markdown to use code-blocks which keep the backslashes)

porres:
Same happens to curly braces by the way - I can't think of other characters, I believe there might be no other examples, but I wouldn't be surprised if we find more.

porres:
This is a minor inconvenience, nothing breaks on the patch, but it'd be nice to get rid of it. I tried my best, but what I see in 'pdtk_textwindow_appendatoms' is that the backslashes aren't added there. I can print that it receives, sends and reads back a proper symbol without escaping.

I don't know what to do. I don't know why a list box manages this correctly but not this. I can't see where it is added.

Now, anyway, what I just realized is that this is related to https://github.com/pure-data/pure-data/issues/2151 !

The problem with this other issue is that reopening re-adds yet more backslashes, so it's really a bug.

porres:
I will reboot this issue to also include handling of these characters in imeguis and atom boxes

Issue #2151, opened 2023-11-30
iemgui labels show escaping backslash for brackets "[]", curly braces "{}" and "\". Atom boxes have similar issues.


porres:
see below
![Screen Shot 2023-11-29 at 01 45 23](https://github.com/pure-data/pure-data/assets/5360142/268b37fd-3029-4d7c-9a0e-d44002850bec)

originally [py4pd] did not have backslashes, and now if you save properites, they get added to the label

The funny thing is that the backslash itself also adds a backslash

porres:
So, this relates to https://github.com/pure-data/pure-data/issues/2794 but at least, when reopening the window over there, more backslashes are not added!

porres:
this is not only for labels, brackets also get scaped in receive names for instance (but curlies simply disappear there)

porres:
This is not only for labels, brackets also get scaped in receive names for instance (but curlies simply disappear there)

This is also not only for iemguis, atom boxes will also have "[]" escaped in labels, but "{}" in labels won't even stick, which is how it goes for send/receive names in iemguis.

porres:
I have a better understanding of the issue now and will simply reboot it

PR #2835, opened 2026-02-11
allow typing "{" and "}" in Pd patches.

### LINKED ISSUE #505: [feature request] curly brace support
hi, 

I saw that backslash are now supported which is great, thanks @millerpuckette 
would it be possible to also support opening and closing curly brace ?
those are really useful in some syntax context like regexp

for example, with libossia, man can control several parameters at once with [ø.remote foo.{1,2,4,15} ]
which is quite convenient.

for now we have a workaround but curly braces are so nice :-)

porres:
Typing these are already possible in PlugData and even in Vanilla's [text]. Support is already given but not fully practical because we can't type it, even though we can generate these symbols with [makefilename %c] and insert it in messages.

- closes https://github.com/pure-data/pure-data/issues/505

The issue above has detoured a bit in my opinion on a parallel discussion about expanding Pd's syntax. In the end, curlies are already partially supported and it's not trivial to use them for a new syntax. Also, it's not like we can't expand Pd's syntax in other ways and this has been discussed in the referenced issue. I guess the discussion ended up being an unnecessary obstacle in the way.

Also, somewhat related, there was another PR about expanding the syntax - see https://github.com/pure-data/pure-data/pull/875

### DIFF
--- a/src/g_editor.c
+++ b/src/g_editor.c
@@ -3165,11 +3165,6 @@ void canvas_key(t_canvas *x, t_symbol *s, int ac, t_atom *av)
     else gotkeysym = gensym("?");
     fflag = (av[0].a_type == A_FLOAT ? av[0].a_w.w_float : 0);
     keynum = (av[1].a_type == A_FLOAT ? av[1].a_w.w_float : 0);
-    if (keynum == '{' || keynum == '}')
-    {
-        post("keycode %d: dropped", (int)keynum);
-        return;
-    }
 #if 0
     post("keynum %d, down %d", (int)keynum, down);
 #endif


porres:
So, yeah, I tested it, and it simply works :) 

Screenshot 2026-02-11 at 09 41 23

Spacechild1:
Yay! Now I could finally type Lua, JS, JSON etc. in Pd text boxes :)

I guess dropping the keycodes came from a time when messages sent to the GUI weren't properly escaped.

porres:
By the way, there are other cases where typing curlies is actually already allowed these days besides the [text] window. For instance, as the label of iemGUIs and other windows, even though the Pd window still triggers things like "keycode 125: dropped".

And we could also do it in externals like [cyclone/coll]. And I didn't know about this, but [else/messbox] already allowed us to type it in a message box like interface!!! This is a GUI that you can type when in run mode by the way.

Screenshot 2026-02-11 at 15 50 24

So yeah, this PR simply removes this last obstacle...

porres:
> I guess dropping the keycodes came from a time when messages sent to the GUI weren't properly escaped.

yeah, and this is well taken care of now, but there are a couple of issues still, see https://github.com/pure-data/pure-data/issues/2794 and https://github.com/pure-data/pure-data/issues/2151

umlaeute:
i'm afraid i cannot resist to detour this again.

i still think that this is wasted opportunity for extending the syntax of messages.

to be honest, i think the proposal of @Spacechild1 to use `${` instead was kind of weak (at least not really stronger than mine): if it is impossible to assign meaning to `{` because "people are already using it", then the same goes for `${` (that is: it is possible to use that right now, and assigning special meaning to it will break this use case).

(but then: my proposal to have `[list unflatten]` and `[list flatten]` can work even when `{`/`}` are ordinary symbols)

porres:
Could we discuss this back into a new discussion or something? One way or another, this is a PR to allow us to type this character, what it'll mean and do is something beyond it I think!

My take is that I think a good use case of "{" and "}" is to generate code blocks for things like lua and I'm now particularly interested on it for a [dict] external, for JSON like syntax. I find it weird that it would be used for a syntax expansion, mostly cause I don't really expect a syntax expansion and new data types. Seems like a big language overhaul and we'd need many objects to be aware of this new data type for nested lists.

> if it is impossible to assign meaning to { because "people are already using it", then the same goes for ${ (that is: it is possible to use that right now, and assigning special meaning to it will break this use case).

I can argue that '${' is definitely safer than '{', which is not only possible, but definitely being used and with a reasonable meaning/use case, while the usage of '$' seems unquestionably something reserved for syntax in Pd, hence there are other parallel discussions for things like "$@", or even the PR https://github.com/pure-data/pure-data/pull/875 with even more stuff and even a suggestion from you to use something like "${2:3}".

> (but then: my proposal to have [list unflatten] and [list flatten] can work even when {/} are ordinary symbols)

yup, and then we'd have a particular object which has its own syntax (and maybe a new object just for nested lists, like [nlist] or something). That makes sense to me as a Pd-like thing. For instance, a symbol that starts with "#" is treated as a hexadecimal value to set colors in iemguis, while Pd has no actual hexadecimal data type, not to mention some externals that handle complex data and already do things like that (ceammc handling nested lists with "(" and ")" with a lisp like notation comes to mind in this context, as it's the exact same idea and didn't require a syntax addition/overhaul).

I could argue we can do a lot with special objects and just keep things as is. MAX has a dictionary object, but no special syntax or a special data type for it. There are externals that deal with nested lists, dicts and other data types without the need of a native syntax expansion.

One way or another, we cannot prevent this to be added just because we can't decide or something and I don't mean to get in the way. Let's just vote, settle and deal with this (in a discussion somewhere else here on on the list) and move on as soon as we're all 'ok' with it :)

Spacechild1:
> i still think that this is wasted opportunity for extending the syntax of messages.

How is this a missed opportunity then? Being able to type these characters is the *prerequisite* for any of these proposals. 

I agree with @porres  that the actual proposals should be discussed in separate issues/PRs. This PR is really just about allowing curly braces keycodes. AFAICT it's a trivial change that should have done years ago (when we started to properly escape them in the Tcl/Tk messages).

danomatika:
I also think it would make sense to add a focused support for these characters *before* extending syntax. I feel like the dictionary discussion went in so many directions that it's not entirely self-evident *what* might actually be needed at this point.

porres:
> I also think it would make sense to add a focused support for these characters before extending syntax

Sure, but only if we're actually extending it :) so we must decide on that for good. And also we can still expand the syntax in other ways as discussed ;)

> I feel like the dictionary discussion went in so many directions that it's not entirely self-evident what might actually be needed at this point.

Yeah, that was noisy, but it just requested a native json file/data structure. The need for native support is just so Flucoma would not bother with externals - here's a request for it anyway https://github.com/pure-data/pure-data/issues/2831 

But yeah, there are externals that handle it, and maybe we don't need to bother with native support. I mentioned ceammc's support for nested lists with "( )" but let me add is also has a dictionary data type (see [data.dict])! Dictionary nesting is then supported by "[ ]". This reinforces how other data types are possible without a special native syntax, and also without relying on this special character.

Spacechild1:
I want to stress again that curly braces have never been a special character in FUDI messages. It's just that you couldn't type them on the patch level. However, we have always been able to create such messages programmatically. That's why I think this PR is really a no-brainer.

ben-wes:
> wasted opportunity for extending the syntax of messages.

^ this made me hesitate to state my opinion on removing the current input filter - but right now, i think that we'd probably profit more from just allowing these characters so that at least externals and abstractions can make use of them.

... concerning the current opportunity, i feel like that ship has already sailed with the option to type them into text windows and (maybe not the strongest argument, but still relevant) with plugdata/purr-data not applying such a filter.

besides that, i think that extended syntax might be great. and that there should possibly be a difference between `{ }` and `\{ \}` (but i admittedly did not think that through).

millerpuckette:
I had to think about whether there was a good reason to use '{' and '}' as syntactic characters, perhaps to allow structured message box contents... but I think it's a can of worms.

OTOH, there's an opening in messages boxes to interpret messages starting with numbers to mean to output to an outlet by its left-to-right position.  Hmm... :)  Anyhow, time to merge this I guess.

PR #2837, opened 2026-02-13
drop backslash in iemgui's labels instead of keeping them

### LINKED ISSUE #2836: iemgui's labels don't manage backslashes properly, they must be dropped as in other cases.
For send and receive names, if you type something like "\hi", the backslash get immediately dropped. The same is true for atom boxes for send, receive and labels. Nonetheless, iemgui's labels don't drop and keep it.

Under the hood, in dialog_iemgui.tcl, send and receive names go through "unspace_text",

see 
https://github.com/pure-data/pure-data/blob/master/tcl/dialog_iemgui.tcl#L272

and see https://github.com/pure-data/pure-data/blob/b35ff302ab3bc422881654fc3dc934694671e7a3/tcl/wheredoesthisgo.tcl#L197

This is not commented on the function, but it also drops backslashes. I haven't really checked how atom boxes deal with it, but it doesn't matter, it also drops and you can't produce a label such as "\hi", but in iemguis you can.

Nonetheless, this generates some issues. Even though you can see "\hi" if you enter it as a label, if you save the patch, you can't actually restore the label as "\hi". I didn't know, but you need to save a backslash as "\\\\" in the label argument so it survives opening the patch.

I am not suggesting to fix how to properly save and restore the backslash, what I want is to simply drop it like it happens for atom boxes labels and everything else, so it is consistent.

Another issue is that when you get to set a label as "\hi", if you recall the properties, another backslash is added, so yo usee "\hi" in the label entry. If you change something else in the properties and save it, or if you save it without changing anything, you now accidentally added another backslash, so your label is now "\\hi". Reopen properties and now you have "\\\\hi", and this will keep happening forever an growing exponentially long as you keep doing it. So, as you see, a lot of trouble happens...

Also, for some reason, a reasonably allowed character like brackets do, for some reason, get a backslash when opening properties, so "[ok]" becomes "\[ok\]". I can argue it shouldn't do that, but this is not a problem for atom boxes or send/receive names in iemguis and atom boxes, since the backslashes are dropped. So dropping the backslash in iemgui's labels is something that manages this reasonably well and keeps things consisten. It is also a minimally intrusive fix.

Anyway, this bug bit me when I tried to so this into a canvas, and I wanted to use brackets to mean that "py4pd" was an object name, as we normally do.

Image

And then I noticed that rditing the canvas and saving would keep the backslashes.

The fix is simple, I see that the label is specially handled in this line of the code

see
https://github.com/pure-data/pure-data/blob/master/tcl/dialog_iemgui.tcl#L258

And the code actually escapes and keeps the backslash, while it should drop it 'unspace_text' does. I tested editing this mapping and it all works nicely now. I am sending a PR right away.

porres:
- closes https://github.com/pure-data/pure-data/issues/2836

### DIFF
--- a/tcl/dialog_iemgui.tcl
+++ b/tcl/dialog_iemgui.tcl
@@ -255,7 +255,7 @@ proc ::dialog_iemgui::apply {mytoplevel} {
     if {$::dialog_iemgui::var_rcv($vid) ne ""} {set receivename $::dialog_iemgui::var_rcv($vid)}
     if {$::dialog_iemgui::var_label($vid) ne ""} {set labelname $::dialog_iemgui::var_label($vid)}
 
-    set labelname [string map { "\\" {\\} {$} {\$} { } {\ } {,} {\,} {;} {\;}  "{" "\{" "}" "\}" } $labelname]
+    set labelname [string map { "\\" "" {$} {\$} { } {\ } {,} {\,} {;} {\;}  "{" "\{" "}" "\}" } $labelname]
 
     # make sure the offset boxes have a value
     if {$::dialog_iemgui::var_label_dx($vid) eq ""} {set ::dialog_iemgui::var_label_dx($vid) 0}


porres:
@umlaeute this seems to me like a simple no brainer fix to be included in the develop branch, what do you think?

PR #2839, opened 2026-02-17
Add methods and warning for [declare]

### LINKED ISSUE #281: [declare] request: warn that the library/path isn't found
When loading anything with [declare], it never tells you if it succeeded or not.

I think that an error message/warning in the console is helpful, maybe even something saying that it was successfully loaded

### LINKED ISSUE #660: feature-request [declare]: print added absolute paths in the verbose console
It would be quite useful for debugging to know the actual absolute paths added to Pd by a `[declare]` object when relative paths are included with the object. These paths could be printed to the console in verbose level.

### LINKED ISSUE #886: Introduce an internal object which queries the patch window's blocksize
As discussed on the mailing list in the past years it seems there is currently no way to query a patches blocksize from within Pd using internal objects. The existing [block~] object could be made to report this at an outlet after having received a bang message.

fdch:
### declare iolets and methods
This PR adds an inlet to [declare], and some methods: a "bang", "paths [0,1]", and "libs":
- The "bang" allows the user to re-run the object without re-creating it (or closing and opening the patch). 
- The "path [0,1]" method outputs the loaded paths (0 or no argument) or the absolute paths (1) on a canvas.
- The "libs" method outputs the loaded libraries (i.e., loaded classes and abstractions) on the instance (whatever libs were loaded (via s_loader.c))

### declare warnings
The warnings are all on the debug level (3), and they are only clickable when triggered by the object. If the canvas triggered the declares (e.g., on patch load, via message), they won't be traceable. 

Please let me know what you think, it's not set in stone. The PR closes https://github.com/pure-data/pure-data/issues/281, in the hopes that the solution is sane and in line with the op.

- Closes: #281
- Closes: #660
- Closes: #886

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -1626,49 +1626,37 @@ typedef struct _declare
     int x_useme;
     t_atom *x_argv; /* store creation arguments for re-triggering */
     int x_argc;
+    t_outlet *x_outlet;
 } t_declare;
 
 int canvas_declare_do(t_declare *z, t_canvas *x, t_symbol *s, int argc, t_atom *argv, t_loglevel level);
 
-static void declare_loaded_absolute_paths(t_declare *z, t_canvas *x, t_symbol *s)
+static void declare_loaded_absolute_paths(t_declare *x, t_symbol *s)
 {
-    t_canvasenvironment *e = canvas_getenv(x);
+    t_canvasenvironment *e = canvas_getenv(x->x_canvas);
     t_namelist *nl = e->ce_path;
         /* prefix canvas-path */
     char strbuf[MAXPDSTRING], *path;
     for (; nl; nl = nl->nl_next)
     {
         path = nl->nl_string;
-        canvas_completepath(path, strbuf, MAXPDSTRING, x);
-        if (s && s->s_thing)
-            pd_symbol(s->s_thing, gensym(strbuf));
-        else
-            logpost(z, PD_NORMAL, "loaded <path>: %s", strbuf);
+        canvas_completepath(path, strbuf, MAXPDSTRING, x->x_canvas);
+        outlet_symbol(x->x_outlet, gensym(strbuf));
     }
 }
-static void declare_loaded_paths(t_declare *z, t_canvas *x, t_symbol *s)
+static void declare_loaded_paths(t_declare *x, t_symbol *s)
 {
-    t_canvasenvironment *e = canvas_getenv(x);
+    t_canvasenvironment *e = canvas_getenv(x->x_canvas);
     t_namelist *nl = e->ce_path;
     for (; nl; nl = nl->nl_next)
-    {
-        if (s && s->s_thing)
-            pd_symbol(s->s_thing, gensym(nl->nl_string));
-        else
-            logpost(z, PD_NORMAL, "loaded <path>: %s", nl->nl_string);
-    }
+        outlet_symbol(x->x_outlet, gensym(nl->nl_string));
 }
 
 static void declare_loaded_libs(t_declare *x, t_symbol *s)
 {
     t_loadlist *ll = sys_getloadlist();
     for (; ll; ll = ll->ll_next)
-    {
-        if (s && s->s_thing)
-            pd_symbol(s->s_thing, ll->ll_name);
-        else
-            logpost(x, PD_NORMAL, "loaded <lib>: %s", ll->ll_name->s_name);
-    }
+        outlet_symbol(x->x_outlet, ll->ll_name);
 }
 
 static void declare_bang(t_declare *x)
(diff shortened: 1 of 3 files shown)


src/g_canvas.c
```diff
@@ -1626,27 +1626,84 @@ typedef struct _declare
...
     x->x_useme = 1;
     x->x_canvas = canvas_getcurrent();
         /* LATER update environment and/or load libraries */
-    if (!x->x_canvas->gl_loading)
+    x->x_argc = argc;
+    if (argc > 0)
     {
+        x->x_argv = (t_atom *)getbytes(argc * sizeof(t_atom));
+        memcpy(x->x_argv, argv, argc * sizeof(t_atom));
+    }
```
fdch, line 1678: I'm not sure about this, though... maybe it's safer to use the ALLOCA/FREEA routines? Or somehow restrict the arguments to those sane for [declare]?

fdch:
In this discussion https://github.com/pure-data/pure-data/issues/660, it seems absolute paths could be an option as well for the query message, so I've included that in https://github.com/pure-data/pure-data/pull/2839/commits/ea8c0adae3fc2b9682c211763680e649f3f017cc

fdch:
After reading https://github.com/pure-data/pure-data/issues/886, I realized this query method for canvas could prove more general than I originally thought. So, I separated it from the declare in https://github.com/pure-data/pure-data/pull/2839/commits/7ab208c10258d823fe8345123eacafe03423064d, and assigned special declare messages "paths [0,1]" and "libs" instead. 

I do realize this PR addresses two very different things. I kept them on this same PR, so apologies for the feature creep! Please let me know if you'd like me to change anything.

NOTE: the PR description was changed accordingly

umlaeute:
> Closes https://github.com/pure-data/pure-data/issues/281, https://github.com/pure-data/pure-data/issues/660, and https://github.com/pure-data/pure-data/issues/886

in order to make your PR close the bugs automatically as soon as it is merged into `master`, i think you must put the `Closes: #...` into the main description of this PR (that is: your first post, rather than some comment). extra points if you do that in an enumeration.

src/g_canvas.c
```diff
@@ -1623,27 +1624,146 @@ typedef struct _declare
...
+    for (; i < argc; i++)
+    {
+        const char *arg = atom_getsymbolarg(i, argc, argv)->s_name;
+        if (!strcmp(arg, "paths"))
+            declare_loaded_paths(0, x, snd_symbol);
+        else if (!strcmp(arg, "libs"))
+            declare_loaded_libs(0, snd_symbol);
+        else if (!strcmp(arg, "absolute"))
+            declare_loaded_absolute_paths(0, x, snd_symbol);
+        else if (!strcmp(arg, "blocksize"))
```
umlaeute: i think that the `blocksize` method is totally out of scope for the `[declare]` object.

umlaeute (changes requested):

src/g_canvas.c
fdch: Yes, I agree. That's why I (later) split things up and gave declare its own separate methods for paths and libs. 

Or do you mean to remove it altogether from this PR?

umlaeute: yes, please.

`[declare]` is all about making external resources available (externals, soundfiles,...).

it has absolutely nothing to do with signal processing (or put otherwise: `[moses]` would have been equally ill-suited)

src/g_canvas.c
```diff
@@ -1640,22 +1640,35 @@ static void declare_loaded_absolute_paths(t_declare *z, t_canvas *x)
     {
         path = nl->nl_string;
         canvas_completepath(path, strbuf, MAXPDSTRING, x);
-        logpost(z, PD_NORMAL, "loaded <path>: %s", strbuf);
+        if (s && s->s_thing)
+            pd_symbol(s->s_thing, gensym(strbuf));
```
umlaeute: why? what's wrong with adding an outlet to get the data?

using connections seems to be the most common way of communication between objects.
i think the only object that does it differently is... `[send]` :-)

umlaeute: one possible reason i can think of, is that you want to be able to distinguish whether the user just wants a printout, or whether they want to do something with the information.

so here's a better idea:
if the user wants a printout, they could just hook up the outlet to a `[print]` object

src/g_canvas.c
umlaeute: adding methods to the `canvas_class` (that are not related to `[declare]`) should be grouped with other canvas-methods (a few lines up north)

src/g_canvas.c
umlaeute: after building the PR and trying it out, I see that the `query` message is now attached to the canvas, rather than the object.

but why? (apart from: "blocksize fits better")

sending messages to the canvas is not something you usually do explicitly (leaving aside dynamic patching; but i don't think this PR is related to dynamic patching).

src/g_canvas.c
umlaeute: nitpick: if find the `<foo>` notation a bit strange.

also technically *declare* is really just "declaring" (as in: "announcing", "making it clear") that this patch would like to load a specific library or look up files in a specific directory.
so i think that "failing to declare" would mean that the communication between the object and Pd's path/libloading mechanism failed.
so "failed to load library: %s" might be better.

umlaeute: personally i like the idea more and more, that the user should be able to build the error message by themselves.

e.g. querying which libs have been loaded might not print anything at all, but send the information about what is loaded and what not to the objects outlet, and the patch author can decide whether they want to complain or not.
(this is of course owned to my reluctance to make a unfulfilled declaration an error, as per my comments on https://github.com/pure-data/pure-data/issues/281)

src/g_canvas.c
fdch: Hah, I will remove it. Would not want `[moses]` to also tell us what time it is while splitting up that stream

src/g_canvas.c
fdch: My reasoning was: the declare object really has nothing in it. It's the canvas what has the paths. More crucially, the canvas has a "declare" method, which is the one used at load time. That's when I just thought: the query interface is really for the canvas. And since the canvas has no outlet, that's when I went with a symbol instead. I was thinking mostly of the send method for pointers. 

The libs are a different story, because those feel basically like globals. And it does not really belong to the canvas. But I'm lost here and just went with asking the canvas as well as the declare object for those.

umlaeute: i see.

but since a canvas doesn't really have a "handle", Pd has a `[declare]` object: the sole purpose of this object is to interface with the canvas.
similarly, the `[block~]` object's sole purpose is to interface with the signal-aspects of a canvas.

both `[declare]` and `[block~]` could have been solved by directly sending messages to the canvas instead of requiring an explicit object. but for $reasons, we have dedicated objects.

now `send`ing messages to an arbitrary receiver is an OKish interface for a *setter*. but it becomes really awkward for *querying* the data.
and since we already do have `[declare]` and `[block~]` for *setting* the data, it seems very weird to me to not use them for *getting* the data.

fdch: Thanks for taking the time to review this PR! I see it clearly now. Super thanks for building and running!

I will make changes so that there's an outlet to declare and keep only relevant paths and libs. 

About printing to console, I'm inclined to do some log level 3 printing on failure. I do like "failed to load library/path: %s" on the debug level 3. My reasoning is that, this would maybe save the user the trouble of patching in order to debug a failed lib. If they see it failed on the log 3 then they found the issue in one click.

I will remove the canvas query. I'm thinking to put that in a different PR with a more carefully thought of handle for the canvas. But I know there's a canvasinfo external (I forget if that's its name) so maybe there's no need for an internal handle.

fdch:
I've addressed all the issues in your comments, please let me know if I missed any. Also, I'm not sure if you or me should mark the opened conversations as resolved. I'll leave those open just in case.  Here's an updated version of the example file that explains a bit the usage and the expected outcome:
- [declare-example.zip](https://github.com/user-attachments/files/25502076/declare-example.zip)

src/g_canvas.c
```diff
@@ -1873,28 +1922,44 @@ void canvas_declare(t_canvas *x, t_symbol *s, int argc, t_atom *argv)
...
+                success = 0;
+            }
             i++;
         }
         else post("declare: %s: unknown declaration", flag);
     }
+    return success;
 }
 
+int canvas_declare_message(t_canvas *x, t_symbol *s, int argc, t_atom *argv)
```
umlaeute: i think this will play havoc with emscripten and the like, where you cannot just arbitrarily return values in callback functions that have been declared as returning `void` (see 32aea205f9d93afb8c0bfc5ba57b67a0f1899704)

fdch: thanks! should be fixed in https://github.com/pure-data/pure-data/pull/2839/commits/6b22ca17a7420f0630e3e2205d424b6c00a15a21

Issue #566, opened 2019-03-04
XYrange is lost when switching [array] from non-GOP back to GOP


fdch:
I attach a small test patch to show this. 

Say you have a GOP array with Xrange 0..N+3 and Yrange 1..-1, and change it to non-GOP and then back to GOP, you get:

- Xrange is 0..1
- Yrange is inverted -1..1 

NOTE: Sinesum message reverts it back to original coordinates when array is in GOP with wrong XY ranges.

[array_gop.txt](https://github.com/pure-data/pure-data/files/2926732/array_gop.txt)

fdch:
I also wonder if non-GOP array should not become [array define array1] instead of [graph], which does not really allow for anything but hiding the array...

umlaeute:
turning GOP on/off doesn't change the object itself. so an array really *is* `[graph]`.

Spacechild1:
first: a graphical array is a GOP subpatch ("graph") containg a data structure array. [array define] and [table] are canvases *containing* a graphical array.

now to your question: if you open the canvas properties you see the fields "X units per pixel" and "Y units per pixel". These are only enabled when GOP is off. The thing is that those really just set the array bounds: 
```
X units per pixel: 100
Y units per pixel: -200
```
actually is:
```
X range: from 0 to 100
Y range from 0 to 200
```
Note that the bounds always start at (0, 0) and that the Y value is inverted (don't ask me why)... 
The array bounds have a slightly different meaning for a non-GOP subpatch (which data structure quantity/size corresponds to 1 pixel on the canvas?) vs.  GOP subpatches (which data structure range is shown in the graphing area?)

so what happens is that by switching to non-GOP you effectively override the array bounds and there's no place in the internal canvas structure to cache the old values. there could be, but I don't think it's worth the complexity.

to your question why sinesum restores the array bounds: because it implicitly resizes the array and resizing automatically adjusts the *X-values* of the array bounds to fit the whole array.

fdch:
I've proposed a simple fix for this on #2841. Maybe the side-effect of a vis function is not something that we want, as it could be a symptom that we might need a more general way to handle arrays and graphing. But anyhow, a band-aid would not hurt for now.

PR #2841, opened 2026-02-18
Restore array size after coming back from non-GOP [graph]

fdch:
This is a simple fix that just puts the size of the array into the size of the graph, no extra storage needed.

It is in fact rather dumb, so dumb that we restore the size of the first array only. If there are more than one arrays, then this may result in one array being larger than the gop area, and thus some unwanted/weird resizing. But, I think that problem is even more of an edge case, as it may be less likely to have 2 arrays on the same graph and do the non-gop -> gop dance.

This could be done smarter: some iteration on all array sizes might be worth the effort to get the max size and restore to that size, but for now the problem in the related issue should be solved.

Closes:
- https://github.com/pure-data/pure-data/issues/566.

### DIFF
--- a/src/g_array.c
+++ b/src/g_array.c
@@ -737,6 +737,8 @@ static void garray_delete(t_gobj *z, t_glist *glist)
 static void garray_vis(t_gobj *z, t_glist *glist, int vis)
 {
     t_garray *x = (t_garray *)z;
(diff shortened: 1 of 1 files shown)


millerpuckette:
I think it's a bug that the range gets reset at all when you un-click "gop" for a garray... but then, perhaps it's best simply to dissallow doing that at all since it's never useful as far as I can see :)

PR #2846, opened 2026-02-25
add getzoom to pdcontrol

solipd:
adds a method to get the zoom factor of a patch window to [pdcontrol]


umlaeute:
what's the use-case for this?

ben-wes:
this seems related to https://forum.pdpatchrepo.info/topic/12334/audiolab-is-now-available-on-deken/66

i'm wondering though whether it's not actually a bug that a struct's `displace` message is reporting movements which don't consider zoomed coordinates?

porres:
> i'm wondering though whether it's not actually a bug that a struct's displace message is reporting movements which don't consider the zoomed in coordinates?

sounds like it to me

solipd:
it would also be useful to adjust the window frame according to the zoom level. Attached is a patch to test this.
[test_getzoom.zip](https://github.com/user-attachments/files/26004698/test_getzoom.zip)

ben-wes:
> adjust the window frame according to the zoom level

this should probably be discussed in #2847

umlaeute:
@solipd, can we close this?

PR #2854, opened 2026-03-12
Flip menu order and move 'Load Meter' entry to 'Tools'

### LINKED ISSUE #2853: reorder "Media" menu in menu  bar
i think the current menu bar has a sub-optimal arrangement:

file edit find media window tools help

i find the location of the `Media` menu especially annoying: its position between `Find` and `Window` somehow suggests that it is about navigating the "running environment" (the totality of all running patches), which it is not.

i would suggest moving it after the `Window` menu:

file edit find window media tools help

or even after the `Tools` menu:

file edit find window tools media help

what do you think?

ben-wes:
mainly wanted to see what the suggestions in #2853 felt like - but why not make this a PR right away ...

tested on macos.

image

image

* closes #2853

(EDIT: not moving the "test audio and midi" entry anymore with this PR)

### DIFF
--- a/tcl/pd_menus.tcl
+++ b/tcl/pd_menus.tcl
@@ -59,7 +59,7 @@ proc ::pd_menus::create_menubar {} {
     $::patch_menubar add cascade -label [_ Put] \
             -underline 0 -menu $::patch_menubar.put
 
-    foreach mymenu "find media window tools help" {
+    foreach mymenu "find window media tools help" {
         if {$mymenu eq "find"} {
             set underlined 3
         } {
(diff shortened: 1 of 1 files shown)


umlaeute:
thanks. I'm still unsure about the "test audio and midi" patch (as this is closely related to the "Media" setup).

the good thing with lumping them into Tools is, that the separator makes sense as shown in your screenshot (it makes less sense if there's only a single item)

ben-wes:
> unsure about the "test audio and midi" patch

i was hesitant there, too - it's certainly valid in both menus.
my reasoning was: it's not used for changing anything about the context or configuration like the other entries in "Media".

but this might be acceptable as well, i guess - also with a single entry above the separator:

image

image

... and obviously, moving entries will cause irritation - no matter how reasonable the decision might be.

umlaeute:
fair enough. 
looking at it, the separator after a single element doesn't seem to be that bad :-)

so: I'm fine with either.

ben-wes:
ok, updated the PR. i guess, we might still decide to move the "test audio and midi" entry in the future. less confusion for now. :)

umlaeute:
@millerpuckette this would be nice to have

PR #2855, opened 2026-03-12
displace message: output zoom-aware canvas coordinates

### LINKED ISSUE #2856: data structures do not take zoom into account when reporting click
the `click` event of datastructures report the (relative) click-position - but the position is scaled according to zoom.
it should be normalized.

patch showing the problem: [ds-zoom.zip](https://github.com/user-attachments/files/25931775/ds-zoom.zip)

---
see https://forum.pdpatchrepo.info/topic/12334/audiolab-is-now-available-on-deken/64
- Related: https://github.com/pure-data/pure-data/pull/2846

ben-wes:
this is an attempt to output canvas coordinates, making them independent of the current zoom level.

* related to #2846
* closes #2856

the fractional handling feels a bit awkward, but switching the displace API to floats felt out of scope here.

test patch: [test-displace.pd.zip](https://github.com/user-attachments/files/25931773/test-displace.pd.zip)

### DIFF
--- a/src/g_scalar.c
+++ b/src/g_scalar.c
@@ -446,17 +446,16 @@ static void scalar_displace(t_gobj *z, t_glist *glist, int dx, int dy)
         goty = 0;
     if (gotx)
         *(t_float *)(((char *)(x->sc_vec)) + xonset) +=
-            glist->gl_zoom * dx * (glist_pixelstox(glist, 1) -
-                glist_pixelstox(glist, 0));
+            glist_dpixtodx(glist, dx * glist_getzoom(glist));
     if (goty)
         *(t_float *)(((char *)(x->sc_vec)) + yonset) +=
-            glist->gl_zoom * dy * (glist_pixelstoy(glist, 1) -
-                glist_pixelstoy(glist, 0));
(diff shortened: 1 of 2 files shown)


ben-wes:
@umlaeute just saw that you created #2856 at the almost exact same time. if you're planning a PR for this, i'll gladly close mine again!

umlaeute:
ha!

anyhow: it seems that with this PR i can only drag the `[pp.xypad]` once.
after I release it, and start dragging again, the knob starts jumping around...

ben-wes:
> the knob starts jumping around...

yep ... but that's related to the implementation of `[pp.xypad]`: the y-coordinate of the `click` message gets inverted in `[pd struct_jumpgrid]`.

this was obviously done since the inverted canvas y-coordinates didn't have an effect on the reported values. now they do and this manual inversion needs to be removed again.

not sure if the previous behavior can be considered a bug, or if we're stuck with it for backwards compatibility reasons?

ben-wes:
(not sure whether github notifies people mentioned in an edited comment, so posting separately now ...)

@solipd — curious what you think about this, since it concerns your library.

solipd:
@ben-wes this works just great, thanks!

PR #2858, opened 2026-03-13
Add Pd colors to preferences tabs

### LINKED ISSUE #2643: Request: Allow setting colors via preferences and save it. Maybe also add some presets for "themes"?
This is a reboot of https://github.com/pure-data/pure-data/issues/931

Now that we have colors ("yay"), it seems natural and reasonable to set and save them via preferences, as Miller noted in the release notes. 

While we're at it, maybe it's also a good idea to allow some presets that users can change, like "dark" or "light" mode.

Here's an example of such presets in Purr Data

Image

porres:
This PR adds a 'colors' section to the 'GUI settings' in the 'Misc' tab of preferences.

It uses the common scheme we know from iemguis with the color presets and radio buttons to set bg/fg/sel/gop plus the compose color button.

the colors are saved in the preferences, but I am not yet retrieving them from it to reset the colors at startup, so this is still a draft and I need more guidance. But it's working nicely and I'd like to share and let people try.

I also added a dropmenu for some presets, I stole a couple from Pd-L2ork. A next step would be to also allow users to save/delete their own items in their gui-preferences.

closes https://github.com/pure-data/pure-data/issues/2643

### DIFF
--- a/tcl/dialog_preferences.tcl
+++ b/tcl/dialog_preferences.tcl
@@ -135,8 +135,8 @@ proc ::dialog_preferences::fill_frame {prefs} {
     # Row 1: Dropdown menu and Compose button
     frame $prefs.guiframe.colors.toprow
     pack $prefs.guiframe.colors.toprow -side top -fill x -pady 5
-
-    # Dropdown on left
+    label $prefs.guiframe.colors.toprow.leftspacer -text "" -width 2
+    pack $prefs.guiframe.colors.toprow.leftspacer -side left -expand 1
     label $prefs.guiframe.colors.toprow.lab -text [_ "Presets:"]
     menubutton $prefs.guiframe.colors.toprow.mb \
         -text "Default" \
@@ -145,7 +145,6 @@ proc ::dialog_preferences::fill_frame {prefs} {
     menu $prefs.guiframe.colors.toprow.mb.menu -tearoff 0
     $prefs.guiframe.colors.toprow.mb configure \
         -menu $prefs.guiframe.colors.toprow.mb.menu
-    # Add preset options
     set presets { "Default" "Dark Mode" "Light Mode" "Pastel" "Dark Contrasted" "Purple Haze" "Misty Rose" "C64" "Strongbad" "Solarized" "Solarized Inverted"}
     foreach preset $presets {
         $prefs.guiframe.colors.toprow.mb.menu add command \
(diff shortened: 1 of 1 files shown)


porres:
a quick silly test


https://github.com/user-attachments/assets/06842452-0cff-4a69-b7c4-81599328b261

charlesneimog:
@porres, that's really cool. Maybe you could add a simple toggle for dark and light mode?

porres:
> @porres, that's really cool. Maybe you could add a simple toggle for dark and light mode?

it's on the dropdown already

porres:
One thought from @ben-wes: I guess we need to immediately see the effect in macOS

porres:
I guess I can have a mid row with "compose color" button (and not just "compose") to conform with iemguis, and the presets dropdown menu could be to the right of it

porres:
> it's on the dropdown already

at least the idea of it is there. I was just adding quick ideas, I actually need to work more on those options, study some nice and cool presets, which, by the way, I am open to :)

porres:
Ok, I fixed applying colors right away and have rearranged things and made other improvements

Screenshot 2026-03-13 at 23 14 51

porres:
I'm settling at offering 10 presets other than the default and I did my best in offering what seems like an interesting set. Like I said, I got some from Pd-L2ork and I added a couple I had already developed and that are available in the Pd docs. The "light mode" is what I actually use when I have the "flux" app on my mac, which filters completely the blue light and is supposed to be good on your eyes...

But anyway, I'd like to call into the room @ben-wes, the competent web designer and others to chime in for what would be a nice set of presets.

Again, for later, i'd like to think of a way to make users add and store their own presets into the preferences

ben-wes:
> competent web designer

not sure i qualify for either "web" or "competent".

i wrote some thoughts a while ago in #2454 though. that was before color changes were actually added to Pd and assumed that different colors might be used for object border, background and text. i guess that these lines of the second comment still hold:

> typical purpose of a default darkmode is probably a stage setup where the screen should ideally be as dark as possible.

> [...] #FFFFFF on #000000 is quite a heavy contrast though - so maybe the white could just be dimmed down a bit.

the issue with dimmed white is that display of grey values on screens differ quite heavily. so in the end, the choice is a bit arbitrary.

personally, i use this little abstraction as a dark mode switcher - this is not in line with the "full black" approach though: [darkmode.zip](https://github.com/user-attachments/files/26002818/darkmode.zip)

we should probably provide both kinds of neutral dark modes - one on full black and one on dark grey.

porres:
"colors #c8d4e2 #222429 #60b2ff" looks great for dark mode, thanks, I will include it

porres:
> the colors are saved in the preferences, but I am not yet retrieving them from it to reset the colors at startup, so this is still a draft and I need more guidance. But it's working nicely and I'd like to share and let people try.

@umlaeute I think this is more mature now with the presets and the way the properties looks, but we need to find a good way to recall this GUI setting saved from preferences so Pd restarts with the previously set colors, and I'm not sure what is the best strategy to do that - so I need your guidance :)

porres:
@umlaeute and @millerpuckette with the new updates, please check this out so we can maybe sort it out and include soon.

We need to find a good way to recall this GUI setting saved from preferences so Pd restarts with the previously set colors, and I'm not sure what is the best strategy to do that, so what do you think?

millerpuckette:
I want eventually to get rid of all "gui settings" except whatever is specific to Tcl/TK, so that things will work correctly in GTK as well.  So I'd argue for putting color preferences in the Pd settings, not in the "gui settings".

umlaeute:
for this I would like to have a more generic settings system in pd-core... an arbitrary key/value store (that can properly handle lists), and with namespaces, so it can be reused by externals. 

and it should also use a more standard conformant location of the settings (the days are long gone, where applications dumped everything into a hidden file in the user's home. these days we really should use [`XDG_CONFIG_HOME`](https://specifications.freedesktop.org/basedir/latest/))

umlaeute:
on linux, clicking on one of the colors appears to not do anything. only after i hit <kbd>Apply</kbd> the color is applied to the windows (but by then i might have already forgotton which color i chose).

this is probably different on macOS, where any color selection is applied immediately, but for the other systems we need some preview. (like the iemgui's does, but hopefully prettier)

PR #2860, opened 2026-03-15
fix 'color' message for arrays that use points

### LINKED ISSUE #2859: new array color feature not working for points
This is a regression since 0.56-2. In 0.56-2 sending a 'color' message to an array changed the color of the array, regardless of display type (polygons, points, bezier). In current master, an array set to 'points' stays black when sending 'color' messages.

Attached patch illustrates the issue.

[colorarray.zip](https://github.com/user-attachments/files/26000011/colorarray.zip)

reduzent:
Fix suggested by @ben-wes 

fixes #2859

### DIFF
--- a/src/g_template.c
+++ b/src/g_template.c
@@ -2058,7 +2058,7 @@ static void plot_vis(t_gobj *z, t_glist *glist,
                 {
                     pdgui_vmess(0, "r crri kk iiii", "pdtk_canvas_create_rect",
                         glist_getcanvas(glist), tag, "-", 0,
-                        THISGUI->i_foregroundcolor, THISGUI->i_backgroundcolor,
+                        color, THISGUI->i_backgroundcolor,
                         ixpix , (int) glist_ytopixels(glist, basey +
                             fielddesc_cvttocoord(yfielddesc, minyval)),
                         inextx, (int)(glist_ytopixels(glist, basey +


umlaeute:
as there were merge conflicts, i've rebased this on the latest changes by miller (where `pdtk_canvas_create_rect` was updated to have a background color) and force pushed it.

i wonder though, what to do with the background color of `pdtk_canvas_create_rect` (which currently is set to `THISGUI->i_backgroundcolor`): shouldn't arrays be just transparent instead of occluding anything behind them?
esp. since the new hardcoded value of "background color" cannot be overridden.

ping @millerpuckette

umlaeute:
ah, i now see that this code is actually for drawing the data points. so it's not as bad as i assumed.

the other question is of course: while this PR fixes the color of the graph when the user explicitly requested one with a `color` message, it obviously ignores any color scheme (even if the user just used the defaults).

we could use some special color values like `-1` or so to flag a "default" color, but i dunno.

porres:
> the other question is of course: while this PR fixes the color of the graph when the user explicitly requested one with a color message, it obviously ignores any color scheme (even if the user just used the defaults).
> 
> we could use some special color values like -1 or so to flag a "default" color, but i dunno.

Not sure what you mean here, so you mean adapting to Pd's color settings? There was a big discussion about it for arrays and iemguis at https://github.com/pure-data/pure-data/issues/2639

But then, array colors are not in hexadecimal and we should allow Data Structures to be able to handle this color format. I worked on a PR for that in https://github.com/pure-data/pure-data/pull/2670 but I now see it has conflicts as well

porres:
>  it has conflicts as well

just merged with master and fixed conflicts, so now arrays and data structures can handle hex colors, next step would be to get the colors from Pd's colors

umlaeute:
> Not sure what you mean here, so you mean adapting to Pd's color settings?

i mean:
1. start Pd
2. issue `[; pd colors #ffffff #000000 #0000ff(`
3. type <kbd>Ctrl</kbd>+<kbd>a</kbd> to create a new array
4. the result is that you get a (white=foreground) array box but the data is invisible (as it is also black)

the same is true if i swap steps 2 and 3 (create the array first, then change the color scheme)

my point is that I did not manually change the colors of the array, and therefore my expectation is that it stays readable.

so i'm not talking about interpolating the colors to auto-adapt to the color-scheme  (as in #2639) , but instead to honour the global foreground/background *as long as nobody changed the graph-color*.


(and i admit that i was not aware that #2639 was not only about iemguis but also about arrays)

porres:
hmmm, geez, I am now better testing what master is doing, and I see that for points it just gets the foreground color, but not the others, right? Also, one cannot set the color of points, but can with the others. 

I didn't test this PR so, what does it do? I guess it just prevents points to reflect Pd's forergound color like polygon/bezier

> but instead to honour the global foreground/background as long as nobody changed the graph-color.

That is a good idea, and one should also undo the change and set to pd's color again, like with the "-1" idea.

ben-wes:
+1 for -1

porres:
It seems miller accepted this, but the PR didn't get closed.

PR #2872, opened 2026-03-24
Add missing tag changes

### LINKED ISSUE #2866: regression: can't select [cnv] by clicking on its rectangle area
In current master I can only select the [cnv] by clicking and dragging to include it into the selection area.

ben-wes:
... following 706ce01

* closes #2866


fdch:
+1 Thanks for catching this! I tested it locally with `doc/5.reference/all_guis.pd` and it definitely fixes the issue.

Issue #2865, opened 2026-03-18
at least 3 different ways to escape strings on the GUI side


umlaeute:
while looking into #2864 , i stumbled upon the many ways how Pd-GUI escapes strings:

we have at least:
- `enquote_path`
- `unspace_text`
- `pdtk_encodedialog`
- `::dialog_gatom::escape`

they all handle things slightly differently, e.g. either escape semicolon to `\;`, or remove it altogether.
most use backslash for escaping (e.g. space "` `" becomes "`\ `"), except for `pdtk_encodedialog` which invents an altogether new mechanism (where space "` `" becomes "`+s`").

empty strings are encoded very differently (`pdtk_encodedialog` uses a single `+`,  `unspace_text` uses the reserved keyword `empty`, and `::dialog_gatom::escape` uses `-´).

i think we should get this down to a single escaping mechanism.
(this should also help with the GTK port)

porres:
> we have at least:
> 
> enquote_path
> unspace_text
> pdtk_encodedialog
> ::dialog_gatom::escape

"at least" indeed, here's one more: pdgui_strnescape in s_print.c

I have been suffering with all this as well and would really like that the whole escape/unescape system could be drastically simplified

umlaeute:
> "at least" indeed, here's one more: `pdgui_strnescape` in s_print.c

but that goes in the other direction. we do need escaping mechanisms in both directions (so one  obviously needs to be implemented on the core-side and the other on the GUI-side), and they are slightly different because some characters (like `[]`) have very special meaning in Tcl/Tk (but not for the core).

but yes,...

PR #2889, opened 2026-04-20
dialogs: use `concat` instead of `list` for pdsend payloads 

### LINKED ISSUE #2864: regression: dialogs no longer handle spaces or special characters
> @umlaeute : when i save any iemgui's properties now with an `$` in a text field, additional embracing `\{` and `\}` get added. it looks like this here is the relevant change. 

 _Originally posted by @ben-wes in [0d0280e](https://github.com/pure-data/pure-data/commit/0d0280e0955f12f020e519835d2ae67e4b2a044c#r179733659)_

ben-wes:
reverts some of the changes in 0d0280e

* closes #2864 (tentative fix)

@umlaeute : based on #2865, i assume that you're aiming at a more elegant solution for this. but with the new validation you added, empty values are avoided anyway AFAICS - so this might be acceptable for now?

### DIFF
--- a/tcl/dialog_array.tcl
+++ b/tcl/dialog_array.tcl
@@ -224,7 +224,7 @@ proc ::dialog_array::listview_edit+paste {arrayName startIndex data} {
         lappend values $value
     }
     if { $values ne {} } {
-        pdsend [list $::dialog_array::listview_id($arrayName) $offset $values]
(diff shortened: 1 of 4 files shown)


umlaeute:
i currently do not have the time to revamp the escaping mess.

so i think for now, let's revert my changes (accepting this PR), and let's try again later.

PR #2900, opened 2026-05-07
Feature: open files in cwd

### LINKED ISSUE #2888: [file cwd] doesn't work with read
If I follow a [file cwd] with an open message, the correct file is opened from the subdirectory. If I follow a [file cwd] with a read message for the same file, pd says that the file is not found.

Image

umlaeute:
with `[file cwd]` it is possible to change the working directory.

however, many objects (including `[soundfiler]`) won't take the CWD into account.

this patch "fixes" this:
- extend `canvas_open()` to just try to open the file without a dir as a last resort
- also extend `[soundfiler]` (read) and `[readsf~]` to fallback to opening the file without a dir as a last resort

:warning: I'm not entirely sure whether we actually want this behaviour, though.

in any case:
- Closes: https://github.com/pure-data/pure-data/issues/2888.

### DIFF
--- a/src/d_soundfile.c
+++ b/src/d_soundfile.c
@@ -417,6 +417,8 @@ int open_soundfile_via_namelist(const char *dirname, const char *filename,
     fd = do_open_via_path(dirname, filename, "", buf, &dummy, MAXPDSTRING,
         1, nl, 0);
     if (fd < 0)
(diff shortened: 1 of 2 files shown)


umlaeute:
i'm not convinced that this is a good idea.

PR #2904, opened 2026-05-12
documentation updates

### LINKED ISSUE #309: outlet~ args reference seems wrong
hope, i'm not getting this wrong ...

[pd reference] for `outlet~` arguments states:

```text
downsampling method: 'hold' for sample/hold (default), 'pad' for zero-padded and 'lin' for linear interpolation).
```

i think this is only true for *upsampling* (as documented in the [pd up/downsampling] subpatch: "There is no corresponding choice of downsampling method - downsampling is done simply by dropping the extra samples.")

also, [pd test-subpatch-downsampled] is obviously a downsampled subpatch, but the `outlet~`s are *upsampling* - so the text there should also be corrected and "downsampling methods:" should say "upsampling methods:".

### LINKED ISSUE #310: inlet~ [pd reference] still mentions outdated fwd argument
i assume this should be omitted: "if 'fwd' argument is given."

porres:
- updated the manual to the current version and added a couple more things to the release notes
- mentioned warning about using [vline~] with DSP off, see https://github.com/pure-data/pure-data/pull/2906
- closes https://github.com/pure-data/pddp/issues/309
- closes https://github.com/pure-data/pddp/issues/310


doc/1.manual/5.current.status.htm
umlaeute: please fix the `h4` tag to read:
~~~html
<h4 id="0.56-3">
~~~

umlaeute: @porres I pushed a fix to your branch

PR #2906, opened 2026-05-13
vline~: clean up outdated segments on insertion of new ones 

ben-wes:
see title and comments.

### DIFF
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -336,6 +336,18 @@ static void vline_tilde_float(t_vline *x, t_float f)
         vline_tilde_stop(x);
         return;
     }
+        /* drain past segments to prevent unbounded queue growth when messages
+        arrive metronomically while DSP is off. */
+    while ((s1 = x->x_list) && s1->s_targettime <= timenow)
+    {
+        x->x_value = x->x_target = s1->s_target;
+        x->x_targettime = 1e20;
+        x->x_inc = 0;
+        x->x_list = s1->s_next;
+        t_freebytes(s1, sizeof(*s1));
+    }
+    if (!x->x_list)
+        x->x_list_tail = 0;
     snew = (t_vseg *)t_getbytes(sizeof(*snew));
         /* check if we append after the last segment.  We append when the new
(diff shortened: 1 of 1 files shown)


porres:
I see now this adds quite a big overhead for scheduloing lines, good! That should free me from all the warnings ;)

anyway, I added a documentation update to [vline~] in https://github.com/pure-data/pure-data/pull/2904

hope this gets included for the next test release or final version

millerpuckette:
This is kind of complicated.  I'm thinking about another idea - the difficulty comes not because segments get added so much as that they get added metronomically (with the passage of scheduler time) without DSP on.  But it's not clear what the semanticas should be in that case at all.  Maybe if time passes between messages, the older messages should be back-dated and then allowed to disappear when they slip into the past.

ben-wes:
> This is kind of complicated.

agreed. all this warning foo can probably be omitted again that way ... i gave it a try. if this looks good, i should probably squash these commits though (and correct the PR description).

porres:
we still have the warning with 0.56-3test2, right?

My documentation remark about the warning needs to be removed if this PR gets accepted or something else happens that prevents the warning...

millerpuckette:
I ended up taking the warning out - it probably would cause more trouble 
with spurious warnings than it would prevent.  Will plan to do something 
better for 0.57 :)

M

porres:
> I ended up taking the warning out - it probably would cause more trouble

ok, @millerpuckette  please take this change also then :) 

https://github.com/pure-data/pure-data/pull/2915

ben-wes:
> I ended up taking the warning out - it probably would cause more trouble with spurious warnings than it would prevent.  Will plan to do something better for 0.57

sounds good to me. thanks! since the tail pointer handling commit is still used, most of the trouble is probably avoided anyway.

i squashed the 2 commits here and rebased on master if relevant for 0.57.

so this PR now only removes outdated segments when adding new ones. certainly not a critical addition (and obviously some minor overhead).

ben-wes:
closing in favor of https://github.com/pure-data/pure-data/commit/1120ac94a7551a45384b6b1632b7fcf8203f2f79

PR #2917, opened 2026-05-29
fix HTML docs

### LINKED ISSUE #2916: html docs are invald HTML
at least the release-notes have dangling open tags...

see https://validator.w3.org/nu/?doc=http%3A%2F%2Fmsp.ucsd.edu%2FPd_documentation%2F5.current.status.htm

while it doesn't *look* bad, the broken HTML is a problem when parsing the page to extract metainformation (such as required for the flatpak distribution of Pd)

umlaeute:
The HTML docs (`doc/1.manual/` and `doc/8.topics`) have a couple of invalid and or outdated HTML.

since I would really like to be able to parse/transform the releasenotes, i need syntactically correct HTML; therefore i have setup a CI-job that runs an HTML validator.

of course the validator found a couple of other issues, some of them overly anal (imho):
- **any tag opened must be closed (in correct order!)**
- case of opening tag must match case of closing tag (e.g. `<a>` must be closed by `</a>`. `</A>` will fail the validation!) ¯\_(ツ)_/¯
- the `<tt>` tag has been deprecated in HTML5 in favour of `<code>` and friends
- angle brackets must be written as `&lt;` (we cannot just write `pdsend <portnumber>` in HTML ....)
-   `doc/8.topics/expr.htm` used a couple of deprecated attributes, which i have replaced with modern CSS (that should look the same)

Closes: https://github.com/pure-data/pure-data/issues/2916

### DIFF
--- a/doc/1.manual/2.theory.of.operation.htm
+++ b/doc/1.manual/2.theory.of.operation.htm
@@ -331,7 +331,7 @@ the parameters of your patch, so you can load text files and make other preparat
 
 <P>Pd searches for files using specified paths. Most objects capable of reading files
 search along these paths, but when Pd writes files, they are saved to the directory
-where the patch itself is located by default. Check <a href="3.installing.configuring.htm#s3.4">3.4. Path and
+where the patch itself is located by default. Check <A href="3.installing.configuring.htm#s3.4">3.4. Path and
 startup</A> for more details.</P>
 
 <H3> <A href="#s2.2" id="s2.2">2.2. Editing Pd patches</A> </H3>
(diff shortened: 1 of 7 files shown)


umlaeute:
@porres this is mostly documentation fixes; so if you intend to work on your documentation branch, it would be great if these changes could be included.

ben-wes:
> case of opening tag must match case of closing tag (e.g. `<a>` must be closed by `</a>`. `</A>` will fail the validation!)

ouch. that's on me, sorry - wasn't aware and didn't care much. i guess, consistency always pays off. :)

umlaeute:
honestly, that was a wtf-moment when looking at the output of the validator. 
I do not remember HTML to be case sensitive, so I would have much preferred a way to tell the validator to ignore these "errors".
it just turned out to be simpler to fix the tags instead...

millerpuckette:
Sorry I didn't manage to merge this in time for 0.56-3 but will update the HTML doc on my own site with this.

umlaeute:
> Sorry I didn't manage to merge this in time for 0.56-3 

this was expected, as I only submitted the PR after the release (I only discovered the problem when trying to update the flatpak package of Pure Data to 0.56-3, for which I need to parse the releasenotes)

since this happened multiple times to me in the past, i've now added a new CI-job (thanks @Lucarda) to monitor the validity of the HTML *before* the releases

ben-wes:
Just curious about that last update: why didn't the validator complain then?

umlaeute:
> Just curious about that last update: why didn't the validator complain then?

i wonder as well.
maybe the validator is just really bad, and we should switch to another one? i'm very open for suggestions

PR #2923, opened 2026-06-10
Quote paths in macOS application makefile

HaHeho:
The recent changes allow build directory paths to include spaces. Previously, using `make` would succeed, but when running `make app` on projects with absolute paths containing spaces, it would fail with an error, due to the path being split within `mac/osx-app.sh`.

This issue has now been resolved. However, I’m not entirely sure if the solutions implemented conform to best practices. I would appreciate any suggestions for alternative approaches.

### DIFF
--- a/mac/Makefile.am
+++ b/mac/Makefile.am
@@ -25,9 +25,9 @@ pkgdata_DATA += \
 
 # build an app bundle on OSX
 app:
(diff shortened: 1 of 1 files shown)


umlaeute:
looks good to me.
including your PR in the `develop` branch

PR #2925, opened 2026-06-11
[siginfo~] to query signal properties

### LINKED ISSUE #1687: [feature] report blocksize (and overlap, and oversampling)
we can set the blocksize, the overlap-factor and the oversampling via `[block~]`.
we can even dynamically change these block settings.

however, there is no way to query the current settings.

i suggest to add an outlet to `[block~]` and allow the `[bang(` message to the object, which would give us the blocksize/overlap/resampling triplet.
`[switch~]`'s bang-method should be left unaltered.

umlaeute:
a new object `[siginfo~]` that reports (for a connected input signal):
- blocksize
- number of channels
- overlap factor
- samples per second (total amount of samples this objects sees per channel per second; that is: `parentsamplerate*resampling*overlap` (as reported by the illnamed `t_signal.s_sr` member)

the object can be banged, but it also emits this information when the DSP is started.

~~~
[block~ 512 2 4]

[snake~ in 3]
|
[siginfo~]
|
[512 3 2 384000\
~~~

this
- Closes: https://github.com/pure-data/pure-data/issues/1687


ben-wes:
cool. should i remove the `count` verb from https://github.com/pure-data/pure-data/pull/2615 then?

umlaeute:
i guess so, as the `count`  verb would be obsoleted by this.
(but then: who knows whether @millerpuckette is going to accept this PR?)

millerpuckette:
I think this should also report sample rate - it could then replace samplerate~ which also should have spat the value out on DSP starts (but can't be fixed compatibly without adding a creation arg).  But then, why not:

[siginfo~ sr]
[siginfo~ blocksize]

etc... so that you don't have to use [unpack] or whatnot to get the value out?

and also, this object could report DSP starting and stopping as in [siginfo~ onoff].  Maybe then dspinfo~ or
even dsp~ would be a better name.  (I think I already proposed a dsp~ object somewhere on the dev list I think but never followed up on it.)

umlaeute:
about the object name

- i used `[siginfo~]` because it also reports the channel count (for (multichannel) signals that are actually connected to the object), which i think makes less sense for a `[dspinfo~]` object.
- iemlib has shipped a s`[dsp~]` abstraction for more than 20 years, so better not take it over

about the output format
- i like the idea of not having to remember the anonymous arguments
  - personally i would prefer [route]able messages on the output, but it's not very idiomatic in Pd-vanilla. maybe we can still use them if the object is instantiated without arguments)
- i'd prefer "samplerate" over "sr"

  should we report the "samples per second" separately? i think it's sometimes important to understand why objects that take e.g. frequencies behave like they do (and I'm not entirely sure whether it can be fully constructed from the canvas "samplerate" and the overlap factor).

porres:
I think this mixes two things/ideas. One is the patch audio signal chain, which is block size, overlap factor and actual sample rate based on the overlap. Basically, it gets the info you set with [block~]...

Another completely separate thing in my opinion is the channel count of an object, for which I have else/nchs~ for and I see that stands out quite much as the odd one out. I also believe it should be something we could have in [snake~] instead, as in [snake~ nchs] or [snake~ count] or whatever...

> I think this should also report sample rate

yeah, global rate could also be possible, but you can derive it from the given local sample rate divided by overlap.

Reporting DSP status is also nice, and I also believe we could include more stuff and change the name of the object to [dspinfo~] or something, leaving the channel count out.

porres:
For reference and inspiration, Pd-L2ork has a special built-in object that spits out a lot of information, including the number of audio inputs and outputs defined in the settings, which might also be a cool thing :) though I do also beliebe it does report too much things anyway...

Screenshot 2026-06-11 at 12 19 57

It also does report block size and dsp-status, which is what we also want in Vanilla. 

Anyway, here's all that I got reported in my system

arch: unknown
audio-api: PortAudio
audio-apilist: PortAudio
audio-apilist-all: none ALSA OSS MMIO PortAudio JACK SGI
audio-inchannels: 2
audio-indev: 0 2
audio-indevlist: Core Audio: Alexandre’s iPhone Microphone Core Audio: BlackHole 2ch Core Audio: MacBook Air Microphone Core Audio: Microsoft Teams Audio Core Audio: ZoomAudioDevice
audio-multidev-support: 1
audio-outchannels: 2
audio-outdev: 0 2
audio-outdevlist: Core Audio: BlackHole 2ch Core Audio: MacBook Air Speakers Core Audio: Microsoft Teams Audio Core Audio: ZoomAudioDevice
blocksize: 64
canvaslist: (pointer) (pointer) (pointer) (pointer)
classlist: 907 classes loaded ("classtable" outputs the full list)
dir: /Applications/Pd-L2Ork.app/Contents/Resources/app.nw/bin/
dsp-status: 0
floatsize: 32
gui: 1
libdir: /Applications/Pd-L2Ork.app/Contents/Resources/app.nw
midi-api: ALSA
midi-apilist: DEFAULT
midi-indev:
midi-indevlist: IAC Driver Bus 1
midi-outdev:
midi-outdevlist: IAC Driver Bus 1
platform: darwin
samplerate: 44100
version: 0 55 2
l2ork_version: 20251203 rev.a71bf063

porres:
> I do also beliebe it does report too much things anyway...

but then, also, some of these stuff could be included in [pdinfo] ;)

umlaeute:
>>  I do also beliebe it does report too much things anyway...

yes.

> but then, also, some of these stuff could be included in [pdinfo] ;)


the **only** overlap between `[siginfo~]` and Pd-l2ork's `[pdinfo]` is the (new) DSP-state.

`[siginfo~]` was *especially* meant to query information about the *local* signal, not about the global state of Pd.
e.g. the number of channels in a multichannel cord. or the blocksize of *this* signal (because Pd enforces this is conincidentally the blocksize of the entire canvas), but not the "*global* blocksize".

umlaeute:
@millerpuckette : I've now implemented selectable info (via args) and dsp-state:


~~~
[siginfo~ dspstate dspstate blocksize]
|
[1 1 64\
~~~

dsp-off events are only generating output, if `[siginfo~]` is asked to report dspstate (that is: `[siginfo~ samplerate dspstate]` will report something like `[48000 0(`, but `[siginfo~ samplerate]` will only report the samplerate when turning DSP on and nothing when DSP is turned off.


providing no arguments (just `[siginfo~]`) will report all info as routable messages. (on DSP-off it will *only* report the `dspstate`)


allowed selectors are:
- `dspstate`
- `blocksize`
- `numchannels`
- `overlap`
- `samplerate`
- `samplespersecond`

porres:
> e.g. the number of channels in a multichannel cord. or the blocksize of this signal (because Pd enforces this is conincidentally the blocksize of the entire canvas), but not the "global blocksize".

I'm confused, how can we have a block size in a signal that is different that the canvas blocksize, and if it's not possible, how is it local?

Anyway, I still believe that the number of channels belongs to the [snake~] family

Now, about samplerate, the [samplerate~] object effectively reports the up/downsampling of a canvas, not a global samplerate state. An actual global state might be nice as we need to derive from the oversampling rate, which is not retrievable, not even in this PR yet... so if we're reporting the overlap number, we can also report the up/downsample factor.

I agree "dspinfo~" should report a global sample rate, which is not what we have here or in [samplerate~]. On the other hand, the dsp status IS global. So an object called [dspinfo~] should have the DSP status and perhaps the global samplerate that we don't have anywhere.

So, well, I am proposing quite an overhaul here. In short, here are my suggestions  in order of preference:

1) Take DSP status and number of channels out and rename this object to [blockinfo], and also include up/downsample. Have a new [dspinfo~] for status and actual global sample rate. Have [snake~] report number of channels.

2) Also include: A) global samplerate; B) up/downsampling factor; and call this whatever you like cause I can't think of a good name as I think this is just a kludge that gathers a lot of separate info :)

porres:
> the only overlap between [siginfo~] and Pd-l2ork's [pdinfo] is the (new) DSP-state.

I think it also reports cool global things, like number of audio signal inputs and outputs, which come from global settings. And this object actually reports the global sample rate, which is something we don't have, as I mentioned in my last message above.

umlaeute:
> So, well, I am proposing quite an overhaul here

this might be so. but then it's not about this PR.

*this* PR is about `[siginfo~]`, an object that describes the signal properties, as can be found in the `t_signal` structure.

that's it. 

there are *zero* plans to turn this into a generic object to query the DSP properties of the system, nor about cool features of Pd-l2ork's `[pdinfo]`, nor about reading soundfiles.
it's also not about existing objects in other libraries that might provide similar (or different) functionality.


if you don't like the object's design, that's fine for me, just say so (well, your already did).
but I'm not going to change this PR fundamentally.

if you want to propose a different object, please open a separate pull request (or issue)

porres:
> this PR is about [siginfo~], an object that describes the signal properties, as can be found in the t_signal structure.

The limited scope around the 't_signal' structure wasn't clear, but then, getting the DSP status already falls out of that scope. I'd expect to get more things from such a general and vaguely named obejct. I also proposed to include the global (aka "system") sample rate, and/or oversampling number with no real overhaul. What about that?

Today, there's simply no way to get that and it always bothered me. I can't think of a use case on the top of my head right now, but my [else/sr~] external does report the global sample rate and I remember I needed it at some point. If no one else bothered to open an issue and propose such a  feature, I'm actually cool... I'm not going farther than proposing it in here.

> if you don't like the object's design, that's fine for me, just say so (well, your already did).

I don't have strong opinions about the design. I do feel [snake~] is a better place to report number of channels, I guess that's the main criticism, but I can live with whatever we get. I'll be glad to have the feature one way or another. And then, there's already another PR for that ( https://github.com/pure-data/pure-data/pull/2615 ) ... so...

umlaeute:
> getting the DSP status already falls out of that scope. 

the object doesn't get the global DSP, but the local DSP state (of the objects signal inlet).
as such I think that it is in scope of the object. 

to be true, this is currently only true for the DSP-on state; the DSP-off state is reported globally, but that's a bug that needs to be fixed.

also note that the object originally did not report the DSP state explicitly. I only added this function after miller requested it; but I now agree that it is a good addition (and the object was already implicitly reporting DSP-on, because this is the event that triggers the report)

umlaeute:
> to be true, this is currently only true for the DSP-on state

actually, it's not not even true for the DSP-on state (since the DSP graph is built when the global DSP state changes).
i'll check if the object react on local DSP state only (so we get an output when toggling `[switch~]`).
it's easier if the automatic output reacts on global DSP state (even if it only reports the local DSP state)

millerpuckette:
My thinking now is to go on and merge this and also to write a "sr" object, not a tilde object at all, that acts like samplerate~ except that it would output the value whenever DSP starts.

And I imagine "sr" could take optional arguments to report, for instance, sample overlap, although by default it should act as if you had sent a bang to samplerate~.

Question though: "dsp~" is an abstraction in iemlib... has someone somewhere written a "sr" abstraction that I should avoid displacing?

umlaeute:
how about spelling it out: `[samplerate]` ?

assuming that most commonly used libraries are available on deken, you can to the webinterface and search for objects: <https://deken.puredata.info/search.html?objects=samplerate>

note that libraries that have not been updated for a while (or that use non-standard means to create the `.dek` package), might not have an object list, so those will be missed.
i haven't checked how many objects this would exclude...

umlaeute:
before you merge: should I try to report the canvas-local dsp state first (rather than the global dsp-state)?

so a `[siginfo~]` in a `[switch~]`ed off canvas would report `dspstate 0` when turning on the DSP?

millerpuckette:
You're welcome to try that :)

umlaeute:
@millerpuckette I think it's ready now (with a couple of crasher bugs fixed ;-))

Spacechild1:
Sorry for being late to the party. In general, I like this, but **please** output the info through individual outlets if the user provides arguments like `[siginfo~ samplerate blocksize]`. Sending everything as a single list requires the user to unpack the values, which I find rather awkward. I don't see many use cases where I would want the info as an actual list. Most of the time I would send the individual outputs to different destinations. If someone really needs the output as a list, this can be easily done with `[pack]`. I think we should optimize for (what I think is) the common case. This would also match the behavior of `[sigmund~]`.

millerpuckette:
+1 :)

porres:
hi, this says it closes https://github.com/pure-data/pure-data/issues/1687 but it only partially does so, as Closes: https://github.com/pure-data/pure-data/issues/1687 also requests for a way to get the oversampling factor.

Like I said on that issue/request, I do have an external for blocksize and I can see why it could be useful, but not essential in my cases as the blocksize (and other block info like overlap for that matter) are set by the user. Number of channels are clearly quite needed and essential. I just wonder what is the use case for the other info, if there's a particular example I may be missing.

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+        siginfo_tilde_outmsg(x, s_numchannels, numchannels);
+        siginfo_tilde_outmsg(x, s_overlap, overlap);
+        siginfo_tilde_outmsg(x, s_samplerate, samplerate);
+        siginfo_tilde_outmsg(x, s_samplespersecond, samplespersecond);
+        return;
+    }
+    for(unsigned int i=x->x_argc; i>0; i--) {
+        t_symbol*s = vec[i-1];
+        t_float value = -1;
+        if(0) {
```
Spacechild1: That's pretty idiosyncratic and not used anywhere in the rest of the code base.

Also, we tend to omit braces for single line if/else-clauses. If you want to keep the braces, please use Allman-style.

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+    if(!vec) {
+        siginfo_tilde_outmsg(x, s_dsp, state);
+        siginfo_tilde_outmsg(x, s_blocksize, blocksize);
+        siginfo_tilde_outmsg(x, s_numchannels, numchannels);
+        siginfo_tilde_outmsg(x, s_overlap, overlap);
+        siginfo_tilde_outmsg(x, s_samplerate, samplerate);
+        siginfo_tilde_outmsg(x, s_samplespersecond, samplespersecond);
+        return;
+    }
+    for(unsigned int i=x->x_argc; i>0; i--) {
```
Spacechild1: operators like `=` or `>` should be separated by whitespace. There are many other cases of this in this PR.

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+    t_symbol*s_samplespersecond = gensym("samplespersecond");
+
+    int state = x->x_globaldspstate && canvas_getswitchedon(x->x_canvas);
+    int blocksize = x->x_blocksize;
+    int numchannels = x->x_numchannels;
+    int overlap = x->x_overlap;
+    float samplerate = canvas_getsr(x->x_canvas);
+    float samplespersecond = x->x_samplespersecond;
+
+    if(!vec) {
```
Spacechild1: There should be a whitespace after `if`. Also, Pd generally uses Allman-style for braces!

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
         gensym("dsp"), A_CANT, 0);
 }
 
+/* -------------------------- siginfo~ ------------------------------ */
+
+static t_class *siginfo_tilde_class, *siginfo_proxy_class;
+
+typedef struct _siginfo_proxy {
```
Spacechild1: Please use Allman-style for struct declarations.

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+    t_pd x_pd;
+    struct _siginfo_tilde*x_parent;
+} t_siginfo_proxy;
```
Spacechild1, line 877: missing newline.

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
         gensym("dsp"), A_CANT, 0);
 }
 
+/* -------------------------- siginfo~ ------------------------------ */
+
+static t_class *siginfo_tilde_class, *siginfo_proxy_class;
+
+typedef struct _siginfo_proxy {
+    t_pd x_pd;
+    struct _siginfo_tilde*x_parent;
```
Spacechild1: missing space before `*`

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+} t_siginfo_proxy;
+typedef struct _siginfo_tilde
+{
+    t_object x_obj;
+    t_outlet **x_outlet;        /* outlets */
+    t_float x_f;                /* scalar inlet */
+    t_siginfo_proxy x_proxy;    /* proxy object to receive "pd-dsp-stopped" events */
+    t_canvas *x_canvas;         /* canvas we live in (for the local dsp state and samplerate) */
+
+    t_symbol**x_vec;            /* object arguments (or NULL) */
```
Spacechild1: missing space before `**`

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+
+static void siginfo_tilde_bang(t_siginfo_tilde*x) {
+    t_symbol**vec = x->x_vec;
+
+    t_symbol*s_dsp = gensym("dspstate");
+    t_symbol*s_blocksize = gensym("blocksize");
+    t_symbol*s_numchannels = gensym("numchannels");
+    t_symbol*s_overlap = gensym("overlap");
+    t_symbol*s_samplerate = gensym("samplerate");
+    t_symbol*s_samplespersecond = gensym("samplespersecond");
```
Spacechild1: missing spaces before `*`

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+
+    int state = x->x_globaldspstate && canvas_getswitchedon(x->x_canvas);
+    int blocksize = x->x_blocksize;
+    int numchannels = x->x_numchannels;
+    int overlap = x->x_overlap;
+    float samplerate = canvas_getsr(x->x_canvas);
+    float samplespersecond = x->x_samplespersecond;
```
Spacechild1: shouldn't these be `t_float`?

src/d_ctl.c
```diff
@@ -866,6 +866,193 @@ static void threshold_tilde_setup(void)
...
+    } else {
+        x->x_outlet = (t_outlet**)getbytes(sizeof(*x->x_outlet));
+        x->x_outlet[0] = outlet_new(&x->x_obj, 0);
+    }
+    t_symbol**vec = x->x_vec;
+    int warned = 0;
+    for(int i=0; i<argc; i++) {
+        x->x_outlet[i] = outlet_new(&x->x_obj, 0);
+        s = atom_getsymbol(argv+i);
+        if (0
```
Spacechild1: Again, this is rather idiosyncratic.

Spacechild1:
@umlaeute I have a few stylistic concerns. Normally, you're following Pd's style, but for some reason in this PR you're not, at least not in `d_ctl.c`. (Strangely, your changes in `d_ugen.c` do follow Pd's style.) Maybe you're using a code formatting tool and you just forgot to apply to `d_ctl.c`?

I won't list every single thing. In general:
- Allman-style for braces
- space before `*` (= pointer)
- space around operators
- space after `if`/`else`

But you know that anyway :)

Spacechild1:
>  I just wonder what is the use case for the other info, if there's a particular example I may be missing.

`overlap` is useful for adjusting the frequency of oscillators or ramps in FFT patches. I'm not sure what's the use case for `samplespersecond`, though.

As a side note: I tend to think that `numchannels` could be shortened to `channels` because `[siginfo~ channels]` is already long enough :) What do others think?

umlaeute:
> @umlaeute I have a few stylistic concerns.
[...]
> But you know that anyway :)

indeed.
thanks for pointing this out. fixed in d6e0fd56bc5817ea09cf6b3c7dd8ee978248fb61

umlaeute:
> I tend to think that `numchannels` could be shortened to `channels`

oh totally.

i was busy expanding the cryptic `nchans` to the verbose `numchannels`, without realizing that `channels` is the obvious choice.

fixed in 259a6f2d9d9be30e76b2045ebc96bec6e4954f9c

src/d_ctl.c
```diff
@@ -866,6 +866,205 @@ static void threshold_tilde_setup(void)
...
+        x->x_outlet = (t_outlet **)getbytes(sizeof(*x->x_outlet)*argc);
+    }
+    else
+    {
+        x->x_outlet = (t_outlet **)getbytes(sizeof(*x->x_outlet));
+        x->x_outlet[0] = outlet_new(&x->x_obj, 0);
+    }
+    t_symbol **vec = x->x_vec;
+    int warned = 0;
+    for(int i = 0; i < argc; i++)
```
Spacechild1: Really sorry for nitpicking, but all the `for` loops are missing a whitespace before the parenthesis.

Another thing: aren't we supposed to write C89, i.e. all variables declared on the top of the block and no variable declarations in for-loop initializers? (Personally, I would *love* to switch to C99, but I think we need the official blessing of Miller :)

src/d_ctl.c
```diff
@@ -866,6 +866,205 @@ static void threshold_tilde_setup(void)
...
+    {
+        x->x_outlet[i] = outlet_new(&x->x_obj, 0);
+        s = atom_getsymbol(argv+i);
```
Spacechild1, line 1028: Again, very nitpicky, but repurposing an unrelated variable feels a bit iffy to me.

src/d_ugen.c
```diff
@@ -219,6 +219,24 @@ static void block_set(t_block *x, t_floatarg fcalcsize, t_floatarg foverlap,
     canvas_resume_dsp(dspstate);
 }
 
+int canvas_getswitchedon(t_canvas *x)
+{
+    t_float srate = sys_getsr();
```
Spacechild1: `srate` is not really used. Probably a leftover.

src/d_ctl.c
umlaeute: > Really sorry for nitpicking, but all the for loops are missing a whitespace before the parenthesis.

that's what you get when you have to fix the coding style manually :-)

> aren't we supposed to write C89

this is something i considered briefly, and then decided that miller has already broken the C99-rule by using anonymous unions in https://github.com/pure-data/pure-data/blob/df6f59497f6440d69f7a8dcdac9c4b017fab2ca9/src/m_pd.h#L710-L714

but you are right: this would be the first occurence of variable declarations in for-loop initializers, so i fixed it.

src/d_ugen.c
umlaeute: yea, copy-paste error.

src/d_ctl.c
umlaeute: well, it also got rid of an otherwise *unused* parameter. ;-)

src/d_ctl.c
```diff
@@ -866,6 +866,211 @@ static void threshold_tilde_setup(void)
...
+        x->x_outlet = (t_outlet **)getbytes(sizeof(*x->x_outlet)*argc);
+    }
+    else
+    {
+        x->x_outlet = (t_outlet **)getbytes(sizeof(*x->x_outlet));
+        x->x_outlet[0] = outlet_new(&x->x_obj, 0);
+    }
+    t_symbol **vec = x->x_vec;
+    int warned = 0;
+    int i;
```
Spacechild1, line 1024: That's hopefully my last nitpick for today. By moving the declaration right above the for-loop, you're only writing C94 :) For proper C89, you'd also have to move all three variables above up to the top of the function.

src/d_ctl.c
Spacechild1: IIRC, Pd has already been using a couple of C99 library features, but I never said anything :)

Here it's really more about the code style. On the one hand, it feels a bit silly to write C89 style in 2026, but on the other hand, all Pd code is currently written in that style. I think we should definitely have a discussion about that because IMO C99-style leads to much cleaner code.

src/d_ctl.c
millerpuckette: OK if I squash-merge this?  (Also I might want to make a small tweak or two to the code :)

umlaeute: go ahead

millerpuckette:
Merged with tweaks.  I should have put the tweaks into the PR itself but didn't know to do that.  This can now be 'closed' if all looks OK.

PR #2926, opened 2026-06-12
Update docs

porres:
for now, just correct the version in the manual

not sure I'll be able to do anything else


porres:
hmmm, there was also an HTML validation error from the current status where an ID was duplicated. I fixed it so I guess all checks pass now...

like I said, I don't have real plans to add more fixes to the docs, so I guess this is it @millerpuckette

porres:
ok, when this is getting squashed and merged, it still stays active. I can keep closing it afterwards, but it's clumsy :)

PR #2940, opened 2026-06-30
Add StartupWMClass to pd-gui.desktop

City-busz:
This allows desktop environments to associate the main window with the desktop file and display the appropriate icon for it.

### DIFF
--- a/linux/org.puredata.pd-gui.desktop
+++ b/linux/org.puredata.pd-gui.desktop
@@ -15,3 +15,4 @@ GenericName=Dataflow IDE
 Type=Application
 Terminal=false
 Icon=puredata
+StartupWMClass=PdWindow


umlaeute:
the fix looks trivial enough to include it without much thought.

however, before i merge this into the `develop` branch, i would like to know how you trigger your issue in the first place (or put otherwise: here (Debian/sid xfce4), my Pd window has the proper icon without further ado)

City-busz:
Without this, the window has no icon in GNOME 50 (Wayland) on the dock and on the Alt-Tab switcher, because it tries to read the icon from the associated .desktop file instead of reading it from the X window property.

| Before  | After|
| ------------- | ------------- |
| gnome-shell-50-before  | gnome-shell-50-after  |

umlaeute:
thx.

how did you install Pd? how do you start Pd? (aka: is there a difference between starting from the terminal, or starting from the menu?)

City-busz:
I'm using Arch Linux (native package). There is no difference if I launch Pd from the menu or start it from the terminal.

PR #2961, opened 2026-08-09
GUI: proof-of-concept backend abstraction with Tcl/Tk and GTK4 implementations

charlesneimog:
Besides being a large PR, this should be considered a **proof of concept**. Most of the changes are related to replacing direct `pdgui_vmess()` calls with a set of semantic GUI functions.

Currently, implementing a substantially different GUI still requires understanding and reproducing Tcl/Tk concepts, or modifying GUI-related code throughout the Pd sources. This is also one of the difficulties encountered in the ongoing GTK work by @millerpuckette. *I think!*

This PR explores a different approach: instead of having the Pd core describe *how* a GUI operation should be performed through Tcl/Tk-specific messages, the core expresses *what* GUI operation it needs. For example:

```c
pdgui_item_destroy(...);
pdgui_item_move(...);
pdgui_text_set_position(...);
pdgui_patchcord_create(...);
pdgui_canvas_set_title(...);
pdgui_canvas_set_cursor(...);
```

For me, this is similar to what Pd already does with the Audio and MIDI APIs: the core defines the operations, while each backend implements them. Tcl/Tk could continue using `pdgui_vmess()` internally, while GTK, WebAssembly, or other backends implement the same interface differently.

WebAssembly is an important motivation. A web backend could eventually allow Pd itself — including patch creation, unlike the current [pd4web](https://github.com/charlesneimog/pd4web/) approach — to run in a browser. This could make Pd more accessible on smartphones and enable installable PWAs without depending on the Apple or Google app stores.

This **proof of concept** was developed **primarily using AI coding agents**. I defined and guided the architecture, while agents performed much of the mechanical refactoring and implementation. The current API, function names, and GTK implementation should therefore not be considered final; the goal of this PR is to demonstrate the approach and discuss whether this abstraction is worth pursuing.


umlaeute:
as this is a POC and meant as a basis for discussions (afaict), i'm converting this to a *draft* (so it doesn't get accidentally accepted).

PR #2962, opened 2026-08-14
Unbind "#A" when a canvas is freed

### LINKED ISSUE #2079: [garray] use-after-free when unbinding from #A
I've been running into this problem for a while when running pure-data with the ```-fsanitize=address ``` compilation flag. This enables Address Sanitizer, which will reveal runtime memory issues that might not always come out normally. This is the only memory issue I'm see popping up often, so it might be worth investigating.

In this bit of code, it detects a use-after-free problem:

https://github.com/pure-data/pure-data/blob/c0a7c16c63e86ccc8ce05c8273ccd8aec7ae044c/src/g_array.c#L553-L554

Specifically, it will crash inside the pd_findbyclass function, when it tries to do ```*s->s_thing```.

<br/>

I don't fully understand why this happens. But replacing it with something like this:

```
   if(gensym("#A")->s_thing == &x->x_gobj.g_pd)
   {
       pd_unbind(&x->x_gobj.g_pd, gensym("#A"));
   }
```

Appears to solve it, and to me looks like it's a more simple method to accomplish the same thing. However, I'm not familiar enough with this part of pure-data's source code to judge if this is completely correct. I don't understand why there is a while-loop in the current code, shouldn't there be only one thing assigned at a time to gensym("#A")->s_thing?

<br/>

Steps to reproduce:

1. Compile pure-data with -fsanitize=address. I've mostly tested it on macOS, but I've seen it happen on Linux too. Both clang and gcc's sanitizer will report this issue.

2. I've found multiple patches affected by this bug, but having some arrays wrapped in graphs seems to be essential. What triggers it every time for me, is loading the "multi.vsl" help file from [ELSE](https://github.com/porres/pd-else/), and then closing the help file window.

### LINKED ISSUE #2777: Use after free resulting in potential crash
I've recently run into a crash using PD 0.55.2, but I believe the same is still valid also for 0.56.1.

This is the minimal test case that causes a crash on my machine (Windows 10). 
Just load **crash_test.pd** and close it to destroy the patch. 
Since this is a use after free it's possible that it won't always crash (or at all).

**crash_test.pd**
```
#N canvas 807 561 450 300 12;
#N canvas 0 50 450 250 (subpatch) 0;
#X array array1 100 float 3;
#A 0 0 0 0 0 0 0 0 0.428571 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0;
#X coords 0 1 100 -1 200 140 1;
#X restore 217 143 graph;
#X obj 75 24 loadbang;
#N canvas 174 634 450 300 nested 1;
#X obj 10 10 crash_test_abs;
#X obj 10 20 crash_test_abs;
#X restore 43 163 pd nested;
#X msg 204 56 obj 10 10 crash_test_abs;
#X obj 41 126 send pd-nested;
#X msg 178 89 obj 10 20 crash_test_abs;
#X obj 68 53 t b b;
#N canvas 0 50 450 250 (subpatch) 0;
#X array array2 100 float 3;
#A 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 -0.185712 -0.599997 -0.414283 -0.285712 -0.114283 -0.042855 0.185716 0.300001 0.514287 0.614286 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0;
#X coords 0 1 100 -1 200 140 1;
#X restore 7 221 graph;
#X connect 1 0 6 0;
#X connect 3 0 4 0;
#X connect 5 0 4 0;
#X connect 6 0 5 0;
#X connect 6 1 3 0;
```

**crash_test_abs.pd**
```
#N canvas 1078 572 450 300 12;
#X obj 24 26 inlet~;
#X obj 154 57 inlet;
#X obj 24 197 outlet~;
#X obj 24 166 *~, f 19;
#X obj 154 139 line~;
#X msg 154 112 \$1 10;
#X obj 154 84 expr if($f1 < 0.5 \, 0 \, 1);
#X connect 0 0 3 0;
#X connect 1 0 6 0;
#X connect 3 0 2 0;
#X connect 4 0 3 1;
#X connect 5 0 4 0;
#X connect 6 0 5 0;
```

After doing some debugging I think I have an idea of what may be going on: if dynamic object creation is used to create an object from an abstraction, after the parsing is done, `#A` will stay bound to the last created object [because this is called](https://github.com/pure-data/pure-data/blob/e3943608ab65d3e28f0a6b9f8a6f61de81fb1095/src/g_canvas.c#L1231).

If the patch is destroyed, and it also contains a garray, we may end up in a situation where the object bound to `#A` is destroyed first. When the garray is also destroyed, [it will try to unbound](https://github.com/pure-data/pure-data/blob/e3943608ab65d3e28f0a6b9f8a6f61de81fb1095/src/g_array.c#L618) any remaining arrays bound to `#A`, resulting in a use after free [here](https://github.com/pure-data/pure-data/blob/12343d6f2d9e4753653319f0e6bcb75cec393d9f/src/m_pd.c#L206).

Here's some debug output I've added that may clear things up a bit:

```
CANVAS FREE: crash_test.pd 000001CCE7D582B0 
GARRAY FREE array1 
#A Thing: 000001CCE7D586F0 
CANVAS FREE: (subpatch) 000001CCE7D58800 
CANVAS FREE DONE: (subpatch) 000001CCE7D58800 
CANVAS FREE: crash_test_abs.pd 000001CCE7D58910 
CANVAS FREE DONE: crash_test_abs.pd 000001CCE7D58910 
CANVAS FREE: crash_test_abs.pd 000001CCE7D585E0 
CANVAS FREE DONE: crash_test_abs.pd 000001CCE7D585E0 
CANVAS FREE: crash_test_abs.pd 000001CCE7D57F80 
CANVAS FREE DONE: crash_test_abs.pd 000001CCE7D57F80 
PD FREE 000001CCE7D586F0
CANVAS FREE: crash_test_abs.pd 000001CCE7D586F0 
/CANVAS FREE DONE: crash_test_abs.pd 000001CCE7D586F0 
CANVAS FREE: nested 000001CCE7D583C0 
/CANVAS FREE DONE: nested 000001CCE7D583C0 
GARRAY FREE array2 
#A Thing: 000001CCE7D586F0
```

As you can see `000001CCE7D586F0` is freed before `array2` tries to access it while being freed.

I'm not sure what a proper fix for this might be, to be honest I'm not sure why the `canvas_popabstraction` binds `#A`. I thought that was reserved for arrays but I'm probably missing something.

jamiebullock:
This PR fixes a bug where freeing a canvas leaves the symbol `#A` pointing at freed memory.

`canvas_popabstraction()` binds every new abstraction to `#A` and nothing ever unbinds it, so once the canvas is freed the binding dangles. The next `garray_free()` or `textbuf_free()` walks it, since both run `while ((x2 = pd_findbyclass(gensym("#A"), ...))) pd_unbind(x2, gensym("#A"));`, and `pd_findbyclass()` dereferences `*e->e_who`.

Steps to reproduce:

`myabs.pd` is an abstraction containing only an inlet wired to an outlet. `repro.pd` has an empty subpatch `holder`, a subpatch `arrays` containing an array, and a `loadbang` firing one message box:

```
; pd-holder obj 20 20 myabs ; pd-holder clear ; pd-arrays clear ; pd quit
```

[pd-A-binding-repro.zip](https://github.com/user-attachments/files/31077819/pd-A-binding-repro.zip)

Run it headless against a build made with `-fsanitize=address`:

```
$ pd -nogui -noaudio -nomidi -path . repro.pd
==28588==ERROR: AddressSanitizer: heap-use-after-free
    #0 pd_findbyclass  m_pd.c:206
    #1 garray_free     g_array.c:618
```

In an ordinary build the read is silent, so it tends to surface later as an unrelated crash.

The fix is to drop the `#A` binding in `canvas_unbind()`, alongside the canvas's own bind symbol:

- `canvas_unbind()` unbinds `#A` when the canvas is bound to it
- adds `pd_isbound()`, since bindlists are private to `m_pd.c` and callers cannot otherwise unbind conditionally

closes #2777
closes #2079

### DIFF
--- a/src/g_canvas.c
+++ b/src/g_canvas.c
@@ -758,6 +758,8 @@ static void canvas_unbind(t_canvas *x)
 {
     if (strcmp(x->gl_name->s_name, "Pd"))
         pd_unbind(&x->gl_pd, canvas_makebindsym(x->gl_name));
+    if (pd_isbound(&x->gl_pd, gensym("#A")))
+        pd_unbind(&x->gl_pd, gensym("#A"));
 }
 
 void canvas_reflecttitle(t_canvas *x)
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -157,6 +157,21 @@ void pd_bind(t_pd *x, t_symbol *s)
     else s->s_thing = x;
 }
 
+int pd_isbound(t_pd *x, t_symbol *s)
+{
+    if (s->s_thing == x)
+        return (1);
+    else if (s->s_thing && *s->s_thing == bindlist_class)
+    {
+        t_bindlist *b = (t_bindlist *)s->s_thing;
+        t_bindelem *e;
+        for (e = b->b_list; e; e = e->e_next)
+            if (e->e_who == x)
+                return (1);
+    }
+    return (0);
+}
+
 void pd_unbind(t_pd *x, t_symbol *s)
 {
 #ifdef VST_CLEANSER
--- a/src/m_pd.h
+++ b/src/m_pd.h
@@ -443,6 +443,7 @@ EXTERN t_pd *pd_new(t_class *cls);
 EXTERN void pd_free(t_pd *x);
 EXTERN void pd_bind(t_pd *x, t_symbol *s);
 EXTERN void pd_unbind(t_pd *x, t_symbol *s);
+EXTERN int pd_isbound(t_pd *x, t_symbol *s);
 EXTERN t_pd *pd_findbyclass(t_symbol *s, const t_class *c);
 EXTERN t_pd *pd_findbyclassname(t_symbol *s, const t_symbol *classname);
 EXTERN void pd_pushsym(t_pd *x);


millerpuckette:
Thanks for figuring this out.  I've adopted this in a slightly more localized way (avoiding adding anything to the API)

porres:
closing in favor of https://github.com/pure-data/pure-data/commit/0dce8b2224886f9bdcdac4a38bcdf8d4908d1cc0

thanks @jamiebullock

PR #2965, opened 2026-08-21
vline~: don't drop pending segments while DSP is running

### LINKED ISSUE #2956: vline~ swallows schedule messages after turning DSP on
1120ac94 introduced a regression that leads to [vline~] omitting the first scheduling message after turning DSP on. Check this patch:

```
#N canvas 679 42 557 395 10;
#X obj 38 14 bng 20 250 50 0 empty empty empty 0 -10 0 12 #fcfcfc #000000 #000000;
#X obj 38 126 t b b b;
#X obj 181 156 t b b, f 9;
#X obj 181 179 del 500;
#X msg 181 202 0;
#X msg 232 201 1;
#X obj 181 286 switch~;
#X obj 38 199 vline~;
#X obj 39 250 env~ 2048 2048;
#X floatatom 39 324 5 0 0 0 - - - 0;
#X obj 181 234 t a a;
#X obj 208 258 tgl 18 0 empty empty empty 0 -9 0 10 #fcfcfc #000000 #000000 0 1;
#X obj 38 88 del 0.5231;
#X msg 110 17 0 30;
#X msg 38 167 1 \, 0 1000;
#X obj 191 13 bng 20 250 50 0 empty empty empty 0 -10 0 12 #fcfcfc #000000 #000000;
#X text 22 16 1;
#X text 89 15 2, f 2;
#X text 171 15 3, f 2;
#X text 79 323 <- should show ~94 \, not 89;
#X text 138 83 it seems the first scheduling message to vline~ is "swallowed" after switch~ing DSP on.;
#X connect 0 0 12 0;
#X connect 1 0 14 0;
#X connect 1 2 2 0;
#X connect 2 0 3 0;
#X connect 2 1 5 0;
#X connect 3 0 4 0;
#X connect 4 0 10 0;
#X connect 5 0 10 0;
#X connect 7 0 8 0;
#X connect 8 0 9 0;
#X connect 10 0 6 0;
#X connect 10 1 11 0;
#X connect 12 0 1 0;
#X connect 13 0 7 0;
#X connect 14 0 7 0;
#X connect 15 0 12 0;
```

Image

The ramp should restart from 1 after hitting '3', but it actually continues going downwards.

ben-wes:
don't drop pending segments while DSP is running. if the list empties, clear the tail but still insert new segment.

* closes #2956 

i do wonder if the additional DSP check might cause other issues - but afaict, segments accumulating during dsp:off were the reason for the original fix, so it should probably be fine.

### DIFF
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -334,17 +334,15 @@ static void vline_tilde_float(t_vline *x, t_float f)
         where lots of them get added while DSP isn't running. We add 0.1 msec
         to the stop time limit just in case of truncation error in case it and
         the current time are logically, but not numerically, equal. */
-    while (x->x_list && timenow > x->x_list->s_stoptime + 0.1)
+    while (!pd_getdspstate() && x->x_list &&
+        timenow > x->x_list->s_stoptime + 0.1)
     {
         t_vseg *was = x->x_list;
         x->x_list = x->x_list->s_next;
         freebytes(was, sizeof(*was));
-        if (!x->x_list)
-        {
-            vline_tilde_stop(x);
-            return;
-        }
     }
+    if (!x->x_list)
+        x->x_tail = 0;
         /* negative delay input means stop and jump immediately to new value */
     if (inlet2 < 0)
     {
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -334,16 +334,16 @@ static void vline_tilde_float(t_vline *x, t_float f)
         where lots of them get added while DSP isn't running. We add 0.1 msec
         to the stop time limit just in case of truncation error in case it and
         the current time are logically, but not numerically, equal. */
-    while (x->x_list && timenow > x->x_list->s_stoptime + 0.1)
+    if (!pd_getdspstate())
     {
-        t_vseg *was = x->x_list;
-        x->x_list = x->x_list->s_next;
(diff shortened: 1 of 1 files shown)


reduzent:
Thanks for providing a fix. Works for me.

src/d_ctl.c
```diff
@@ -334,17 +334,15 @@ static void vline_tilde_float(t_vline *x, t_float f)
         where lots of them get added while DSP isn't running. We add 0.1 msec
         to the stop time limit just in case of truncation error in case it and
         the current time are logically, but not numerically, equal. */
-    while (x->x_list && timenow > x->x_list->s_stoptime + 0.1)
+    while (!pd_getdspstate() && x->x_list &&
```
umlaeute: is there any specific reason why `pd_getdspstate()` is queried in each iteration of the loop?
can the DSP state actually change while the loop is running?

ben-wes: no. actually i first had it as:

```c
    if (!pd_getdspstate())
    {
        while (x->x_list && timenow > x->x_list->s_stoptime + 0.1)
```

... only put it in the while-condition to avoid the extra indentation and expecting that check to be cheap enough. but it's certainly a redundant check.

ben-wes: decided to add the separate check. seems cleaner - although the diff looks a bit ugly.

PR #2968, opened 2026-08-27
Import deken v0.10.22

umlaeute:
https://github.com/pure-data/deken/commit/785862399822b6d12f4804cf7b12f5ed55ee10c4

- integrate with the new keyboard shortcut system
- fix binding to <kbd>Escape</kbd> (it was emitting an message at debug-level)


umlaeute:
i think this needs another thought.

umlaeute:
there was an issue if you assigned a keyboard shortcut to the plugin manager, it would not be displayed in the menu after a restart.
this has now been resolved.

PR #2970, opened 2026-08-28
add inlet to receive

### LINKED ISSUE #603: request: inlet for [receive] to set receive symbol

porres:
reboot of https://github.com/pure-data/pure-data/pull/604

this makes [receive] behave just like [send] - when you load it without arguments, it creates an inlet that you can use to set a receive name by giving it a symbol message.

closes https://github.com/pure-data/pure-data/issues/603

this was my oldest PR and it required the recently merged https://github.com/pure-data/pure-data/pull/1760

### DIFF
--- a/src/x_connective.c
+++ b/src/x_connective.c
@@ -316,15 +316,37 @@ static void send_setup(void)
     class_addanything(send_class, send_anything);
     class_sethelpsymbol(send_class, gensym("send-receive"));
 }
+
 /* -------------------- receive ------------------------------ */
 
 static t_class *receive_class;
+static t_class *receive_proxy_class;
+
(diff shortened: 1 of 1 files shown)


umlaeute:
instead of (or in addition to) detailing your personal history with this pr, it would be nice if you could: 
- explain what this PR does
- mention any issues this PR closes

(it's probably enough to copy the original PR description)

millerpuckette:
You did this more carefully than I would have - I would have just used an unreachable symbol like " set" (with a space) - but it's barely possible although probably always a terrible idea to actually generate a " set" message and confuse things.

PR #2977, opened 2026-08-30
a few bugfixes about zooming

### LINKED ISSUE #2976: array drawing glitches when zooming in/out
_disclaimer: i consider this new zoom feature a great improvement (having heard multiple requests about it in the past)!_

testing on macos with current master at 621423908ab9fa005045ead73047d910d2e4a8fd

the issue only occurs when the array is drawn with points. zooming in and out (from/to 100%) makes the graph disappear completely (i also just noticed another glitch where my message text gets replaced by its previous text on zoom, which should possibly be another issue - see screenshot, where the actual text is `resize 48000`)

Image

Ant1r:
fixes #2976:
- array with 'points' style is badly drawn at zoom level other than 100%. The solution here is to draw it as lines instead of rectangles (which had to use a white outline color, hence the glitches). 
- when a text content is modified at 100% zoom, an optimization was preventing the text tag (that is added by zoom processing) to be updated. 

This PR also fixes:
- when a text is created at different zoom levels (excepting integer levels like 200% or 400%), the actual font size (stored in a font tag) is different because of rounding error. The solution is to always add the font tag during text creation (excepting if current zoom is 100%, because no rounding problem in this case).

### DIFF
--- a/src/g_template.c
+++ b/src/g_template.c
@@ -2042,7 +2042,6 @@ static void plot_vis(t_gobj *z, t_glist *glist,
 
     if (glist->gl_isgraph)
         linewidth *= glist_getzoom(glist);
-
     if (tovis)
     {
          /* we use t_word because pdgui_vmess() has a convenient FLOATWORDS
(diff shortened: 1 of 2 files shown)


millerpuckette:
I fixed a couple of conflicts, notably in pd_canvaszoom.tcl, "scale font if any" - I don't know why auto-merge failed.  It might be worth checking that I got it right :)

Ant1r:
All seems to be good, thanks a lot!

Just one thing: this block has been uselessly re-added in g_template.c:

https://github.com/pure-data/pure-data/blob/dcbb8d0f95ca499e5a8775a82cdf028aedaf2ef3/src/g_template.c#L2050-L2051

but it's not a big deal.

millerpuckette:
cool - I took it back out.  I think it would be OK to take the declaration out of g_canvas.h too since anything using
it is doubtless broken anyway.

Ant1r:
well it shouldn't break anything (think of externals, that should continue to compile normally), it just doesn't do anything anymore and always returns "1".

PR #2979, opened 2026-08-31
bugfix/default zoom scale

### LINKED ISSUE #2978: scrollbar in zoom preferences eats chars
playing around with zoom on multiple platforms, i noticed that in my Win11 VM, the "default zoom" settings in the preferences lack the actual zoom factor:

Image

obviously this is because `scale`'s label is clipped to the size of the scale-widget, which is hardcoded to `200` (and for some fonts this is not enough).

umlaeute:
- fix typo in procname `defaut_zoom_callback` (should read: `default_zoom_callback`)
- make the string "Default zoom level" in the preferences translatable
- separatee widgets for the slider and the associated label "Default zoom level"
   - Closes: https://github.com/pure-data/pure-data/issues/2978
- prefer `::ttk::scale` over `scale` (if available)

### `::ttk::scale`

image

### normal `scale`
image

i don't really care much about `::ttk:scale` vs `::scale` but the former usually has a more *native* look and feel.

maybe @Ant1r wants to have a look before this gets accepted...

### DIFF
--- a/tcl/pd_canvaszoom.tcl
+++ b/tcl/pd_canvaszoom.tcl
@@ -38,18 +38,27 @@ proc ::pd_canvaszoom::init_default_zoom {} {
 after idle ::pd_canvaszoom::init_default_zoom
 
 proc ::pd_canvaszoom::default_zoom_callback {widget value} {
+    set value [expr 20 * int($value / 20.)]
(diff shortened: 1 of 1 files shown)


Ant1r:
fine :-)

PR #2989, opened 2026-09-01
[zoom] minor fixes

### LINKED ISSUE #2984: [zoom] splash can be very small
on my windows VM (which uses some weird scaling; eg see https://github.com/pure-data/pure-data/pull/2978), the overlay indicating the current zoom factor can be very small (practically unreadable).
I currently do not have access to the VM, so there's no screenshot yet. 

probably the don't should be `TkFixedFont` (or `TkDefaultFont`), which should scale automatically with the Font menu (untested).

### LINKED ISSUE #2985: [zoom] allow higher values for default zoom?
I wonder whether the "default zoom factor" in the properties isn't limited too much: currently the maximum value you get here is 200%.

what if you prefer larger fonts (because you are getting older :-)) and then you also get a HiDPI screen (which mandates yet another zoom in)?

I do not have a HiDPI screen myself, so this is a bit theoretical, but then: now is probably a good time to think about it.

I don't think that the current minimum (50%) needs adjusting

### LINKED ISSUE #2988: [zoom] splash does not have a constant position
as @Ant1r wrote in https://github.com/pure-data/pure-data/issues/2984#issuecomment-5484471343
>  there's that position glitch (sometimes the splash is not placed at the NW corner)

and
> The most obvious is when you manually scroll the patch while the splash is visible... But I guess this one could be fixed in a way or another.
>
> More random are the cases related to implicit patch scrolling, when the system tries to stick the content under the pointer, or at moment the scrollbars appear or disappear.

### LINKED ISSUE #2990: [zoom] scrollbars are wrong
consider a trivial non-empty patch, like this:
~~~
#N canvas 2655 476 450 300 12;
#X floatatom 51 38 5 0 0 0 - - - 0;
~~~

this patch apparently does not need any scollbars.

however, if i *zoom out* to e.g. 75% i suddenly get scrollbars out of nowhere for no apparent reason.

umlaeute:
- Closes: https://github.com/pure-data/pure-data/issues/2985
- Closes: https://github.com/pure-data/pure-data/issues/2988
- Closes: https://github.com/pure-data/pure-data/issues/2984
- Closes: https://github.com/pure-data/pure-data/issues/2990

ping @Ant1r for a review.

also ping @ben-wes as this still uses the splash for showing the zoom level instead of the window title (and he expressed preference over the latter in https://github.com/pure-data/pure-data/issues/2984#issuecomment-5485689188)

### DIFF
--- a/tcl/pd_canvaszoom.tcl
+++ b/tcl/pd_canvaszoom.tcl
@@ -238,30 +238,36 @@ proc ::pd_canvaszoom::scroll_point_to {c xcanvas ycanvas xwin ywin} {
     $c yview moveto $scrolly
 }
 
-proc ::pd_canvaszoom::delete_toastzoom {c} {
-    if {[info commands $c] == {}} return
-    $c delete _zoomtoast_
+proc ::pd_canvaszoom::delete_toastzoom {w} {
+    if [winfo exists $w] {
+        wm withdraw $w
+    }
 }
 
 proc ::pd_canvaszoom::toastzoom {c} {
     variable zdepth
     if { ! [info exists zdepth($c)] } {return}
-    set zoom [expr int($zdepth($c) * 100)]
-    set c ::pd_canvaszoom::canvas::$c
-    set scrollregion [$c cget -scrollregion]
-    set x0 [lindex $scrollregion 0]
-    set y0 [lindex $scrollregion 1]
-    set W [expr [lindex $scrollregion 2] - [lindex $scrollregion 0]]
-    set H [expr [lindex $scrollregion 3] - [lindex $scrollregion 1]]
-    set xT [expr $x0 + $W * [lindex [$c xview] 0] + 3]
-    set yT [expr $y0 + $H * [lindex [$c yview] 0] + 3]
-    after cancel ::pd_canvaszoom::delete_toastzoom $c
-    delete_toastzoom $c
-    $c create rectangle $xT $yT [expr $xT + 50] [expr $yT + 16] -tags _zoomtoast_ -fill "#E7E7E7"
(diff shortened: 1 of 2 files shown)


ben-wes:
> expressed preference

thanks for the ping! i don't really mind - it was just a minor aesthetic preference. i see that this PR now changes the appearance anyway! looks good to me - maybe border+shadow is a bit much to separate the splash from patch content. but works well.

Ant1r:
yes thanks @umlaeute ! I couldn't find the way to do it like that, even if I looked for it for a while.

umlaeute:
border and shadow is something that macOS adds...
(well, I do see a border in both the old implementation and the new one; but the shadow is new)

## old look 
image

## new look
image

umlaeute:
apparently i used another offset of the zoom splash (10 instead of 3). this was rather accidental and I have no opinion about it. if you want I can revert it to the original offset.

Ant1r:
There's also some shadow here on Linux, see (I played with the gamma to make it more visible):
Capture d’écran du 2026-09-01 11-48-01
But I quite like it.

OTOH I think I 'd prefer it with less margin/offset, maybe 3 or 4 would be better than 10.

umlaeute:
> There's also some shadow here on Linux

well yeah. it's obviously something that is in control of the window manager (as this is a toplevel window, the wm starts decorating a bit). mine does not do any such fancy things¹.
but if everybody likes it (or doesn't care), then it's all for the better.


i've changed the offset back to 3 (and force-pushed)



¹ ah, i just found the "Show shadows under popup windows" setting under "Window Manager Tweaks" in my settings; but i guess i'll forget about that again :-)

Issue #2991, opened 2026-09-01
[zoom] finer step size


ben-wes:
currently, the zoom octave has ~6~ 5 steps.

with font size 10, this rarely hits a good combination of font size and scale factor.

maybe we could just divide the octave by 12 steps instead? that looks alright here when testing with font sizes 10 and 12 (which seem to be the most common cases to me) ... obviously not in all cases, but there are many steps where text and boxes align quite well.

Ant1r:
actually there are 5 steps to reach the octave, not 6, imo:
let's start from 100%, we have:
114% 131% 151% 174% 200% (5 steps).

That's because internally the octave is divided by 100 units, and, currently, each step is 20 units.

So do you mean to divide the octave in 12 parts? Do you think it would improve the font size accuracy?
Or do you just mean to halve the steps, so that we have 10 steps?

ben-wes:
> actually there are 5 steps to reach the octave, not 6, imo:

ah, you're totally right of course, sorry! hmm ... i was actually thinking of 12 steps. for musical reasons. :) ... but it probably doesn't make a huge difference. i guess, we should prefer what works best with the two common font sizes 10 and 12? either way, i think that 12 is still not too many.

Ant1r:
I'm experimenting on this.
BTW I think it would be better to store the `default zoom level` setting to file in *percent* instead of *steps*.  That would make this setting independent from the actual number of steps per octave, so we'd be allowed to change it to suit our need (currently we can't divide the octave in 12 steps, because 12 does not divide 100).

ben-wes:
> `default zoom level` setting to file in _percent_ instead of _steps_

sounds like a good decision for future compatibility to me.

PR #2995, opened 2026-09-01
Zoom improvement

Ant1r:
fixes #2991
This PR divides the zoom "octave" in 12 steps instead of 5 before.
Also now the `default zoom level` preference is stored in *percent*(%) to file, which allows to safely change the resolution of the zoom.

Also fixes an annoying jumping behavior of the `default zoom level` setting slider.

### DIFF
--- a/tcl/pd_canvaszoom.tcl
+++ b/tcl/pd_canvaszoom.tcl
@@ -14,17 +14,26 @@ namespace eval ::pd_canvaszoom:: {
     variable default_zoom
 }
 
+set ::pd_canvaszoom::steps_by_octave 100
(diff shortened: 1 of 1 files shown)


umlaeute:
sorry for the comment in https://github.com/pure-data/pure-data/commit/2cd69c7da623866bb9afd4cf3fe3ee18f71e0a45 about `::pdtk_canvas::pdtk_canvas_popup`.
for whatever reasons i'm unable to view your replies (but it seems that there *are* replies).

in any case, i see that you've fixed this and then noticed that i had a conflicting fix in `develop` and then you undid your fix. so i guess it works either way.

PR #2996, opened 2026-09-01
pd_free: don't free anything past freemethod if `c_size` is zero

myQwil:
If a developer wants total control over the memory management of their external, such that they also free the pd object, they could set the class size to zero.

Currently, this might already work, but if an allocator deliberately poisons the freed memory, pd will crash when it tries to free patchable items.

### DIFF
--- a/src/m_pd.c
+++ b/src/m_pd.c
@@ -34,6 +34,7 @@ void pd_free(t_pd *x)
 {
     t_class *c = *x;
     if (c->c_freemethod) (*(t_freemethod)(c->c_freemethod))(x);
+    if (!c->c_size) return;
     if (c->c_patchable)
     {
         while (((t_object *)x)->ob_outlet)
@@ -43,7 +44,7 @@ void pd_free(t_pd *x)
         if (((t_object *)x)->ob_binbuf)
             binbuf_free(((t_object *)x)->ob_binbuf);
     }
-    if (c->c_size) t_freebytes(x, c->c_size);
+    t_freebytes(x, c->c_size);
 }
 
 void gobj_save(t_gobj *x, t_binbuf *b)


Spacechild1:
I'm not sure what this is trying to achieve. The current code only calls `t_freebytes` if `c_size` is not 0. If the object is patchable, wouldn't you want to free the inlets, outlets and the binbuf?

In general, the only time I have seen people manually allocating Pd objects is for proxy objects, but in that case you wouldn't call `pd_new` and neither would Pd call `pd_free`.

Can you provide a use case where you would manually allocate a Pd object and then hand it off to Pd?

myQwil:
> Can you provide a use case where you would manually allocate a Pd object and then hand it off to Pd?

When you don't want the Pd object to be the first field in the struct.
The external's free method receives the Pd object pointer, and then instead of casting directly to the containing struct, use a field parent ptr pattern.

> wouldn't you want to free the inlets, outlets and the binbuf?

Yes, but the external's free method would do it instead.

Spacechild1:
Aha, now I understand:

1. you manually allocate the struct with the embedded Pd object and return a pointer to the latter
2. when Pd wants to free the object, it first calls the external's free method. Since this method is responsible for deallocating the whole struct, `pd_free` might access invalid memory.

~~But here's the thing: if you deallocate the struct in your free method, the whole struct becomes invalid, including the class pointer! `pd_free` needs to follow the class pointer to check the size. This just doesn't work.~~

EDIT: I just realized that you're getting the class pointer before calling the free method. My bad!

> When you don't want the Pd object to be the first field in the struct.

Why would you want to do that in the first place?

Spacechild1:
I think you should add a comment that explains the early return!

You might also want to document (in `m_pd.h`) that a class size of 0 is supported and prevents `pd_free` from doing *any* cleanup!

I have yet to see a use case for manually managing the memory of patchable Pd objects, but maybe you can provide one.

myQwil:
> pd_free needs to follow the class pointer to check the size

That part should be fine actually, because of this:
```c
t_class *c = *x;
```
So we've already got the class pointer before calling the free method. We're holding on to the dereferenced value, not the pointer itself, so when we ask about the class's size, pd doesn't crash.

Spacechild1:
Yeah, I realized that shortly after I wrote that comment... See my edit.

Spacechild1:
One case I can think of is putting a `t_object` inside a C++ class with virtual methods. In that case, the `t_object` would always come after the vtable. However, everytime I ended up in that situation, I just rearranged my code and put the C++ class inside the Pd struct. Problem solved.

myQwil:
Modern languages like Rust and Zig have moved away from guaranteed field ordering. They still offer C structs, but particularly in the case of Zig, you can't put Zig structs inside of C structs.

Spacechild1:
> Modern languages like Rust and Zig have moved away from guaranteed field ordering. 

In Rust you have `#[repr(C)]` and in Zig there is `extern struct`. If there was no way of enforcing field ordering, you couldn't interface with other languages or C APIs. That would be pretty pointless.

> you can't put Zig structs inside of C structs.

Looks like you can't embed a Zig struct by value, but you can surely store a pointer to it.

myQwil:
> In Rust you have `#[repr(C)]` and in Zig there is `extern struct`.

Right, that's what I mean when I say they still offer C structs.
Their native struct type is preferable when a well-defined in-memory layout isn't required.
I also like the design philosophy of pure composition over the subtyping pattern that's used a lot in C.

Spacechild1:
> I also like the design philosophy of pure composition over the subtyping pattern that's used a lot in C.

I assume you mean C++? C does not have subtyping.

Anyway, just define your Pd object as a C struct and add a member that points to your Zig/Rust struct. It's still composition. There is really no need to manually allocate and deallocate the Pd object.

myQwil:
> I assume you mean C++? C does not have subtyping.

The pseudo-subtyping pattern, where the first field in a struct shares the same address as its parent.
Pure composition doesn't typically rely on shared addresses. Instead, you would subtract the parent-child offset from the child pointer to get the parent pointer.

Spacechild1:
Ok, but what I'm suggesting is actually composition:
```c
struct pd_foo
{
    t_object x_obj;
    t_float x_f;
    Foo* x_foo; // Foo is your Zig/Rust type
};
```
That's the typical pattern for wrapping libraries in Pd object. Just treat your Rust/Zig struct like any other library type.

Instead of subtracting member offsets to get to the base pointer, you just access `x_foo` in the Pd methods. It's dead simple.

myQwil:
It would work, but I would prefer it be local instead of a separate allocation.

I just want to say that having total control over the memory management of my external is a feature that really appeals to me. That really is what I'm most interested in.

The conversation kind of turned into one about Zig and Rust interoperability with C, but all I really care about is total control over my external's memory. You don't have to free the pd object for me, You can trust me to handle that on my end.

Spacechild1:
I just wanted to understand why such a feature is needed because it sounded like an XY-problem to me.

> It would work, but I would prefer it be local instead of a separate allocation.

Can't you just pass the total size (Pd C struct + Zig/Rust struct) to `class_new` and then initialize your Zig/Rust struct in place after the C struct? (Pretty much like you would use placement-new in C++.) Every systems programming language should be able to do this, otherwise you couldn't even implement a memory allocator.

I really don't think that one should have to reimplement `pd_new` and `pd_free` just to write an external in Zig or Rust. This just seems wrong to me.

(Note that there are environments were you don't even get to control the allocation. In SuperCollider, for example, a UGen is allocated as part of the whole Graph from an internal realtime-memory-pool. There surely must be an established pattern for this.)

Spacechild1:
I'm not entirely against allowing people to manually (de)allocate their patchable Pd objects, but then we would also need functions for (de)initializing the `t_object` part. For example:
```c
/* initializes a Pd object (patchable or not) that has been allocated by the user */
void pd_object_init(t_class *c, t_pd *x);

/* deinitializes a Pd object without freeing the underlying memory */
void pd_object_deinit(t_pd *x);
```

myQwil:
> I really don't think that one should have to reimplement `pd_new` and `pd_free` just to write an external in Zig or Rust.

You can use those functions, it's just less idiomatic for those languages.
I don't know about Rust, but Zig separates allocation from initialization. Functions like `new` and `free` are replaced with `init` and `deinit`.

The difference is that these functions don't self-create or self-destruct. `init` initializes an already-allocated object, while `deinit` releases resources owned by that object.

And yes, I want to deallocate the pd object, but that's only because I was the one who allocated it in the first place.

Spacechild1:
> I don't know about Rust, but Zig separates allocation from initialization

Yes, but that does not mean that you have to be in charge of the allocation! The very point of this separation is that you can put objects in existing memory regions. You can let Pd allocate the memory for you and then initialize your Zig object in that memory (after the C struct part). That's what I was trying to say in https://github.com/pure-data/pure-data/pull/2996#issuecomment-5507784024.

> The difference is that these functions don't self-create or self-destruct. init initializes an already-allocated object, while deinit releases resources owned by that object.

I know. It's just like placement-new in C++. Here's how you typically write a Pd object in C++ (when using C++ types as members):

<details>

<summary>Click me</summary>

```cpp
t_class* foo_class;

struct t_foo
{
    t_object x_obj;
    std::vector<t_float> x_buffer;
};

t_foo::t_foo()
{
    // initialization logic
}

t_foo::~t_foo()
{
    // deinitialization logic
}

void* foo_new()
{
    // pd_new() returns the memory for a t_foo instance and initializes the x_obj member
    auto x = pd_new(foo_class);
    // use placement-new to (default-)initialize the remaining members of t_foo.
    return new (x) t_foo();
}

void foo_free(t_foo* x)
{
    // manually call the destructor to destroy the C++ members.
    x->~t_foo();
}

void foo_setup()
{
    clone_class = class_new(gensym("foo"), (t_newmethod)foo_new,
        (t_method)foo_free, sizeof(t_foo), 0, 0);
}
```
</details>

I'm sure you can do a similar thing in Zig or Rust.

> You can use those functions, it's just less idiomatic for those languages.

What I meant was: if you don't call `pd_new` and `pd_free` you have to manually init and deinit the `t_object` struct, which shouldn't be your business. That's why I proposed `pd_object_init` and `pd_object_deinit` that do this for you.

However, I still stand by my point that you should just use `pd_new` and init your Zig object in the provided memory block.

myQwil:
`pd_object_init` and `pd_object_deinit` sound like a good idea, and I would definitely make use of those functions.

But why does Pd allow me to set the class size to zero and therefore skip the deallocation? Is there a real intended use case for a class size of zero, or is it simply a bug?

If it is just a bug, that's a shame because it seems to lend itself perfectly to the situation that I find myself in.

Here's how I currently write a Zig external:
<details>
<summary>Click me</summary>

```zig
//! Reverse `[moses]`. Outputs numbers to the left if they're greater than control value.

const Sesom = @This();
const pd = @import("pd");
const Pd = pd.Pd;
const Float = pd.Float;

/// Probably the first field, but it's not a promise Zig makes (unless it's a C struct).
obj: pd.Object,
out_l: *pd.Outlet,
out_r: *pd.Outlet,
f: Float,

const name = "sesom";
var class: *pd.Class = undefined;

fn createC(f: Float) callconv(.c) ?*Pd {
	// If an error occurs, `wrap` prints the error name to pd's console and returns null.
	return pd.wrap(*Pd, create(f), name);
}
inline fn create(f: Float) pd.Oom!*Pd {
	// `pd.gpa` is a Zig `std.mem.Allocator` that uses pd's allocation functions.
	// Sometimes it's useful to swap this out with a debug allocator to check for leaks.
	// The debug allocator also poisons freed memory.
	const self: *Sesom = try pd.gpa.create(Sesom);
	errdefer pd.gpa.destroy(self);

	// This is where inlets and outlets are initialized to null.
	// Default field values have been specified in the `pd.Object` struct declaration.
	// `pd.Pd` is a struct so we can use the namespace to associate functions with it.
	self.obj = .{ .g = .{ .pd = .{ .class = class } } };
	const obj: *pd.Object = &self.obj;
	errdefer obj.deinit(); // releases any inlets/outlets that may have been allocated.

	_ = try obj.inletFloat(&self.f);
	self.* = .{
		.obj = self.obj,
		.out_l = try obj.outlet(pd.s.float()),
		.out_r = try obj.outlet(pd.s.float()),
		.f = f,
	};
	return &obj.g.pd;
}

// Returns a function that subtracts the parent-child offset from `*Pd`.
const parentPtr = pd.parentPtr(Sesom, "obj");

fn floatC(p: *Pd, f: Float) callconv(.c) void {
	const self = parentPtr(p);
	(if (f > self.f) self.out_l else self.out_r).float(f);
}

fn destroyC(p: *Pd) callconv(.c) void {
	const self = parentPtr(p);
	defer pd.gpa.destroy(self);

	self.obj.deinit();
}

inline fn setup() pd.Class.Error!void {
	// Creation args are placed before the new method.
	// This allows Zig to check at compile time if `createC` has the correct parameters.
	class = try .create(name, &.{ .deffloat }, createC, destroyC, 0, .{
		// Options go in here.
	});
	class.addFloat(floatC);
}

export fn sesom_setup() void {
	_ = pd.wrap(void, setup(), @src().fn_name);
}
```
</details>

Spacechild1:
> But why does Pd allow me to set the class size to zero and therefore skip the deallocation? Is there a real intended use case for a class size of zero, or is it simply a bug?

That's a good question! After some digging I think I found the answer: This seems to be an internal workaround for `t_scalar`. Since `t_word sc_vec[1];` is a flexible struct member, the overall size of `t_scalar` is not fixed. As a consequence, the memory cannot be freed by `pd_free` because it wouldn't know the size.

Your fix in this PR is certainly correct. It just didn't matter for `t_scalar` because it's not a patchable object.

---

Thanks for the code snippet! Would you mind trying out my approach and compare it with yours?

myQwil:
I have a workaround now. It looks like this:
<details>
<summary>Click me</summary>

```zig
//! Reverse `[moses]`. Outputs numbers to the left if they're greater than control value.

const Sesom = @This();
const pd = @import("pd");

const Pd = pd.Pd;
const Float = pd.Float;
const Object = pd.Object;

out_l: *pd.Outlet,
out_r: *pd.Outlet,
f: Float,

const name = "sesom";
var class: *pd.Class = undefined;

fn createC(f: Float) callconv(.c) ?*Pd {
	return pd.wrap(*Pd, create(f), name);
}
inline fn create(f: Float) pd.Oom!*Pd {
	const obj: *Object = @ptrCast(try class.pd());
	errdefer obj.g.pd.destroy();
	const self: *Sesom = @ptrFromInt(@intFromPtr(obj) + @sizeOf(Object));
	_ = try obj.inletFloat(&self.f);
	self.* = .{
		.out_l = try .create(obj, pd.s.float()),
		.out_r = try .create(obj, pd.s.float()),
		.f = f,
	};
	return &obj.g.pd;
}

fn floatC(p: *const Pd, f: Float) callconv(.c) void {
	const self: *Sesom = @ptrFromInt(@intFromPtr(p) + @sizeOf(Object));
	(if (f > self.f) self.out_l else self.out_r).float(f);
}

inline fn setup() pd.Class.Error!void {
	const size = @sizeOf(Object) + @sizeOf(Sesom);
	class = try .create(name, &.{ .deffloat }, createC, null, size, .{});
	class.addFloat(floatC);
}

export fn sesom_setup() void {
	_ = pd.wrap(void, setup(), @src().fn_name);
}
```
</details>

So basically, the external no longer contains the pd object. To get the external, we jump ahead in memory by the size of one pd object. It hasn't been tried out extensively yet, and there might also be some alignment issues that need to be worked out first, but it looks like this approach will allow me to use the native zig struct without making any changes to `pd_free`.

EDIT: I use this now, which hopefully fixes any alignment issues:
```zig
pub fn Box(T: type) type { return extern struct {
	obj: Object,
	data: [@sizeOf(T)]u8 align(@alignOf(T)),

	pub fn fromObject(p: *const Pd) *T {
		return @ptrFromInt(@intFromPtr(p) + @offsetOf(@This(), "data"));
	}
};}
```

Spacechild1:
> So basically, the external no longer contains the pd object. To get the external, we jump ahead in memory by the size of one pd object.

Yes, that's exactly what I meant!

myQwil:
This can probably be closed now. Thank you @Spacechild1 for your help.

Spacechild1:
You're welcome!

PR #2999, opened 2026-09-03
snake~ improvements

### LINKED ISSUE #2998: [snake~ pick]: indexes should be zero-based
I've noticed that the indices for `[snake~ pick]` seem to start at one instead of zero:

Image

This is inconsistent with the rest of Pd, which generally uses zero-based indexing (with a few exceptions like `[adc~]` and `[dac~]`).

@ben-wes Judging from your code comments, this behavior is on purpose, but I'm really wondering what the rationale was.

I also see a few problems with the implementations of `[snake~ pick]` *). If everyone agrees to switch the indexing to zero-based, I can make a PR that fixes everything in one go.

Spacechild1:
- `[snake~ in]`: optimize for common case where all inputs only have a single channel.
  since single-channel signals never alias multi-channel signals (since they never fall in the same size class), we don't have to make a temporary copy of the inputs.

- `[snake~ split]`: only update DSP if the index changes

- `[snake~ pick]`: channel indices now start at 0

- `[snake~ pick]`: DSP is only updated if the number of picked channels changes. Individual channels can be changed dynamically without interruption.

- `[print~]`: channel numbers start at 0

Closes https://github.com/pure-data/pure-data/issues/2998.


ben-wes:
@millerpuckette @umlaeute whatever the decision on channel/index counting will be - i assume that the improving commits here should certainly make it into 0.57 (258cdabe7a114135390212d18ce55d4424efdaf2 combines 2 changes though)?

millerpuckette:
Actually it's the channel/index count that needs to be frozen in its final state since changing that later would cause future incompatibliities.  The improvements could come now or later as we wish.  My feeling its to go on and merge the whole thing, since I don't have an opinion about channel/index and if the rest introduces bugs it won't be my problem :)

Spacechild1:
> and if the rest introduces bugs it won't be my problem :)

I think I did enough testing, but if I broke something that's on me :)

millerpuckette:
ok, merging...

ben-wes:
ok, thx both! 

fixing my broken [snake~ pick]s now - and trying to get used to calling the first channel "0". ;)

PR #3003, opened 2026-09-07
[zoom] fix text width

### LINKED ISSUE #3002: [zoom] handle text `-width` when zooming
this is not an issue for pd itself with text width always being 0. i stumbled upon it when zooming pdlua gui objects. they mostly work quite well - but zooming doesn't affect the text's `-width` parameter. it's handled correctly on creation, but not when updating the text on zoom.

text width can be set on the pdlua side (forcing text to wrap). i wonder whether it would be a reasonable fix to handle it on the pd side? obviously, a redraw from pdlua fixes it. but it feels a bit incomplete to me to have only partial handling of `-width` within pd.

Ant1r:
fixes #3002 

when zooming, the `-width` option of text items needs to be refreshed.
This is only needed by externals such as `pdlua`, since Pd itself doesn't use the `-width` option.

Also this PR allows to display smaller texts (down to font size 2, instead of 4 before).


Ant1r:
I just tweaked that a little, I think it's good now (despite failing CI build due to osuosl.org being to slow...).

@ben-wes can you confirm that it fixes #3002? thx!

ben-wes:
> @ben-wes can you confirm that it fixes #3002? thx!

thank you - this certainly fixes the wrapping issue for me!

there's one more minor issue now though that might be fixed here as well: pdlua sets the line width to 0 for shapes without border. the following line will add a border on zoom though:

https://github.com/pure-data/pure-data/blob/d9639d2b55736354082f1cd7c1247ac34dcd92ac/tcl/pd_canvaszoom.tcl#L426

do you think it would be valid to add a check for `0 < $newwidth` here?

Ant1r:
yes sure! Thanks for finding that. In fact I think the whole line is useless.

OTOH I just realized that my "fix text width" solution is actually buggy, because Tk stores the width value as integer...
You can see it on the pdlua "graphics" demo, if you zoom out to 6% then re-zoom in, back and forth, widths eventually get damaged.

I'll need to add a new tag I'm afraid.

Ant1r:
Hope it's OK now.
@ben-wes plz test!

ben-wes:
> test!

thanks - lgtm!

Ant1r:
Cool. @umlaeute maybe this can be merged to develop?

Ant1r:
There was another bug, only visible with horizontal slider at moderate zoom out, here at 89%:
Capture d’écran du 2026-09-16 17-51-53

The solution is to round the width to nearest integer. Last commit implements that.

Ant1r:
BTW I fixed a rounding bug with default_zoom when is less than 100%.
Sorry for being off-topic, but I discovered this when trying to configure default zoom to 89%, to help fixing the hslider bug...

PR #3005, opened 2026-09-15
use vector size when choosing reverse scalar perf8

ben-wes:
seems like a minor slip and the case (scalar left, vector right) is probably rather uncommon anyway:

this fixes the check to right-inlet size for use of reverse scalar perf8.


ben-wes:
merged via 460477b

PR #3007, opened 2026-09-17
Fix non-ASCII path corruption in sys_expandpath() on Windows

Zatmcg:
## Problem

`sys_expandpath()` uses the ANSI `ExpandEnvironmentStrings()` on Windows, which decodes the input byte stream with the active code page (ACP). But the paths libpd deals with internally are **UTF-8**. On any path containing non-ASCII characters this pairing is fundamentally broken: UTF-8 bytes are re-decoded as ACP, and the result is either corrupted (mojibake) or replaced with `?` - there is no correct outcome on the ANSI path.

Concrete case on zh-CN (ACP = GBK/936): a UTF-8 multibyte tail byte (`0x80-0xBF`) lands in the GBK lead-byte range; when followed by an ASCII byte below `0x40` (such as the `/` in `D:/...`), the pair is an invalid GBK sequence and the byte is replaced with `?`. Absolute paths containing non-ASCII characters were silently broken: abstract files loaded via `libpd_openfile` with a Chinese-named base dir, and `[open]`/`[text]` inside patches, failed to load with **no error message**.

## Fix

Replace the ANSI call with the wide-character chain already used by `sys_open()`/`sys_fopen()` in the same file:

`u8_utf8toucs2 -> ExpandEnvironmentStringsW -> u8_ucs2toutf8`

This is byte-for-byte lossless for ASCII and correct for all non-ASCII forms.

## Why this stayed hidden (for context)

With the backslash form (`0x5C` is a valid GBK tail byte), `D:\...\??\` paths happened to survive the ACP decode **by coincidence of the GBK layout**; only the forward-slash form (`0x2F < 0x40`) triggers the `?` replacement. This coincidence masked the underlying bug, it did not make the ANSI path correct.

## Verification (Windows zh-CN, hex byte-level)

3-state matrix:

| variant | input bytes | result |
|---|---|---|
| A + backslash | `E9 94 81 5C` | survives (GBK coincidence, not correctness) |
| A + forward slash | `E9 94 3F` | **corrupted** - invalid GBK pair replaced with `?` |
| W + forward slash | `E9 94 81 2F` | intact (the fix) |


umlaeute:
@Zatmcg shouldn't your fix also be applied to this? https://github.com/pure-data/pure-data/blob/6490cb00a41db6dd8a77ab4d69252a7e62c7a50c/src/x_file.c#L155-L159

it would be nice if you could test and update your PR accordingly.

millerpuckette:
I think yes it would be good to fix the other spot too but that can be a separate PR if anone wants to do it :)

Zatmcg:
I didn't intentionally only fix isolated cases. I wasn't confident in my understanding of how the library worked, so I only fixed the bugs I encountered in the most cautious way I could think of.

umlaeute:
fair enough. 

(nevertheless we should strive to avoid code "duplication" where virtually the same thing is done slightly differently)

PR #3008, opened 2026-09-17
stable development branch

### LINKED ISSUE #2960: crash when downsizing upsampled clone with [inlet~ lin] via [siginfo~ channels]
tested on macos arm64, master @ https://github.com/pure-data/pure-data/commit/ee89088440a0c5b2f970bb564e7c0c4c059441d0

EDIT: keeping the example and description below just to reproduce this crash - but noting that there's no need to resize clone like this anymore with https://github.com/pure-data/pure-data/commit/5fe2b284337f996499762281b622f050df9593d7 (via [clone -d voice 0] in this case). 

---

using `[siginfo~ channels]` for setting clone instance count. since resizing works with a manual `[resize 1(` message, it seems to be a timing issue. here's a test patch + abstraction:

* [siginfo-clone-resize.zip](https://github.com/user-attachments/files/30785988/siginfo-clone-resize.zip)

sidenote: without the `[inlet~ lin]`, pd hangs.

### LINKED ISSUE #3002: [zoom] handle text `-width` when zooming

### LINKED ISSUE #3011: garray: wave envelopes look skewed in point mode
garrays in 'points' mode are drawn as lines since the fix in 3b945218dca0a9cd0defa7a8927985438bad9f2b ... this becomes a minor issue though when there's more than 1 point per horizontal pixel (Pd's way of displaying wave envelopes): a slightly diagonal line gets drawn in this case (from lower left to upper right corner) and at some zoom levels, it renders out of bounds on the right.

@Ant1r: i think, it might be an option to stick with rectangles there and use a border of 0?

Image

test patch: [issue-array-envelope.pd.zip](https://github.com/user-attachments/files/32392770/issue-array-envelope.pd.zip)

umlaeute:

it supersedes https://github.com/pure-data/pure-data/pull/2994

---

## GUI
- Closes: https://github.com/pure-data/pure-data/issues/3002
- Closes: https://github.com/pure-data/pure-data/issues/3011

## Core
- Closes: https://github.com/pure-data/pure-data/issues/2960
- Improve/fix `[list join]` and `[list unjoin]`

### DIFF
--- a/src/d_ctl.c
+++ b/src/d_ctl.c
@@ -900,6 +900,7 @@ typedef struct _siginfo_tilde
     t_float x_f;                /* scalar inlet */
     t_siginfo_proxy x_proxy;    /* proxy to receive "pd-dsp-stopped" events */
     t_canvas *x_canvas;         /* canvas we live in (for canvas-local info) */
+    t_clock *x_clock;           /* clock for scheduling messages */
 
     t_symbol **x_vec;           /* object arguments (or NULL) */
     unsigned int x_argc;
@@ -978,7 +979,8 @@ static void siginfo_tilde_dsp(t_siginfo_tilde *x, t_signal **sp)
     x->x_channels = sp[0]->s_nchans;
     x->x_overlap = sp[0]->s_overlap;
     x->x_globaldspstate = 1;
-    siginfo_tilde_bang(x);
+
+    clock_delay(x->x_clock, 0);
 }
 
 static void siginfo_proxy_bang(t_siginfo_proxy *p)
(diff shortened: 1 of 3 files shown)


umlaeute:
@Spacechild1 it seems your commits 5412892a9a4397dd17ab93d3d47b4a7c0a99f5ab (here in `develop`) and c07a4d85fa66ea04dbb0753ee24f0ccac9fe706a (in current `master`) conflict.

would you mind have a look?

umlaeute:
> would you mind have a look?

for now, I have reverted 5412892a9a4397dd17ab93d3d47b4a7c0a99f5ab.

umlaeute:
i have no idea why this PR cannot be merged via the webinterface.
merging works nicely on the cmdline...

